PR #4, opened 2019-04-24
Feature/corbett/umpire sidre GitHub

corbett5:
Here is my first pass at adding Umpire support to sidre. This PR allows the specification of an allocator to all of the Allocate methods in `Buffer`, `View` and `Group` and gives each `Group` a default allocator.

### DIFF
--- a/src/axom/core/memory_management.hpp
+++ b/src/axom/core/memory_management.hpp
@@ -28,6 +28,7 @@ constexpr int INVALID_ALLOCATOR_ID = -1;
 /// @{
 
 #ifdef AXOM_USE_UMPIRE
+
 /*!
  * \brief Returns the umpire allocator associated with the given ID.
  * \param [in] allocatorID the ID of the allocator to get.
@@ -110,6 +111,19 @@ inline void deallocate( T*& p ) noexcept;
  */
 template < typename T >
 inline T* reallocate( T* p, std::size_t n ) noexcept;
+  
+/*!
+ * \brief Copies memory from the source to the destination.
+ *
+ * \param [in/out] dst the destination to copy to.
+ * \param [in] src the source to copy from.
+ * \param [in] numbytes the number of bytes to copy.
+ * 
+ * \note When using Umpire if either src or dst is not registered with the
+ *  ResourceManager then the default host allocation strategy is assumed for
+ *  that pointer.
+ */
+inline void copy( void* dst, void* src, std::size_t numbytes ) noexcept;
 
 /// @}
 
@@ -186,6 +200,32 @@ inline T* reallocate( T* pointer, std::size_t n ) noexcept
 #endif
 
   return pointer;
+}
+
+inline void copy( void* dst, void* src, std::size_t numbytes ) noexcept
+{
+#ifdef AXOM_USE_UMPIRE
+  umpire::ResourceManager & rm = umpire::ResourceManager::getInstance();
+  umpire::op::MemoryOperationRegistry & op_registry = umpire::op::MemoryOperationRegistry::getInstance();
+
+  auto dstStrategy = rm.getAllocator( "HOST" ).getAllocationStrategy();
+  auto srcStrategy = dstStrategy;
+
+  if (rm.hasAllocator(dst))
+  {
+    dstStrategy = rm.findAllocationRecord( dst )->m_strategy;
+  }
+
+  if (rm.hasAllocator(src))
+  {
+    srcStrategy = rm.findAllocationRecord( src )->m_strategy;
+  }
+
+  auto op = op_registry.find( "COPY", srcStrategy, dstStrategy );
+  op->transform( src, &dst, nullptr, nullptr, numbytes );
+#else
+  std::memcpy( dst, src, numbytes );
+#endif
 }
 
 } // namespace axom
--- a/src/axom/core/memory_management.hpp
+++ b/src/axom/core/memory_management.hpp
@@ -86,14 +86,17 @@ inline umpire::Allocator getDefaultAllocator()
  */
 template < typename T >
 #ifdef AXOM_USE_UMPIRE
-inline T* allocate( std::size_t n, umpire::Allocator allocator=getDefaultAllocator() ) noexcept;
+inline T* allocate( std::size_t n,
+                    umpire::Allocator allocator=
+                      getDefaultAllocator() ) noexcept;
 #else
 inline T* allocate( std::size_t n ) noexcept;
 #endif
 
 /*!
  * \brief Frees the chunk of memory pointed to by the supplied pointer, p.
- * \param [in/out] p a pointer to memory allocated with allocate/reallocate or a nullptr.
+ * \param [in/out] p a pointer to memory allocated with allocate/reallocate or a
+ * nullptr.
  * \post p == nullptr
  */
 template < typename T >
@@ -102,7 +105,8 @@ inline void deallocate( T*& p ) noexcept;
 /*!
  * \brief Reallocates the chunk of memory pointed to by the supplied pointer.
  *
- * \param [in] p pointer to memory allocated with allocate/reallocate, or a nullptr.
+ * \param [in] p pointer to memory allocated with allocate/reallocate, or a
+ * nullptr.
  * \param [in] n the number of elements to allocate.
  *
  * \tparam T the type pointer p points to.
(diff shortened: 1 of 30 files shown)


src/axom/core/memory_management.hpp
```diff
@@ -179,6 +201,34 @@ inline T* realloc( T* pointer, std::size_t n ) noexcept
...
 
+inline void copy( void* dst, void* src, std::size_t numbytes ) noexcept
+{
+#ifdef AXOM_USE_UMPIRE
+  umpire::ResourceManager & rm = umpire::ResourceManager::getInstance();
+  umpire::op::MemoryOperationRegistry & op_registry = umpire::op::MemoryOperationRegistry::getInstance();
+
+  // Note auto is used because the types will change when we update umpire.
+  // At that point we should make them concrete.
+  auto dstStrategy = rm.getAllocator( "HOST" ).getAllocationStrategy();
```
davidbeckingsale: Every strategy has the same type, since they are pointers to the base class `umpire::strategy::AllocationStrategy`.

src/axom/core/tests/core_memory_management.cpp
```diff
@@ -195,6 +315,55 @@ TEST( core_memory_management, alloc_realloc_free )
...
+  axom::copy(host_array, src_array, SIZE * sizeof(int));
+
+  for (int i = 0; i < SIZE; ++i)
+  {
+    ASSERT_EQ(host_array[i], i);
+  } 
+}
+
+const std::string copy_locations[] = {
+  "NEW", "MALLOC", "STATIC"
```
davidbeckingsale: 👍

src/axom/core/memory_management.hpp
corbett5: So in the version of Umpire we're using the return type here I believe is `std::shared_ptr<AllocationStrategy>` but this has changed in develop to `AllocationStrategy *` which is why I used auto here.

src/axom/quest/CMakeLists.txt
```diff
@@ -87,21 +87,27 @@ blt_list_append(TO quest_depends_on IF SPARSEHASH_FOUND ELEMENTS sparsehash)
 blt_list_append(TO quest_depends_on IF ENABLE_OPENMP ELEMENTS openmp)
 blt_list_append(TO quest_depends_on IF ENABLE_CUDA   ELEMENTS cuda )
 
+# Add sparsehash when available
+if(SPARSEHASH_FOUND)
```
gzagaris: I would use `blt_list_append` here. Note, there was a recent fix to that macro, so, you may have to fast-forward the version of BLT

src/axom/core/memory_management.hpp
davidbeckingsale: Ah, gotcha!

src/axom/sidre/core/View.hpp
```diff
@@ -383,6 +384,12 @@ class View
    */
   bool isEquivalentTo(const View* other) const;
 
+  /*!
+   * \brief Returns true if both Views are either associated with a buffer or
+   * external, they span the same number of bytes and have unit stride.
```
kennyweiss: Should the data type matter for update-ability ?  E.g. moving floats to ints?

kennyweiss:
Also, our CI has not yet been set up on github.  

Please run against all host-configs using the `scripts/llnl_scripts/build_src.py` on each `SYS_TYPE`.

src/axom/core/tests/core_memory_management.cpp
```diff
@@ -17,27 +17,129 @@
 // HELPER METHODS
 //------------------------------------------------------------------------------
 
-constexpr int ALLOCATION_SIZE = 5345;
+constexpr int SIZE = 5345;
```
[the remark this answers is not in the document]
corbett5: The umpire limit for device constant memory is 64Kb and the `check_alloc_realloc_free` method makes an allocation of `3 * SIZE * sizeof(int)` which works out to 62.6Kb.
https://github.com/LLNL/Umpire/blob/47f3cf087fcfe89b09577fefe474e502a92736c6/src/umpire/resource/CudaConstantMemoryResource.hpp#L25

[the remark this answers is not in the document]
kennyweiss: Thanks @corbett5.  

Could you please add a comment to that effect?  
Also, is 3 related a vector in space?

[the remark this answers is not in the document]
corbett5: Will do.
No it's just a factor to grow the allocation by.

src/axom/core/memory_management.hpp
corbett5: Done.

src/axom/quest/CMakeLists.txt
corbett5: Done. Someone already changed to using `blt_list_append` in develop.

src/axom/core/tests/core_memory_management.cpp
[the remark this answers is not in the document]
corbett5: Done.

src/axom/sidre/core/Group.cpp
```diff
@@ -478,12 +478,20 @@ View* Group::createView( const std::string& path,
  */
 View* Group::createViewAndAllocate( const std::string& path,
                                     TypeID type,
-                                    IndexType num_elems )
+                                    IndexType num_elems,
+                                    int allocID )
 {
+#ifdef AXOM_USE_UMPIRE
```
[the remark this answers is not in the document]
corbett5: Done.

src/axom/sidre/core/View.cpp
```diff
@@ -64,8 +64,15 @@ std::string View::getPathName() const
  *
  *************************************************************************
  */
-View* View::allocate()
+View* View::allocate(int allocID)
 {
+#ifdef AXOM_USE_UMPIRE
```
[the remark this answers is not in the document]
corbett5: Done.

src/axom/sidre/core/Buffer.cpp
```diff
@@ -51,8 +50,15 @@ Buffer* Buffer::describe(TypeID type, IndexType num_elems)
  *
  *************************************************************************
  */
-Buffer* Buffer::allocate()
+Buffer* Buffer::allocate(int allocID)
 {
+#ifdef AXOM_USE_UMPIRE
+  if (allocID == INVALID_ALLOCATOR_ID)
```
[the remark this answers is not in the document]
corbett5: I don't think so, it's defined right at the top of `memory_management.hpp` and we're not prefixing any of the other memory management functions.

src/axom/sidre/core/View.hpp
corbett5: I have a warning in `View::updateFrom` but I'm not sure if we want to prevent it outright.

src/axom/sidre/CMakeLists.txt
```diff
@@ -138,6 +125,12 @@ blt_add_library( NAME       sidre
                  OBJECT     TRUE
                 )
 
+# Disable warning introduced in gcc@8.1 related to how sparsehash casts memory
```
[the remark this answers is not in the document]
corbett5: For whatever reason the flag is now propagating without the changes in the sidre and quest `CMakeLists.txt` so I removed them. I swear I needed them before...

corbett5:
I believe I've addressed all the comments, so I will merge this by the end of the day unless I get any more suggestions.

src/axom/mint/tests/mint_execution_cell_traversals.cpp
```diff
@@ -127,21 +135,28 @@ void check_for_all_cells_ijk( )
...
-  IndexType* jcoords = test_mesh->template createField< IndexType >( "j", CELL_CENTERED );
-  IndexType* kcoords = test_mesh->template createField< IndexType >( "k", CELL_CENTERED );
+  IndexType* icoords = test_mesh->template createField< IndexType >( "i",
+                                                                     CELL_CENTERED );
+  IndexType* jcoords = test_mesh->template createField< IndexType >( "j",
+                                                                     CELL_CENTERED );
+  IndexType* kcoords = test_mesh->template createField< IndexType >( "k",
+                                                                     CELL_CENTERED );
 
   for_all_cells< ExecPolicy, xargs::ijk >( test_mesh,
```
gzagaris: @corbett5 , can you please restrict the ``uncrustify`` changes to Sidre and Core which are related to this PR and not any other file changes in Mint. Unfortunately, ``uncrustify`` doesn't deal well with the formatting of Lambdas and until I have time to tweak what that looks like, I prefer we don't apply formatting to Mint. I'll deal with this in another PR.

corbett5: Done.

PR #7, opened 2019-04-29
Small RTD build fixes

white238:
* Add RTD config file


white238:
RTD uses this settings file for the branch it is currently building but once we (kenny and i) get this working we will create a master release.

white238:
This passed build.  My guess is its not very deterministic on finding the conf.py it is supposed to build, which sounds like a bad idea but i got nothing else.  On a side note, setting your RTD version to private makes it so you can't see your own branch docs.

PR #21, opened 2019-05-03
Feature/elliott/reduce ranks

nselliott:
This enables SPIO reads when running on a smaller number of processors.  It has a restriction that the original files must have been produced in a file-per-processor mode.  Removing this restriction will require some further development.

### DIFF
--- a/src/axom/sidre/docs/sphinx/parallel_io_concepts.rst
+++ b/src/axom/sidre/docs/sphinx/parallel_io_concepts.rst
@@ -46,17 +46,29 @@ management (such as using burst buffers if available).
 
 In typical usage, a run that calls ``read()`` on a certain set of files
 should be executed on the same number of MPI ranks as the run that created
-those files with a ``write()`` call.  Still a ``read()`` call, if using
-the "sidre_hdf5" protocol, can work when called from a greater number of
-processors.  If ``write()`` was executed on N ranks and ``read()`` is called
-while running on M ranks (M > N), then data will be read into ranks 0 to N-1,
-and all ranks higher than N-1 will receive no data.
+those files with a ``write()`` call.  However, if using the "sidre_hdf5"
+protocol, there are some usage patterns that do not have this limitation.
+
+A ``read()`` call using "sidre_hdf5" will work when called from a greater
+number of processors.  If ``write()`` was executed on N ranks and ``read()``
+is called while running on M ranks (M > N), then data will be read into ranks
+0 to N-1, and all ranks higher than N-1 will receive no data.
+
+If ``read()`` is called using "sidre_hdf5" to read data that was created on
+a larger number of processors, this will work only in the case that the data
+was written in a file-per-processor mode (M ranks to M files).  In this case
+the data in the Group being filled with file input will look a bit different
+than in other usage patterns, since a Group on one rank will end up with data
+from multiple ranks.  An integer scalar View named ``reduced_input_ranks``
+will be added to the Group with the value being the number of ranks that
+wrote the files.  The data from each output rank will be read into subgroups
+located at ``rank_*******/sidre_input`` in the input Group's data hierarchy.
 
 .. warning::
-   If ``read()`` is called in an attempt to read data that was created on a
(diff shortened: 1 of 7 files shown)


nselliott:
Domain-overloading is still ultimately a task for the calling code.  This can get them the data they need for it, but it's up to them to determine what they do with it.

src/axom/sidre/docs/sphinx/parallel_io_concepts.rst
```diff
@@ -46,17 +46,29 @@ management (such as using burst buffers if available).
...
+
+If ``read()`` is called using "sidre_hdf5" to read data that was created on
+a larger number of processors, this will work only in the case that the data
+was written in a file-per-processor mode (M ranks to M files).  In this case
+the data in the Group being filled with file input will look a bit different
+than in other usage patterns, since a Group on one rank will end up with data
+from multiple ranks.  An integer scalar View named ``reduced_input_ranks``
+will be added to the Group with the value being the number of ranks that
+wrote the files.  The data from each output rank will be read into subgroups
+located at ``rank_*******/sidre_input`` in the input Group's data hierarchy.
```
kennyweiss: Is ``*******`` a stand-in for a number?  If so, would something like ``{%07d}`` be clearer?

cyrush:
Ok reading more not I am a bit confused. This changes the partitioning and the resulting trees in a strange way.

Can we white board this a bit (as a larger team)?

src/axom/sidre/docs/sphinx/parallel_io_concepts.rst
nselliott: Yes, that would be better.

nselliott:
> Ok reading more not I am a bit confused. This changes the partitioning and the resulting trees in a strange way.
> 
> Can we white board this a bit (as a larger team)?

I did bring this up for discussion in a meeting a few weeks ago.  In this usage a single processor is getting data that was produced by multiply outputting processors, and that output may consist of identically-structured Sidre hierarchies, so there needed to be an extra layer added to keep the data from each processor separate.  The "reduced_input_ranks" View is an indicator that something different has happened and the user needs to be aware.

src/axom/sidre/tests/spio/spio_parallel.cpp
```diff
@@ -809,6 +810,145 @@ TEST(spio_parallel, parallel_increase_procs)
...
+
+  int top_input_rank = 1;
+  if (num_ranks <= 2)
+  {
+    top_input_rank = 0;
+  }
+
+  // Split the communicator so that ranks up to and including
+  // top_input_rank have their own communicator for the input step.
+  MPI_Comm split_comm;
```
gzagaris: I think you need to call `MPI_Comm_free( split_comm )` at the end to free that communicator and properly cleanup any MPI state.

cyrush:
@nselliott sorry I wasn't there, I do understand the challenge with unique names. I am wondering how this will compose with domain overloaded case that we will need to support. Seems like we should clarify the use cases.

`reduced_input_ranks` is documented, which is great but it doesn't seem very intuitive to me.

nselliott:
@cyrush I suppose I see domain-overloading as tangential to the scope of this feature addition.  Someone can create a domain-overloaded mesh without ever using this feature, as long as they run from start to finish without changing the size of their MPI allocation.  IOManager itself doesn't know anything about domains--it does I/O of Groups and Views that have data that could be anything.

cyrush:
@nselliott sounds like you have a specific use case in mind for this --  maybe we can talk about it.

Seems to low bandwidth here, maybe at a future meeting?

nselliott:
The use case is that a restart dump exists and the user wants to re-launch the run on a smaller processor allocation.  In order for this to work, the application code needs to be programmed to know what to do with data that once existed on the higher ranks that no longer exist.  It could handle this with a domain overloading approach, or with a repartition and communication of data.  IOManager has no knowledge of what the calling code wants to do with the data--all that it's doing here is getting the data from the files into memory.

PR #22, opened 2019-05-06
New component spin

agcapps:
Spin contains spatial acceleration data structures (spatial indexes) which previously resided in quest and primal components.


agcapps:
After discussion with @gzagaris , it seems that a pointer + length-based API would be desirable instead of (or in addition to?) the current Primal object-based API.  The idea is to lower the barrier to adoption by host codes by not requiring them to box data up into `Point` or other structs they're not accustomed to. This is a considerable change to the code as it sits, so I'd like to get everyone's opinion.  I also need to do some more work on the documentation.

Once I've brought the docs to an acceptable state and the Axom developers have discussed the possible new API, I'll click "Ready for review" to change this PR from "draft" to "real" PR.

kennyweiss:
**Re: pointer-based interface.**  Sounds like a good idea to have an additional interface that does not directly use the (templated) primal types.  However, since it is orthogonal to this PR and there are already a lot of changes, it would probably be better to hold off for now.

src/axom/quest/docs/sphinx/point_mesh_query_cpp.rst
```diff
@@ -83,12 +67,15 @@ Class header:
    :language: C++
 
 The constructor takes several arguments.  Here, ``surface_mesh`` is a pointer to
-a triangle surface mesh.  The second argument indicates the mesh is watertight.
+a triangle surface mesh.  The second argument indicates the mesh is a watertight
+mesh, a manifold.  The signed distance from a point to a manifold is
+mathematically well-defined.  When the input is not a closed surface mesh, the
+mesh span the entire computational mesh domain, dividing it into two regions.
```
cyrush: must span, or spans?

agcapps:
Docs available at https://axom.readthedocs.io/en/feature-capps2-spin/

src/axom/spin/docs/sphinx/helper.rst
```diff
@@ -0,0 +1,62 @@
...
+.. literalinclude:: ../../examples/spin_introduction.cpp
+   :start-after: _morton_use_start
+   :end-before: _morton_use_end
+   :language: C++
+
+
+The Morton index or Z-index curve is a fractal space-filling curve.  The first
+four iterations of the 2D curve's construction are shown below, and a 3D
+analogue exists as well.
+
```
kennyweiss: @agcapps -- this is really nice!

Could you please discuss the following limitations of the Morton index class:
* The input point type must have integral coordinates. It does not work with floating point coordinates.
* Since the `MortonIndex` interleaves bits from each coordinate of the input point, it has a limited resolution.  Specifically, if the ``MortonIndexType`` is 64 bits, then, in 2D, it can only uniquely index the 32 least significant bits of the `x-` and `y-` coordinates.  Similarly, in 3D, it can only uniquely index the 21 least significant bits of the `x-`, `y-` and `z-` coordinates.

src/axom/quest/MeshTester.cpp
```diff
@@ -27,10 +28,16 @@ using Triangle3 = primal::Triangle<double, 3>;
 
 using Point3 = primal::Point<double, 3>;
 using SpatialBoundingBox = primal::BoundingBox<double, 3>;
-using UniformGrid3 = primal::UniformGrid<int, 3>;
+using UniformGrid3 = spin::UniformGrid<int, 3>;
 using Vector3 = primal::Vector<double, 3>;
 using Segment3 = primal::Segment<double, 3>;
 
+using Point3 = primal::Point<double, 3>;
```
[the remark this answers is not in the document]
agcapps: thanks---got it

src/axom/quest/docs/sphinx/point_mesh_query_cpp.rst
agcapps: Fixed.

src/axom/spin/docs/sphinx/helper.rst
agcapps: done.  Thanks for the suggestions.

src/axom/spin/Primitives.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include "axom/primal/geometry/Vector.hpp"
+
+namespace axom
+{
+namespace spin
+{
+  using primal::BoundingBox;
+  using primal::NumericArray;
+  using primal::Point;
+  using primal::Vector;
```
gzagaris: @agcapps , as we discussed, the fact that you are "importing" primal datatypes within the ``spin`` namespace here is perhaps the biggest argument for why the spatial acceleration data structures are tightly coupled to the geometric primitive types and therefore belong together.

I understand why you are doing it, but creating these aliases here and exposing them in application is at least confusing. For example, the application can now use ``spin::BoundingBox`` but, that class is not part of ``spin`` since it is in ``primal`` that creates confusion. 

It would be better to put all this in an ``internal`` namespace to discourage ``spin::BoundingBox`` to

src/axom/spin/SpatialOctree.hpp
```diff
@@ -29,18 +29,18 @@ class SpatialOctree : public OctreeBase<DIM, BlockDataType>
 {
 public:
 
-  typedef primal::BoundingBox<double,DIM> GeometricBoundingBox;
-  typedef primal::Point<double,DIM> SpacePt;
-  typedef primal::Vector<double,DIM> SpaceVector;
+  using GeometricBoundingBox = BoundingBox<double,DIM> ;
+  using SpacePt = Point<double,DIM> ;
+  using SpaceVector = Vector<double,DIM> ;
```
gzagaris: So, ``SpacePt`` is now an alias to ``spin::Point<double,DIM>``, which is an alias to ``primal::Point< double, DIM>.

All this aliasing can get very confusing, which makes it harder to grok and follow the code. I don't see a strong technical reason to justify all this indirection. It would be better to map aliases directly to the actual type.

src/axom/spin/docs/sphinx/conf.py
```diff
@@ -0,0 +1,282 @@
+# -*- coding: utf-8 -*-
```
gzagaris: This file is no longer needed, since all components use the top-level ``conf.py`` file.

src/axom/spin/tests/spin_bvhtree.cpp
```diff
@@ -5,19 +5,18 @@
...
-TEST( primal_bucket_tree, insert_object )
+TEST( spin_bucket_tree, insert_object )
 {
   // some typedef short-cuts
-  typedef primal::BVHTree< int,3 > TreeType;
-  typedef primal::BoundingBox< double,3 > BoxType;
-  typedef primal::Point< double,3 > PointType;
+  using TreeType = spin::BVHTree< int,3 > ;
+  using BoxType = spin::BoundingBox< double,3 > ;
+  using PointType = spin::Point< double,3 > ;
```
gzagaris: Here is another instance: ``BoxType`` is an alias ``spin::BoundingBox``, which is an alias to ``primal::BoundingBox``. This indirection would be confusing. 

For this and in all other similar instances, it would be better if you the alias maps directly to the actual type instead of another alias. For example do this instead:
```
using BoxType = primal::BoundingBox< double, 3 >;
```

gzagaris:
I don't see a strong technical reason for why spatial acceleration data-structures are not part of primal.    
Given the fact that the spatial acceleration data-structures are tightly coupled to the geometric primitives, it makes more sense to me that it's all inside primal.

That said, I am OK, with moving those to ``spin`` if:

* The public API is defined **only** by primitive types and pointers to primitive types, e.g., 
   ``double``, ``int``, etc.

    We want to avoid:
    ```
    spin::UniformGrid( primal::BoundingBox& bb );
    ``` 
    and avoid
    ```
    spin::UniformGrid( spin::BoundingBox& bb );
    ```
    where ``spin::BoundingBox`` is an alias to ``primal::BoundingBox``

* Hide all primal usage inside the implementation and don't expose ``primal`` through the ``spin`` API.

cyrush:
I also think we should avoid aliasing primal types into the spin namespace -- at least in as far as they are part of the interface. It would help avoid confusion in the implementation as well.

Not sure we need spin to have a primal-free interface, but I do think it would also be useful to have a pointer style interface to spin features. Providing this would also help simplify C and Fortran APIs.

kennyweiss:
I agree that we should not directly add the primal aliases as spin aliases. Instead, we should use the ``primal::`` namespace for these.

Other than that, could we please hold off on any major API changes for this PR, and try to limit the scope to the changes necessary to create the new spin component, as @agcapps has done?

After this, we should have a discussion about the design of primal's interface.

cyrush:
@kennyweiss  - agreed,  i don't think we need the pointer style API as part of this PR

gzagaris:
I think we need to have more fundamental discussions about this. 

I am not sure of what the technical merits of having ``spin`` are vs. putting the spatial acceleration data-structures in ``primal``, as we currently do. What are the benefits of putting those in ``spin``. So far, I see technical reasons of why we shouldn't do that. For example, the confusion with aliases.

We should work towards having the right and agreed upon API first and then create a new component. 

Pushing this now, will cause some turbulence, since application codes would need to be updated to use spin,  and then update again when we make additional API changes. 

Is there any urgency that spin must happen now?

src/axom/spin/MortonIndex.hpp
```diff
@@ -21,8 +21,7 @@
 #include "axom/core/Types.hpp"
 #include "axom/core/Macros.hpp"    // defines AXOM_STATIC_ASSERT
 
-#include "axom/primal/geometry/Point.hpp"
-#include "axom/primal/geometry/Vector.hpp"
+#include "axom/spin/Primitives.hpp"
 
 #ifdef AXOM_USE_CXX11
```
kennyweiss: I created #33 as a reminder to remove guards like ``#ifdef AXOM_USE_CXX11`` that deal with non-`C++11` compilers.

src/axom/spin/docs/sphinx/octree.rst
```diff
@@ -0,0 +1,73 @@
...
+.. literalinclude:: ../../examples/spin_introduction.cpp
+   :start-after: _octree_start
+   :end-before: _octree_end
+   :language: C++
+
+Unlike the ``BVHTree`` class, the ``SpatialOctree`` is intended as a building
+block for further specialization.  Please see the ``quest::InOutOctree`` as an
+example of this.
+
+Some ancillary classes used in the implementation of ``SpatialOctree`` include
```
kennyweiss: This paragraph is really nice!

src/axom/spin/docs/sphinx/octree.rst
```diff
@@ -0,0 +1,73 @@
+SpatialOctree
+^^^^^^^^^^^^^
+
+Axom provides an implementation of the octree spatial index.  The
+``SpatialOctree`` recursively divides a bounding box into a hierarchy of
+non-intersecting bounding boxes.  Each level of subdivision divides the bounding
+box of interest along each of its dimensions, so 2D ``SpatialOctree`` objects
+contain four child bounding boxes at each level, while 3D objects contain eight
+children at each level.  The process stops when each a bounding box contains
```
kennyweiss: > The process stops when....

This sentence is not strictly true.  The `SpatialOctree` (through is base class `OctreeBase`) provides a means to refine a leaf block, but neither of these classes define when this refinement should happen.

See: https://github.com/LLNL/axom/blob/d23ed293165c16f58a301ba2f387efd527416c8d/src/axom/spin/OctreeBase.hpp#L817-L823

kennyweiss: It might be best to remove this sentence.

The sentence that you wrote below is accurate:

> Unlike the ``BVHTree`` class, the ``SpatialOctree`` is intended as a building
block for further specialization.  Please see the ``quest::InOutOctree`` as an
example of this.

src/axom/spin/docs/sphinx/uniformgrid.rst
```diff
@@ -12,7 +12,7 @@ checking if a triangle mesh intersects itself.  The following naive
 implementation is straightforward but runs in :math:`O(n^2)` time, where
```
kennyweiss: (This comment is related to a sentence that was not changed in this PR)

> The ``UniformGrid`` implements the Cell list...

I think it would be more accurate to say that `UniformGrid` can be used to implement a Cell list data structure.
The `UniformGrid` itself does not have a notion of "region of interest", and (I think) the Cell list adds some additional requirements on the spacings for the bins.  (Or at least strongly implies a spacing for optimal performance).

src/axom/spin/Primitives.hpp
agcapps: I removed the spin alias to primal types, as recommended by yourself and by @kennyweiss .

src/axom/spin/SpatialOctree.hpp
agcapps: see previous; resolved

src/axom/spin/tests/spin_bvhtree.cpp
agcapps: as before

src/axom/spin/examples/spin_introduction.cpp
```diff
@@ -0,0 +1,697 @@
...
+// some functions we'll use
+bool expensiveTest(ISpacePt & query, Triangle2DType & tri) ;
+void makeTriangles(std::vector<Triangle2DType> & tris) ;
+// _igrid_header_end
+
+// _rectlattice_header_start
+#include "axom/spin/RectangularLattice.hpp"
+// We'll be using a 2D lattice with space coordinates of type double
+// and cell coordinates of type int.
+using RectLatticeType = axom::spin::RectangularLattice<in2D, double, int> ;
```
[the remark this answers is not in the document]
agcapps: I called out `in2d` in the text.

src/axom/spin/docs/sphinx/octree.rst
agcapps: Thanks.

src/axom/spin/docs/sphinx/octree.rst
agcapps: I removed the sentence "The process stops..."

src/axom/spin/docs/sphinx/uniformgrid.rst
agcapps: I rephrased to avoid the strong statement.

src/axom/spin/docs/sphinx/conf.py
agcapps: thanks; deleted

src/axom/spin/MortonIndex.hpp
agcapps: Thanks.

I removed Primitives.hpp.

white238:
Approved on the condition you update BLT.

PR #30, opened 2019-05-13
Feature/white238/updatetoss3compilers

white238:
Removes gcc4.9.3 (old and also tested on blueos), clang3.9.1, icc16, and icc17
Adds clang6 and icc19


scripts/uberenv/spack_configs/toss_3_x86_64_ib/compilers.yaml
```diff
@@ -6,11 +6,24 @@ compilers:
     modules: []
     operating_system: rhel7
     paths:
-      cc:  /usr/tce/packages/gcc/gcc-4.9.3/bin/gcc
-      cxx: /usr/tce/packages/gcc/gcc-4.9.3/bin/g++
+      cc:  /usr/tce/packages/clang/clang-4.0.0/bin/clang
+      cxx: /usr/tce/packages/clang/clang-4.0.0/bin/clang++
       f77: /usr/tce/packages/gcc/gcc-4.9.3/bin/gfortran
```
kennyweiss: Should the gfortran version be updated as well?

(If the answer is yes, it could also be done separately)

scripts/uberenv/specs.json
```diff
@@ -14,15 +14,13 @@
     "__comment__":"##############################################################################",
 
     "toss_3_x86_64_ib":
-    [ "clang@3.9.1",
-      "clang@4.0.0+devtools",
-      "gcc@4.9.3",
+    [ "clang@4.0.0+devtools",
+      "clang@6.0.0",
```
corbett5: Do we want to change the devtools to clang6? If so it could be done in another PR.

white238: It doesn't matter what compiler they are using.  Soon someone will rip them out of our axom package completely into their own spack package so its not tied to our compiler suites at all.

scripts/uberenv/spack_configs/toss_3_x86_64_ib/compilers.yaml
white238: <shrugs>

PR #31, opened 2019-05-15
Remote unneeded headers, remove string utils

white238:
Played around with include-what-you-use. Lots of false positives, like 90+%, so I only did code I was familiar with.  While I was at it I removed StringUtilities which is not needed now that we require C++11.


cyrush:
Looks good!

PR #37, opened 2019-05-17
Enable Slam's Sets to represent Handle types

kennyweiss:
This PR generalizes the `PositionType` and `ElementType`s that Slam's Set classes can represent.
Notably, we can now have a Set over handle objects, i.e. a class that wraps an index without directly exposing it.  See the new ``HandleMesh`` example in the `slam/examples` directory.

This PR touches a lot of files due to the new template parameters added to Slam's Set, Relation and Map classes.

Misc:
* Slam's OrderedSet iterator is now an stl-compliant random access interator
* There is also some code cleanup to replace typedefs with type aliases
* Added some new tests and improved others
* Adds a function to file utilities to change the cwd. This was helpful before we had the `AXOM_DATA_DIR`, and seemed useful enough to keep in the code.


src/axom/slam/docs/sphinx/index.rst
```diff
@@ -3,14 +3,16 @@
...
+=======================
 Slam User Documentation
 =======================
 
 Axom's Set-theoretic Lightweight API for Meshes (SLAM) component provides high performance
 building blocks for distributed-memory mesh data structures in HPC simulation codes.
 
-Introduction
-------------
+.. raw:: html
```
kennyweiss: This is how I "fixed" the Slam user docs section headings.
(The nested `toc` calls in our unified docs promoted the heading in the index.rst file)

I'd be happy to hear if there is a cleaner way to do this.
If people are happy with this approach, I can apply this to the other affected components.

src/axom/slam/examples/HandleMesh.cpp
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2017-2019, Lawrence Livermore National Security, LLC and
```
kennyweiss: New Slam example where the `PositionType` is an integer, but the `ElementType` is a "handle" type that wraps an integer.

src/axom/slam/examples/HandleMesh.cpp
```diff
@@ -0,0 +1,99 @@
...
+  const int sz = 5;
+  HandleSet::IndirectionBufferType vecHandle(sz);
+
+  // Create a set of handles
+  HandleSet hSet = HandleSet::SetBuilder()
+                   .size( sz )
+                   .data( &vecHandle );
+
+  // Add handles with (somewhat) arbitrary IDs to the set
+  for(auto i : hSet.positions() )
```
kennyweiss: This is also new -- I added a `positions()` function to `slam::OrderedSet`, which makes it easy to iterate over the indices (from `0` to `set.size()`). We can then access the set element using the subscript operator, e.g. `hSet[i]`.

src/axom/quest/InOutOctree.hpp
```diff
@@ -17,8 +17,7 @@
 #include "axom/slam.hpp"
 #include "axom/primal.hpp"
 #include "axom/mint.hpp"
-
-#include "axom/spin/SpatialOctree.hpp"
+#include "axom/spin.hpp"
```
white238: I might be remembering our conversation wrongly but I thought we decided not to use the unified headers in our code due to compilation time concerns.  Am I or have we decided against that decision?

kennyweiss: My recollection was that within a component, we should try to use the unified headers for the dependent components, but not for itself.

(and perhaps we should revisit this if compilation time increases, but it certainly simplifies header inclusions until then...  It is also possible that using unified headers can accelerate compilation times if the compiler is always including things in the same order.).

src/axom/slam/examples/HandleMesh.cpp
cyrush: Just so I understand `slam::PositionType` is a default type, that is an integer?

kennyweiss: Yes.  They are defined here:
https://github.com/LLNL/axom/blob/6704667d9981e6d3cf9aaffb8c92a560f7809a1a/src/axom/slam/Utilities.hpp#L22-L24

kennyweiss:
@cyrush -- I added the fancier user-defined `ElementType` parameters to support a user code that has `Handle` types which wrap an integer index but does not directly expose the underlying integer.  

As you suggested, they use handles for improved compile time type safety. In this use case, providing a handle to a vertex handle when one expected an edge handle will result in a compilation error.

The default `ElementType` is for it to be the same as `PositionType`.

src/axom/slam/examples/HandleMesh.cpp
cyrush: Thanks. I was a little confused b/c use of PosType throughout the implementation of the templates and slam::PositionType as a concrete type. 

For example, cases like:
PosType = slam::PositionType

It might help to call these things like:

`slam::DefaultPositionType`

cyrush:
@kennyweiss thanks for clarifying. 

To provide context for users it would be good to write up when to leverage the complex paths (to support type safety) vs when to use the default types. Also an explicit call out of these default types in the docs would help. (Sorry if this is there, can't really tell from the diffs) Doesn't have to be part of this PR, but I recommend adding an issue as a reminder. 

One thought was to try to make the default case easy to wield - if we expect those to be the common cases (some way that doesn't require the extra template params).  Maybe that is already there, not quite sure?  If it's not there, there simply might not be a sane way to do that w/o adding more templating.

src/axom/slam/examples/HandleMesh.cpp
kennyweiss: Great suggestion @cyrush -- I'll try to make that change.

kennyweiss:
Thanks for the great feedback @cyrush.
I will add a note to update the docs.  There is not really a good place to add it right now, but there will be in my next docs PR for slam.

Slam does have a few simple utility classes to help set up set classes when you don't need the extra complexity. They all derive from ``slam::OrderedSet`` (which has lots of template parameters) but they fix default values for many of the template parameters that do not need to be exposed to the user, and provide defaults for the remaining ones.

Here are the most useful:

```cpp
// in slam/RangeSet.hpp
slam::PositionSet<P,E>  // runtime size, compile time stride and offset, no indirection
slam::RangeSet<P,E>    //  runtime size and offset, compile time stride, no indirection

// in slam/IndirectionSet.hpp
slam::ArrayIndirectionSet<P,E> // runtime size, compile time stride and offset, array indirection
slam::VectorIndirectionSet<P,E> // runtime size, compile time stride and offset, vector indirection
```

All of them have default template arguments, but after this PR, they must be declared with the empty angle brackets.  I.e.   `PositionSet` becomes `PositionSet<>`.

Do you think empty angle brackets for the default case are too burdensome for users? 
I think we'd still want these classes to have the ability to be templated on `PosType` (and possibly on `ElemType`) in some cases, so I don't think we'd want to remove the angle brackets...

src/axom/slam/examples/HandleMesh.cpp
kennyweiss: Done. I think this is a lot cleaner.
...and it also uncovered a few places where the code was using the wrong type!

Thanks @cyrush.

Issue #42, opened 2019-05-20
SLIC enumerator Info is not prefixed properly


white238:
A code ran into a conflicting symbol with SLIC's enumerator "Info" from wrapSLIC.h.

After talking to @gzagaris, we came to the conclusion that Shroud is not prefixing the individual emerators in the enumeration.  I assume this is just an oversight in Shroud and should be fixed at that level.  @ltaylor16, thoughts?

For example, this:

enum SLIC_Level {
  Info
}

should be this:

enum SLIC_Level {
  SLIC_Info
}

ltaylor16:
I was able to update slic_shroud.yaml to change the option C_enum_member_template to include more scoping information.  I'll make this the default value in Shroud.

ltaylor16:
Does this mean that they were using the C wrapper?

gzagaris:
I don't think the C wrapper was used. They were just running into compilation issues since that code had symbols with the same name as the levels we use in Slic.

ltaylor16:
I see that wrapSLIC.h is included by `include/axom/slic.hpp`, which I think is generated.  I think that the files in `interface/c_fortran` should not be included since they're intended for C.

gzagaris:
Yes, ``include/axom/slic.hpp`` is the generated unified header. I could certainly exclude the generated C headers from being included in that file since they are intended for C/Fortran.

gzagaris:
@ltaylor16  fixed this in 9017f6c2ff279d7a4c4037783311ba70dcd6af33

PR #44, opened 2019-05-22
Fix enum values in slic

ltaylor16:
Change C_enum_member_template to add more scope into the name.  A user
was getting a conflict from 'Info', it is now 'SLIC_Level_Info'.

Other changes are whitespace changes from generating with new
copyright text.

Fixes issue #42


src/axom/slic/interface/c_fortran/wrapSLIC.h
```diff
@@ -26,11 +26,11 @@ extern "C" {
...
-  Error,
-  Warning,
-  Info,
-  Debug,
-  Num_Levels
+  SLIC_Level_Error,
+  SLIC_Level_Warning,
+  SLIC_Level_Info,
+  SLIC_Level_Debug,
+  SLIC_Level_Num_Levels
```
gzagaris: @ltaylor16 could you explain how a Fortran code would access these enums? Would it be possible to name them `SLIC_Error` instead of `SLIC_Level_Error` or

ltaylor16: I changed the names to remove Level.  I'm still working on Shroud changes which need to include namespace and class names into the member names.  This is to deal with the case where the same enum name is used in several scoped enums. C needs to flatten the scopes. This change should get the user going without having to update TPLs as part of the update.

Fortran generates `integer(C_INT), parameter :: level_error = 0`.  Like C, the `level` here is probably redundant since it's not possible to have another symbol at the same level named `error`.  Notice that it does not have a prefix like C does.  Since this is in a module, the user can resolve conflicts by controlling which symbol is used.  For example, `use slic_mod, only : level_error`.

gzagaris: Great -- thanks for the explanation.

PR #48, opened 2019-05-28
Some updates for uberenv TPL setup on macos

kennyweiss:
* Resolves #46 
* Resolves #47

PR #49, opened 2019-05-30
Adds openmp variant to Axom's local umpire spack package

kennyweiss:
Looks like I merged my previous macos uberenv PR (#48) too soon...

This PR allows disabling openmp in our local umpire spack package for compilers that do not support it (e.g. the built-in clang on macos).  Umpire enables openmp by default, so we need to be able to explicitly disable it.

After this PR, I was able to configure uberenv with the default compiler using the following spec
```bash
--spec=%clang~openmp^conduit@master
```

Notes:
* I had to use a newer version of conduit due to an incompatibility between 32-bit and 64-bit int types with the current Axom and conduit version 0.3.1 on mac (#50)
* My laptop's Fortran compiler is currently not working, so my test build set the Fortran compiler to `None` and my spec also disabled mpi, e.g. ``--spec=%clang~openmp~mpi^conduit@master``

### DIFF
--- a/scripts/uberenv/packages/uberenv-axom/package.py
+++ b/scripts/uberenv/packages/uberenv-axom/package.py
@@ -81,8 +81,10 @@ class UberenvAxom(Package):
     depends_on("raja~openmp+cuda", when="+raja~openmp+cuda")
     depends_on("raja+openmp+cuda", when="+raja+openmp+cuda")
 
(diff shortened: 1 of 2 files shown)


davidbeckingsale:
Just FYI, Umpire no longer enable OpenMP by default soon: https://github.com/LLNL/Umpire/pull/110

Issue #23, opened 2019-05-07
Incorrect blueprint index generation


kennyweiss:
In sidre and spio, there is a function to automatically generate the "blueprint index" from a "mesh blueprint"-conforming group in the sidre hierarchy.  This is used by Visit's mesh blueprint plugin, to visualize mesh data output by sidre. 

In the current implementation, the function takes the name of the mesh as one of its parameters ("mesh_name"), but for hdf5 files, it should really be using the path to the mesh. 

There might need to be some logic in the function's implementation to correct the provided path for hdf5 files.  (Possibly because hdf5 renames the top-level group?)

See: 
* https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html
* https://github.com/LLNL/axom/blob/b8d1ae4cdfd6698c428025b362de2faa463b7501/src/axom/sidre/core/DataStore.hpp#L375-L400
* https://github.com/LLNL/axom/blob/b8d1ae4cdfd6698c428025b362de2faa463b7501/src/axom/sidre/spio/IOManager.hpp#L165-L187

cyrush:
@kennyweiss for Sidre meshes, this should be the path in Sidre to the root of the mesh.

nselliott:
I recall that I've produced a blueprint using the SPIO interface that is readable by VisIt.  I'll try to dig that up.

kennyweiss:
Thanks @cyrush  and @nselliott.

I was planning to make the examples a bit more complicated as well.  E.g. where the mesh is not located at the root of the hierarchy.

nselliott:
@kennyweiss I was mistaken--the example I have was from calling conduit's generate_index directly, not using sidre at all.  Nonetheless it is a working example of using this kind of feature to generate VisIt-readable files, so we should get SPIO to produce effectively the same thing if it can't already do so.

nselliott:
@kennyweiss and @cyrush I just gave you on our network a tarball of a simple blueprint-based json vis dump produced by SAMRAI calling conduit's generate_index.  It contains only one field with a linear profile on a single-level mesh.  If SPIO can't do this, then we need to fix it, which is probably a matter of getting the names and string arguments correct.

kennyweiss:
Thanks @nselliott -- I was able to open that dataset in visit.

The problem might be restricted to cases where the mesh isn't located at the root of the hierarchy.

kennyweiss:
[...]

PR #51, opened 2019-06-04
Make blueprint index generation accept mesh path instead of mesh name

nselliott:
This addresses issue #23.  The mesh_name argument in IOManager:: writeBlueprintIndexToRootFile is changed to mesh_path to allow for cases where the mesh is not at the root of the SPIO-generated HDF5 file.  mesh_path can still be a simple name if the mesh is at the root.  The method signature is not changed, so existing calling code does not need to be modified.


src/axom/sidre/spio/IOManager.cpp
```diff
@@ -1086,10 +1086,18 @@ void IOManager::writeBlueprintIndexToRootFile(DataStore* datastore,
   AXOM_DEBUG_VAR(root_file_id);
   SLIC_ASSERT(root_file_id >= 0);
 
-  std::string bp_index("blueprint_index/" + mesh_name);
+  std::string blueprint_name;
+  std::string path_to_mesh;
+  std::string slash = "/";
```
kennyweiss: [I don't imagine this will change anytime soon, but] 
this function should probably be using ``sidre::Group::s_path_delimiter`` rather than "slash".

src/axom/sidre/spio/IOManager.hpp
```diff
@@ -172,19 +172,31 @@ class IOManager
...
+   * sidre_hdf5.  This must be called after calling write().
+   *
+   * The parameters domain_path and mesh_path are related. domain_path is the
+   * path from the root of the DataStore to the domain that is being used to
+   * generate the index.  mesh_path is the path from the group that was
+   * passed into the preceding write() call to the same domain.  If write()
+   * was called using the root group of the DataStore, then domain_path and
+   * mesh_path will be identical.  Otherwise mesh_path is a sub-path of
+   * domain_path.
+   *
```
kennyweiss: @nselliott -- would it be possible to add a simple, non-trivial example to the doxygen comment?

src/axom/sidre/spio/IOManager.hpp
```diff
@@ -172,19 +172,31 @@ class IOManager
...
+   * This is not an MPI collective call.  One rank writes to blueprint index
+   * to one root file.
    *
    * \param datastore     DataStore containing Groups that hold domains
    *                      that adhere to the Blueprint format
-   * \param domain_path   path to the domain in the DataStore that will be
+   * \param domain_path   path in the the DataStore to the domain that will be
```
kennyweiss: typo: the the

src/axom/sidre/core/DataStore.cpp
```diff
@@ -574,7 +574,9 @@ bool DataStore::generateBlueprintIndex(const std::string& domain_path,
   conduit::Node mesh_node;
   domain->createNativeLayout(mesh_node);
 
-  Group* bpindex = getRoot()->createGroup(index_path);
+  Group* bpindex = getRoot()->hasGroup(index_path)
```
kennyweiss: thanks

src/axom/sidre/examples/sidre_createdatastore.cpp
```diff
@@ -415,6 +415,60 @@ void generate_blueprint(DataStore* ds) {
...
 
+void generate_blueprint_to_path(DataStore* ds) {
+  // _blueprint_generate_path_start
+  // Conduit needs a specific hierarchy.
+  // For this example, we locate the domain data at a path
+  // that is not the top level of its Sidre heirarchy.
+  DataStore cds;
+  std::string domain_name = "domain";
+  std::string domain_location = "domain_data/level/domains/" + domain_name;
+  std::string mesh_name = "mesh";
```
kennyweiss: Do we have a test/example where the mesh in not named "mesh"?  
If not, could you please change this name to something else?

src/axom/sidre/examples/sidre_createdatastore.cpp
```diff
@@ -415,6 +415,60 @@ void generate_blueprint(DataStore* ds) {
...
+  setup_blueprint_coords(ds, coords);
+
+  setup_blueprint_topos(ds, topos);
+
+  setup_blueprint_fields(ds, fields);
+
+  // _blueprint_generate_path_save_start
+  conduit::Node info, mesh_node, root_node;
+  cds.getRoot()->createNativeLayout(mesh_node);
+  if (conduit::blueprint::verify("mesh", mesh_node[domain_mesh], info))
```
kennyweiss: The "mesh" here is the protocol, right?
Could you please make this a variable, rather than a string literal, to highlight this?

src/axom/sidre/examples/sidre_createdatastore.cpp
```diff
@@ -415,6 +415,60 @@ void generate_blueprint(DataStore* ds) {
...
+  {
+    std::string bp("rootfile_data/blueprint_index/automesh");
```
kennyweiss: How does the name "automesh" relate the the mesh in the hierarchy named "mesh"?

Will this show up in visit as "automesh", even though it is named "mesh" in the actual hierarchy?

src/axom/sidre/examples/sidre_createdatastore.cpp
```diff
@@ -415,6 +415,60 @@ void generate_blueprint(DataStore* ds) {
   // _blueprint_generate_save_end
 }
 
+void generate_blueprint_to_path(DataStore* ds) {
```
kennyweiss: @nselliott -- could you please add a description to the top of the file that explains the different meshes that are generated/output by this example?

src/axom/sidre/spio/IOManager.hpp
kennyweiss: Clarification: I was asking for something like this: 
https://github.com/mfem/mfem/blob/edfb62d8c27061a0de2b495eceffcee266b3d2cb/fem/sidredatacollection.hpp#L50

(although, not necessarily as fancy).

src/axom/sidre/spio/IOManager.cpp
cyrush: @kennyweiss the access code is non trivial -- best way to validate is using VisIt to read. 

I guess what you are after is some check for the generic path construction. I wan't to move that to conduit but its not there yet (it's in visit and ascent)?

cyrush: https://github.com/visit-dav/visit/blob/develop/src/databases/Blueprint/avtBlueprintTreeCache.C

src/axom/sidre/examples/sidre_createdatastore.cpp
nselliott: This will show up as "automesh" in VisIt.  As I understand it, there is no requirement that the name held by the blueprint index has to match the name of the mesh in the hierarchy.  Some of the other examples in this file do have the names matching.

src/axom/sidre/spio/IOManager.cpp
kennyweiss: Thanks @cyrush -- so, should we just wait for this functionality to be available in conduit and then add a call here?

src/axom/sidre/examples/sidre_createdatastore.cpp
nselliott: I've given them some different names now.

src/axom/sidre/examples/sidre_createdatastore.cpp
nselliott: I've added a brief description of every function called from main().

src/axom/sidre/examples/sidre_createdatastore.cpp
nselliott: Got it.

src/axom/sidre/spio/IOManager.hpp
nselliott: I have added something here.

kennyweiss: Thanks @nselliott -- this is perfect.

src/axom/sidre/spio/IOManager.cpp
nselliott: I've changed the delimiter--to access it I have to call a method on the root group, even though it's static data.  Should Group have a static accessor method for this char?

kennyweiss: I noticed that too -- it definitely should.

Right now, it has a member function to access the static variable.  This function should probably just be made static.

https://github.com/LLNL/axom/blob/1043eeed4973a464b395cd6af93c5e2562015238/src/axom/sidre/core/Group.hpp#L131-L137

nselliott:
If anyone can give this a second approval, please do so.

src/axom/sidre/spio/interface/c_fortran/typesSPIO.h
```diff
@@ -4,7 +4,8 @@
 // Copyright (c) 2017-2019, Lawrence Livermore National Security, LLC and
 // other Axom Project Developers. See the top-level COPYRIGHT file for details.
 //
-// SPDX-License-Identifier: (BSD-3-Clause)
+// SPDX-License-Identifier (BSD-3-Clause)
```
ltaylor16: It looks like other uses of SPDX have the colon.

nselliott: This is what was generated running shroud.  I don't know if there's inconsistent behavior somewhere.

ltaylor16: The YAML file has the colon.  I'll check it out.  Don't let it hold up the pull request.

PR #53, opened 2019-06-18
bugfix/zagaris2/fix-bqq-build

gzagaris:
The BG/Q build was failing when CMake was trying to configure
due to using undefined variables in conditional logic.

Specifically, checks of the form:
```
if ( ${RAJA_FOUND} AND ${UMPIRE_FOUND} )
    ...
endif()
```

Would cause CMake to fail when ${RAJA_FOUND} and/or ${UMPIRE_FOUND}
were not defined. These variables are set when calling
```
find_package_handle_standard_args()
```

However, when RAJA_DIR or UMPIRE_DIR are not specified, the code
will not set those variables. Consequently, downstream CMake code that
relies on checking for those variables would fail to configure.

This commit fixes the issue encountered on BG/Q where we don't
specify `UMPIRE_DIR`.

### DIFF
--- a/src/axom/spin/CMakeLists.txt
+++ b/src/axom/spin/CMakeLists.txt
@@ -39,6 +39,7 @@ if ( RAJA_FOUND AND UMPIRE_FOUND )
   set( spin_headers
        ${spin_headers}
        BVH.hpp
+       execution_space.hpp
 
        ## internal
        internal/linear_bvh/aabb.hpp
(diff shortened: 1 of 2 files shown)


white238:
This has to due with how CMake parses variables in the if statements. This can be fixed by changing the line to:

if(RAJA_FOUND AND UMPIRE_FOUND)

This will fail if either of the variables are undefined or if not a false constant.

We are already following this practice in all other places and i recommend we continue to follow it.

gzagaris:
@white238  I agree. What you proposed is the appropriate treatment for this. Thanks!

PR #54, opened 2019-06-21
Feature/mir

MarkoSterbentz:
This branch contains a working implementation of the Equi-Z algorithm for 2D material interface reconstruction. In order to improve the code and make it more extensible, I plan to split up the InterfaceReconstructor class into a few different sub-classes in order to make the code easier to read and easier to write unit tests.


kennyweiss:
Thanks @MarkoSterbentz.

First order of business -- it looks like you included the `uberenv_libs` in your PR.  We will need to remove this from your git history since it adds thousands of large files.  I can help you with this on Monday.

src/axom/mir/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+    MIRMesh.hpp
+    MIRMeshTypes.hpp
+    ZooClippingTables.hpp
+    InterfaceReconstructor.hpp
+    CellData.hpp
+    MeshTester.hpp
+    )
+
+set(mir_sources
+    ../Axom.cpp
```
kennyweiss: Please remove ``../Axom.cpp`` from the source list -- it is no longer necessary

src/axom/mir/InterfaceReconstructor.cpp
```diff
@@ -0,0 +1,715 @@
...
+//--------------------------------------------------------------------------------
+
+InterfaceReconstructor::~InterfaceReconstructor()
+{
+
+}
+
+//--------------------------------------------------------------------------------
+
+mir::MIRMesh InterfaceReconstructor::computeReconstructedInterface(mir::MIRMesh& inputMesh)
```
kennyweiss: The resulting mesh should probably be passed in as an OUT parameter rather than returned.

src/axom/mir/MIRMesh.hpp
```diff
@@ -0,0 +1,200 @@
...
+       * \brief Initializes a mesh with the provided data and topology.
+       * 
+       * \param _verts  The set of vertices of the mesh.
+       * \param _elems  The set of elements of the mesh.
+       * \param _numMaterials  The number of materials present in the mesh.
+       * \param _topology  The topology/connectivity of the mesh.
+       * \param _mapData  The data used to initialized the maps associated with the vertex and element sets.
+       * \param _elementVF  The volume fractions of each element. Note that this is an optional parameter.
+       */
+      void  initializeMesh(VertSet _verts, ElemSet _elems, int _numMaterials, CellTopologyData _topology, CellMapData _mapData, std::vector<std::vector<axom::float64> > _elementVF = {});
```
kennyweiss:
```cpp
 std::vector<std::vector<axom::float64> > _elementVF
```

Please make sure that larger/more complex objects are passed by (const) reference.

BradWhitlock:
@rhornung67, @kennyweiss - Yes, the iterative scheme needs to be added to EquiZ. This mir work lives in Axom's MIR component in the "reference" directory.

PR #59, opened 2019-06-27
Feature/gunderman/bezier

davidgunderman:
This is a draft pull request that is nearly the same as the pull request from yesterday. Bezier Curve class has been added, along with unit tests for the class and methods. This is not intended to be merged in-- it is simply a draft pull request so that Kenny can provide feedback.


src/axom/primal/operators/intersect.hpp
```diff
@@ -299,7 +306,49 @@ bool intersect(const OrientedBoundingBox< T, 3 >& b1,
...
+ *
+ * Finds all intersection points between the two curves.
+ *
+ * \note This function assumes that the curves are in general position.
+ * Specifically, we assume that all intersections are at points and that
+ * the curves don't overlap.
+ *
+ * \note This function assumes that the curves are half-open, i.e. they
+ * contain their first endpoint, but not their last endpoint. Thus, the
+ * curves do not intersect at \f$ s==1 \f$ or at \f$ t==1 \f$.
```
kennyweiss: This PR does not handle the case where the curves intersect at their last endpoint.  
We will address this in a forthcoming PR.

src/axom/primal/operators/detail/intersect_bezier_impl.hpp
```diff
@@ -0,0 +1,204 @@
...
+template < typename T, int NDIMS>
+bool intersect_2d_linear( const Point<T,NDIMS> &a, const Point<T,NDIMS> &b,
+                          const Point<T,NDIMS> &c, const Point<T,NDIMS> &d,
+                          T &s, T &t)
+{
+  // Implementation inspired by Section 5.1.9.1 of
+  // C. Ericson's Real-Time Collision Detection book
+
+  // Note: Uses exact floating point comparisons since the subdivision algorithm
+  // provides both sides of the line segments for interior curve points.
```
kennyweiss: Note -- we are intentionally using exact floating point comparisons here, rather than fuzzy comparisons.  This prevents the subdivision algorithm from finding multiple occurrences of the same intersection.

src/axom/primal/operators/detail/intersect_bezier_impl.hpp
```diff
@@ -0,0 +1,204 @@
...
+ * \param t_scale The scale in parameter space for \a c2
+ *
+ * Bezier curves can only intersect when their bounding boxes intersect.
+ * The base case of the recursion is when we can approximate the curves as
+ * line segments, where we directly find their intersections. Otherwise,
+ * check for intersections recursively after bisecting one of the curves.
+ *
+ * \note A BezierCurve is parametrized in [0,1). The scale and offset parameters
+ * are used to track the local curve parameters during subdivisions
+ *
```
agcapps: Please note what happens when the Bezier curves are identical or have sub-sections that are identical.  Also note what happens when the Bezier curves have sections that are very close, such as a tangent intersection.

src/axom/primal/tests/primal_bezier_intersect.cpp
```diff
@@ -0,0 +1,364 @@
...
+                                   0.6475905573955213,
+                                   0.872202573839462,
+                                   0.9370948395920837,
+                                   0.9853581324900483};
+
+  checkIntersections(curve1, curve2,
+                     exp_s, exp_t,
+                     eps, eps_test);
+}
+
```
agcapps: Consider adding
1. Intersection tests between Bezier curves of different orders
2. Test non-intersecting curves
3. Test curves embedded in 3-space
4. Tangent intersections; identical (sections of) curve (if the algorithm will handle this)

davidgunderman: I think commit https://github.com/LLNL/axom/pull/59/commits/1d8a83ddc3468fd8d5054f1894db9d4efc741e42 addresses 1. and 2. and 3. I've also created issue #71 , which will require more thought to deal with 4.

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -0,0 +1,335 @@
...
+#include "axom/primal/geometry/NumericArray.hpp"
+#include "axom/primal/geometry/Point.hpp"
+#include "axom/primal/geometry/Vector.hpp"
+#include "axom/primal/geometry/Segment.hpp"
+#include "axom/primal/geometry/BoundingBox.hpp"
+#include "axom/primal/geometry/OrientedBoundingBox.hpp"
+
+#include "axom/primal/operators/squared_distance.hpp"
+
+#include "fmt/fmt.hpp"
```
gzagaris: Is this header needed here? It doesn't look like `fmt` is being used anywhere. Please, remove.

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -0,0 +1,335 @@
...
+ * The curve is approximated by the control points,
+ * parametrized from t=0 to t=1.
+ */
+
+template < typename T,int NDIMS >
+class BezierCurve
+{
+public:
+  using PointType = Point< T,NDIMS >;
+  using VectorType = Vector< T,NDIMS >;
```
gzagaris: It looks like `VectorType` is not used anywhere

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -0,0 +1,335 @@
...
+  using NumArrayType = NumericArray< T,NDIMS >;
+  using SegmentType = Segment< T, NDIMS >;
+  using CoordsVec = std::vector< PointType >;
```
gzagaris: I understand the convenience of STL, but, we should avoid it.

Could we just use `axom::Array`, or `numerics::Matrix` with dimensions `ndims x numPoints`, i.e., each column in the matrix is a control point (node), or just flat arrays?

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -0,0 +1,335 @@
...
+public:
+
+  /*!
+   * \brief Constructor for a Bezier Curve that reserves space for
+   *  the given order of the curve
+   *
+   * \param [in] order the order of the resulting Bezier curve
+   * \pre order is greater than or equal to -1.
+   */
+  explicit BezierCurve(int ord = -1)
```
gzagaris: Could the default be set to zero and simplify the code below?

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -0,0 +1,335 @@
...
+  {
+    SLIC_ASSERT(pts != nullptr);
+    SLIC_ASSERT(ord >= 0);
+
+    const int sz = utilities::max(0, ord+1);
+    m_controlPoints.resize(sz);
+
+    for ( int p = 0 ; p <= ord ; p++)
+    {
+      auto& pt = m_controlPoints[p];
```
gzagaris: This should just be `T` -- there is no strong need to use `auto` here.

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  friend inline bool operator!=(const BezierCurve< T, NDIMS>& lhs,
+                                const BezierCurve< T, NDIMS>& rhs)
+  {
+    return !(lhs == rhs);
+  }
+
+  /*! Returns a copy of the Bezier curve's control points */
+  CoordsVec getControlPoints() const
```
gzagaris: Likewise, I think using the `numerics::Matrix` here to store the control points would be better from the API standpoint and from a

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -0,0 +1,335 @@
...
+  {
+    return m_controlPoints;
+  }
+
+  /*! Returns an axis-aligned bounding box containing the Bezier curve */
+  BoundingBoxType boundingBox() const
```
gzagaris: For consistency, I would rename this method to ``getBoundingBox()``

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -0,0 +1,335 @@
...
+  {
+    return BoundingBoxType(m_controlPoints.data(), m_controlPoints.size());
+  }
+
+  /*! Returns an oriented bounding box containing the Bezier curve */
+  OrientedBoundingBoxType orientedBoundingBox() const
```
gzagaris: For consistency, I would rename this method to ``getOrientedBoundingBox()``

src/axom/primal/operators/detail/intersect_bezier_impl.hpp
```diff
@@ -0,0 +1,211 @@
...
+  // Implementation inspired by Section 5.1.9.1 of
+  // C. Ericson's Real-Time Collision Detection book
+
+  // Note: Uses exact floating point comparisons since the subdivision algorithm
+  // provides both sides of the line segments for interior curve points.
+
+  AXOM_STATIC_ASSERT( NDIMS==2);
+
+  // compute signed areas of endpoints of segment (c,d) w.r.t. segment (a,b)
+  auto area1 = twoDcross(a,b,c);
```
gzagaris: It is more clear to use `T` instead of `auto`

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -0,0 +1,335 @@
...
+ * \brief Represents a Bezier curve defined by an array of control points
+ * \tparam T the coordinate type, e.g., double, float, etc.
+ * \tparam NDIMS the number of dimensions
+ *
+ * The order of a Bezier curve with N+1 control points is N.
+ * The curve is approximated by the control points,
+ * parametrized from t=0 to t=1.
+ */
+
+template < typename T,int NDIMS >
```
gzagaris: Should probably add a static assert here for `T` and `NDIMS`

src/axom/primal/geometry/BezierCurve.hpp
kennyweiss: I believe it is used for derivatives in a later commit (not part of this PR).

src/axom/primal/geometry/BezierCurve.hpp
kennyweiss: Good suggestion @gzagaris.

Let's save stl changes for a future PR since @davidgunderman has another branch with further changes.

src/axom/primal/geometry/BezierCurve.hpp
kennyweiss: I don't think this is a good idea.

An order-zero Bezier curve has a single control point, i.e. it is a point.
If the user does not provide this point, then it will be filled with garbage data.

gzagaris: Do you think it's a good idea to create a Bezier curve that is order less than zero?

src/axom/primal/geometry/BezierCurve.hpp
gzagaris: If it is used for shape function derivatives, we should be using the ``numerics::Matrix`` class.

src/axom/primal/geometry/BezierCurve.hpp
kennyweiss: That could very well be true.  I think it depends on what you want to do with the returned control points.  

Let's hold off on this change for now until we have a better understanding of how the control points will be used.

src/axom/primal/geometry/BezierCurve.hpp
kennyweiss: Yes, if the user doesn't have data to fill it. 

It would be arbitrary and error prone to set this to zero, which would imply that there is a valid control point.

src/axom/primal/geometry/BezierCurve.hpp
kennyweiss: `auto` is the recommended return type for modern `C++`.  

E.g. see Scott Meyers' "Effective Modern C++" --> "Item 5: Prefer auto to explicit type declarations"

kennyweiss: Also, the return type would be ``PointType&`` since it is returning a point, rather than `T&`, a coordinate of a point.

gzagaris: Then, this should definitely be `PointType&` to avoid confusion. I believe our coding guidelines discourage the use of `auto` unless it makes code cleaner, e.g., when using iterators etc.

src/axom/primal/geometry/BezierCurve.hpp
kennyweiss: I think we did it this way since ``getBoundingBox()`` seems to imply that the bounding box is precomputed and that the function would simply be returning this, rather than having to compute the bounding box of the control points.

src/axom/primal/geometry/BezierCurve.hpp
gzagaris: This type of changes are best handled while the code is still small and doesn't bleed out. 

Moreover, since we have better alternatives now in Axom, which are also GPU friendly, we should ensure we don't add technical debt to the code with every new feature that is added. 

That said I am fine with doing this change in a separate PR, but, please create a Github Issue for this.

src/axom/primal/geometry/BezierCurve.hpp
gzagaris: Perhaps, `computeBoundingBox()` would be more clear in this case.

src/axom/primal/geometry/BezierCurve.hpp
gzagaris: OK, I see, so this serves as a sentinel that the Bezier curve is not properly initialized...

src/axom/primal/tests/primal_bezier_intersect.cpp
agcapps: Thanks!  Looks good.

src/axom/primal/operators/detail/intersect_bezier_impl.hpp
agcapps: I see that you documented the user-facing functions to point out that they handle non-tangent intersections of multiplicity 1.  That is OK with me.

PR #64, opened 2019-07-15
Quiet lots of warnings from compile flag

white238:
# Summary

- This PR is a bugfix
  Quiets a lot of warnings out of GNU 8.1+ due to our flag added to sparsehash

### DIFF
--- a/src/thirdparty/CMakeLists.txt
+++ b/src/thirdparty/CMakeLists.txt
@@ -125,14 +125,13 @@ if (AXOM_ENABLE_SPARSEHASH)
     target_include_directories(sparsehash SYSTEM INTERFACE
                 $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/thirdparty>)
 
-    # Disable warning introduced in gcc@8.1 related to how sparsehash casts memory
+    # Disable warning introduced in gcc@8.1+ related to how sparsehash casts memory
+    # The double guarding for compiler family helps when compiling libraries with GNU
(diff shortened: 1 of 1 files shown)


white238:
I am doing one more test to make sure this solves the problem completely instead of just one case.

src/thirdparty/CMakeLists.txt
```diff
@@ -125,14 +125,13 @@ if (AXOM_ENABLE_SPARSEHASH)
...
+    # then compiling the main code with another compiler (this is happens for example when
+    # running clang-query)
     if(C_COMPILER_FAMILY_IS_GNU AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "8.1")
-      if (ENABLE_FORTRAN)
         blt_add_target_compile_flags(TO sparsehash FLAGS
-            $<$<NOT:$<COMPILE_LANGUAGE:Fortran>>:-Wno-class-memaccess>)
-      else()
-        blt_add_target_compile_flags(TO sparsehash FLAGS -Wno-class-memaccess)
-      endif()
+            $<$<AND:$<CXX_COMPILER_ID:GNU>,$<COMPILE_LANGUAGE:CXX>>:-Wno-class-memaccess>)
```
kennyweiss: So we only want this as a `CXX` flag, for gnu 8.1 or above, right?

white238: Correct. Even C throws a warning about the flag.

white238:
This fixes both warnings.

Issue #45, opened 2019-05-23
external vs view into shared buffer?


cyrush:
@nselliott pointed out an issue with reading external buffers in the VisIt Blueprint plugin. 

The files were created with `src/axom/sidre/examples/sidre_createdatastore.cpp`
I found the issue with the VisIt plugin, however I am still a bit confused b/c I did not expect these fields to be written out as "external".

Here is the snippet that creates the views:

```
  // "x", "y", and "z" are three views into a shared Sidre buffer object that
  // holds 3 * nodecount doubles.  These views might describe the location of
  // each node in a 16 x 16 x 16 hexahedron mesh.  Each view is described by
  // number of elements, offset, and stride into that data.
  Buffer* buff = ds->createBuffer(sidre::DOUBLE_ID, 3*nodecount)->allocate();
  nodes->createView("x", buff)->apply(sidre::DOUBLE_ID, nodecount, 0, 3);
  nodes->createView("y", buff)->apply(sidre::DOUBLE_ID, nodecount, 1, 3);
  nodes->createView("z", buff)->apply(sidre::DOUBLE_ID, nodecount, 2, 3);
  // _first_example_nodes_end
```

(With the fix) in VisIt I am definitely reading them as external views from the HDF5 file:

```
loading mesh/coordsets/coords/values/x as sidre external view
relay:io::hdf5_read domain 0 : datagroup_0000000/sidre/external/mesh/coordsets/coords/values/x
```

Is this expected? I expected them to simply be a view into a known buffer. 

@kennyweiss @agcapps any thoughts?

cyrush:
I think the issue is actually here:

```
void setup_blueprint_coords(DataStore* ds, Group* coords)
{
  // _blueprint_restructure_coords_start
  // Set up the coordinates as Mesh Blueprint requires
  coords->createViewString("type", "explicit");
  // We use prior knowledge of the layout of the original datastore
  View* origv = ds->getRoot()->getView("nodes/xs");
  Group* conduitval = coords->createGroup("values");
  conduitval->createView("x", sidre::DOUBLE_ID,
                         origv->getNumElements(),
                         static_cast<double*>(origv->getArray()));
  origv = ds->getRoot()->getView("nodes/ys");
  conduitval->createView("y", sidre::DOUBLE_ID,
                         origv->getNumElements(),
                         static_cast<double*>(origv->getArray()));
  origv = ds->getRoot()->getView("nodes/zs");
[...]

PR #69, opened 2019-07-24
Feature/elliott/bp example

nselliott:
# Summary

- This PR is a modification of some example code.
- It does the following:
  - Modifies sidre_createdatastore.cpp by simplifying the process of setting up a blueprint
    for a simple mesh.
  - Makes some of the examples in this file use Views that see pre-existing Buffers as Buffers, while
    other examples have Views see the pre-existing data as described external pointers, thus providing
    test coverage for both use cases.
  - Addresses issue #45 
  - Documentation in sidre_conduit.rst uses code snippets from this file, so the documentation text has
    been edited to be consistent with these code changes.


nselliott:
@agcapps I've edited the documentation but left the figures untouched.  The key change is that there is only one DataStore, so the `tinymesh` subtree should now be a child of `ds.root` parallel to `nodes` and `fields`.

src/axom/sidre/docs/sphinx/sidre_conduit.rst
```diff
@@ -61,9 +60,9 @@ First build top-level groups required by the Blueprint.
    :end-before: _blueprint_restructure_toplevel_end
    :language: C++
 
-Add the node coordinates.  The Groups in ``cds`` will use the Buffers within ``ds``
-as external pointers that have a description: while Sidre and Conduit will use the
-array type and shape information, ``cds`` will not deallocate the memory.
+Add the node coordinates.  The Views in ``tinymesh`` will point to the Buffers within ``nodes``
```
agcapps: Maybe rephrase to something like "The Views under ``tinymesh`` will point to the same Buffers as the Views under ``nodes`` and ``fields`` so that ``tinymesh`` can use the data ...".  The intent being to emphasize that the Buffers are shared, in the same Dataset, but not contained or owned by any Groups.

nselliott:
The figures look good, thanks @agcapps

PR #86, opened 2019-08-30
Update Conduit to master

white238:
Updates Conduit to the master branch for all specs and platforms.  This is required for any codes that rely on Sidre and use Conduit's find_package logic. Basically hdf5 and silo were backwards on the link line.

Host-config's incoming next week....

- [ Check-in new built host-configs]

### DIFF
--- a/src/cmake/thirdparty/SetupAxomThirdParty.cmake
+++ b/src/cmake/thirdparty/SetupAxomThirdParty.cmake
@@ -11,40 +11,38 @@
 # UMPIRE
 ################################
 if (UMPIRE_DIR)
-  include(cmake/thirdparty/FindUmpire.cmake)
-  blt_register_library( NAME umpire
-                        INCLUDES ${UMPIRE_INCLUDE_DIRS}
-                        TREAT_INCLUDES_AS_SYSTEM ON
-                        LIBRARIES umpire )
+    include(cmake/thirdparty/FindUmpire.cmake)
+    blt_register_library( NAME      umpire
+                          INCLUDES  ${UMPIRE_INCLUDE_DIRS}
+                          LIBRARIES umpire
+                          TREAT_INCLUDES_AS_SYSTEM ON)
 else()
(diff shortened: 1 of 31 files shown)


src/cmake/thirdparty/SetupAxomThirdParty.cmake
```diff
@@ -11,40 +11,38 @@
...
     include(cmake/thirdparty/FindConduit.cmake)
-    blt_register_library( NAME conduit
-                          INCLUDES ${CONDUIT_INCLUDE_DIRS}
-                          LIBRARIES conduit
-                          TREAT_INCLUDES_AS_SYSTEM ON)
-    blt_register_library( NAME conduit_relay
-                          INCLUDES ${CONDUIT_INCLUDE_DIRS}
-                          LIBRARIES conduit_relay
+    blt_register_library( NAME      conduit
+                          INCLUDES  ${CONDUIT_INCLUDE_DIRS}
```
cyrush: you shouldn't have to set the includes, but it also wont hurt.

white238:
> Thanks @white238
> 
> The code still works with the old conduit and host-configs, right?

No.  Checked in new host-configs.

src/cmake/thirdparty/SetupAxomThirdParty.cmake
white238: Apparently the registered target "conduit" was conflicting with your under the hood exported target "conduit" and caused info to be lost.  I removed the registered target and switched completely to your convience target "conduit::conduit".  This solved the fact that it wasn't inheriting things properly.  This does lose the system header flag though.

cyrush: ahh ok, yes I don't see how to easily propgate treat as system.

If folks see annoying warnings from Conduit --  I'l just have to fix them. I know they exist on Windows but I have tried to keep them at bay on other platforms.

PR #87, opened 2019-09-03
Change error to warning when trying to import conduit lists into Sidre

nselliott:
# Summary

- This PR is a small change
  - Changes an error to a warning at the request of @kennyweiss

In the previous pull request #82, an attempt to import a conduit list resulted in a SLIC error, since Sidre Groups do not have an analogous concept.  @kennyweiss requested that this be changed to a warning, but I merged the pull request without addressing the comment.

Here the error is changed to a warning, and the import will still import all other parts of the condiut tree that it can import, while leaving out the lists and anything descending from lists.  A boolean return value is added, true if the full import is successful, false if there was a list that had to be left out.

### DIFF
--- a/src/axom/sidre/core/Group.cpp
+++ b/src/axom/sidre/core/Group.cpp
@@ -2046,8 +2046,9 @@ void Group::importFrom(conduit::Node& node,
  *************************************************************************
  */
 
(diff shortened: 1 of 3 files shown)


kennyweiss:
Thanks @nselliott!

PR #95, opened 2019-09-07
Slic Updates

gzagaris:
# Summary

This PR improves the internal organization of Slic. Specifically, it moves all Slic Macros in a seperate header and also adds unit tests for all slic macros.


gzagaris:
@rhornung67 , at your convenience, could you take a quick look at this. These are basic cleanup changes that I did while putting together the SLIC documentation.

PR #112, opened 2019-10-02
Feature/gunderman/bezier polygon

davidgunderman:
# Summary

- This PR adds intersection capabilities for polygons with curved edges

This PR adds a ``CurvedPolygon`` class and implements an algorithm for computing the intersections of curved polygons and their areas and centroids.  It leaves out some special cases, including tangent polygon intersection and intersections of polygons whose component curves overlap for the following reasons:

- Typically tangent polygon intersection won't come up often in practice.
- Intersecting polygons whose component curves overlap could come up, but we haven't gotten to this yet.

It also leaves out a full mesh-mesh intersection example, because I'd like to add arbitrary polynomial field transfer (not just areas and centroids) functionality before pushing a full example.


kennyweiss:
@davidgunderman -- I pushed some changes/bugfixes to this branch. 
Here is a brief summary:
* I pushed an mfem nurbs mesh to the ``axom_data`` submodule. 
  The quest example no longer depends on a missing mesh.
  (This should be converted to a command line argument in the future)
* The code now builds and passes all tests on MSVC.
* ``BezierCurve::dt()`` now returns a ``primal::Vector``, rather than a ``primal::Point``
* Some additional minor improvements/corrections in the API
* Looks like there is still a compilation problem with macos 
  (related to sorting? -- https://dev.azure.com/axom/axom/_build/results?buildId=604 )

src/axom/primal/operators/detail/intersect_curved_poly_impl.hpp
```diff
@@ -0,0 +1,333 @@
...
+ */
+template < typename T, int NDIMS>
+bool intersect_polygon(  const CurvedPolygon< T, NDIMS>& p1,
+                         const CurvedPolygon< T, NDIMS>& p2,
+                         std::vector<CurvedPolygon< T, NDIMS>>& pnew,
+                         double sq_tol
+                       )
+{
+  // Object to store intersections
+  std::vector<std::vector<IntersectionInfo<T>>> E1IntData(p1.numEdges());
```
kennyweiss: @davidgunderman -- I think the macos problem is getting triggered here. 
For some reason, xcode is not using ``IntersectionInfo<T>::operator<()`` to sort.

```cpp
   std::vector<std::vector<IntersectionInfo<T>>> E1IntData(p1.numEdges()); 
   std::vector<std::vector<IntersectionInfo<T>>> E2IntData(p2.numEdges()); 
   ...
   std::sort( E1IntData[i].begin(), E1IntData[i].end() );
   std::sort( E2IntData[i].begin(), E2IntData[i].end() ); 
```

davidgunderman: I'm really not sure why this would be happening. I'll spend some time today looking around.

davidgunderman: @kennyweiss I think this is fixed now. It looks like custom operator< definition needed to be const and/or needed to take a reference as a parameter

davidgunderman: @kennyweiss I think this is fixed now.

kennyweiss:
Superseded by #590, which brings this PR up to date with our develop branch and touches up the code a bit

PR #119, opened 2019-11-04
Consolidate Execution Policies

### LINKED ISSUE #56: Consolidate mint::policy and spin::execution_space in axom::core
The ``spin::execution_space`` defines a set of high-level execution policies, e.g., ``spin::SEQ_EXEC``, ``spin::CUDA_EXEC``, etc. and a traits class that binds the high-level  execution policies to the underlying ``RAJA`` execution policies and to a corresponding ``Umpire`` memory space.

A similar strategy is employed by ``mint::policy`` for Mint's mesh-aware execution model.

Although there is overlap between ``mint::policy`` and ``spin::execution_space`` the two implementations also complement each other. For example, ``mint::policy_traits`` has mappings to corresponding ``RAJA`` kernel and synchronization policies that are not implemented by ``spin::execution_space``. Likewise, the ``spin::execution_space`` has mappings to  ``RAJA`` atomic policies and default memory space which are not implemented by``mint::policy_traits``.

As more Axom components, e.g., ``Quest``, begin to support fine-grain parallelism on heterogeneous architectures, using ``RAJA``'s parallel execution model and ``Umpire`` for memory management, a similar traits class and policies will most likely be needed.

This issue proposes to unify ``spin::execution_space`` and ``mint::policy`` in ``axom::core`` so that the net functionality and future extensions are broadly available to all Axom components.

gzagaris:
# Summary

Consolidates the execution policies in axom/core. See #56 for details.

Resolves #56


src/axom/core/execution/internal/cuda_exec.hpp
```diff
@@ -0,0 +1,140 @@
...
+ * \tparam BLOCK_SIZE the number of CUDA threads to launch
+ *
+ */
+template < int BLOCK_SIZE >
+struct execution_space< CUDA_EXEC< BLOCK_SIZE, SYNCHRONOUS > >
+{
+  using loop_policy   = RAJA::cuda_exec< BLOCK_SIZE >;
+
+  /* *INDENT-OFF* */
+  using loop2d_policy =
```
rhornung67: Are these policy definitions too constraining, potentially? They might be suboptimal depending on the kernel.

src/axom/core/execution/internal/omp_exec.hpp
```diff
@@ -0,0 +1,73 @@
...
+/*!
+ * \brief execution_space traits specialization for OMP_EXEC
+ */
+template < >
+struct execution_space< OMP_EXEC >
+{
+  using loop_policy = RAJA::omp_parallel_for_exec;
+
+  /* *INDENT-OFF* */
+  using loop2d_policy = RAJA::KernelPolicy<
```
rhornung67: For most kernels, using OpenMP collapse on all loop levels performs poorer than applying OpenMP to a subset.

src/axom/core/tests/core_execution_space.cpp
```diff
@@ -0,0 +1,309 @@
...
+//------------------------------------------------------------------------------
+#ifdef AXOM_USE_RAJA
+TEST( core_execution_space, check_seq_exec )
+{
+  check_valid< axom::SEQ_EXEC >( );
+
+  constexpr bool IS_ASYNC = false;
+
+  /* *INDENT-OFF* */
+  using raja_loop2d_policy =
```
rhornung67: Why use RAJA policies here and not the axom ones you defined?

gzagaris: This code tests the mapping of the `execution_space` traits class to the corresponding RAJA policies. So, the expected RAJA policies are hard-coded in this test. Makes sense?

src/axom/core/execution/internal/cuda_exec.hpp
gzagaris: The loop 2d/3d policies are intended for generic i-j-k loops. Is there something better and generic that we should do in this case instead?

These type of policies should not be used for things like, loop over the DOFs of each element and process each DOF independently.  Indeed, that would be suboptimal in this case. My expectation is that for specific use cases, different policies will be defined at the application level. 

That said, exposing these loop policies here, may be misleading application developers to use them regardless.

Perhaps, we shouldn't define these policies here at all. We could leave it to the application, perhaps, by inheriting and extending the `execution_space` traits class here. Thoughts?

rhornung67: That last part seems better, I think.

gzagaris: Sounds good -- I'll remove the predefined kernel policies from here.

src/axom/mint/execution/internal/structured_exec.hpp
```diff
@@ -0,0 +1,157 @@
...
+  /* *INDENT-OFF* */
+   using loop2d_policy = RAJA::KernelPolicy<
+       RAJA::statement::Collapse< RAJA::omp_parallel_collapse_exec,
+                                  RAJA::ArgList< 1,0 >,
+                                  RAJA::statement::Lambda< 0 > > >;
+
+   using loop3d_policy = RAJA::KernelPolicy<
+       RAJA::statement::Collapse< RAJA::omp_parallel_collapse_exec,
+                                  RAJA::ArgList< 2,1,0 >,
+                                  RAJA::statement::Lambda< 0 > > >;
```
gzagaris: Is there something better, yet general (i.e., without knowledge of the mesh dimensions), that we can do here?

gzagaris: For example, we could predefine some Tiling policies, and at runtime, we could choose if we can use that or fall-back to the collapse?

gzagaris: I changed this to parallelize with OpenMP in the outer-loop and used tiling for CUDA.

src/axom/mint/docs/sphinx/sections/execution_model.rst
```diff
@@ -82,32 +82,31 @@ Execution Policy
...
+| **SEQ_EXEC**                     |    None.                       | Sequential execution on         |
+|                                  |                                | the CPU.                        |
++----------------------------------+--------------------------------+---------------------------------+
+| **OMP_EXEC**                     |  `RAJA`_ + OpenMP              | Parallel execution on           |
+|                                  |                                | the CPU using OpenMP.           |
++----------------------------------+--------------------------------+---------------------------------+
+| **CUDA_EXEC< NBLOCKS >**         |  `RAJA`_ + CUDA +              | Parallel execution on           |
+|                                  |  `Umpire`_ (memory management) | CUDA-enabled GPUs.              |
++----------------------------------+--------------------------------+---------------------------------+
+| **CUDA_EXEC< NBLOCKS, ASYNC >**  |  `RAJA`_ + CUDA                | Asynchronous parallel           |
```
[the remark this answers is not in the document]
gzagaris: Yes -- good catch! I pushed a fix.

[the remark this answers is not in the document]
rhornung67: @gzagaris if your CUDA_EXEC< NBLOCKS > dipatches to something using a RAJA::cuda_exec policy the template argument is wrong. It is # threads / block, not # blocks. Also, RAJA provides a reasonable default of 256 if not specified by a user. This should be noted as well.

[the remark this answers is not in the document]
gzagaris: Thanks -- I renamed the template parameter to `BLOCKSIZE`, which is what we actually call it in code.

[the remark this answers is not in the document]
rhornung67: 👍

PR #122, opened 2019-11-21
Various Build Issues

white238:
* Update BLT
* Desperately attempt to sort out a usable solution to the MPI Fortran wrapper on blue_os
* Clean up unused variables in tests...
* Unify a bunch of blueos/blueos_p9 host-config generation logic
* Add CUDA_SEPARABLE_COMPILATION to cuda builds

PR #123, opened 2019-11-21
Sets CUDA_SEPARABLE_COMPILATION flag in blueos9 host-config

kennyweiss:
# Summary 

It seems that there was a conflict with setting the ``CUDA_SEPARABLE_COMPILATION``
flag on one of axom's libraries (primal) and compiling axom's libraries as object libraries.

This fixes the weird compiler bug that @bmhan12 reported:
```
ptxas /var/tmp/han12/tmpxft_00002929_00000000-5_intersect_impl.ptx, line 435; error   : Illegal operand type to instruction 'ld'
ptxas /var/tmp/han12/tmpxft_00002929_00000000-5_intersect_impl.ptx, line 435; error   : Unknown symbol '_param_0'
ptxas /var/tmp/han12/tmpxft_00002929_00000000-5_intersect_impl.ptx, line 531; error   : Illegal operand type to instruction 'ld'
ptxas /var/tmp/han12/tmpxft_00002929_00000000-5_intersect_impl.ptx, line 531; error   : Unknown symbol '_param_0'
ptxas fatal   : Ptx assembly aborted due to errors
```
Specifically, I was able to compile your branch and pass all tests using the ``rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.0_nvcc_xlf.cmake `` host-config.

Note: This PR is into @bmhan12's feature branch, not into develop.
I am adding others to this PR in case they see a better solution and/or object to setting this flag in our main GPU host-config for blueos9.

### DIFF
--- a/host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.0_nvcc_xlf.cmake
+++ b/host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.0_nvcc_xlf.cmake
@@ -110,6 +110,7 @@ set(AXOM_CUDA_ARCH "sm_70" CACHE PATH "")
 
 set(CMAKE_CUDA_FLAGS "-restrict -arch ${AXOM_CUDA_ARCH} -std=c++11 --expt-extended-lambda -G" CACHE PATH "")
 
+set(CUDA_SEPARABLE_COMPILATION ON CACHE BOOL "" )
 set(CMAKE_CUDA_HOST_COMPILER "${MPI_CXX_COMPILER}" CACHE PATH "")
 
 # nvcc does not like gtest's 'pthreads' flag
--- a/src/axom/primal/CMakeLists.txt
+++ b/src/axom/primal/CMakeLists.txt
@@ -76,10 +76,6 @@ blt_add_library(
     OBJECT                TRUE
     )
 
-if (ENABLE_CUDA)
-  set_target_properties(primal PROPERTIES CUDA_SEPARABLE_COMPILATION ON)
-endif()
-
 axom_write_unified_header(NAME    primal
                           HEADERS ${primal_headers})
 
--- a/src/axom/spin/examples/CMakeLists.txt
+++ b/src/axom/spin/examples/CMakeLists.txt
@@ -33,6 +33,3 @@ blt_add_executable(
     FOLDER      axom/spin/examples
     )
 
-if (ENABLE_CUDA)
-  set_target_properties(naive_intersect_raja_ex PROPERTIES CUDA_SEPARABLE_COMPILATION ON)
-endif()


kennyweiss:
@bmhan12 --  Regarding the failing CI -- it looks like it is due to not guarding RAJA usage in your ``naive_intersect_raja``  example.

```
cd /home/axom/azure-linux_gcc6-build/axom/spin/examples && /usr/bin/g++   -I/home/axom/azure-linux_gcc6-build/include -I/home/axom/src -isystem /home/axom/src/thirdparty  -Wall -Wextra  -fPIE   -std=c++11 -o CMakeFiles/naive_intersect_raja_ex.dir/naive_intersect_raja.cpp.o -c /home/axom/src/axom/spin/examples/naive_intersect_raja.cpp
/home/axom/src/axom/spin/examples/naive_intersect_raja.cpp:28:25: fatal error: RAJA/RAJA.hpp: No such file or directory
 #include "RAJA/RAJA.hpp"
                         ^
```

I realize it is a work-in-progress, but you should be able to get around that by disabling that example when RAJA is not available.
Specifically, this block:
https://github.com/LLNL/axom/blob/42b772a9675c5f9a2640932c7dea01bfe679af61/src/axom/spin/examples/CMakeLists.txt#L28-L38

should be guarded by
```cmake
if(RAJA_FOUND)
   ...
endif()
```

bmhan12:
@kennyweiss  Got it.  I'll merge this PR now, and add the guard in.

host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.0_nvcc_xlf.cmake
```diff
@@ -110,6 +110,7 @@ set(AXOM_CUDA_ARCH "sm_70" CACHE PATH "")
 
 set(CMAKE_CUDA_FLAGS "-restrict -arch ${AXOM_CUDA_ARCH} -std=c++11 --expt-extended-lambda -G" CACHE PATH "")
 
+set(CUDA_SEPARABLE_COMPILATION ON CACHE BOOL "" )
```
white238: This is going to be lost when we regenerate the host-configs..... It needs to be added to scripts/uberenv/packages/uberenv-axom/package.py

kennyweiss: Thanks @white238 -- I forgot to point that out.

Would you mind adding it to your PR -- https://github.com/LLNL/axom/pull/122 ?
Or should we do this separately, e.g. in @bmhan12's branch?

It should go in here:
https://github.com/LLNL/axom/blob/134082dfad44a90b1a90ecb5b4e3e9be0bc1bc65/scripts/uberenv/packages/uberenv-axom/package.py#L427-L449

PR #151, opened 2020-01-22
Update Umpire to v2.0.0

gzagaris:
# Summary

This PR updates Umpire to v2.0.0. In addition, `axom::reallocate()` is updated since umpire-v2.0.0 natively handles zero byte reallocations consistently.  The previously implemented workaround remains in place for backwards compatibility.

### DIFF
--- a/src/axom/core/memory_management.hpp
+++ b/src/axom/core/memory_management.hpp
@@ -198,10 +198,14 @@ inline T* reallocate( T* pointer, std::size_t n ) noexcept
 {
   const std::size_t numbytes = n * sizeof( T );
 
-#ifdef AXOM_USE_UMPIRE
+#if defined(AXOM_USE_UMPIRE) && !defined(UMPIRE_VERSION_MAJOR)
 
   // Workaround for bug in Umpire's handling on reallocate(0)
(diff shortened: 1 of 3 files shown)


src/axom/core/memory_management.hpp
```diff
@@ -198,7 +198,7 @@ inline T* reallocate( T* pointer, std::size_t n ) noexcept
 {
   const std::size_t numbytes = n * sizeof( T );
 
-#ifdef AXOM_USE_UMPIRE
+#if defined(AXOM_USE_UMPIRE) && !defined(UMPIRE_VERSION_MAJOR)
```
kennyweiss: Please add a comment here that ``UMPIRE_VERSION_MAJOR`` was added in Umpire v2.0.0, so if it is not defined, then we're assuming that we need the workaround.

gzagaris: Thanks, I added a comment to clarify that.

gzagaris:
> Are you updating the host-configs as well?

I'll update the host-configs in the release candidate branch, which will include the RAJA TPL update, as well.

PR #155, opened 2020-01-27
Support for CI Testing using Azure Pipelines/Docker

keithhealy:
# Summary

- This PR is a feature
- It adds support for CI testing of Axom using Azure Pipelines/Docker.
- Currently enables testing on Linux, Windows, and Mac.
- Linux testing uses following compilers: Clang 4,5,6 and GCC 6,7,8
- Dockerfiles located in axom/scripts/docker were used to generate docker images stored on Docker Hub.  See README.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,38 +13,54 @@ strategy:
   matrix:
     linux_gcc6:
       VM_ImageName: 'ubuntu-16.04'
-      Compiler_ImageName: 'axom/compilers:gcc-6'
-      CMAKE_EXTRA_FLAGS: '-DAXOM_ENABLE_SIDRE=Off'
+      Compiler_ImageName: 'axom/tpls:gcc-6'
+      CMAKE_EXTRA_FLAGS: ''
       COMPILER: 'g++'
       TEST_TARGET: 'linux_gcc6'
-    linux_gcc7:
-      VM_ImageName: 'ubuntu-16.04'
-      Compiler_ImageName: 'axom/compilers:gcc-7'
-      CMAKE_EXTRA_FLAGS: '-DAXOM_ENABLE_SIDRE=Off'
-      COMPILER: 'g++'
-      TEST_TARGET: 'linux_gcc7'
+      HOST_CONFIG: 'docker-linux-ubuntu16.04-x86_64-gcc@6.1.0'
     linux_gcc8:
       VM_ImageName: 'ubuntu-16.04'
-      Compiler_ImageName: 'axom/compilers:gcc-8'
-      CMAKE_EXTRA_FLAGS: '-DAXOM_ENABLE_SIDRE=Off'
+      Compiler_ImageName: 'axom/tpls:gcc-8'
+      CMAKE_EXTRA_FLAGS: ''
       COMPILER: 'g++'
       TEST_TARGET: 'linux_gcc8'
+      HOST_CONFIG: 'docker-linux-ubuntu16.04-x86_64-gcc@8.1.0'
+    linux_gcc7:
+      VM_ImageName: 'ubuntu-16.04'
+      Compiler_ImageName: 'axom/tpls:gcc-7'
+      CMAKE_EXTRA_FLAGS: ''
+      COMPILER: 'g++'
+      TEST_TARGET: 'linux_gcc7'
+      HOST_CONFIG: 'docker-linux-ubuntu16.04-x86_64-gcc@7.3.0'
     linux_clang4:
-      VM_ImageName: 'ubuntu-16.04'
-      Compiler_ImageName: 'axom/compilers:clang-4'
-      CMAKE_EXTRA_FLAGS: '-DAXOM_ENABLE_SIDRE=Off'
-      COMPILER: 'clang++'
-      TEST_TARGET: 'linux_clang4'
+       VM_ImageName: 'ubuntu-16.04'
+       Compiler_ImageName: 'axom/tpls:clang-4'
+       CMAKE_EXTRA_FLAGS: ''
+       COMPILER: 'clang++'
+       TEST_TARGET: 'linux_clang4'
+       HOST_CONFIG: 'docker-linux-ubuntu16.04-x86_64-clang@4.0.0'
+    linux_clang5:
+       VM_ImageName: 'ubuntu-16.04'
+       Compiler_ImageName: 'axom/tpls:clang-5'
+       CMAKE_EXTRA_FLAGS: ''
+       COMPILER: 'clang++'
+       TEST_TARGET: 'linux_clang5'
+       HOST_CONFIG: 'docker-linux-ubuntu16.04-x86_64-clang@5.0.0'
     linux_clang6:
-      VM_ImageName: 'ubuntu-16.04'
-      Compiler_ImageName: 'axom/compilers:clang-6'
-      CMAKE_EXTRA_FLAGS: '-DAXOM_ENABLE_SIDRE=Off'
-      COMPILER: 'clang++'
-      TEST_TARGET: 'linux_clang6'
+       VM_ImageName: 'ubuntu-16.04'
+       Compiler_ImageName: 'axom/tpls:clang-6'
+       CMAKE_EXTRA_FLAGS: ''
+       COMPILER: 'clang++'
+       TEST_TARGET: 'linux_clang6'
+       HOST_CONFIG: 'docker-linux-ubuntu16.04-x86_64-clang@6.0.0'
     osx_gcc:
       VM_ImageName: 'macos-10.13'
       CMAKE_EXTRA_FLAGS: '-DAXOM_ENABLE_SIDRE=Off'
       TEST_TARGET: 'osx_gcc'
+    windows:
+      VM_ImageName: 'windows-2019'
+      CMAKE_EXTRA_FLAGS: '-DAXOM_ENABLE_SIDRE=Off'
+      TEST_TARGET: 'win_vs'
 
 pool:
   vmImage: $(VM_ImageName)
@@ -57,7 +73,13 @@ steps:
   inputs:
     workingDir: 'build'
     cmakeArgs: '$(CMAKE_EXTRA_FLAGS) ../src'
-  condition: eq( variables['Agent.OS'], 'Darwin')
+  condition: or( eq( variables['Agent.OS'], 'Windows_NT'), eq( variables['Agent.OS'], 'Darwin'))
+- task: VSBuild@1
+  inputs:
+    solution: 'build/*.sln'
+    vsVersion: 'latest'
+  condition: eq( variables['Agent.OS'], 'Windows_NT')
+  displayName: 'Visual Studio Build'
 - script: |
     cd build
     make
@@ -69,8 +91,10 @@ steps:
   displayName: '$(TEST_TARGET) Test'
   condition: eq( variables['Agent.OS'], 'Darwin')
 - script:  |
-    docker run --rm --user='root' -v `pwd`:/home/axom $(Compiler_ImageName) chown -R axom /home/axom
-    docker run --rm  -v `pwd`:/home/axom -e TEST_TARGET -e COMPILER -e DO_BUILD -e DO_TEST -e CMAKE_EXTRA_FLAGS $(Compiler_ImageName) ./scripts/azure-pipelines/linux-build_and_test.sh
+    echo " -e $TEST_TARGET -e $COMPILER -e $DO_BUILD -e $DO_TEST -e $CMAKE_EXTRA_FLAGS $(Compiler_ImageName) ./scripts/azure-pipelines/linux-build_and_test.sh"
+    docker run --rm --user='root' -v `pwd`:/home/axom/axom $(Compiler_ImageName) chown -R axom /home/axom
+    docker run --rm -v `pwd`:/home/axom/axom -e TEST_TARGET -e COMPILER -e DO_BUILD -e DO_TEST -e HOST_CONFIG -e CMAKE_EXTRA_FLAGS $(Compiler_ImageName) ./axom/scripts/azure-pipelines/linux-build_and_test.sh
+
   condition: eq( variables['Agent.OS'], 'Linux')
   displayName: '$(TEST_TARGET) Build & Test'
 - task: PublishTestResults@2
--- a/host-configs/docker/docker-linux-ubuntu16.04-x86_64-gcc@8.1.0.cmake
+++ b/host-configs/docker/docker-linux-ubuntu16.04-x86_64-gcc@8.1.0.cmake
@@ -1,0 +1,94 @@
+##################################
+# !!!! This is a generated file, edit at own risk !!!!
+##################################
+
+# Copyright (c) 2017-2019, Lawrence Livermore National Security, LLC and
+# other Axom Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (BSD-3-Clause)
+##################################
+
+##################################
+
+# SYS_TYPE: linux-ubuntu16.04-x86_64
+# Compiler Spec: gcc@8.1.0
+##################################
+
+# CMake executable path: /home/axom/axom_tpls/gcc-8.1.0/cmake-3.9.6/bin/cmake
+
+##############
+# Compilers
+##############
+
+# C compiler used by spack
+set(CMAKE_C_COMPILER "/usr/bin/gcc" CACHE PATH "")
+
+# C++ compiler used by spack
+set(CMAKE_CXX_COMPILER "/usr/bin/g++" CACHE PATH "")
+
+# Fortran compiler used by spack
+set(ENABLE_FORTRAN ON CACHE BOOL "")
+
+set(CMAKE_Fortran_COMPILER "/usr/bin/gfortran" CACHE PATH "")
+
+##############
+# TPLs
+##############
+
+# Root directory for generated TPLs
+set(TPL_ROOT "/home/axom/axom_tpls/gcc-8.1.0" CACHE PATH "")
+
+# conduit from uberenv
+set(CONDUIT_DIR "${TPL_ROOT}/conduit-master" CACHE PATH "")
+
+# mfem from uberenv
+set(MFEM_DIR "${TPL_ROOT}/mfem-4.0" CACHE PATH "")
+
+# hdf5 from uberenv
+set(HDF5_DIR "${TPL_ROOT}/hdf5-1.8.19" CACHE PATH "")
+
+# scr not built by uberenv
+
+# raja from uberenv
+set(RAJA_DIR "/usr/local/share/raja/cmake" CACHE PATH "")
+
+# umpire from uberenv
+set(UMPIRE_DIR "${TPL_ROOT}/umpire-1.0.0/share/umpire/cmake" CACHE PATH "")
+
+# python not built by uberenv
+
+set(ENABLE_DOCS OFF CACHE BOOL "")
+
+# shroud not built by uberenv
+
+# uncrustify not built by uberenv
+
+# lcov and genhtml not built by uberenv
+
+# cppcheck not built by uberenv
+
+##############
+# MPI
+##############
+
+set(ENABLE_MPI ON CACHE BOOL "")
+
+set(MPI_C_COMPILER "/home/axom/axom_tpls/gcc-8.1.0/mpich-3.2.1/bin/mpicc" CACHE PATH "")
+
+set(MPI_CXX_COMPILER "/home/axom/axom_tpls/gcc-8.1.0/mpich-3.2.1/bin/mpic++" CACHE PATH "")
+
+set(MPI_Fortran_COMPILER "/home/axom/axom_tpls/gcc-8.1.0/mpich-3.2.1/bin/mpif90" CACHE PATH "")
+
+set(MPIEXEC "/home/axom/axom_tpls/gcc-8.1.0/mpich-3.2.1/bin/mpiexec" CACHE PATH "")
+
+set(MPIEXEC_NUMPROC_FLAG "-n" CACHE PATH "")
+
+##############
+# Other machine specifics
+##############
+
+set(ENABLE_GTEST_DEATH_TESTS ON CACHE BOOL "")
+
+set(ENABLE_OPENMP ON CACHE BOOL "")
+
+
--- a/scripts/azure-pipelines/linux-build_and_test.sh
+++ b/scripts/azure-pipelines/linux-build_and_test.sh
@@ -6,7 +6,8 @@
 # SPDX-License-Identifier: (BSD-3-Clause)
 ##############################################################################
 
-env
+set -x
+
 function or_die () {
     "$@"
     local status=$?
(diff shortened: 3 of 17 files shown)


azure-pipelines.yml
```diff
@@ -13,38 +13,54 @@ strategy:
   matrix:
     linux_gcc6:
       VM_ImageName: 'ubuntu-16.04'
-      Compiler_ImageName: 'axom/compilers:gcc-6'
-      CMAKE_EXTRA_FLAGS: '-DAXOM_ENABLE_SIDRE=Off'
+      Compiler_ImageName: 'axom/tpls:gcc-6'
+      CMAKE_EXTRA_FLAGS: '-DAXOM_ENABLE_SIDRE=ON -DBUILD_SHARED_LIBS_ON'
```
gzagaris: You don't have to explicitly set `AXOM_ENABLE_SIDRE=ON` here. It's `ON" by default. Also, there is a typo in `-DBUILD_SHARED_LIBS_ON`. I think the intent probably was `-DBUILD_SHARED_LIBS=ON`, but we shouldn't need this.

keithhealy:
I have removed these flags.

scripts/docker/README
```diff
@@ -0,0 +1,37 @@
+Dockerfiles are used to create Docker images.  Docker images are the 'recipes' used to build virtual environments called containers.
+These dockerfiles all use a base Linux image that has a particular compiler installed.  These base images were created by David
+Beckinsale and are located on Docker Hub at this location:
+https://hub.docker.com/r/axom/compilers/tags
+
+The base images are accessed in these dockerfiles by the use of the 'FROM' command:
+FROM axom/compilers:<compiler>
+
+The remaining dockerfile commands attempt to build the Axom third party libraries using the specified compiler.
+The following commands update the compiler and add Fortran compiler support:
```
kennyweiss: @davidbeckingsale -- should we fold these commands into the compiler docker images that way we do not need to rerun them each time?

(or am I reading this wrong?)

scripts/docker/README
```diff
@@ -0,0 +1,37 @@
...
+RUN sudo apt-get update -y
+RUN sudo apt-get install curl -fy
+RUN sudo apt-get -qq install -y --no-install-recommends gfortran-7 && sudo update-alternatives --install /usr/bin/gfortran gfortran /usr/bin/gfortran-7 100
+#(For Clang Builds - Need to restore stdc++ lib)
+RUN sudo apt-get install libstdc++-7-dev
+
+Then the Axom and RAJA source is retrieved and RAJA is built and installed: (WORKDIR functions as a 'cd' command)
+RUN git clone --recursive https://github.com/LLNL/axom.git
+RUN git clone --recursive https://github.com/LLNL/RAJA.git
```
kennyweiss: I think we should be able to build ``RAJA`` directly from uberenv.  

Let's take care of that in a separate PR, after this is merged.

scripts/docker/README
```diff
@@ -0,0 +1,37 @@
+Dockerfiles are used to create Docker images.  Docker images are the 'recipes' used to build virtual environments called containers.
```
kennyweiss: For readability on github, it might be better to convert this to a markdown file.
(we can take care of this separately).

This is what it looks like now on github: https://github.com/keithhealy/axom/tree/develop/scripts/docker

scripts/azure-pipelines/linux-build_and_test.sh
```diff
@@ -16,21 +18,40 @@ function or_die () {
...
+git submodule init 
+git submodule update 
+
+echo HOST_CONFIG
+echo $HOST_CONFIG
+
+./config-build.py -hc /home/axom/axom/host-configs/docker/${HOST_CONFIG}.cmake -DENABLE_GTEST_DEATH_TESTS=ON
+
+or_die cd build-$HOST_CONFIG-debug
+or_die make -j 3
```
kennyweiss: Do you know how many threads we can use on azure pipelines?

It looks like we're building with 3 threads and runing tests with 8.

scripts/azure-pipelines/linux-build_and_test.sh
```diff
@@ -16,21 +18,40 @@ function or_die () {
...
-    fi
-fi
+or_die cd axom
+git submodule init 
+git submodule update 
+
+echo HOST_CONFIG
+echo $HOST_CONFIG
+
+./config-build.py -hc /home/axom/axom/host-configs/docker/${HOST_CONFIG}.cmake -DENABLE_GTEST_DEATH_TESTS=ON
```
kennyweiss: > -DENABLE_GTEST_DEATH_TESTS=ON

Please remove this cmake option -- it is set in the individual ``host-configs``

scripts/docker/README
```diff
@@ -0,0 +1,37 @@
...
+RUN mkdir build
+WORKDIR "/home/axom/RAJA/build"
+RUN cmake ../. -DENABLE_TESTS=OFF
+RUN make 
+RUN sudo make install
+
+Finally, the uberenv script is invoked to build the libraries and store them in the axom_tpls directory.  (The host config files
+produced during this process had been added to the Axom repo under "axom/host-configs/docker")
+WORKDIR "/home/axom/axom"
+RUN ./scripts/uberenv/uberenv.py --spack-config-dir=./scripts/uberenv/spack_configs/docker --spec=%clang@4.0.0^mpich@3.0.4 --prefix=/home/axom/axom_tpls -k
```
kennyweiss: @keithhealy -- can you please add a line about requiring ``mpich`` for docker.  This will document the trouble you had with the default spack MPI.

scripts/docker/dockerfile_clang-4
```diff
@@ -0,0 +1,30 @@
...
+
+WORKDIR "/home/axom/RAJA"
+RUN mkdir build
+
+WORKDIR "/home/axom/RAJA/build"
+
+RUN cmake ../. -DENABLE_TESTS=OFF
+RUN make 
+RUN sudo make install
+
```
kennyweiss: This probably doesn't matter since we're going to fix the RAJA build problem, 
but should we delete the ``raja/build`` directory?

host-configs/docker/docker-linux-ubuntu16.04-x86_64-gcc@8.1.0.cmake
```diff
@@ -0,0 +1,94 @@
...
+
+set(MPIEXEC "/home/axom/axom_tpls/gcc-8.1.0/mpich-3.2.1/bin/mpiexec" CACHE PATH "")
+
+set(MPIEXEC_NUMPROC_FLAG "-n" CACHE PATH "")
+
+##############
+# Other machine specifics
+##############
+
+set(ENABLE_GTEST_DEATH_TESTS ON CACHE BOOL "")
```
kennyweiss: Note: The ``ENABLE_GTEST_DEATH_TESTS`` option is being set in the host-configs, so we can remove it from the ``linux-build_and_test`` shell script.

scripts/docker/dockerfile_clang-4
```diff
@@ -0,0 +1,30 @@
+FROM axom/compilers:clang-4
```
kennyweiss: Thanks for adding the dockerfiles @keithhealy !

scripts/azure-pipelines/linux-build_and_test.sh
```diff
@@ -16,21 +18,40 @@ function or_die () {
...
+or_die cd build-$HOST_CONFIG-debug
+or_die make -j 3
+make CTEST_OUTPUT_ON_FAILURE=1 test ARGS='-T Test -VV -j8'
+
+
+find ./axom/sidre -type d -exec chmod 755 {} \;
+find ./axom/sidre -type f -exec chmod 644 {} \;
+find . -name "*.xml"
+
+#set -x
```
kennyweiss: @keithhealy -- should we fold the ``DO_BUILD``, ``DO_TEST`` logic back into the above script and delete the stuff below?

scripts/azure-pipelines/linux-build_and_test.sh
```diff
@@ -16,21 +18,40 @@ function or_die () {
...
+
+./config-build.py -hc /home/axom/axom/host-configs/docker/${HOST_CONFIG}.cmake -DENABLE_GTEST_DEATH_TESTS=ON
+
+or_die cd build-$HOST_CONFIG-debug
+or_die make -j 3
+make CTEST_OUTPUT_ON_FAILURE=1 test ARGS='-T Test -VV -j8'
+
+
+find ./axom/sidre -type d -exec chmod 755 {} \;
+find ./axom/sidre -type f -exec chmod 644 {} \;
```
kennyweiss: @keithhealy -- are these ``chmod``s necessary? Or was it part of your testing?

scripts/docker/README
```diff
@@ -0,0 +1,37 @@
...
+
+Once completed, this container can then be used to build and test the Axom application on a Docker-using CI web tool such as Azure Pipelines.
+Already built Docker images containing these third party libraries can be accessed on Docker Hub at this location:
+https://hub.docker.com/r/axom/tpls/tags
+
```
kennyweiss: @keithhealy -- what is the process for updating these TPLs on dockerhub?  Could you please document this?

scripts/azure-pipelines/linux-build_and_test.sh
```diff
@@ -16,21 +18,40 @@ function or_die () {
...
+git submodule update 
+
+echo HOST_CONFIG
+echo $HOST_CONFIG
+
+./config-build.py -hc /home/axom/axom/host-configs/docker/${HOST_CONFIG}.cmake -DENABLE_GTEST_DEATH_TESTS=ON
+
+or_die cd build-$HOST_CONFIG-debug
+or_die make -j 3
+make CTEST_OUTPUT_ON_FAILURE=1 test ARGS='-T Test -VV -j8'
```
kennyweiss: Please add ``or_die`` at the beginning of this line.

scripts/azure-pipelines/linux-build_and_test.sh
```diff
@@ -16,21 +18,40 @@ function or_die () {
...
+
+
+find ./axom/sidre -type d -exec chmod 755 {} \;
+find ./axom/sidre -type f -exec chmod 644 {} \;
+find . -name "*.xml"
```
kennyweiss: Should this also be wrapped with ``or_die``?

rhornung67:
@kennyweiss I agree. I didn't mean for those changes to be in this PR. I just don't think we should spend time testing compilers that folks don't use.

azure-pipelines.yml
```diff
@@ -57,7 +73,13 @@ steps:
...
+  condition: or( eq( variables['Agent.OS'], 'Windows_NT'), eq( variables['Agent.OS'], 'Darwin'))
+- task: VSBuild@1
+  inputs:
+    solution: 'build/*.sln'
+    vsVersion: 'latest'
+  condition: eq( variables['Agent.OS'], 'Windows_NT')
+  displayName: 'Visual Studio Build'
 - script: |
     cd build
     make
```
kennyweiss: Note for future -- it looks like we're building on windows, but not yet testing.  We should fix that in a separate PR.

Tag: @agcapps

scripts/azure-pipelines/linux-build_and_test.sh
keithhealy: Sorry I didn't clean up this file very well.
The find/chmod commands were necessary because some of the sidre files didn't have their permissions set correctly.  The 'find *.xml' is for my own debugging and I'll remove it along with the commented out code.

scripts/azure-pipelines/linux-build_and_test.sh
keithhealy: Sorry I commented in the wrong place.  The chmod commands were necessary.  The find "*.xml" is not and I'll remove it.

scripts/azure-pipelines/linux-build_and_test.sh
keithhealy: As for the DO_BUILD and DO_TEST, I assumed there wouldn't be a case where we didn't want to do both (build and test) so I removed the logic.  I can restore it if that is preferable.

Again I apologize for not doing a better job cleaning up this file.

scripts/docker/README
keithhealy: I'll add to the README.

scripts/azure-pipelines/linux-build_and_test.sh
kennyweiss: Not a problem.  I agree that this seems unnecessary.

@white238, or others, do we need ``DO_BUILD`` and ``DO_TEST`` for our linux CI builds?
Perhaps we had it there to support our coverage tests? (which we'd like to add back in the near future).

rhornung67: I think we need those since we will be able to do build tests for some compilers but may not be able to run tests; e.g., CUDA.

scripts/docker/README
keithhealy: I changed the file to markdown format.

scripts/azure-pipelines/linux-build_and_test.sh
keithhealy: I think we can run up to 10 parallel jobs.
I don't know how many threads of operation there are in a job.
https://docs.microsoft.com/en-us/azure/devops/pipelines/licensing/concurrent-jobs?view=azure-devops

scripts/docker/README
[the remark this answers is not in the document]
keithhealy: I have added to the README to discuss creating images and uploading to Docker Hub.
Regarding the failed tests.  It's a little confusing but it looks like the results you are seeing with the failures was produced by an automatically triggered test run that uses the merged pull request changes. You can see these results here:
https://dev.azure.com/axom/axom/_build/results?buildId=839&view=ms.vss-test-web.build-test-results-tab
Whereas the test results of the unmerged pull request are here:
https://dev.azure.com/healy22/Axom%20CI/_build/results?buildId=206&view=ms.vss-test-web.build-test-results-tab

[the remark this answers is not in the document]
kennyweiss: Thanks @keithhealy that's interesting -- one in is axom's azure space and the other is in your azure space.

I think we want to indicate failure  if either of these has test failures -- premerge or postmerge, since the latter will break tests for others.

Also, could you please merge develop into your branch (probably after fixing the test failure reporting), since I think it is out of date?

scripts/azure-pipelines/linux-build_and_test.sh
kennyweiss: Thanks -- in that case, should we increase this to ``j10``?

scripts/docker/README
kennyweiss: Thanks @keithhealy -- the writeup is really helpful.

Could you please rename it to ``README.md`` ?

azure-pipelines.yml
agcapps: Tag noted.

scripts/docker/README
agcapps: I agree.

scripts/docker/README
keithhealy: Changed to README.md and fixed some formatting errors.

scripts/azure-pipelines/linux-build_and_test.sh
keithhealy: Restored the DO_BUILD and DO_TEST logic.

scripts/azure-pipelines/linux-build_and_test.sh
keithhealy: If the permissions change fails some number of sidre tests will fail but the whole run would not be affected.  Adding or_die would cause these test results to be lost, I believe, as the whole job would be marked as failed.  I should test this.

scripts/uberenv/spack_configs/docker/compilers.yaml
```diff
@@ -0,0 +1,80 @@
+compilers:
+- compiler:
+    environment: {}
+    extra_rpaths: []
+    flags: {}
+    modules: []
+    operating_system: ubuntu16.04
+    paths:
+      cc:  /usr/bin/gcc
```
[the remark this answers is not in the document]
keithhealy: Yes.  Each Docker image has one version installed.

azure-pipelines.yml
```diff
@@ -69,8 +91,10 @@ steps:
   displayName: '$(TEST_TARGET) Test'
   condition: eq( variables['Agent.OS'], 'Darwin')
 - script:  |
-    docker run --rm --user='root' -v `pwd`:/home/axom $(Compiler_ImageName) chown -R axom /home/axom
-    docker run --rm  -v `pwd`:/home/axom -e TEST_TARGET -e COMPILER -e DO_BUILD -e DO_TEST -e CMAKE_EXTRA_FLAGS $(Compiler_ImageName) ./scripts/azure-pipelines/linux-build_and_test.sh
+    echo " -e $TEST_TARGET -e $COMPILER -e $DO_BUILD -e $DO_TEST -e $CMAKE_EXTRA_FLAGS $(Compiler_ImageName) ./scripts/azure-pipelines/linux-build_and_test.sh"
+    docker run --rm --user='root' -v `pwd`:/home/axom/axom $(Compiler_ImageName) chown -R axom /home/axom
```
cyrush: You can actually run the docker container directly with the azure job spec, here is an example:

```
  - job: Ubuntu_Cuda
    pool:
      vmImage: 'ubuntu-16.04'
    container: alpinedav/ascent-ci:cuda-9.2-dev
```

Not sure if that really simplifies things or not in this case, but I wanted to mention it, since I saw we are calling docker ourselves to launch the container.

azure-pipelines.yml
```diff
@@ -13,38 +13,54 @@ strategy:
...
+       COMPILER: 'clang++'
+       TEST_TARGET: 'linux_clang6'
+       HOST_CONFIG: 'docker-linux-ubuntu16.04-x86_64-clang@6.0.0'
     osx_gcc:
       VM_ImageName: 'macos-10.13'
       CMAKE_EXTRA_FLAGS: '-DAXOM_ENABLE_SIDRE=Off'
       TEST_TARGET: 'osx_gcc'
+    windows:
+      VM_ImageName: 'windows-2019'
+      CMAKE_EXTRA_FLAGS: '-DAXOM_ENABLE_SIDRE=Off'
```
gzagaris: Is there a particular reason to turn off Sidre on OS X and windows?

scripts/docker/dockerfile_clang-6
```diff
@@ -0,0 +1,28 @@
...
+RUN sudo apt-get install curl -fy
+
+RUN sudo apt-get -qq install -y --no-install-recommends gfortran-7 && sudo update-alternatives --install /usr/bin/gfortran gfortran /usr/bin/gfortran-7 100
+ 
+#(For Clang Builds - Need to restore stdc++ lib)
+RUN sudo apt-get install libstdc++-7-dev
+
+RUN git clone --recursive https://github.com/LLNL/axom.git
+
+RUN git clone --recursive https://github.com/LLNL/RAJA.git
```
gzagaris: Is the RAJA clone and build step needed? Doesn't the call to `uberenv` below (line 24) install RAJA? (Ditto for all the other docker files.)

azure-pipelines.yml
kennyweiss: @gzagaris -- I think the issue is that we're not building TPLs on these platforms and ``sidre`` requires ``conduit``.  

We'll have to fix this in a separate PR.  
* @agcapps is working on the Windows part.
* For OS X -- does anyone know if docker/azure supports this?

scripts/docker/dockerfile_clang-6
kennyweiss: Keith had trouble getting this to work through docker.  

The plan is to fix this is a separate PR, after this one has merged.

kennyweiss:
#### RE: docker in the host-config name.

@gzagaris -- I am pretty sure this is the name that uberenv generates.
In the short term, it is probably best to keep things as they are so we can just copy them without renaming. Longer term, perhaps we can modify uberenv to omit the ``docker-`` prefix in the generated host-config.

kennyweiss:
At this point, the only unresolved issue is that the ``sidre_lulesh`` tests are timing out on the gcc docker configurations, which I think is due to the interaction between MPI and OpenMP on a constrained azure resource (our node only has 10 threads, and the sidre_lulesh task is being called on 8 ranks with 2 threads each.  

Oddly enough:
* the slam_lulesh test is being called with 8 ranks and 4 threads each and is passing on the gcc configs
* the same sidre_lulesh tests are passing on the clang configs

See: https://dev.azure.com/axom/axom/_build/results?buildId=877&view=ms.vss-test-web.build-test-results-tab

Since this isn't currently showing up as a test failure, I think we should merge the current PR and resolve this separately.  And also figure out how to get azure to indicate a "check" failure when some of the tests are failing.

Any objections to merging as is?

kennyweiss:
I created a ticket to track the open issues and suggested improvements for our CI plan: https://github.com/LLNL/axom/issues/162

Please add to it if I missed anything.

gzagaris:
Thanks for creating the ticket @kennyweiss -- I agree we should merge this as is now and fine tune things in a subsequent PRs.

kennyweiss:
Thanks @keithhealy!

keithhealy:
A couple of comments.
1. Uberenv adds the hostname to the beginning of the host config file name when it is created.  When you run an instance of a Docker image the hostname that is assigned is a random 12 digit alpha-numeric string.  I removed these and added 'Docker' because it was easier to remember (and wouldn't require changing the test scripts each time).  Removing it should not be a problem. 
2. Azure does recognize that the sidre-lulesh tests fail and reports this fact in the published results.  It does not, however, mark the CI jobs as failed when there are failed tests. The rationale seems to be that the jobs have succeeded (in running the tests and publishing the results) regardless of the contents of those results. I believe there are two options to remedy this.
   A. Fail the 'Build and Test' job when tests fail.  Then set the 'Publish Test Results' job to always run.  Normally a job failing will prevent subsequent jobs from running.  In this case we want the test results to be published even if the previous job fails due to failed tests.  Otherwise we get a notification of a failure but no record of which tests failed other than what's in the log.
  B. Fail the 'Publish Test Results' job when tests fail.  This is probably the easiest solution even if it is slightly misleading.  That is, the job succeeded in publishing the test results but has been marked as failed due to a failed result.
Let me know what you think.

gzagaris:
@keithhealy , if there is a test failure, we'd want to make sure that merging of the branch is not going to be allowed. Would this be the case with what you propose in 2.B?

kennyweiss:
@keithhealy -- I agree with @gzagaris -- We want "check" jobs to fail when one or more unit tests fail, or at the very least, we'd want the github "check" reason to indicate that there were failed tests.  Right now, the only way to see that a test failed is to click on the link that says "All tests passed" and click through a "Tests" tab.

Do we have flexibility over the name "Publish Test Results" ?  Or can we add a task after that named something like "Verify Test Results" ?

keithhealy:
There is a way to define checks that must be passed before a pull request can be merged. This is located under Settings/branches of the repo.  I believe it should be possible to check for failing CI tasks.  I'll investigate.
As for flexibility over the tasks itself, the existing ones have a lot of optional settings.  For example, the PublishTestResults task has an option called 'failTaskOnFailedTests' which will fail the task if any tests fail.  For the most flexibility we could just add a script task which finds and parses the results file.  I don't think we could rename this though.

kennyweiss:
> For example, the PublishTestResults task has an option called 'failTaskOnFailedTests' which will fail the task if any tests fail. 

Thanks @keithhealy -- That sounds like the one we want.

cyrush:
It seems the docker run is not returning the resulting return code?

It may simplify things to have azure use the docker container to run the job, instead of running docker ourselves if there are are issues with it not propagating the return codes. 

Some tests not failing is only one example, what about when a compile fails?

Note - for each script block, azure pipelines will only check the return code of the last executed task. This caught me off guard a bit back, so I wanted to mention it.

In my experience -- A ctest azure step should stop if any tests fail.  (ctest will return non-zero)

In fact if you look at Ascent,  we wanted to make sure to post test results even when ctest failed.  To do this, we had to use a `condition: always()` specifier to the PublishTestResults.



```
      - script: |
           #################################
           # test
           #################################
           # find spack installed cmake
           export ROOT_DIR=`pwd`
           export CMAKE_BIN_DIR=`ls -d ${ROOT_DIR}/uberenv_libs/spack/opt/spack/*/*/cmake*/bin`
           export PATH=${CMAKE_BIN_DIR}:$PATH
           which ctest
           cd build
           # run ctest
           ctest -T test --output-on-failure -V
        displayName: 'Run Unit Tests'
      - task: PublishTestResults@2
        inputs:
          testResultsFormat: 'cTest'
          testResultsFiles: '**/Test.xml'
        displayName: 'Unit Test Results'
        condition: always()
```

gzagaris:
Yes -- having an indicator to easily know which actual stage has failed (build or test) is useful. 

I took a look at the ascent setup. I like how it gives a summary of each of the jobs(or configurations) in the pull request. That way we can know whether the Windows or OS X build/ or tests have failed in the Pull Request page, without having to click on the "Details" to go to the azure-pipelines page.  Would that be easy to setup for Axom?

cyrush:
It should be straight forward, you just use top level stages + jobs.

Another nice thing to fall out of this is that you don't need a bunch of optional steps. The per platform steps are simply written for each platform specific job.

keithhealy:
The compile steps are run through the 'or_die' function which will cause the script to exit with a non-zero return code if the compile fails.  The 'make test' step is not run through 'or_die' because I wanted to make sure the permissions was set correctly on the test results and that the 'PublisTestResults' task was always run.
As Cyrus mentioned, setting the 'always()' condition on PublishTestResults is one way to fix this.  Another is to set the PublishTestResults task to fail if any tests fail.  

You can add checks that must be passed when merging by going to Settings/Branches for the repo and selecting 'Require status checks to pass before merging'.  Next click 'Add Rule' and click the checks you want to require before a merge is allowed.  You should get something like this (Scroll down to the merge pane):
https://github.com/keithhealy/MyAxom/pull/1

keithhealy:
Sorry, that should be "going to Settings/Branches for the repo and selecting 'Add Rule' under Branch Protection Rules"

gzagaris:
Thanks -- I think doing the same for Axom would be useful.

PR #156, opened 2020-01-28
RAJA implementation of naive triangle intersection algorithm 

bmhan12:
This PR adds a RAJA implementation of the naiveIntersectionAlgorithm in `mesh_tester.cpp`.
The naive algorithm tests each 3D triangle against each other triangle in a triangle mesh for intersection.
 
- `__host__ __device__` decorates (`AXOM_HOST_DEVICE`) the necessary functions to call `intersect_tri3D_tri3D` from the naïve algorithm.
- Disables the SLIC Check and Assert macros globally when using RAJA.

### DIFF
--- a/src/axom/core/numerics/Determinants.hpp
+++ b/src/axom/core/numerics/Determinants.hpp
@@ -9,6 +9,7 @@
 #include "axom/core/numerics/LU.hpp"     // for lu_decompose()
 #include "axom/core/numerics/Matrix.hpp" // for Matrix
 
+#include "axom/core/Macros.hpp"
 
 namespace axom
 {
@@ -27,7 +28,8 @@ namespace numerics
  * \return det the determinant of the 2X2 matrix
  */
 template < typename real >
-inline real determinant( const real& a00, const real& a01,
+inline AXOM_HOST_DEVICE 
+real determinant( const real& a00, const real& a01,
                          const real& a10, const real& a11 )
 {
   const real det = a00*a11 - a10*a01;
--- a/src/axom/core/numerics/Determinants.hpp
+++ b/src/axom/core/numerics/Determinants.hpp
@@ -30,7 +30,7 @@ namespace numerics
 template < typename real >
 inline AXOM_HOST_DEVICE 
 real determinant( const real& a00, const real& a01,
-                         const real& a10, const real& a11 )
+                  const real& a10, const real& a11 )
 {
   const real det = a00*a11 - a10*a01;
   return det;
--- a/src/axom/core/numerics/matvecops.hpp
+++ b/src/axom/core/numerics/matvecops.hpp
@@ -575,9 +575,9 @@ inline bool linspace( const T& x0, const T& x1, T* v, int N )
 template < typename T >
 inline void cross_product( const T* u, const T* v, T* w )
 {
-    assert( "pre: u pointer is null" && (u != nullptr) );
-    assert( "pre: v pointer is null" && (v != nullptr) );
-    assert( "pre: w pointer is null" && (w != nullptr) );
+  assert( "pre: u pointer is null" && (u != nullptr) );
+  assert( "pre: v pointer is null" && (v != nullptr) );
+  assert( "pre: w pointer is null" && (w != nullptr) );
 
   w[ 0 ] = numerics::determinant( u[1], u[2], v[1], v[2] );
 
@@ -590,9 +590,9 @@ inline void cross_product( const T* u, const T* v, T* w )
 template < typename T >
 inline T dot_product( const T* u, const T* v, int dim)
 {
-    assert("pre: u pointer is null" && (u != nullptr));
-    assert("pre: v pointer is null" && (v != nullptr));
-    assert("pre: dim >= 1" && (dim >= 1));
+  assert("pre: u pointer is null" && (u != nullptr));
+  assert("pre: v pointer is null" && (v != nullptr));
+  assert("pre: dim >= 1" && (dim >= 1));
 
   T res = u[0]*v[0];
   for (int i = 1 ; i < dim ; ++i)
--- a/src/axom/core/numerics/matvecops.hpp
+++ b/src/axom/core/numerics/matvecops.hpp
@@ -575,11 +575,9 @@ inline bool linspace( const T& x0, const T& x1, T* v, int N )
 template < typename T >
 inline void cross_product( const T* u, const T* v, T* w )
 {
-  #ifndef AXOM_USE_RAJA
     assert( "pre: u pointer is null" && (u != nullptr) );
     assert( "pre: v pointer is null" && (v != nullptr) );
     assert( "pre: w pointer is null" && (w != nullptr) );
-  #endif
 
   w[ 0 ] = numerics::determinant( u[1], u[2], v[1], v[2] );
 
@@ -592,11 +590,9 @@ inline void cross_product( const T* u, const T* v, T* w )
 template < typename T >
 inline T dot_product( const T* u, const T* v, int dim)
 {
-  #ifndef AXOM_USE_RAJA
     assert("pre: u pointer is null" && (u != nullptr));
     assert("pre: v pointer is null" && (v != nullptr));
     assert("pre: dim >= 1" && (dim >= 1));
-  #endif
 
   T res = u[0]*v[0];
   for (int i = 1 ; i < dim ; ++i)
--- a/src/axom/core/numerics/matvecops.hpp
+++ b/src/axom/core/numerics/matvecops.hpp
@@ -24,6 +24,8 @@
 #include <cassert> // for assert()
 #include <cmath>   // for sqrt()
 
+#include "axom/core/Macros.hpp"
+
 namespace axom
 {
 namespace numerics
@@ -85,7 +87,8 @@ inline bool linspace( const T& x0, const T& x1, T* v, int N );
  *
  */
 template < typename T >
-inline void cross_product( const T* u, const T* v, T* w );
+inline AXOM_HOST_DEVICE
+void cross_product( const T* u, const T* v, T* w );
 
 /*!
  * \brief Computes the dot product of the arrays u and v.
@@ -103,7 +106,8 @@ inline void cross_product( const T* u, const T* v, T* w );
  * \pre v has at least dim entries
  */
 template < typename T >
-inline T dot_product( const T* u, const T* v, int dim);
+inline AXOM_HOST_DEVICE
+T dot_product( const T* u, const T* v, int dim);
 
 /*!
  * \brief Makes u orthogonal to v.
@@ -571,9 +575,11 @@ inline bool linspace( const T& x0, const T& x1, T* v, int N )
 template < typename T >
 inline void cross_product( const T* u, const T* v, T* w )
 {
-  assert( "pre: u pointer is null" && (u != nullptr) );
-  assert( "pre: v pointer is null" && (v != nullptr) );
-  assert( "pre: w pointer is null" && (w != nullptr) );
+  #ifndef AXOM_USE_RAJA
+    assert( "pre: u pointer is null" && (u != nullptr) );
+    assert( "pre: v pointer is null" && (v != nullptr) );
+    assert( "pre: w pointer is null" && (w != nullptr) );
+  #endif
 
   w[ 0 ] = numerics::determinant( u[1], u[2], v[1], v[2] );
 
@@ -586,9 +592,11 @@ inline void cross_product( const T* u, const T* v, T* w )
 template < typename T >
 inline T dot_product( const T* u, const T* v, int dim)
 {
-  assert("pre: u pointer is null" && (u != nullptr));
-  assert("pre: v pointer is null" && (v != nullptr));
-  assert("pre: dim >= 1" && (dim >= 1));
+  #ifndef AXOM_USE_RAJA
+    assert("pre: u pointer is null" && (u != nullptr));
+    assert("pre: v pointer is null" && (v != nullptr));
+    assert("pre: dim >= 1" && (dim >= 1));
+  #endif
 
   T res = u[0]*v[0];
   for (int i = 1 ; i < dim ; ++i)
--- a/src/axom/primal/geometry/NumericArray.hpp
+++ b/src/axom/primal/geometry/NumericArray.hpp
@@ -12,6 +12,8 @@
 
 #include "axom/slic/interface/slic.hpp"
 
+#include "axom/core/Macros.hpp"
+
 // C/C++ includes
 #include <cstring>    // For memcpy()
 #include <algorithm>  // For std:: copy and fill
@@ -66,6 +68,7 @@ NumericArray< T,SIZE > operator+( const NumericArray< T,SIZE >& lhs,
  * \result C resulting numeric array from component-wise subtraction.
  */
 template < typename T,int SIZE >
+AXOM_HOST_DEVICE
 NumericArray< T,SIZE > operator-( const NumericArray< T,SIZE >& lhs,
                                   const NumericArray< T,SIZE >& rhs  );
 
@@ -201,6 +204,7 @@ public:
    * The rest will be set to zero.  Defaults is SIZE.
    * If sz is greater than SIZE, we set all coordinates to val
    */
+  AXOM_HOST_DEVICE
   explicit NumericArray( T val = T(), int sz = SIZE);
 
   /*!
@@ -209,17 +213,20 @@ public:
    * \param [in] sz number of coordinates. Defaults to SIZE.
    * \note If sz is greater than SIZE, we only take the first SIZE values.
    */
+  AXOM_HOST_DEVICE
   NumericArray(const T* vals, int sz = SIZE);
 
   /*!
    * \brief Copy constructor.
    * \param [in] other The numeric array to copy
    */
+  AXOM_HOST_DEVICE
   NumericArray( const NumericArray& other ) { *this = other; };
 
   /*!
    * \brief Destructor.
    */
+  AXOM_HOST_DEVICE
   ~NumericArray() { }
 
   /*!
@@ -233,6 +240,7 @@ public:
    * \brief Assignment operator.
    * \param [in] rhs a numeric array instance on the right hand side.
    */
+  AXOM_HOST_DEVICE
   NumericArray& operator=(const NumericArray& rhs);
 
   /*!
@@ -241,14 +249,14 @@ public:
    * \return \f$ p_i \f$ the value at the given component index.
    * \pre \f$  0 \le i < SIZE \f$
    */
-  const T& operator[](int i) const;
-  T& operator[](int i);
+  AXOM_HOST_DEVICE const T& operator[](int i) const;
+  AXOM_HOST_DEVICE T& operator[](int i);
 
   /*!
    * \brief Returns a pointer to the underlying data.
    */
-  const T* data() const;
-  T* data();
+  AXOM_HOST_DEVICE const T* data() const;
+  AXOM_HOST_DEVICE T* data();
 
   /*!
    *
@@ -280,6 +288,7 @@ public:
    * Subtracts the numeric array arr from this instance (component-wise).
    * \return A reference to the NumericArray instance after subtraction.
    */
+  AXOM_HOST_DEVICE
   NumericArray< T,SIZE >& operator-=( const NumericArray< T,SIZE >& arr );
 
   /*!
@@ -289,6 +298,7 @@ public:
    * \return A reference to the NumericArray instance after scalar
    * multiplication.
    */
+  AXOM_HOST_DEVICE
   NumericArray< T,SIZE >& operator*=(double scalar);
 
   /*!
@@ -298,6 +308,7 @@ public:
    * Each element of the numeric array is divided by scalar
    * \return A reference to the NumericArray instance after scalar division.
    */
+  AXOM_HOST_DEVICE
   NumericArray< T,SIZE >& operator/=(double scalar);
 
   /*!
@@ -307,6 +318,7 @@ public:
    * \return A reference to the NumericArray instance after cwise
    * multiplication.
    */
+  AXOM_HOST_DEVICE
   NumericArray< T,SIZE >& operator*=( const NumericArray< T,SIZE >& arr );
 
   /*!
@@ -379,6 +391,7 @@ public:
   int argMin() const;
 
 private:
+  AXOM_HOST_DEVICE
   void verifyIndex(int AXOM_DEBUG_PARAM(idx)) const
   {
     SLIC_ASSERT(idx >= 0 && idx < SIZE);
@@ -410,12 +423,13 @@ NumericArray< T,SIZE >::NumericArray(T val, int sz)
 
   // Fill first nvals coordinates with val ( 0 <= nvals <= SIZE )
   const int nvals = axom::utilities::clampVal(sz, 0, SIZE);
-  std::fill( m_components, m_components+nvals, val );
+  for (int i = 0; i < nvals; i++) {
+    m_components[i] = val;    
+  }
 
   // Fill any remaining coordinates with zero
-  if ( nvals < SIZE )
-  {
-    std::fill( m_components+nvals, m_components+SIZE, T() );
+  for (int j = nvals; j < SIZE; j++) {
+    m_components[j] = T();    
   }
 }
 
@@ -428,19 +442,19 @@ NumericArray< T, SIZE >::NumericArray(const T* vals, int sz)
   const int nvals = axom::utilities::clampVal(sz, 0, SIZE);
 
   // Copy first nvals coordinates from vals array ( 0 <= nvals <= SIZE )
-  std::copy( vals, vals+nvals, m_components);
-
-  // Fill any remaining coordinates with zero
-  if ( nvals < SIZE)
-  {
-    std::fill( m_components+nvals, m_components+SIZE, T());
+  for (int i = 0; i < nvals; i++) {
+    m_components[i] = vals[i];    
   }
 
+  // Fill any remaining coordinates with zero
+  for (int j = nvals; j < SIZE; j++) {
+    m_components[j] = T();    
+  }
 }
 
 //------------------------------------------------------------------------------
 template < typename T,int SIZE >
-inline NumericArray< T,SIZE >&
+inline  AXOM_HOST_DEVICE NumericArray< T,SIZE >&
 NumericArray< T,SIZE >::operator=( const NumericArray< T,SIZE >& rhs )
 {
 
--- a/src/axom/primal/geometry/NumericArray.hpp
+++ b/src/axom/primal/geometry/NumericArray.hpp
@@ -7,12 +7,8 @@
 #define NUMERIC_ARRAY_HPP_
 
 #include "axom/core/Macros.hpp"
-
 #include "axom/core/utilities/Utilities.hpp"
-
 #include "axom/slic/interface/slic.hpp"
-
-#include "axom/core/Macros.hpp"
 
 // C/C++ includes
 #include <cstring>    // For memcpy()
@@ -423,12 +419,14 @@ NumericArray< T,SIZE >::NumericArray(T val, int sz)
 
   // Fill first nvals coordinates with val ( 0 <= nvals <= SIZE )
   const int nvals = axom::utilities::clampVal(sz, 0, SIZE);
-  for (int i = 0; i < nvals; i++) {
+  for (auto i = 0; i < nvals; i++)
+  {
     m_components[i] = val;    
   }
 
   // Fill any remaining coordinates with zero
-  for (int j = nvals; j < SIZE; j++) {
+  for (auto j = nvals; j < SIZE; j++)
+  {
     m_components[j] = T();    
   }
 }
@@ -442,19 +440,22 @@ NumericArray< T, SIZE >::NumericArray(const T* vals, int sz)
   const int nvals = axom::utilities::clampVal(sz, 0, SIZE);
 
   // Copy first nvals coordinates from vals array ( 0 <= nvals <= SIZE )
-  for (int i = 0; i < nvals; i++) {
+  for (auto i = 0; i < nvals; i++)
+  {
     m_components[i] = vals[i];    
   }
 
   // Fill any remaining coordinates with zero
-  for (int j = nvals; j < SIZE; j++) {
+  for (auto j = nvals; j < SIZE; j++) 
+  {
     m_components[j] = T();    
   }
 }
 
 //------------------------------------------------------------------------------
 template < typename T,int SIZE >
-inline  AXOM_HOST_DEVICE NumericArray< T,SIZE >&
+inline  AXOM_HOST_DEVICE 
+NumericArray< T,SIZE >&
 NumericArray< T,SIZE >::operator=( const NumericArray< T,SIZE >& rhs )
 {
 
--- a/src/axom/primal/geometry/NumericArray.hpp
+++ b/src/axom/primal/geometry/NumericArray.hpp
@@ -245,14 +245,20 @@ public:
    * \return \f$ p_i \f$ the value at the given component index.
    * \pre \f$  0 \le i < SIZE \f$
    */
-  AXOM_HOST_DEVICE const T& operator[](int i) const;
-  AXOM_HOST_DEVICE T& operator[](int i);
+  AXOM_HOST_DEVICE 
+  const T& operator[](int i) const;
+  
+  AXOM_HOST_DEVICE 
+  T& operator[](int i);
 
   /*!
    * \brief Returns a pointer to the underlying data.
    */
-  AXOM_HOST_DEVICE const T* data() const;
-  AXOM_HOST_DEVICE T* data();
+  AXOM_HOST_DEVICE 
+  const T* data() const;
+  
+  AXOM_HOST_DEVICE 
+  T* data();
 
   /*!
    *
@@ -419,13 +425,13 @@ NumericArray< T,SIZE >::NumericArray(T val, int sz)
 
   // Fill first nvals coordinates with val ( 0 <= nvals <= SIZE )
   const int nvals = axom::utilities::clampVal(sz, 0, SIZE);
-  for (auto i = 0; i < nvals; i++)
+  for (int i = 0; i < nvals; i++)
   {
     m_components[i] = val;    
   }
 
   // Fill any remaining coordinates with zero
-  for (auto j = nvals; j < SIZE; j++)
+  for (int j = nvals; j < SIZE; j++)
   {
     m_components[j] = T();    
   }
@@ -440,13 +446,13 @@ NumericArray< T, SIZE >::NumericArray(const T* vals, int sz)
   const int nvals = axom::utilities::clampVal(sz, 0, SIZE);
 
   // Copy first nvals coordinates from vals array ( 0 <= nvals <= SIZE )
-  for (auto i = 0; i < nvals; i++)
+  for (int i = 0; i < nvals; i++)
   {
     m_components[i] = vals[i];    
   }
 
   // Fill any remaining coordinates with zero
-  for (auto j = nvals; j < SIZE; j++) 
+  for (int j = nvals; j < SIZE; j++) 
   {
     m_components[j] = T();    
   }
--- a/src/axom/primal/geometry/Triangle.hpp
+++ b/src/axom/primal/geometry/Triangle.hpp
@@ -14,6 +14,8 @@
 
 #include "axom/slic/interface/slic.hpp"
 
+#include "axom/core/Macros.hpp"
+
 #include <cmath>   // for acos()
 #include <ostream> // for std::ostream
 
@@ -56,6 +58,7 @@ public:
   /*!
    * \brief Default constructor. Creates a degenerate triangle.
    */
+  AXOM_HOST_DEVICE
   Triangle() { }
 
   /*!
@@ -64,6 +67,7 @@ public:
    * \param [in] B point instance corresponding to vertex B of the triangle.
    * \param [in] C point instance corresponding to vertex C of the triangle.
    */
+  AXOM_HOST_DEVICE
   Triangle( const PointType& A,
             const PointType& B,
             const PointType& C );
@@ -71,6 +75,7 @@ public:
   /*!
    * \brief Destructor
    */
+  AXOM_HOST_DEVICE
   ~Triangle() { }
 
   /*!
@@ -78,6 +83,7 @@ public:
    * \param idx The index of the desired vertex
    * \pre idx is 0, 1 or 2
    */
+  AXOM_HOST_DEVICE
   PointType& operator[](int idx)
   {
     SLIC_ASSERT(idx >=0 && idx < NUM_TRI_VERTS);
@@ -89,6 +95,7 @@ public:
    * \param idx The index of the desired vertex
    * \pre idx is 0, 1 or 2
    */
+  AXOM_HOST_DEVICE
   const PointType& operator[](int idx) const
   {
     SLIC_ASSERT(idx >=0 && idx < NUM_TRI_VERTS);
@@ -100,6 +107,7 @@ public:
    * \pre This function is only valid when NDIMS = 3
    * \return n triangle normal when NDIMS=3, zero vector otherwise
    */
+  AXOM_HOST_DEVICE
   VectorType normal() const
   {
     SLIC_CHECK_MSG(NDIMS==3, "Triangle::normal() is only valid in 3D.");
@@ -114,6 +122,7 @@ public:
    * \brief Returns the area of the triangle
    * \pre Only defined when dimension NDIMS is 2 or 3
    */
+  AXOM_HOST_DEVICE
   double area() const
   {
     SLIC_CHECK_MSG( NDIMS == 2 || NDIMS == 3,
@@ -241,6 +250,7 @@ public:
    * \return true iff the triangle is degenerate (0 area)
    * \see primal::Point
    */
+  AXOM_HOST_DEVICE
   bool degenerate(double eps = 1.0e-12) const
   {
     return axom::utilities::isNearlyEqual(area(),  0.0, eps);
@@ -305,12 +315,9 @@ namespace primal
 template < typename T, int NDIMS >
 Triangle< T,NDIMS >::Triangle( const PointType& A,
                                const PointType& B,
-                               const PointType& C  )
-{
-  m_points[0] = A;
-  m_points[1] = B;
-  m_points[2] = C;
-}
+                               const PointType& C  ):
+  m_points{A,B,C}
+{}
 
 //------------------------------------------------------------------------------
 template < typename T, int NDIMS >
--- a/src/axom/primal/geometry/Triangle.hpp
+++ b/src/axom/primal/geometry/Triangle.hpp
@@ -6,15 +6,15 @@
 #ifndef TRIANGLE_HPP_
 #define TRIANGLE_HPP_
 
-#include "axom/core/numerics/Determinants.hpp" // For numerics::determinant()
+#include "axom/core/Macros.hpp"
+#include "axom/core/numerics/Determinants.hpp"
 #include "axom/core/utilities/Utilities.hpp"
+
+#include "axom/slic/interface/slic.hpp"
 
 #include "axom/primal/geometry/Point.hpp"
 #include "axom/primal/geometry/Vector.hpp"
 
-#include "axom/slic/interface/slic.hpp"
-
-#include "axom/core/Macros.hpp"
 
 #include <cmath>   // for acos()
 #include <ostream> // for std::ostream
@@ -315,8 +315,8 @@ namespace primal
 template < typename T, int NDIMS >
 Triangle< T,NDIMS >::Triangle( const PointType& A,
                                const PointType& B,
-                               const PointType& C  ):
-  m_points{A,B,C}
+                               const PointType& C)
+   : m_points{A,B,C} 
 {}
 
 //------------------------------------------------------------------------------
--- a/src/axom/sidre/examples/lulesh2/CMakeLists.txt
+++ b/src/axom/sidre/examples/lulesh2/CMakeLists.txt
@@ -37,11 +37,11 @@ blt_add_executable(
 if(AXOM_ENABLE_TESTS)
     if(ENABLE_MPI)
         blt_add_test(NAME sidre_lulesh2
-                     COMMAND sidre_lulesh2_ex
+                     COMMAND sidre_lulesh2_ex -s 5
                      NUM_MPI_TASKS 1
                      NUM_OMP_THREADS 2 )
     else()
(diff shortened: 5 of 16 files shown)


src/tools/mesh_tester.cpp
```diff
@@ -250,6 +269,124 @@ std::vector< std::pair<int, int> > naiveIntersectionAlgorithm(
...
+std::vector< std::pair<int, int> > naiveIntersectionAlgorithm(
+  mint::Mesh* surface_mesh,
+  std::vector<int> & degenerate)
+{
+  SLIC_INFO("Execution Space is " <<
+    axom::execution_space< ExecSpace >::name());
+
+  // Get allocator
+  int allocatorID = axom::execution_space< ExecSpace >::allocatorID();
+  umpire::Allocator allocator = axom::getAllocator(allocatorID);
```
kennyweiss: Umpire usage should be guarded by ``AXOM_USE_UMPIRE`` since you are not strictly requiring umpire as a precondition for the ``mesh_tester`` app.

@gzagaris -- is there a clean way to set this up with umpire allocators when umpire is available and using the default allocation when it is not?

src/tools/mesh_tester.cpp
```diff
@@ -250,6 +269,124 @@ std::vector< std::pair<int, int> > naiveIntersectionAlgorithm(
...
+    if (t1.degenerate())
+    {
+      degenerate.push_back(i);
+    }
+  }
+
+  RAJA::RangeSegment row_range(0, ncells);
+  RAJA::RangeSegment col_range(0, ncells);
+
+  // Kernel policy
```
kennyweiss: @gzagaris -- is there a cleaner way to define these (nested) kernels using the execution policies that you set up?

src/tools/CMakeLists.txt
```diff
@@ -53,8 +53,11 @@ set(mesh_tester_depends
 
 blt_list_append(TO mesh_tester_depends ELEMENTS lumberjack mpi IF AXOM_ENABLE_LUMBERJACK)
 blt_list_append(TO mesh_tester_depends ELEMENTS openmp IF ENABLE_OPENMP)
+blt_list_append( TO mesh_tester_depends ELEMENTS umpire IF ${UMPIRE_FOUND} )
+blt_list_append( TO mesh_tester_depends ELEMENTS openmp  IF ${ENABLE_OPENMP} )
+blt_list_append( TO mesh_tester_depends ELEMENTS cuda IF ${ENABLE_CUDA} )
 
-if(AXOM_ENABLE_QUEST)
+if(AXOM_ENABLE_QUEST AND RAJA_FOUND)
```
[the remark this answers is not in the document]
bmhan12: Reverted. I did add additional guards to `mesh_tester` for `raja` and` umpire`.

src/axom/core/numerics/matvecops.hpp
```diff
@@ -571,9 +575,11 @@ inline bool linspace( const T& x0, const T& x1, T* v, int N )
 template < typename T >
 inline void cross_product( const T* u, const T* v, T* w )
 {
-  assert( "pre: u pointer is null" && (u != nullptr) );
-  assert( "pre: v pointer is null" && (v != nullptr) );
-  assert( "pre: w pointer is null" && (w != nullptr) );
+  #ifndef AXOM_USE_RAJA
+    assert( "pre: u pointer is null" && (u != nullptr) );
```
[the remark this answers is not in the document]
gzagaris: CUDA has a native `assert()` -- this shouldn't be ifdefed out.

src/axom/core/numerics/Determinants.hpp
```diff
@@ -27,7 +28,7 @@ namespace numerics
  * \return det the determinant of the 2X2 matrix
  */
 template < typename real >
-inline real determinant( const real& a00, const real& a01,
+inline AXOM_HOST_DEVICE real determinant( const real& a00, const real& a01,
```
gzagaris: newline after `AXOM_HOST_DEVICE` for consistency

src/axom/primal/geometry/Triangle.hpp
```diff
@@ -305,12 +315,9 @@ namespace primal
...
                                const PointType& B,
-                               const PointType& C  )
-{
-  m_points[0] = A;
-  m_points[1] = B;
-  m_points[2] = C;
-}
+                               const PointType& C  ):
+  m_points{A,B,C}
+{}
```
gzagaris: Nice find @bmhan12 -- compilers, trust them you should not!

src/axom/slic/interface/slic_macros.hpp
```diff
@@ -106,7 +106,8 @@
 
 /// @}
 
-#ifdef AXOM_DEBUG
+// Disable Asserts when using RAJA kernels
+#if defined(AXOM_DEBUG) && !defined(__CUDA_ARCH__)
```
gzagaris: I think it is better to have `asserts()` for errors when when running on the device, but we shouldn't exclude all Slic macros

src/tools/mesh_tester.cpp
gzagaris: Based on conversation with @rhornung67 , we opted not to include convenience aliases for RAJA::Kernel since it can be too restrictive and lead to less performant implementations, which would allow users to use RAJA in a way they are not supposed to. Instead we push that to the application. 

We should revisit this and think about this more. If nothing else, we should provide the means to define those using our predefined axom::* policies

src/axom/core/numerics/matvecops.hpp
```diff
@@ -571,9 +575,11 @@ inline bool linspace( const T& x0, const T& x1, T* v, int N )
 template < typename T >
 inline void cross_product( const T* u, const T* v, T* w )
 {
-  assert( "pre: u pointer is null" && (u != nullptr) );
-  assert( "pre: v pointer is null" && (v != nullptr) );
-  assert( "pre: w pointer is null" && (w != nullptr) );
+  #if !defined(__CUDA_ARCH__)
```
gzagaris: Instead of using guards with `__CUDA_ARCH__`, I suggest we add the following macro in `axom/Macros.hpp`:
```
#if defined(__CUDA__ARCH__)
#define AXOM_DEVICE_CODE
#endif 
```

That way this can be extended to different device types in one place while we more uniformly use `AXOM_DEVICE_CODE` throughout the code.

src/axom/slic/interface/slic_macros.hpp
kennyweiss: @bmhan12 -- what I had in mind was something more like this:

```cpp
#if defined(AXOM_DEBUG)

#define SLIC_ASSERT( EXP )                                                    \
  do {                                                                        \
#  ifdef __CUDA_ARCH__       // or the new macro that @gzagaris suggested     \
    assert( !EXP);  // assuming we can get assert to work on the device as suggested by @gzagaris  \
#  else                                                                       \
    if ( !(EXP) ) {                                                           \
      std::ostringstream oss;                                                 \
      oss << "Failed Assert: " << # EXP << std::ends;                         \
      axom::slic::logErrorMessage(oss.str(),__FILE__,__LINE__ );              \
    }                                                                         \
# endif                                                                       \
  } while ( axom::slic::detail::false_value )

...
#endif // defined(AXOM_DEBUG)
```

I.e. where the device check is folded into the macro definition.
The benefit of this is that the device check is embedded in the caller side, so the slic macros will still run on the host side, if that is where the code is being run

src/axom/core/numerics/matvecops.hpp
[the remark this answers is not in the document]
bmhan12: I had these `assert()` guarded because of a compiler bug I found with CUDA 10.2 on rzansel. I have not ran into the issue with CUDA 10.1, so it should be fine to leave them unguarded.

src/axom/core/numerics/matvecops.hpp
```diff
@@ -571,9 +575,11 @@ inline bool linspace( const T& x0, const T& x1, T* v, int N )
 template < typename T >
 inline void cross_product( const T* u, const T* v, T* w )
 {
-  assert( "pre: u pointer is null" && (u != nullptr) );
-  assert( "pre: v pointer is null" && (v != nullptr) );
-  assert( "pre: w pointer is null" && (w != nullptr) );
+  #ifndef AXOM_USE_RAJA
```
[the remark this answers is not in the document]
bmhan12: Had these assert() unguarded for reason above.

src/axom/slic/interface/slic_macros.hpp
bmhan12: Done. Slic assert and check macros become `assert()` when running on the device.

gzagaris: Thanks @bmhan12!

src/axom/core/numerics/matvecops.hpp
bmhan12: Added.

src/axom/primal/geometry/NumericArray.hpp
```diff
@@ -428,19 +442,19 @@ NumericArray< T, SIZE >::NumericArray(const T* vals, int sz)
...
   }
-
 }
 
 //------------------------------------------------------------------------------
 template < typename T,int SIZE >
-inline NumericArray< T,SIZE >&
+inline  AXOM_HOST_DEVICE NumericArray< T,SIZE >&
 NumericArray< T,SIZE >::operator=( const NumericArray< T,SIZE >& rhs )
 {
```
kennyweiss: @bmhan12 -- I am making some small edits as I review this PR.

You replaced all the ``std::fill`` and ``std::copy`` calls above.
Is the ``memcpy`` below ok on the device? Or does it just not get called in any of our code paths?

bmhan12: Yes,` memcpy` is ok on device.

kennyweiss:
@bmhan12 -- I am seeing a number of warnings of the following form with the ``rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.1_nvcc_xlf.cmake`` host-config:

```
<axom>/src/axom/primal/geometry/NumericArray.hpp(424): warning: calling a __host__ function from a __host__ __device__ function is not allowed
          detected during:
            instantiation of "axom::primal::NumericArray<T, SIZE>::NumericArray(T, int) [with T=axom::spin::BlockData, SIZE=8]" 
<axom>/src/axom/spin/DenseOctreeLevel.hpp(155): here
            instantiation of "axom::spin::DenseOctreeLevel<DIM, BlockDataType, MortonIndexType>::DenseOctreeLevel(int) [with DIM=3, BlockDataType=axom::spin::BlockData, MortonIndexType=axom::uint16]" 
<axom>/src/axom/spin/OctreeBase.hpp(499): here
            instantiation of "axom::spin::OctreeBase<DIM, BlockDataType>::OctreeBase() [with DIM=3, BlockDataType=axom::spin::BlockData]" 
<axom>/src/axom/spin/tests/spin_octree.cpp(77): here

<axom>/src/axom/primal/geometry/NumericArray.hpp(430): warning: calling a __host__ function from a __host__ __device__ function is not allowed
          detected during:
            instantiation of "axom::primal::NumericArray<T, SIZE>::NumericArray(T, int) [with T=axom::spin::BlockData, SIZE=8]" 
<axom>/src/axom/spin/DenseOctreeLevel.hpp(155): here
            instantiation of "axom::spin::DenseOctreeLevel<DIM, BlockDataType, MortonIndexType>::DenseOctreeLevel(int) [with DIM=3, BlockDataType=axom::spin::BlockData, MortonIndexType=axom::uint16]" 
<axom>/src/axom/spin/OctreeBase.hpp(499): here
            instantiation of "axom::spin::OctreeBase<DIM, BlockDataType>::OctreeBase() [with DIM=3, BlockDataType=axom::spin::BlockData]" 
<axom>/src/axom/spin/tests/spin_octree.cpp(77): here

<axom>/src/axom/primal/geometry/NumericArray.hpp(430): warning: calling a __host__ function from a __host__ __device__ function is not allowed
          detected during:
            instantiation of "axom::primal::NumericArray<T, SIZE>::NumericArray(T, int) [with T=axom::spin::BlockData, SIZE=8]" 
<axom>/src/axom/spin/DenseOctreeLevel.hpp(155): here
            instantiation of "axom::spin::DenseOctreeLevel<DIM, BlockDataType, MortonIndexType>::DenseOctreeLevel(int) [with DIM=3, BlockDataType=axom::spin::BlockData, MortonIndexType=axom::uint16]" 
<axom>/src/axom/spin/OctreeBase.hpp(499): here
            instantiation of "axom::spin::OctreeBase<DIM, BlockDataType>::OctreeBase() [with DIM=3, BlockDataType=axom::spin::BlockData]" 
<axom>/src/axom/spin/tests/spin_octree.cpp(77): here
```

I think we discussed this, but is there a way to resolve/suppress these?

src/tools/mesh_tester.cpp
```diff
@@ -79,15 +110,36 @@ struct Input
   void fixOutfilePath();
 };
 
+const std::set<rajaPolicy> Input::s_validPolicies({
```
kennyweiss: @bmhan12 -- This interface seems a bit awkward to me.

What do you think about changing this to ``runtime policy``, with a default of ``sequential`` (non-raja) ?

The user can then set the policy to one of the following:
* seq           // default -- always available
* raja-seq    // if AXOM_USE_RAJA
* raja-omp  // if AXOM_USE_RAJA and AXOM_USE_OPENMP
* raja-cuda  // if AXOM_USE_RAJA and AXOM_USE_CUDA

bmhan12: I think that is a good idea!

rhornung67:
@kennyweiss those warnings are a common occurrence when compiling with nvcc.  You can try using the RAJA_SUPPRESS_HD_WARN macro we have in RAJA to kill these. Poke around int the RAJA source to see how we use them.

src/tools/mesh_tester.cpp
kennyweiss: About to push this change -- running some tests locally to make sure I got everything right.

kennyweiss:
|  You can try using the RAJA_SUPPRESS_HD_WARN macro we have in RAJA to kill these

Thanks @rhornung67. 
@bmhan12 -- should we address this in the current PR, or separately?

bmhan12:
@kennyweiss Separately might be better. I tried out the macro `RAJA_SUPPRESS_HD_WARN ` in Axom and it works. When you pointed out the warnings, I also noticed there were some other warnings with MFEM and inside `linear_bvh.`

MFEM:
```
/usr/bin/ld: /usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2020_02_18_10_07_58/clang-8.0.1_nvcc_xlf/mfem-4.0/lib/libmfem.a(densemat.o): In function `mfem::DenseMatrix::CalcEigenvalues(double*, double*) const':
densemat.cpp:(.text+0xa62c): toc optimization is not supported for 0xe7e30003 instruction.
/usr/bin/ld: /usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2020_02_18_10_07_58/clang-8.0.1_nvcc_xlf/mfem-4.0/lib/libmfem.a(fe.o): In function `mfem::BiCubic2DFiniteElement::CalcDShape(mfem::IntegrationPoint const&, mfem::DenseMatrix&) const':
fe.cpp:(.text+0x10168): toc optimization is not supported for 0xe4440003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x10184): toc optimization is not supported for 0xe4630003 instruction.
/usr/bin/ld: /usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2020_02_18_10_07_58/clang-8.0.1_nvcc_xlf/mfem-4.0/lib/libmfem.a(fe.o): In function `mfem::BiCubic2DFiniteElement::CalcHessian(mfem::IntegrationPoint const&, mfem::DenseMatrix&) const':
fe.cpp:(.text+0x103d4): toc optimization is not supported for 0xe4030003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x103e8): toc optimization is not supported for 0xe4230003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x10414): toc optimization is not supported for 0xe4c30003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x104d8): toc optimization is not supported for 0xe5630003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x1053c): toc optimization is not supported for 0xe4060003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x1054c): toc optimization is not supported for 0xe4e60003 instruction.
/usr/bin/ld: /usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2020_02_18_10_07_58/clang-8.0.1_nvcc_xlf/mfem-4.0/lib/libmfem.a(fe.o): In function `mfem::Cubic2DFiniteElement::CalcDShape(mfem::IntegrationPoint const&, mfem::DenseMatrix&) const':
fe.cpp:(.text+0x10d2c): toc optimization is not supported for 0xe4a60003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x10d34): toc optimization is not supported for 0xe4c60003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x10d48): toc optimization is not supported for 0xe4e60003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x10d64): toc optimization is not supported for 0xe5060003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x10d90): toc optimization is not supported for 0xe4c40003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x10d94): toc optimization is not supported for 0xe5260003 instruction.
/usr/bin/ld: /usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2020_02_18_10_07_58/clang-8.0.1_nvcc_xlf/mfem-4.0/lib/libmfem.a(fe.o): In function `mfem::Cubic2DFiniteElement::CalcHessian(mfem::IntegrationPoint const&, mfem::DenseMatrix&) const':
fe.cpp:(.text+0x11010): toc optimization is not supported for 0xe4470003 instruction.
/usr/bin/ld: /usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2020_02_18_10_07_58/clang-8.0.1_nvcc_xlf/mfem-4.0/lib/libmfem.a(fe.o): In function `mfem::Cubic3DFiniteElement::CalcDShape(mfem::IntegrationPoint const&, mfem::DenseMatrix&) const':
fe.cpp:(.text+0x115d0): toc optimization is not supported for 0xe4230003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x115e8): toc optimization is not supported for 0xe4460003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x115fc): toc optimization is not supported for 0xe4680003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x11604): toc optimization is not supported for 0xe4a30003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x11618): toc optimization is not supported for 0xe4870003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x11710): toc optimization is not supported for 0xe4280003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x11728): toc optimization is not supported for 0xe5280003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x118e4): toc optimization is not supported for 0xe4660003 instruction.
/usr/bin/ld: /usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2020_02_18_10_07_58/clang-8.0.1_nvcc_xlf/mfem-4.0/lib/libmfem.a(fe.o): In function `mfem::RT2TriangleFiniteElement::CalcVShape(mfem::IntegrationPoint const&, mfem::DenseMatrix&) const':
fe.cpp:(.text+0x1588c): toc optimization is not supported for 0xe5240002 instruction.
/usr/bin/ld: fe.cpp:(.text+0x15894): toc optimization is not supported for 0xe5440002 instruction.
/usr/bin/ld: fe.cpp:(.text+0x1589c): toc optimization is not supported for 0xe5640002 instruction.
/usr/bin/ld: fe.cpp:(.text+0x158bc): toc optimization is not supported for 0xe5640002 instruction.
/usr/bin/ld: fe.cpp:(.text+0x158d4): toc optimization is not supported for 0xe5640002 instruction.
/usr/bin/ld: fe.cpp:(.text+0x158e4): toc optimization is not supported for 0xe5640003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x158f4): toc optimization is not supported for 0xe5640002 instruction.
/usr/bin/ld: fe.cpp:(.text+0x15904): toc optimization is not supported for 0xe5640002 instruction.
/usr/bin/ld: fe.cpp:(.text+0x15920): toc optimization is not supported for 0xe5440002 instruction.
/usr/bin/ld: fe.cpp:(.text+0x15928): toc optimization is not supported for 0xe5840003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x15948): toc optimization is not supported for 0xe5840002 instruction.
/usr/bin/ld: fe.cpp:(.text+0x15960): toc optimization is not supported for 0xe5840002 instruction.
/usr/bin/ld: /usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2020_02_18_10_07_58/clang-8.0.1_nvcc_xlf/mfem-4.0/lib/libmfem.a(fe.o): In function `mfem::RT2QuadFiniteElement::CalcVShape(mfem::IntegrationPoint const&, mfem::DenseMatrix&) const':
fe.cpp:(.text+0x15ed0): toc optimization is not supported for 0xe4440003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x15ed8): toc optimization is not supported for 0xe4640003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x15f24): toc optimization is not supported for 0xe4a30002 instruction.
/usr/bin/ld: /usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2020_02_18_10_07_58/clang-8.0.1_nvcc_xlf/mfem-4.0/lib/libmfem.a(fe.o): In function `mfem::RT1HexFiniteElement::CalcDivShape(mfem::IntegrationPoint const&, mfem::Vector&) const':
fe.cpp:(.text+0x1dba4): toc optimization is not supported for 0xe4250003 instruction.
/usr/bin/ld: fe.cpp:(.text+0x1dbc0): toc optimization is not supported for 0xe4240003 instruction.
```

linear_bvh:
```
/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(231): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<float,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed

/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(232): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<float,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed

/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(235): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<float,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed

/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(236): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<float,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed

/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(231): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<double,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed

/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(232): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<double,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed

/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(235): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<double,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed

/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(236): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<double,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed

/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(189): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<float,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed

/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(190): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<float,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed

/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(191): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<float,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed

/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(193): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<float,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed

/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(194): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<float,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed

/g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(195): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<float,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed
```

kennyweiss:
> Separately might be better

Sounds good, @bmhan12 -- could you please create an Issue to track this?

gzagaris:
> linear_bvh:
> 
> ```
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(231): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<float,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed
> 
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(232): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<float,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed
> 
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(235): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<float,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed
> 
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(236): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<float,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed
> 
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(231): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<double,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed
> 
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(232): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<double,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed
> 
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(235): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<double,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed
> 
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(236): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<double,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed
> 
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(189): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<float,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed
> 
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(190): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<float,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed
> 
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(191): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMin<float,  ::RAJA::detail::ReduceSeq> ::min const") from a __host__ __device__ function(" const") is not allowed
> 
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(193): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<float,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed
> 
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(194): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<float,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed
> 
> /g/g12/han12/axom/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp(195): warning: calling a __host__ function("RAJA::reduce::detail::BaseReduceMax<float,  ::RAJA::detail::ReduceSeq> ::max const") from a __host__ __device__ function(" const") is not allowed
> ```

The linear BVH warnings are coming from the RAJA reducers. They can be suppressed with the macro as @rhornung67 explained.

src/tools/CMakeLists.txt
```diff
@@ -53,6 +53,9 @@ set(mesh_tester_depends
 
 blt_list_append(TO mesh_tester_depends ELEMENTS lumberjack mpi IF AXOM_ENABLE_LUMBERJACK)
 blt_list_append(TO mesh_tester_depends ELEMENTS openmp IF ENABLE_OPENMP)
+blt_list_append(TO mesh_tester_depends ELEMENTS umpire IF UMPIRE_FOUND)
+blt_list_append(TO mesh_tester_depends ELEMENTS RAJA IF RAJA_FOUND)
+blt_list_append(TO mesh_tester_depends ELEMENTS cuda IF ENABLE_CUDA)
 
 if(AXOM_ENABLE_QUEST)
```
kennyweiss: @white238 -- I am getting an installation error on ``rzgenie -- clang@4`` (among others) with this PR (I split up the paths for easier readability):

```
CMake Error at tools/cmake_install.cmake:65 (file):
  file RPATH_CHANGE could not write new RPATH:

  to the file:

    <axom>/_axom_build_and_test_2020_02_21_17_11_43/install-rzgenie-toss_3_x86_64_ib-clang@4.0.0/bin/mesh_tester

  The current RPATH is:

   /usr/tce/packages/clang/clang-4.0.0/lib:                     \
   /usr/tce/packages/gcc/gcc-4.9.3/lib64:                       \
   /usr/tce/packages/gcc/gcc-4.9.3/lib:                         \
   /usr/tce/packages/mvapich2/mvapich2-2.3-clang-4.0.0/lib:     \
   /usr/tce/packages/mvapich2/mvapich2-2.3-clang-4.0.0/lib:

  which does not contain:

   /usr/tce/packages/mvapich2/mvapich2-2.3-clang-4.0.0/lib:     \
   /usr/WS1/axom/libs/toss_3_x86_64_ib/2020_02_18_10_11_15/clang-4.0.0/zlib-1.2.11/lib:

  as was expected.
Call Stack (most recent call first):
  cmake_install.cmake:53 (include)


make: *** [install] Error 1
```

I verified that ``zlib`` is in the path that it is listing, but am not sure why ``mesh_tester``'s installation would want ``zlib`` and why it wouldn't be there if it did need it.

Thoughts?

src/tools/mesh_tester.cpp
```diff
@@ -250,6 +328,109 @@ std::vector< std::pair<int, int> > naiveIntersectionAlgorithm(
...
+    int allocatorID = axom::execution_space< ExecSpace >::allocatorID();
+    umpire::Allocator allocator = axom::getAllocator(allocatorID);
+  #endif
+
+  std::vector< std::pair<int, int> > retval;
+
+  const int ncells = surface_mesh->getNumberOfCells();
+  SLIC_INFO("Checking mesh with a total of "<< ncells<< " cells.");
+
+  #ifdef AXOM_USE_UMPIRE
```
kennyweiss: @gzagaris -- This is what I mentioned the other day.  We should update the ``axom::allocate`` interface to allow a dummy (integer) allocator ID when Umpire is not available.

See: 
https://github.com/LLNL/axom/blob/146f2b12fa1ffe4d430a449e13adece30d1a5de2/src/axom/core/memory_management.hpp#L98-L105

gzagaris: Yes -- I'll take care of this. Again, we should just take an integer ID and don't expose the `umpire` objects in the API.

bmhan12:
@kennyweiss The changes look good! Thanks for the help!

kennyweiss:
@bmhan12 -- Please update the code to reflect the changes in #176, once they are merged
(either during this PR or afterwards).

src/axom/sidre/examples/lulesh2/CMakeLists.txt
```diff
@@ -37,11 +37,11 @@ blt_add_executable(
 if(AXOM_ENABLE_TESTS)
     if(ENABLE_MPI)
         blt_add_test(NAME sidre_lulesh2
-                     COMMAND sidre_lulesh2_ex
+                     COMMAND sidre_lulesh2_ex -s 5
                      NUM_MPI_TASKS 1
                      NUM_OMP_THREADS 2 )
     else()
-        blt_add_test(NAME sidre_lulesh2
+        blt_add_test(NAME sidre_lulesh2 -s 5
```
gzagaris: Do you intend to merge these changes or were these temporary changes to get things passing in your branch? Should the changes to the `sidre_lulesh` test be reverted to whatever is in develop now?

src/tools/mesh_tester.cpp
```diff
@@ -250,6 +328,109 @@ std::vector< std::pair<int, int> > naiveIntersectionAlgorithm(
...
+  std::vector<int> & degenerate)
+{
+  SLIC_INFO("Running naive intersection algorithm "
+    << " in execution Space: " 
+    << axom::execution_space< ExecSpace >::name());
+
+  // Get allocator
+  #ifdef AXOM_USE_UMPIRE
+    int allocatorID = axom::execution_space< ExecSpace >::allocatorID();
+    umpire::Allocator allocator = axom::getAllocator(allocatorID);
```
gzagaris: You should be able to just use `allocatorID` here and get rid of all the `AXOM_USE_UMPIRE` ifdefs

src/axom/slic/interface/slic_macros.hpp
```diff
@@ -223,6 +224,14 @@
 
 /// @}
 
+// Use assert when on device
+#elif defined(AXOM_DEBUG) && defined(AXOM_DEVICE_CODE)
+
+#define SLIC_ASSERT( EXP )  assert (EXP)                                    
+#define SLIC_ASSERT_MSG( EXP, msg ) assert (EXP) 
+#define SLIC_CHECK( EXP ) assert (EXP)
+#define SLIC_CHECK_MSG( EXP, msg ) assert (EXP)
```
gzagaris: Nice! We are currently not handling `SLIC_ERROR`, `SLIC_ERROR_IF` and `SLIC_WARNING` right? (This can be done in a different PR)

src/tools/mesh_tester.cpp
gzagaris: Fixed in 0ed1af3682b6fdf71bc08c4788f94aa0d54993f1

src/tools/mesh_tester.cpp
```diff
@@ -250,6 +328,109 @@ std::vector< std::pair<int, int> > naiveIntersectionAlgorithm(
...
+
+  // Allocation to hold intersection pairs and counter to know where to store
+  #ifdef AXOM_USE_UMPIRE
+    int * intersections =
+      axom::allocate <int> (numIntersect.get() * 2, allocator);
+    int * counter = axom::allocate <int> (1, allocator);
+  #else
+    int * intersections =
+      axom::allocate <int> (numIntersect.get() * 2);
+    int * counter = axom::allocate <int> (1);
```
gzagaris: Would it be simpler to make `counter` a RAJA ReduceSum operation? 

Another, perhaps better alternative would be to use our `axom::StackArray` for this since the size is known a priori.

src/axom/primal/geometry/NumericArray.hpp
```diff
@@ -410,12 +419,15 @@ NumericArray< T,SIZE >::NumericArray(T val, int sz)
 
   // Fill first nvals coordinates with val ( 0 <= nvals <= SIZE )
   const int nvals = axom::utilities::clampVal(sz, 0, SIZE);
-  std::fill( m_components, m_components+nvals, val );
+  for (auto i = 0; i < nvals; i++)
```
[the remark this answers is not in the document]
kennyweiss: I disagree -- ``auto`` is idiomatic ``C++11`` and the compiler knows the type, so it should not influence optimization.  

The merit for ``for`` loops is in the developer not having to worry about if the loop bounds are signed or unsigned.  

Perhaps we should discuss this at our next team meeting.

src/axom/sidre/examples/lulesh2/CMakeLists.txt
kennyweiss: I made this change, and yes, it was intentional.  
Please see the commit message (https://github.com/LLNL/axom/pull/156/commits/9344bbe9dae30bca67943476695306fe5d7ab987)

The default (``30``) was running too slow in debug builds.

src/tools/mesh_tester.cpp
kennyweiss: Thanks @gzagaris.

``counter`` needs to be at a unique number across all threads since it allows each thread to add its data to a shared array, so I don't think a ``ReduceSum`` will work here.

What would be the benefit of a ``StackArray``, other than not needing to ``deallocate``?

src/axom/primal/geometry/NumericArray.hpp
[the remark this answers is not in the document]
cyrush: @kennyweiss -- I don't see how auto helps, the type is being inferred from a literal -- not from the value used in the for boolean clause?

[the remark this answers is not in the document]
kennyweiss: Thanks @cyrush 

I am not suggesting that it helps or hurts in this case (where the line above explicitly defines ``nvals`` as a ``const int``).


FWIW, I made this change not @bmhan12, and don't particularly care if it lists ``auto`` or ``int`` as the type of the loop bounds here.

kennyweiss:
I've created https://github.com/LLNL/axom/issues/178 to track some updates to ``slic`` macros to support calls within GPU/cuda/raja scopes.  Please check it out and add anything that I've missed.

kennyweiss:
@gzagaris -- are there any remaining issues with the current PR preventing this from merging?

src/axom/slic/interface/slic_macros.hpp
kennyweiss: Tracked in #178

src/axom/primal/geometry/NumericArray.hpp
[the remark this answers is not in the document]
gzagaris: @kennyweiss I understand your viewpoint. However, there are many things that are idiomatic `C++11`, but that doesn't mean we should use them all. 

We have had discussions about this, see the confluence page here: https://lc.llnl.gov/confluence/pages/viewpage.action?pageId=403275802

My concern is that sprinkling `auto` everywhere will make it harder for newer developers to understand when it is OK to use it or not to use it. So, we've agreed to be more conservative about this and only use it when it truly adds value to the code and makes the  code easier to read and understand.

The type is known and there is no reason not to use the type explicitly.

Please, revert this change to the code that @bmhan12 before.

src/axom/slic/interface/slic_macros.hpp
gzagaris: Thanks for adding the issue.

src/tools/mesh_tester.cpp
```diff
@@ -79,15 +114,44 @@ struct Input
   void fixOutfilePath();
 };
 
+const std::set<RuntimePolicy> Input::s_validPolicies({
```
gzagaris: I thought I commented on this before. 

Is the whole purpose of storing the policies in a set here to validate the command line input?

Couldn't you get by with just the enum:
```
enum RuntimePolicy { 
    seq = 0,
#ifdef AXOM_USE_RAJA 
    raja_seq = 1, 
#ifdef AXOM_USE_OPENMP
    raja_omp = 2,
#endif
#ifdef AXOM_USE_CUDA
    raja_cuda = 3,
#endif
#endif /* AXOM_USE_RAJA */
    NUM_RUNTIME_POLICIES
};
```
Then you can validate the input based on the `NUM_RUNTIME_POLICIES` ? Would that be sufficient?

src/tools/mesh_tester.cpp
```diff
@@ -250,6 +328,97 @@ std::vector< std::pair<int, int> > naiveIntersectionAlgorithm(
...
+      {
+        auto idx = RAJA::atomicAdd<ATOMIC_POL>(counter, 2);
+        intersections[idx] = row;
+        intersections[idx+1] = col;
+      }
+    }
+  });
+
+  // Initialize pairs of clashes
+  for (auto i = 0 ; i < numIntersect.get() * 2 ; i += 2)
```
gzagaris: I would have hoisted out the `int N = numIntersect.get()*2;` into a variable. The compiler should do that and not do the multiplication at every iteration, but, I tend to do that anyways...

src/tools/mesh_tester.cpp
kennyweiss: No -- it actually validates the options at runtime. 

E.g. option '3' is only valid when the code is configured with ``raja`` and ``cuda`` and if you try to use that option in other builds, it throws an error with a nice message about why that option wasn't valid.

gzagaris: You would be able to do runtime validation with just the `enum` by just adding the `NUM_RUNTIME_POLICIES` entry right?

src/tools/mesh_tester.cpp
```diff
@@ -250,6 +328,97 @@ std::vector< std::pair<int, int> > naiveIntersectionAlgorithm(
...
+  counter[0] = 0;
+
+  // RAJA loop to populate with intersections
+  RAJA::kernel<KERNEL_POL>( RAJA::make_tuple(col_range, row_range),
+    AXOM_LAMBDA(int col, int row) {
+    if (row > col)
+    {
+      if (checkTT (tris[row], tris[col]))
+      {
+        auto idx = RAJA::atomicAdd<ATOMIC_POL>(counter, 2);
```
gzagaris: It seems that you could store a `counts` and `offsets` array during the kernel that computes the number of intersections and get rid of the atomic add here. Thoughts?

gzagaris: Or rather do `scan` after you calculate the `counts` array to compute the `offsets`....

src/tools/mesh_tester.cpp
kennyweiss: The runtime validation is handled for us by ``cli11`` via the ``std::set`` and automatically produces good error messages for us with very little code on our side.

What would be a benefit of having a count of the number of runtime policies?

src/tools/mesh_tester.cpp
kennyweiss: It is not clear that that would help in this case.

For a triangle mesh with ``N`` triangles, that would cost ``2N`` ints.
For most triangle meshes, we'd expect to have very few intersections.

src/tools/mesh_tester.cpp
gzagaris: oh, OK, nevermind, I didn't realize that the `cli11` required an `std::set`. 

I tend to use the count to check manually if the supplied argument is within range. Doesn't `cli11` support a valid range without having to specify a `set`?

src/tools/mesh_tester.cpp
gzagaris: It would remove the need for an atomic. 

There is already a kernel launch prior to this to compute the total count, so you might as well fill in the `counts` array and alleviate the need for an atomic. I'd expect it to be faster.

bmhan12: > Or rather do a scan after you calculate a counts array and compute the offsets....

@gzagaris I do not think that I am following correctly and am missing something. The `offsets` would give me the index to write one pair of intersecting triangles. If I had multiple intersections for a given triangle, wouldn't I still need an atomic  to increment the offset for intersecting pairs after the first?

kennyweiss: Since the current version is working, I think we should defer this sort of optimization to a separate pass.

src/tools/mesh_tester.cpp
kennyweiss: @gzagaris -- ``cli11`` would support a range-based validation, but the set-based one is better in this case, since it provides automatic error messages for invalid options that are within the range.

src/tools/mesh_tester.cpp
gzagaris: > @gzagaris I do not think that I am following correctly and am missing something. The `offsets` would give me the index to write one pair of intersecting triangles. If I had multiple intersections for a given triangle, wouldn't I still need an atomic to increment the offset for intersecting pairs after the first?

@bmhan12 , yes you are correct. You would need an atomic as the code currently is. 

I thought that in your first kernel you were excluding candidate triangles based on a bounding box intersection and that in your second kernel you were processing an abbreviated list of candidate triangles serially within each thread, in which case, you could just keep a integer counter for each thread....

src/tools/mesh_tester.cpp
gzagaris: Thanks @kennyweiss good to know....

PR #157, opened 2020-01-31
TPL Updates

gzagaris:
# Summary

- This PR updates the following TPLs:
   * RAJA is updated to v0.11.0
   * Umpire is updated to v2.1.0, and
   * Conduit is updated to v0.5.1

The host configs for `rzgenie` and `rzansel` are updated accordingly.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -22,7 +22,9 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 ### Deprecated
 
 ### Changed
-- Updated Umpire TPL to v2.0.0, which, natively handles zero byte re-allocations consistently. Workaround
+- Updated Conduit to v0.5.1
+- Updated RAJA to v0.11.0
+- Updated Umpire to v2.1.0, which, natively handles zero byte re-allocations consistently. Workaround
   for older releases of Umpire is in place for backwards compatibility. 
 - Updated BLT to develop (f0ab9a4) as of Jan 15, 2020
 - Set CUDA_SEPARABLE_COMPILATION globally instead of just in a few components.
--- a/src/axom/core/memory_management.hpp
+++ b/src/axom/core/memory_management.hpp
@@ -231,9 +231,10 @@ inline T* reallocate( T* pointer, std::size_t n ) noexcept
 
   pointer = static_cast< T* >( rm.reallocate( pointer, numbytes ) );
 
-#elif defined(AXOM_USE_UMPIRE) && (UMPIRE_VERSION_MAJOR >= 2)
+#elif defined(AXOM_USE_UMPIRE) && (UMPIRE_VERSION_MAJOR >= 2) && \
+      (UMPIRE_VERSION_MINOR >= 1)
 
-  // Umpire 2.0.0 and above handles reallocate(0) natively
+  // Umpire 2.1.0 and above handles reallocate(0) natively
   umpire::ResourceManager& rm = umpire::ResourceManager::getInstance();
   pointer = static_cast< T* >( rm.reallocate( pointer, numbytes ) );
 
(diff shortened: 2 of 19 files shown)


gzagaris:
> Specifically, I think we need the `reallocate(0)` workaround for umpire if it's version is less than `2.1` (rather than just `2.0`), since we were seeing failures in our nvcc builds before 

Yes -- we should definitely not call `reallocate(0)` when using `umpire-2.0` since there was a bug on this. I tested the workaround on `rzansel` and it works as expected.

src/axom/sidre/tests/sidre_buffer.cpp
```diff
@@ -542,11 +545,24 @@ TEST_P(UmpireTest, reallocate_zero)
...
+#ifdef UMPIRE_ENABLE_PINNED
   , axom::getResourceAllocatorID( umpire::resource::Pinned )
+#endif
+
+#ifdef UMPIRE_ENABLE_DEVICE
   , axom::getResourceAllocatorID( umpire::resource::Device )
+#endif
+
+#ifdef UMPIRE_ENABLE_CONST
   , axom::getResourceAllocatorID( umpire::resource::Constant )
```
agcapps: At core/tests/core_memory_management.cpp near (new) line 400, there is a comment with some commented-out code saying "Umpire doesn't allow reallocation of Constant memory."  Some of the tests in this file and in sidre_view.cpp perform reallocation.  Will those tests need to be guarded against reallocation with a Constant memory allocator?

src/axom/sidre/tests/sidre_view.cpp
```diff
@@ -1824,12 +1826,27 @@ TEST_P(UmpireTest, reallocate_zero)
...
+#ifdef UMPIRE_ENABLE_PINNED 
   , axom::getResourceAllocatorID( umpire::resource::Pinned )
+#endif
+
+#ifdef UMPIRE_ENABLE_DEVICE
   , axom::getResourceAllocatorID( umpire::resource::Device )
+#endif
+
+#ifdef UMPIRE_ENABLE_CONST
   , axom::getResourceAllocatorID( umpire::resource::Constant )
```
agcapps: See my comment at sidre_buffer.cpp (new) line 558.

RELEASE-NOTES.md
```diff
@@ -22,7 +22,9 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 ### Deprecated
 
 ### Changed
-- Updated Umpire TPL to v2.0.0, which, natively handles zero byte re-allocations consistently. Workaround
+- Updated Conduit to v0.5.1
+- Updated Axom to v0.10.1
```
ltaylor16: I think this should be RAJA- 0.11.0

gzagaris: Yup, definitely, not sure how that happened...

src/axom/sidre/tests/sidre_buffer.cpp
gzagaris: Yes and the original implementation was taking that in to account by:
```
#ifdef AXOM_USE_CUDA
 if (allocID == axom::getResourceAllocatorID( umpire::resource::Constant ) )
  {
    return;
  }
#endif
```

The more precise `ifdef` guard would be:
```
#if defined(AXOM_USE_CUDA) && defined(UMPIRE_ENABLE_CONST)
  if (allocID == axom::getResourceAllocatorID( umpire::resource::Constant ) )
  {
    return;
  }
#endif
```

I made a fix. Good catch!

src/axom/core/memory_management.hpp
```diff
@@ -231,9 +231,10 @@ inline T* reallocate( T* pointer, std::size_t n ) noexcept
 
   pointer = static_cast< T* >( rm.reallocate( pointer, numbytes ) );
 
-#elif defined(AXOM_USE_UMPIRE) && (UMPIRE_VERSION_MAJOR >= 2)
+#elif defined(AXOM_USE_UMPIRE) && (UMPIRE_VERSION_MAJOR >= 2) && \
+      (UMPIRE_VERSION_MINOR >= 1)
```
kennyweiss: Thanks @gzagaris

PR #159, opened 2020-01-31
Release Candidate for v0.3.3 

gzagaris:
# Summary

Release candidate for v0.3.3


src/axom/primal/examples/primal_introduction.cpp
```diff
@@ -157,11 +157,11 @@ PolygonType showClip()
     "{0}"
     "// axes\n"
     "draw(O -- 1.7X, arrow=Arrow3(DefaultHead2), "
-      "L=Label(\"$x$\", position=EndPoint, align=W));\n"
+    "L=Label(\"$x$\", position=EndPoint, align=W));\n"
```
kennyweiss: I've been leaving this one alone since we don't currently have an automatic way of regenerating the figures and I wasn't sure how sensitive asymptote is to white space changes, if at all.

@agcapps ?

RELEASE
```diff
@@ -1,6 +1,6 @@
 *******************************************************************************
 
-Axom: ................................, version 0.3.2
+Axom: ................................, version 0.3.3
```
kennyweiss: I'd vote for bumping us to Axom ``v0.4.0`` after all the library changes and associated bugfixes.
(It's been more than 1 year since ``v0.3.0``).

src/axom/primal/examples/primal_introduction.cpp
agcapps: well, these are all C++ strings that get concatenated.  The whitespace changes should be fine.  They already got broken up because of a strict max-80-column formatting rule...

kennyweiss: Thanks @agcapps

PR #165, opened 2020-02-11
Set boolean option so job fails when tests fail

keithhealy:
Setting the option 'failTaskOnFailedTests' of the PublishTestResults task to true will cause that task to be marked as failed when any tests fail.  This addresses the confusion caused when all the Azure build and test tasks succeed (and show as green) even when one or more tests has failed.

- This PR is a bugfix
- It does the following:
  - Fails the PublishTestResults task when test failures are detected (issue 162)

PR #167, opened 2020-02-14
Attempt to fix CI tests of sidre_lulesh by changing number of ctest threads

kennyweiss:
# Summary

Our azure job is failing the ``sidre_lulesh`` tests for ``gcc`` configurations.
Those tests use ``MPI`` and ``OpenMP``.

This is an attempt to fix it by reducing the number of threads.

### DIFF
--- a/scripts/azure-pipelines/linux-build_and_test.sh
+++ b/scripts/azure-pipelines/linux-build_and_test.sh
@@ -33,7 +33,7 @@ if [[ "$DO_BUILD" == "yes" ]] ; then
         or_die make -j 10 VERBOSE=1
     fi
     if [[ "${DO_TEST}" == "yes" ]] ; then
-        ctest -j8 -T test --output-on-failure -V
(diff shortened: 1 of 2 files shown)


kennyweiss:
Looks like I finally got it!

This fixes the ``sidre_lulesh`` CI failures by setting the number of MPI ranks to 1.
I think the issue is somehow related to the fact that Azure only gives us 10 threads, although I am not sure why the slam_lulesh test works with 8 ranks and 4 omp tasks.

After we merge this, we can merge @keithhealy's PR ( #165 ).

I will squash merge.

gzagaris:
> Looks like I finally got it!
> 
> This fixes the `sidre_lulesh` CI failures by setting the number of MPI ranks to 1.
> I think the issue is somehow related to the fact that Azure only gives us 10 threads, although I am not sure why the slam_lulesh test works with 8 ranks and 4 omp tasks.
> 

Do we know how many physical cores we're getting with Azure? I suspect there is a lot more time-haring of actual physical resources hence why things run slow with both MPI and OpenMP.

Issue #126, opened 2019-11-22
New spatial query to find elements intersecting a Ray


kennyweiss:
A user has requested a new spatial query to find a set of candidate elements from a mesh whose [``BoundingBox``](https://github.com/LLNL/axom/blob/develop/src/axom/primal/geometry/BoundingBox.hpp)es intersect a given [``Ray``](https://github.com/LLNL/axom/blob/develop/src/axom/primal/geometry/Ray.hpp).

#### Request Details
* The input is a mesh containing a custom type of high order elements (and do not need us to implement this).
* They are able to provide the bounding box of each element
* They are also able to compute the exact intersection of a ray with one of their elements
* They are looking for us to provide a simple API that allows them to 
    * build a spatial index over a collection of element bounding boxes
    * pose a query to the index that finds the indices of the elements whose bounding boxes intersect a given ray.
* For the purposes of this query, we can assume that the entire mesh is resident on a single rank. I.e. we do not need to worry about MPI communications.

#### Available functionality
* The ``spin`` component provides several spatial indices that can be built over a collection of axis-aligned bounding boxes. E.g. [BVH tree](https://github.com/LLNL/axom/blob/develop/src/axom/spin/BVHTree.hpp), [ImplicitGrid](https://github.com/LLNL/axom/blob/develop/src/axom/spin/ImplicitGrid.hpp), ...
* ``primal`` already has an [algorithm](url) to determine if a ``Ray`` intersects an axis-aligned ``BoundingBox``
  https://github.com/LLNL/axom/blob/975fc516e6b4dfd4bb82f9c8495db90a7150f8fc/src/axom/primal/operators/intersect.hpp#L77-L88

#### Missing functionality
* We **do not** currently have the capability to traverse a spatial index to find the set of candidate elements that intersect a ``Ray``. This will have to be implemented to satisfy this request.

gzagaris:
@kennyweiss, since we do not get have access to the "custom high order mesh", it would be cleaner if they just use our spatial acceleration data-structures to prune their search, which takes the supplied bounding boxes, which they will be computing.

As I mentioned earlier, I am planning to do this for the BVH, which is also needed for the probe package.

kennyweiss:
Thanks @gzagaris -- that is exactly what I had in mind.  
I think the BVH will be a great fit for their needs.

So, are you claiming this ticket?

PR #177, opened 2020-03-05
Initial Support for Ray Queries using the BVH

gzagaris:
# Summary

This PR adds initial support for ray queries using the BVH and associated unit tests.


src/axom/spin/tests/spin_bvh.cpp
```diff
@@ -801,23 +1350,43 @@ AXOM_CUDA_TEST( spin_bvh, contruct3D_cuda )
...
+
+//------------------------------------------------------------------------------
+AXOM_CUDA_TEST( spin_bvh, find_points_3d_cuda )
 {
   constexpr int BLOCK_SIZE = 256;
   using exec  = axom::CUDA_EXEC< BLOCK_SIZE >;
 
-  check_find3d< exec, float >( );
-  check_find3d< exec, double >( );
+  check_find_points3d< exec, float >( );
```
gzagaris: @mclarsen, @rhornung67 for some reason my tests are failing on the GPU when using  single precision, i.e.,`float`. They are passing with `double`.  

Any ideas? I saw some discussion about the RAJA scans no working properly with single precision. Could this be related to that?

rhornung67: Are you using RAJA scan operations?  These may not work properly in parallel (CUDA or OpenMP) due to low precision and/or non-associativity of plus operations.

gzagaris: Yes, we are using scan operations. They do not work properly with single precision, but, they work properly when using double precision. Correct?

rhornung67: Correct. Not sure how to fix this. We are using CUB internally in the scans.

gzagaris: Thanks @rhornung67 , I'll just remove the tests with single precision -- I guess we shouldn't be thoroughly testing things :)

rhornung67: Well, I just learned about this issue recently as I was beefing up the RAJA scan tests. So at least we know about it now.

gzagaris: Yeah, I saw Will's comment. That's what prompted me to see if the tests are passing with just double precision. 

@mclarsen you may in particular want to take note of that since I think you are intrinsically using single precision in your code as I recall...

rhornung67: @mclarsen do you have a solution for parallel scans that works for float data?

gzagaris:
Alas, I haven't been able to figure out why this is failing on `linux_gcc6`. All the other jobs are passing including manual tests with various compilers on LC....any ideas?

gzagaris:
I booted the docker image and I was able to reproduce the issue there (Thanks to @davidbeckingsale, @cyrush and @keithhealy for their help with docker). It seems pretty random, the tests sometimes pass and sometimes they fail.

One of the caveats in our docker image is that `RAJA` was built manually outside `uberenv`. 

I'll try making a fresh build of `RAJA` to see if that fixes things...

gzagaris:
I can reproduce the issue with a fresh build of RAJA as well. Are there any known issues with docker and the OpenMP runtime etc. that could lead to issues like this?

cyrush:
@gzagaris docker is just another path that is likely to tease out issues. We once found something b/c a docker instance was run with an odd number of threads vs more usually event number. My guess, it could be some quirk, but where there is smoke fire is likely.  It would be good to dig in and try to understand what is going wrong.

rhornung67:
@gzagaris I haven't seen this sort of thing before. FWIW, we have docker images for our RAJA tutorials and they work fine.

gzagaris:
> @gzagaris docker is just another path that is likely to tease out issues. We once found something b/c a docker instance was run with an odd number of threads vs more usually event number. My guess, it could be some quirk, but where there is smoke fire is likely. It would be good to dig in and try to understand what is going wrong.

Yup, @mclarsen mentioned that issue as well. `OMP_NUM_THREADS=4` in this case....

gzagaris:
> @gzagaris I haven't seen this sort of thing before. FWIW, we have docker images for our RAJA tutorials and they work fine.

I ran the RAJA tests and they are all passing within my docker session as well.

gzagaris:
I am also seeing the same random failure when running with a single thread. So, I don't think the error is due to a race condition.

There's something really strange going on. I am also getting internal compiler errors whenever I do `make -j`. Is that a known issue or could we be dealing with with a wonky compiler? I am using the `gcc-6` image. 

Thoughts?

mclarsen:
Internal compiler errors inside a container in my experience mean that the vm is running out of memory. Which makes sense if it only happens on a `make -j`. What is failing with one thread?

gzagaris:
> Internal compiler errors inside a container in my experience mean that the vm is running out of memory. Which makes sense if it only happens on a `make -j`. What is failing with one thread?

Thanks @mclarsen -- that's good to know.

The `findRays` test is failing even when I run it with a single thread. Maybe I need to relax the tolerances, but, that doesn't explain why the random behavior we are seeing, i.e., some runs fail and others are passing...

gzagaris:
> Normally, you would want to scale the bounding boxes conservatively, but since I don't think you control this, you may want to consider setting the the actual epsilon to the `aabb.magnitute() * epsilon`. This will take into account larger ranges of spatial bounds. For example, if the bounds were `1e-10` or `1e200` this will still have a chance of working.

We do have a to set the scale factor. There is a default scale factor of the bounding boxes is set to 1.001. I am still a bit mystified why there are sporadic failures. If it was a tolerance issue, all runs would fail the same way, right? Am I missing something?

mclarsen:
What is the failure? I don't think I have permission to view it.

gzagaris:
> What is the failure? I don't think I have permission to view it.

Specifically, the `find_rays_2d_omp` and `find_rays_3d_omp` tests, which both use `openmp` execution policy, randomly fail (i.e., there is no ray hit).

I am now also able to reproduce this with `intel19` on `rzgenie` as well and I can reproduce even if I set the number of threads to `1`...

The test is consistently passing with Clang and NVCC with OpenMP and CUDA respectively.

I am going try to zoom in on `rzgenie` and see which abstraction layer is falling apart. Please, let me know if you have any further thoughts or suggestions.

gzagaris:
I am also adding to the ability to set tolerances at the level of the calling application which is useful regardless

gzagaris:
Some more info:

When the test fails, the BVH traversal does not happen, the code falls apart here:
https://github.com/LLNL/axom/blob/1a147111b7915790a6539541258bc9a10158d7bf/src/axom/spin/internal/linear_bvh/bvh_traverse.hpp#L92-L100

The `L` and `R` above are functors (lambdas) that define the traversal predicates. In this case, the ray-box intersection checks. These are evaluated at the root node, i.e., `current_node=0`. Whenever the tests fails, both `in_left` and `in_right` evaluate to `false` and the code returns immediately from the BVH traversal routine.

Again, this is pretty random, but, I suspect that the way these lambda captures are defined and used, I may be venturing into some uncharted territory of the C++ 11 standard...

gzagaris:
OK -- I totally understand what was going on with this now.

There was a typo (Ugh!) in the `QueryAccessor` see fix in 29b1b27.

Due to this typo, the query ray components were not being properly initialized. The Clang compiler was initializing the entries to zero, which happened to make my test happy, but, `Intel` and `GCC` were not initializing, consequently the ray data was incorrect random garbage, which also explains the random behavior we observed.

I added additional unit tests for this in ef8383d.

Sigh, this should be ready to go.

rhornung67:
D'oh!!

gzagaris:
> D'oh!!

I am going to go sit in a corner and cry....at least, I learned how to use Docker

rhornung67:
Hey, at least you found the issue. 🥇

cyrush:
good hunting! Better to solve the mystery than face it again in the future.

mclarsen:
download
What were you using the scan on floating point types for?

gzagaris:
> What were you using the scan on floating point types for?

You are correct. We only use the parallel scans on the `counts` array (integer data) to compute the `offsets`. It was the same bug all along that was tormenting me....

src/axom/spin/BVH.hpp
```diff
@@ -230,11 +253,53 @@ class BVH
...
-             const FloatType* y,
-             const FloatType* z=nullptr ) const;
+  void findPoints( IndexType* offsets, IndexType* counts,
+                   IndexType*& candidates, IndexType numPts,
+                   const FloatType* x,
+                   const FloatType* y,
+                   const FloatType* z=nullptr ) const;
+
+  /*!
+   * \brief Finds the candidate bins that intersect the given rays.
```
kennyweiss: Just to clarify -- this function finds **all** boxes that could potentially intersect the ray, not just the candidates for the first intersection, right?

src/axom/spin/internal/linear_bvh/BVH_impl.hpp
```diff
@@ -6,13 +6,13 @@
 #ifndef AXOM_SPIN_BVH_IMPL_HPP_
 #define AXOM_SPIN_BVH_IMPL_HPP_
 
-#include "axom/core/memory_management.hpp"          // for memory functions
-#include "axom/core/Types.hpp"                      // for fixed bitwidth types
-
-#include "axom/core/execution/for_all.hpp"          // for generic for_all()
+#include "axom/core/Types.hpp"                          // fixed bitwidth types
```
kennyweiss: @gzagaris -- what is the relationship between ``spin/BVH`` and ``spin/internal/linear_bvh`` at this point?
Is the latter the implementation for the former?

src/axom/spin/internal/linear_bvh/BVH_impl.hpp
```diff
@@ -171,6 +258,7 @@ IndexType bvh_get_counts(
 template< int NDIMS, typename ExecSpace, typename FloatType >
 BVH< NDIMS, ExecSpace, FloatType >::BVH( const FloatType* boxes,
                                          IndexType numItems ) :
+  m_Tolernace( floating_point_limits<FloatType>::epsilon() ),
```
kennyweiss: Typo: ``m_Tolernace`` -> ``m_Tolerance``

src/axom/spin/BVH.hpp
```diff
@@ -230,11 +253,53 @@ class BVH
    * \pre y != nullptr if dimension==2 || dimension==3
    * \pre z != nullptr if dimension==3
    */
-  void find( IndexType* offsets, IndexType* counts,
-             IndexType*& candidates, IndexType numPts,
-             const FloatType* x,
-             const FloatType* y,
-             const FloatType* z=nullptr ) const;
+  void findPoints( IndexType* offsets, IndexType* counts,
```
kennyweiss: Suggestion:  Would ``findCandidatesForPoints()`` and ``findCandidatesForRays()`` be clearer?

src/axom/spin/internal/linear_bvh/QueryAccessor.hpp
```diff
@@ -68,6 +68,40 @@ class QueryAccessor
...
+   * \param [in] x0 array of the x-coordinate source point locations.
+   * \param [in] nx array of x-components of the ray normal directions.
+   * \param [in] y0 array of the y-coordinate source point locations.
+   * \param [in] ny array of y-components of the ray normal directions.
+   * \param [in] z0 array of of the z-coorindate source point locations.
+   * \param [in] nz array of z-components of the ray normal directions.
+   *
+   * \tparam RayType the data-structure to hold a ray.
+   *
+   * \note The RayType must overload the `[]` operator for accessing the
```
kennyweiss: Does ``primal::Ray`` do this? It doesn't look like you are modifying ``primal::Ray`` in this PR.

src/axom/spin/internal/linear_bvh/BVH_impl.hpp
```diff
@@ -55,6 +59,9 @@ using vec4_t = internal::linear_bvh::Vec< FloatType, 4 >;
 template < typename FloatType, int NDIMS >
 using point_t = internal::linear_bvh::Vec< FloatType, NDIMS >;
 
+template < typename FloatType, int NDIMS >
+using ray_t = internal::linear_bvh::Vec< FloatType, NDIMS*2 >;
```
kennyweiss: Is this the ``RayType`` for ``getRay()`` below?

src/axom/spin/internal/linear_bvh/TraversalPredicates.hpp
```diff
@@ -127,6 +172,54 @@ class TraversalPredicates< DIMENSION_2, FloatType >
...
+      FloatType TOL=numerics::floating_point_limits<FloatType>::epsilon()
+      ) noexcept
+  {
+    const FloatType& x0 = r[ 0 ];
+    const FloatType& y0 = r[ 1 ];
+    const FloatType& nx = r[ 2 ];
+    const FloatType& ny = r[ 3 ];
+
+    // extract left bin, see BVHData.hpp for the internal BVH layout
+    const FloatType& xmin = s1[ 0 ];
```
kennyweiss: Perhaps a utility function in ``BVHData.hpp`` could be used to hide the implementation details?

src/axom/spin/internal/linear_bvh/TraversalPredicates.hpp
```diff
@@ -179,6 +274,64 @@ class TraversalPredicates< DIMENSION_3, FloatType >
...
+  {
+    const FloatType& x0 = r[ 0 ];
+    const FloatType& y0 = r[ 1 ];
+    const FloatType& z0 = r[ 2 ];
+    const FloatType& nx = r[ 3 ];
+    const FloatType& ny = r[ 4 ];
+    const FloatType& nz = r[ 5 ];
+
+    // extract left bin, see BVHData.hpp for the internal BVH layout
+    const FloatType& xmin = s1[ 0 ];
```
kennyweiss: Similarly to my comment in the 2D case -- 
Perhaps a utility function in ``BVHData.hpp`` could be used to hide the implementation details?

src/axom/spin/tests/spin_bvh.cpp
```diff
@@ -836,11 +1470,10 @@ AXOM_CUDA_TEST( spin_bvh, single_box3d_cuda )
   constexpr int BLOCK_SIZE = 256;
   using exec  = axom::CUDA_EXEC< BLOCK_SIZE >;
 
-  check_single_box3d< exec, float >( );
   check_single_box3d< exec, double >( );
```
kennyweiss: Does the order of ``float`` and ``double`` matter?

RELEASE-NOTES.md
```diff
@@ -10,12 +10,20 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 ## [Unreleased] - Release date yyyy-mm-dd
 
 ### Added
+- Added a portable floating_point_limits traits class, to return min(), max(), lowest() 
+  and epsilon() of a `float` or `double` type. The functionality is equivalent to that provided by 
+  std::numeric_limits, but, the code is host/device decorated accordinlgy such that it
```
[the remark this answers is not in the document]
gzagaris: Done. Thanks.

src/axom/core/tests/numerics_floating_point_limits.cpp
```diff
@@ -0,0 +1,65 @@
...
+  const T STD_LOWEST = std::numeric_limits< T >::lowest();
+  const T STD_MIN    = std::numeric_limits< T >::min();
+  const T STD_MAX    = std::numeric_limits< T >::max();
+
+  EXPECT_TRUE( axom::utilities::isNearlyEqual( LOWEST, STD_LOWEST, EPS ) );
+  EXPECT_TRUE( axom::utilities::isNearlyEqual( MIN, STD_MIN, EPS ) );
+  EXPECT_TRUE( axom::utilities::isNearlyEqual( MAX, STD_MAX, EPS ) );
+  EXPECT_TRUE( axom::utilities::isNearlyEqual( EPS, STD_EPS, EPS ) );
+}
+
```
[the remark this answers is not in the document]
gzagaris: We cannot make GTEST calls inside the device. The purpose of this test is to ensure that our `traits` class is consistent with `std::numeric_limits`

src/axom/spin/internal/linear_bvh/QueryAccessor.hpp
gzagaris: I opted not to use `primal::Ray` in this PR since I have something that already works and is sufficient for my use.

src/axom/spin/internal/linear_bvh/BVH_impl.hpp
gzagaris: Yes, that's what I use for `RayType`.

src/axom/spin/internal/linear_bvh/TraversalPredicates.hpp
gzagaris: Yes, I have been thinking about something like that. I'll mull it over some more and add that in a future PR.

src/axom/spin/internal/linear_bvh/TraversalPredicates.hpp
gzagaris: Ditto.

src/axom/spin/tests/spin_bvh.cpp
gzagaris: No, the order doesn't matter.

src/axom/primal/tests/primal_ray_intersect.cpp
```diff
@@ -0,0 +1,399 @@
...
+    for ( int i=0 ; i < N ; ++i )
+    {
+      double tmin = 0.0;
+      double tmax = 0.0;
+
+      const double sx = x[i];
+      const double sy = y0;
+      const double nx = n0[0];
+      const double ny = n0[1];
+      EXPECT_TRUE( intersect_ray(sx,nx,sy,ny,TEST_BOX2D,tmin,tmax) );
```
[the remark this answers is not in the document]
gzagaris: I agree the macro is not great, but, it did simplify writing these tests. 

I am not worried users won't under how to use this method in their application. I think the doxygen docs do a good job in that and I doubt anyone would need to look any further than that.

kennyweiss:
Forgot to ask earlier: Does this resolve https://github.com/LLNL/axom/issues/126 ?  It seems like it does.

gzagaris:
> Forgot to ask earlier: Does this resolve #126 ? It seems like it does.

Yes, it does. Do they need a C-style/Fortran interface or could that be done on the application side, if needed?

kennyweiss:
> Do they need a C-Style/Fortran interface?

Thanks @gzagaris.
I don't remember.  I think their code was in Fortran. When I wrote up the ticket, I indicated that they would need an interface like ``quest_inout``, so perhaps they will. 
I suppose we can reach out to them...

Let's keep the ticket open until we get a definite answer.

Thoughts @rhornung67 ?

gzagaris:
> > Do they need a C-Style/Fortran interface?
> 
> Thanks @gzagaris.
> I don't remember. I think their code was in Fortran. When I wrote up the ticket, I indicated that they would need an interface like `quest_inout`, so perhaps they will.
> I suppose we can reach out to them...
> 
> Let's keep the ticket open until we get a definite answer.
> 

Sounds good. I think that's easy enough to add later or even implement on the application side. I think the latter would be my preference.

rhornung67:
@gzagaris and @kennyweiss on the Fortran interface question, let's go with this and keep the ticket open in case the need arises.

PR #180, opened 2020-03-09
Don't use gcc's non-standandard abi::__cxa_demangle with clang's libc++

### LINKED ISSUE #179: cxxabi is not found when compling with clang++ -stdlib=libc++
In the file `axom/slic/internal/stacktrace.cpp`, there is a include `#include <cxxabi.h>` which is a GCC specific extension.  It is already protected against for Microsoft compilers.  It also should be protected against for using Clang's libc++.

brunner6:
# Summary

- This PR is a bugfix
- It enables building with a pure clang compile using ``libc++``
  - The calls to the ``gcc``'s nonstandard ``abi::__cxa_demangle`` have been protected against.
  - Fixes #179

### DIFF
--- a/src/axom/slic/internal/stacktrace.cpp
+++ b/src/axom/slic/internal/stacktrace.cpp
@@ -13,7 +13,10 @@
   #include <DbgHelp.h>
 #else
   #include <execinfo.h> // for backtrace()
-  #include <cxxabi.h>   // for abi::__cxa_demangle
+  #include <ciso646>
+  #if !defined(_LIBCPP_VERSION)
+     #include <cxxabi.h>   // for abi::__cxa_demangle
+  #endif
 #endif
 
 constexpr int MAX_FRAMES = 25;
@@ -122,9 +125,13 @@ std::string demangle( char* backtraceString, int frame )
     *returnOffset = 0;
     ++returnOffset;
 
-    int status;
+    int status = false;
+#if !defined(_LIBCPP_VERSION)
     char* realName = abi::__cxa_demangle( mangledName, nullptr, nullptr,
                                           &status );
+#else
+    char* realName = mangledName;
+#endif
 
     // if demangling is successful, output the demangled function name
     if (status == 0)
--- a/src/axom/slic/internal/stacktrace.cpp
+++ b/src/axom/slic/internal/stacktrace.cpp
@@ -144,7 +144,9 @@ std::string demangle( char* backtraceString, int frame )
       oss << "Frame " << frame << ": " << mangledName << std::endl;
     }
 
+#if !defined(_LIBCPP_VERSION)
     free(realName);
+#endif
   }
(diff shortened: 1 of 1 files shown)


src/axom/slic/internal/stacktrace.cpp
```diff
@@ -13,7 +13,10 @@
   #include <DbgHelp.h>
 #else
   #include <execinfo.h> // for backtrace()
-  #include <cxxabi.h>   // for abi::__cxa_demangle
+  #include <ciso646>
+  #if !defined(_LIBCPP_VERSION)
```
kennyweiss: Thanks @brunner6.

Looks like there is an issue with our mac CI test.
```
Start  34: slic_logging

34: Test command: /Users/runner/runners/2.165.1/work/1/s/build/examples/slic_logging_ex
34: Test timeout computed to be: 1500
34: slic_logging_ex(3313,0x7fffb6314380) malloc: *** error for object 0x7ff8bf402ce0: pointer being freed was not allocated
34: *** set a breakpoint in malloc_error_break to debug
34: Mon Mar  9 17:27:44 2020
```
See: https://dev.azure.com/axom/axom/_build/results?buildId=993&view=logs&j=8dab7c2a-824f-53c4-ff36-17cdfa82e5d3&t=d13b7ead-2e4c-51d6-8176-f474971cfc15&l=6790

I'm guessing it has to do with the ``free(realName)`` in line 147 when ``status==false``

brunner6:
OK, I can't make sense of the logs, beyond that theytell me OSX failed, and given the change, that might be legitimate.

src/axom/slic/internal/stacktrace.cpp
```diff
@@ -122,9 +125,13 @@ std::string demangle( char* backtraceString, int frame )
...
     ++returnOffset;
 
-    int status;
+    int status = false;
+#if !defined(_LIBCPP_VERSION)
     char* realName = abi::__cxa_demangle( mangledName, nullptr, nullptr,
                                           &status );
+#else
+    char* realName = mangledName;
+#endif
```
kennyweiss: One obvious place to start is:

```diff
// at line 147
-     free(realName);
+ #if !defined(_LIBCPP_VERSION)
+     free(realName);
+ #endif
```

i.e. only free the string if we've allocated it (via ``_cxa_demangle``)

kennyweiss:
Thanks @brunner6.  

The CI failure the second time was really weird -- at some point between your two commits, Azure seemed to have decided it will no longer run the ``macOS-10.13`` plan -- the exact plan that was failing for us.

I updated it to ``macOS-10.14`` in #182. Could you please merge that into your branch and try again?  Alternatively, I'd be happy to do this for you, if you'd like.

kennyweiss:
Thanks @brunner6!

PR #186, opened 2020-03-19
Fix minor warnings on windows

gzagaris:
# Summary

Fix various warnings on Windows due to unused variables
and type conversion from double to integer.


gzagaris:
There are a few more warnings on windows that showed up ... I'm going to clean them up in this PR and let you know when this is ready to go

PR #187, opened 2020-03-20
Small build system changes

white238:
* Update to develop BLT
* Guard code checks and global options against including Axom as a subdirectory from another project

This is mainly because I want to play with clang-format (which i just added to BLT) in a few projects.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -43,10 +43,16 @@ else()
     endif()
 endif()
 
-if (NOT BLT_CXX_STD)
-    set(BLT_CXX_STD "c++11" CACHE STRING "")
+if (“${PROJECT_SOURCE_DIR}” STREQUAL “${CMAKE_SOURCE_DIR}”)
+    # Set some default BLT options before loading BLT only if not included in
+    # another project
+    if (NOT BLT_CXX_STD)
+        set(BLT_CXX_STD "c++11" CACHE STRING "")
+    endif()
+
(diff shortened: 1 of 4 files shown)


white238:
This passes all BlueOS and TOSS3 host-configs.

src/CMakeLists.txt
```diff
@@ -43,10 +43,16 @@ else()
     endif()
 endif()
 
-if (NOT BLT_CXX_STD)
-    set(BLT_CXX_STD "c++11" CACHE STRING "")
+if (“${PROJECT_SOURCE_DIR}” STREQUAL “${CMAKE_SOURCE_DIR}”)
```
agcapps: Are those double-quote `"` characters, or MS Word style open-double-quote and close-double-quote characters?  Is GitHub messing with me?

Edit: Compare to (new) line 50.  Also check all the other comparisons of PROJECT_SOURCE_DIR with CMAKE_SOURCE_DIR.

kennyweiss: Thanks for pointing these out @agcapps.

I saw it too, but for some reason thought it was just how github was rendering them.

white238: its definitely a different character... i can copy and paste it to another screen and type double-quote and they are different. so weird.

agcapps: Chris, are you writing code in MS Word again?

white238: you caught me.... my secret shame. (it was a copy and paste from readthedocs)

PR #193, opened 2020-03-27
Simple Sidre Examples

### LINKED ISSUE #188: Sidre:  external described data
I am trying to apply a view to a 2d array that I own. (correct my verbiage if it seems off)
I am following the example 
[sidre_createdatastore.cpp](https://github.com/LLNL/axom/blob/81f88f60729b0bdf9923a02df8e0ac5e9f6f7e49/src/axom/sidre/examples/sidre_createdatastore.cpp
)
I create a datastore with a group "nodes", then I create a view in "nodes" and apply it to he data I own.
```
nodes->createView("node_t_2", node_coords )->apply( axom::sidre::DOUBLE_ID, 
                                                            4*node_quantity,
                                                            0,
                                                            4 );
```
I then save the datastore to a file and read it back in a separate function. 
When I try to grab that view, I get a seg fault when I try to access the Array.
```
double* node_t_ptr = nodes->getView("node_t_2")->getArray();
for (int pos = 0 ; pos < node_quantity ; ++pos)
      {
         printf(" %d  line %d \n", pos , __LINE__ );
         node_coords[pos][0] = node_t_ptr[pos];  
         // causes a segfault when pos = 0
      }
```
Is it sufficient to apply the view to an array I own as above? Did I miss a step?

gzagaris:
# Summary

Adds some simple Sidre examples in response to a user request.

Resolves #188

### DIFF
--- a/src/axom/sidre/examples/sidre_external_array.cpp
+++ b/src/axom/sidre/examples/sidre_external_array.cpp
@@ -74,12 +74,6 @@ void sidre_read( MPI_Comm comm,
   sidre::IOManager sidre_io( comm );
   sidre_io.read( root, file );
 
-// DEBUG
-  SLIC_INFO( "Here is the data that was read back:" );
-  root->print();
-  std::cout << std::endl;
-// DEBUG
-
   SLIC_ASSERT( root->hasChildView("data") );
   sidre::View* view = root->getView( "data" );
 
@@ -93,8 +87,17 @@ void sidre_read( MPI_Comm comm,
 
   data = axom::allocate< int >(nelems);
   SLIC_ASSERT( data != nullptr );
+  SLIC_ASSERT( view->isDescribed() );
+  SLIC_ASSERT( view->isExternal() );
 
-  memcpy( data, view->getVoidPtr(), nelems*sizeof(int) );
+  view->setExternalDataPtr( data );
+  sidre_io.loadExternalData(root,file);
+
+// DEBUG
+  SLIC_INFO( "Here is the data that was read back:" );
+  root->print();
+  std::cout << std::endl;
+// DEBUG
 }
 
 //------------------------------------------------------------------------------
--- a/src/axom/sidre/examples/sidre_external_array.cpp
+++ b/src/axom/sidre/examples/sidre_external_array.cpp
@@ -1,0 +1,151 @@
+// Copyright (c) 2017-2020, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+// Axom includes
+#include "axom/core.hpp"  // for Axom types and macros
+#include "axom/sidre.hpp" // for sidre
+#include "axom/slic.hpp"  // for logging with slic
+
+// MPI includes
+#include <mpi.h>
+
+// C/C++ includes
+#include <iostream>
+
+// aliases
+namespace sidre = axom::sidre;
+namespace slic  = axom::slic;
+
+//------------------------------------------------------------------------------
+void sidre_write( MPI_Comm comm,
+                  const std::string& file,
+                  int* data,
+                  axom::IndexType numTuples,
+                  axom::IndexType numComponents )
+{
+  SLIC_ASSERT( comm != MPI_COMM_NULL );
+  SLIC_ASSERT( data != nullptr );
+  SLIC_ASSERT( !file.empty( ) );
+
+  int nranks = -1;
+  MPI_Comm_size( comm, &nranks );
+
+  sidre::DataStore ds;
+  sidre::Group* root = ds.getRoot();
+  sidre::View*  view = root->createView( "data" );
+
+  sidre::IndexType shape[2];
+  shape[ 0 ] = numTuples;
+  shape[ 1 ] = numComponents;
+
+  view->setExternalDataPtr( sidre::INT32_ID, 2, shape, data );
+
+// DEBUG
+  SLIC_INFO( "Here is the data that is begin dumped:" );
+  root->print();
+  std::cout << std::endl;
+// DEBUG
+
+
+  // STEP 2: save the array data in to a file
+  sidre::IOManager sidre_io( comm );
+  sidre_io.write( root, nranks, file, "sidre_hdf5" );
+}
+
+//------------------------------------------------------------------------------
+void sidre_read( MPI_Comm comm,
+                 const std::string& file,
+                 int*& data,
+                 axom::IndexType& numTuples,
+                 axom::IndexType& numComponents )
+{
+  SLIC_ASSERT( comm != MPI_COMM_NULL );
+  SLIC_ASSERT( data == nullptr );
+  SLIC_ASSERT( !file.empty( ) );
+
+  int nranks = -1;
+  MPI_Comm_size( comm, &nranks );
+
+  sidre::DataStore ds;
+  sidre::Group* root = ds.getRoot();
+
+  sidre::IOManager sidre_io( comm );
+  sidre_io.read( root, file );
+
+// DEBUG
+  SLIC_INFO( "Here is the data that was read back:" );
+  root->print();
+  std::cout << std::endl;
+// DEBUG
+
+  SLIC_ASSERT( root->hasChildView("data") );
+  sidre::View* view = root->getView( "data" );
+
+  sidre::IndexType shape[2];
+  view->getShape( 2, shape );
+  numTuples     = shape[ 0 ];
+  numComponents = shape[ 1 ];
+
+  axom::IndexType nelems = view->getNumElements();
+  SLIC_ASSERT( nelems==(numTuples*numComponents) );
+
+  data = axom::allocate< int >(nelems);
+  SLIC_ASSERT( data != nullptr );
+
+  memcpy( data, view->getVoidPtr(), nelems*sizeof(int) );
+}
+
+//------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/axom/sidre/examples/sidre_external_array.cpp
```diff
@@ -0,0 +1,151 @@
...
+  sidre::IndexType shape[2];
+  shape[ 0 ] = numTuples;
+  shape[ 1 ] = numComponents;
+
+  view->setExternalDataPtr( sidre::INT32_ID, 2, shape, data );
+
+// DEBUG
+  SLIC_INFO( "Here is the data that is begin dumped:" );
+  root->print();
+  std::cout << std::endl;
```
gzagaris: The output from this is as I would expect the following:
```
{
  "name": "",
  "views":
  {
    "data":
    {
      "name": "data",
      "schema": "{\"dtype\":\"int32\", \"number_of_elements\": 40, \"offset\": 0, \"stride\": 4, \"element_bytes\": 4, \"endianness\": \"little\"}",
      "value": "[10, 20, 30, 40, 50, 60, 70, 80, 90, 100, 110, 120, 130, 140, 150, 160, 170, 180, 190, 200, 210, 220, 230, 240, 250, 260, 270, 280, 290, 300, 310, 320, 330, 340, 350, 360, 370, 380, 390, 400]",
      "state": "EXTERNAL",
      "is_applied": 1
    }
  }
}
```

src/axom/sidre/examples/sidre_external_array.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+  sidre::DataStore ds;
+  sidre::Group* root = ds.getRoot();
+
+  sidre::IOManager sidre_io( comm );
+  sidre_io.read( root, file );
+
+// DEBUG
+  SLIC_INFO( "Here is the data that was read back:" );
+  root->print();
```
gzagaris: However, when I read the data back, I get this:
```
{
  "name": "",
  "views":
  {
    "data":
    {
      "name": "data",
      "schema": "{\"dtype\":\"int32\", \"number_of_elements\": 40, \"offset\": 0, \"stride\": 4, \"element_bytes\": 4, \"endianness\": \"little\"}",
      "value": "null",
      "state": "EXTERNAL",
      "is_applied": 0
    }
  }
}
```

The view is there and has the correct shape information, but, where did the data go?

Do I need to mark external data some how before I write them to a file?

@nselliott, @cyrush , @rhornung67  Thoughts?

kennyweiss: > Do I need to mark external data some how before I write them to a file?

Actually the special steps are on the reading side, not the writing side.

Specifically, there is a three-step process that you have to do to load in external data:
1. Read in the data store
2. The host application then allocates any necessary external arrays and informs sidre about their locations
3. We then load the external data into the supplied arrays.

See the example here: https://github.com/LLNL/axom/blob/develop/src/axom/sidre/tests/spio/spio_parallel.cpp

// 1. Load datastore from file:
https://github.com/LLNL/axom/blob/1827916b4741f3d7402397a4bfd98d132cd3b656/src/axom/sidre/tests/spio/spio_parallel.cpp#L327-L335

// 2a. Host code allocates storage
https://github.com/LLNL/axom/blob/1827916b4741f3d7402397a4bfd98d132cd3b656/src/axom/sidre/tests/spio/spio_parallel.cpp#L337-L342

// 2b. Host code tells sidre about these pointers
https://github.com/LLNL/axom/blob/1827916b4741f3d7402397a4bfd98d132cd3b656/src/axom/sidre/tests/spio/spio_parallel.cpp#L344-L348

// 3. Sidre loads data into these arrays
https://github.com/LLNL/axom/blob/1827916b4741f3d7402397a4bfd98d132cd3b656/src/axom/sidre/tests/spio/spio_parallel.cpp#L350

This somewhat complicated process is necessary since sidre does not own or allocate storage associated with the external views.

gzagaris: Thanks for the clarification @kennyweiss 

Honestly, this is very confusing and unintuitive. I would have expected that when the data is read back is no longer `EXTERNAL`

How do things work with just `conduit`?

kennyweiss: I don't disagree about it being complicated, but it is not unnecessarily complicated.

It was designed to follow an expected use case:  If a host code is using ``sidre`` for checkpoints/restarts on EXTERNAL data, it wants to control the data management for these arrays.  Converting EXTERNAL data to ``sidre``-managed data would break this design.

This whole process is described in the sidre user docs: 
https://axom.readthedocs.io/en/develop/axom/sidre/docs/sphinx/parallel_io_concepts.html#iomanager-class-use

gzagaris: Would it suffice if `EXTERNAL` was an attribute ? Wouldn't that be cleaner ? 

I find this way too complicated for what is necessary. I am surprised no one else has complained about this. Most likely though not a lot of folks have run into this

cyrush: the three step process is necessary if you want the host code to own the external pointers:

 - read in all sidre data + descriptions of external views
 - allow host code  to alloc pointers for the external views and attach them
 - read in all external data in one operation

The second step has to be up to the host code - the other two are each one function call.

The external case is more complex in that it gives the host code extra control over allocate.

nselliott: An attribute still doesn't tell where the application-owned data allocation is.  There would still need to be three steps:  The DataStore loads its tree, which it has to do before it knows there are any external Views at all; the application associates its allocated adrresses with the external Views; something puts the correct data values at the application-owned adresses.

nselliott: If a code doesn't care about preserving the external ownership of the pointers, the protocols that don't have "sidre" in their names load the data into DataStore-owned memory, if I recall correctly.

gzagaris: So, in order to write a general reader without making any assumptions, you would need to traverse the tree, check for all any "external" views, allocate data and set external pointers and then do another I/O call to load the external data.

I find that very perplexing.

 I guess all I am saying is that I think there could be potential for improvement here.

gzagaris: > If a code doesn't care about preserving the external ownership of the pointers, the protocols that don't have "sidre" in their names load the data into DataStore-owned memory, if I recall correctly.

Could there be an 'option' in the `read` function to load the external data and let sidre assume ownership? Maybe that is a good comprise to maintain both the flexibility of the 3-step process and ease of usability. Thoughts?

gzagaris:
> Thanks for adding these examples @gzagaris.
> 
> Please update your `sidre_read` function for the external data as described in my comment.

Done.

src/axom/sidre/examples/sidre_external_array.cpp
cyrush: @gzagaris if you are using external, that is correct. It's there to support use cases where the host code wants extra control. 

It's there in support fortran applications for LLNL where they already allocate their own data.

Again, step 2  is where the work is, and it happens in the host code. 
Codes are already describing their data to sidre, this is a small change beyond that to accomplish restart. 

If there is a pattern or approach that can simplify, then I am all for it.

gzagaris: Thanks -- I'll create a Github issue to discuss this further and revisit this.

cyrush: @gzagaris we could potentially add another read function `readEverythingPlease` (not an actual name suggestion, just to frame the discussion)

gzagaris: @cyrush  yeah, that would work too.

src/axom/sidre/examples/sidre_external_array.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+  axom::IndexType nelems = view->getNumElements();
+  SLIC_ASSERT( nelems==(numTuples*numComponents) );
+
+  data = axom::allocate< int >(nelems);
+  SLIC_ASSERT( data != nullptr );
+  SLIC_ASSERT( view->isDescribed() );
+  SLIC_ASSERT( view->isExternal() );
+
+  view->setExternalDataPtr( data );
```
kennyweiss: Since this was not obvious to you when developing it, please add some comments.
E.g. something like
> // Inform Sidre of the locations to the external data 

and 

> // load the data into these pointers

src/axom/sidre/examples/sidre_external_array.cpp
```diff
@@ -0,0 +1,154 @@
...
+  sidre_io.loadExternalData(root,file);
+
+// DEBUG
+  SLIC_INFO( "Here is the data that was read back:" );
+  root->print();
+  std::cout << std::endl;
```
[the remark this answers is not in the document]
gzagaris: I think that's fine for this example, `root->print()` already calls `cout` and that's what `SLIC` uses internally too.

`slic::flushStream()` doesn't print a newline, which is what I want here to make the debug output easier to parse..

src/axom/sidre/examples/sidre_external_array.cpp
gzagaris: Yeah, definitely not obvious -- I'll add some comments.

src/axom/sidre/examples/sidre_external_array.cpp
gzagaris: See #198

PR #200, opened 2020-03-29
Fix Sidre Array examples

gzagaris:
# Summary

Add MPI barriers to synchronize all ranks after write() and read().


gzagaris:
I was monitoring the CI tests while they were running and I noticed that this was showing that "All checks have passed", but, not all the CI jobs were finished.

PR #212, opened 2020-04-03
Rzmanta devtool/TPLs/specs update

white238:
This basically copies all of rzansel Spack configs and specs to rzmanta and fixes the path to point to the correct sys_type


white238:
We are currently running out of space in the Azure pipeline jobs... I don't think Kenny or I have enough permissions in Azure to fix this.

PR #215, opened 2020-04-03
Resolve out-of-space issue in azure pipelines CI

kennyweiss:
# Summary

- Attempt to resolve 'out-of-space' issue in azure pipelines CI (#214)
- Sets 'clean' step to 'true' in azure-pipelines

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -67,7 +67,7 @@ pool:
 
 steps:
 - checkout: self
-  clean: boolean
+  clean: true
   submodules: recursive
 - task: CMake@1
   inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -103,3 +103,6 @@ steps:
     testResultsFiles: '**/Test.xml'
     testRunTitle: '$(TEST_TARGET) Tests'
     failTaskOnFailedTests: true
+
+workspace:
+  clean: outputs


azure-pipelines.yml
```diff
@@ -67,7 +67,7 @@ pool:
 
 steps:
 - checkout: self
-  clean: boolean
+  clean: true
```
white238, line 70: I did this in my PR and it didnt solve the problem.

kennyweiss: Thanks @white238 -- I didn't see that you already tried.

The next step might be to wrap our ``steps`` task under a ``jobs`` task, and add a ``clean:output`` to the ``job``.

E.g.:

```
jobs:
- job: MyJob
  displayName: My First Job
  continueOnError: true
  workspace:
    clean: outputs                         # <-- This
  steps:
  - script: echo My first job
```
https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=example%2Cparameter-schema#job

white238: I also added a make clean in my repo that dropped the build directory from 220megs to 9.. and that made no difference.

kennyweiss:
@white238 resolved this in #212 by disabling the ``gcc-6`` build. 
We might have found (part of) the problem in #216.

Issue #134, opened 2019-12-16
Allow BVH to use a supplied (pool) allocator for all temporaries


gzagaris:
The use of pool allocators (through umpire) is fairly common in a host application. The BVH will create temporary buffers for certain operations (e.g., construction). In this case, it is desirable to point the BVH to the pool allocator the application is using or create its own internal pool allocator to minimize the overheads of memory allocation/deallocation. 

This issue proposes extending the BVH API, i.e., the constructor, to optionally take the ID of a supplied pool allocator and change all internal memory allocation/de-allocation code to use the pool allocator.

PR #219, opened 2020-04-08
Bounding Box Query for BVH

bmhan12:
This PR adds a `findBoundingBoxes` query for BVH.

### DIFF
--- a/src/axom/primal/geometry/BoundingBox.hpp
+++ b/src/axom/primal/geometry/BoundingBox.hpp
@@ -443,10 +443,10 @@ bool BoundingBox< T,NDIMS >::intersectsWith(
   for ( int i=0 ; i < NDIMS ; ++i )
   {
 
-    status = status && detail::intersect_bbox_bbox_test( m_min[ i ],
-                                                         m_max[ i ],
-                                                         otherBB.m_min[ i ],
-                                                         otherBB.m_max[ i ] );
+    status = status && detail::intersect_bbox_bbox( m_min[ i ],
+                                                    m_max[ i ],
+                                                    otherBB.m_min[ i ],
+                                                    otherBB.m_max[ i ] );
   } // END for all dimensions
 
   return status;
--- a/src/axom/primal/geometry/BoundingBox.hpp
+++ b/src/axom/primal/geometry/BoundingBox.hpp
@@ -13,6 +13,8 @@
 #include "axom/primal/geometry/Point.hpp"
 #include "axom/primal/geometry/Vector.hpp"
 
+#include "axom/primal/operators/detail/intersect_bounding_box_impl.hpp"
+
 namespace axom
 {
 namespace primal
@@ -435,21 +437,19 @@ template < typename OtherType  >
 bool BoundingBox< T,NDIMS >::intersectsWith(
   const BoundingBox< OtherType, NDIMS >& otherBB ) const
 {
+  bool status = true;
+
   // AABBs cannot intersect if they are separated along any dimension
   for ( int i=0 ; i < NDIMS ; ++i )
   {
 
-    if ( (m_max[ i ] < otherBB.m_min[ i ]) ||
-         (m_min[ i ] > otherBB.m_max[ i ]) )
-    {
-
-      return false;
-
-    }  // END if
-
+    status = status && detail::intersect_bbox_bbox_test( m_min[ i ],
+                                                         m_max[ i ],
+                                                         otherBB.m_min[ i ],
+                                                         otherBB.m_max[ i ] );
   } // END for all dimensions
 
-  return true;
+  return status;
 }
 
 //------------------------------------------------------------------------------
--- a/src/axom/primal/operators/detail/intersect_bounding_box_impl.hpp
+++ b/src/axom/primal/operators/detail/intersect_bounding_box_impl.hpp
@@ -29,15 +29,16 @@ namespace detail
  *
  * \note This routine is called by the intersect AABB/AABB methods for each
  *  spatial dimension.
+ * \note This routine does not have a tolerance parameter for fuzzy
+ *  intersection, but the AABBs can be scaled to achieve the same thing.
  */
 template < typename T >
 AXOM_HOST_DEVICE
-inline bool intersect_bbox_bbox_test( const T& min1,
-                                      const T& max1,
-                                      const T& min2,
-                                      const T& max2 )
+inline bool intersect_bbox_bbox( const T& min1,
+                                 const T& max1,
+                                 const T& min2,
+                                 const T& max2 )
 {
-  AXOM_STATIC_ASSERT( std::is_floating_point< T >::value );
 
   bool status = ( ( max1 < min2 || min1 > max2 ) ? false : true );
   return status;
@@ -69,11 +70,10 @@ inline bool intersect_bounding_box( const T& xmin1,
                                     const T& ymin2,
                                     const T& ymax2 )
 {
-  AXOM_STATIC_ASSERT( std::is_floating_point< T >::value );
 
   bool status = true;
-  status = status && intersect_bbox_bbox_test( xmin1, xmax1, xmin2, xmax2 );
-  status = status && intersect_bbox_bbox_test( ymin1, ymax1, ymin2, ymax2 );
+  status = status && intersect_bbox_bbox( xmin1, xmax1, xmin2, xmax2 );
+  status = status && intersect_bbox_bbox( ymin1, ymax1, ymin2, ymax2 );
   return status;
 }
 
@@ -111,12 +111,11 @@ inline bool intersect_bounding_box( const T& xmin1,
                                     const T& zmin2,
                                     const T& zmax2 )
 {
-  AXOM_STATIC_ASSERT( std::is_floating_point< T >::value );
 
   bool status = true;
-  status = status && intersect_bbox_bbox_test( xmin1, xmax1, xmin2, xmax2 );
-  status = status && intersect_bbox_bbox_test( ymin1, ymax1, ymin2, ymax2 );
-  status = status && intersect_bbox_bbox_test( zmin1, zmax1, zmin2, zmax2 );
+  status = status && intersect_bbox_bbox( xmin1, xmax1, xmin2, xmax2 );
+  status = status && intersect_bbox_bbox( ymin1, ymax1, ymin2, ymax2 );
+  status = status && intersect_bbox_bbox( zmin1, zmax1, zmin2, zmax2 );
   return status;
 }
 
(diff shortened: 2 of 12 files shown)


src/axom/primal/operators/detail/intersect_bounding_box_impl.hpp
```diff
@@ -0,0 +1,127 @@
...
+  status = status && intersect_bbox_bbox_test( ymin1, ymax1, ymin2, ymax2 );
+  status = status && intersect_bbox_bbox_test( zmin1, zmax1, zmin2, zmax2 );
+  return status;
+}
+
+} /* namespace detail */
+} /* namespace primal */
+} /* namespace axom */
+
+#endif
```
gzagaris, line 127: missing newline at the end of file?

src/axom/primal/operators/detail/intersect_bounding_box_impl.hpp
```diff
@@ -0,0 +1,127 @@
...
+                           			    const T& ymin1,
+                           			    const T& ymax1,
+                           			    const T& zmin1,
+                           			    const T& zmax1,
+                           			    const T& xmin2,
+                           			    const T& xmax2,
+                           			    const T& ymin2,
+                           			    const T& ymax2,
+                           		    	const T& zmin2,
+                           		    	const T& zmax2 )
```
gzagaris, line 112: Ditto.

src/axom/primal/operators/detail/intersect_bounding_box_impl.hpp
```diff
@@ -0,0 +1,127 @@
...
+template < typename T >
+AXOM_HOST_DEVICE
+inline bool intersect_bounding_box( const T& xmin1,
+                              			const T& xmax1,
+                           	    		const T& ymin1,
+                               			const T& ymax1,
+                           			    const T& xmin2,
+                           			    const T& xmax2,
+                           			    const T& ymin2,
+                           			    const T& ymax2 )
```
gzagaris, line 70: nitpick: code formatting of the arguments looks odd. Do you have any tabs in this file?

You may want to run this through uncrustify but don't let uncrustify do anything weird with the lambdas

src/axom/primal/operators/detail/intersect_bounding_box_impl.hpp
```diff
@@ -0,0 +1,127 @@
...
+ * \param [in] max2 the 2nd AABB max coordinate along a drection.
+ *
+ * \return status true if the AABB's intersect, otherwise false.
+ *
+ * \note This routine is called by the intersect AABB/AABB methods for each
+ *  spatial dimension.
+ */
+template < typename T >
+AXOM_HOST_DEVICE
+inline bool intersect_bbox_bbox_test( const T& min1,
```
gzagaris, line 35: `BoundingBox::intersectsWith()` should also call this method now -- This would avoid duplicating code and also expose this method to the `BoundingBox` tests

gzagaris:
Also, please add something to the RELEASE-NOTES regarding this new query.

src/axom/primal/operators/detail/intersect_bounding_box_impl.hpp
bmhan12: Done!

src/axom/primal/operators/detail/intersect_bounding_box_impl.hpp
bmhan12: Done!

src/axom/spin/tests/spin_bvh.cpp
```diff
@@ -476,6 +476,244 @@ void check_build_bvh3d( )
...
+  FloatType min[ NDIMS ];
+  FloatType max[ NDIMS ];
+  bvh.getBounds( min, max );
+  for ( int i=0 ; i < NDIMS ; ++i )
+  {
+    EXPECT_DOUBLE_EQ( min[ i ], lo[ i ] );
+    EXPECT_DOUBLE_EQ( max[ i ], hi[ i ] );
+  }
+
+  // traverse the BVH to find the candidates for all the centroids
```
kennyweiss, line 530: > centroids

Possible copy/paste error. Should this be "bounding boxes" ?

src/axom/spin/tests/spin_bvh.cpp
```diff
@@ -476,6 +476,244 @@ void check_build_bvh3d( )
...
+  IndexType* offsets    = axom::allocate< IndexType >( N );
+  IndexType* counts     = axom::allocate< IndexType >( N );
+  IndexType* candidates = nullptr;
+  bvh.findBoundingBoxes( offsets, counts, candidates, N, xmin, xmax,
+                         ymin, ymax, zmin, zmax );
+  EXPECT_TRUE( candidates != nullptr );
+
+  // flag cells that are found by the ray ID
```
kennyweiss, line 538: "ray ID" -> "bounding box ID" ?

src/axom/spin/internal/linear_bvh/BVH_impl.hpp
```diff
@@ -522,6 +620,99 @@ void BVH< NDIMS, ExecSpace, FloatType >::findRays( IndexType* offsets,
...
+  SLIC_ASSERT( candidates == nullptr );
+  SLIC_ASSERT( xmin != nullptr );
+  SLIC_ASSERT( xmax != nullptr );
+  SLIC_ASSERT( ymin != nullptr );
+  SLIC_ASSERT( ymax != nullptr );
+
+  // STEP 0: set the default memory allocator to use for the execution space.
+  const int currentAllocatorID = axom::getDefaultAllocatorID();
+  const int allocatorID = axom::execution_space< ExecSpace >::allocatorID();
+  axom::setDefaultAllocator( allocatorID );
```
kennyweiss, line 646: Changing the global default allocator could be a problem for other libraries that are using a shared axom.
E.g. this could lead to concurrency issues which would be very difficult to track down.

Can you please create a ticket to modify the design of this class so that all allocators are local?

Tag: @gzagaris

src/axom/primal/geometry/BoundingBox.hpp
```diff
@@ -435,21 +437,19 @@ template < typename OtherType  >
...
 
-    if ( (m_max[ i ] < otherBB.m_min[ i ]) ||
-         (m_min[ i ] > otherBB.m_max[ i ]) )
-    {
-
-      return false;
-
-    }  // END if
-
+    status = status && detail::intersect_bbox_bbox_test( m_min[ i ],
```
kennyweiss, line 446: It might not matter, but this function *might* be more efficient in its original form:
E.g.
```cpp
 for( ... )
 {
  if(detail::intersect_bbox_bbox_test( m_min[ i ], m_max[ i ], otherBB.m_min[ i ], otherBB.m_max[ i ] ) )
     return false;
 }
  return true;
```
I suppose we'd need to profile it to find out, (which is probably not worth spending time on)...

src/axom/spin/internal/linear_bvh/BVH_impl.hpp
gzagaris: We are setting it back to the `currentAllocatorID` at the end of this function. I am addressing this in #134.

src/axom/primal/geometry/BoundingBox.hpp
gzagaris: I would expect the compiler to inline this function. I agree. It's not worth spending time on this...

src/axom/primal/operators/detail/intersect_bounding_box_impl.hpp
```diff
@@ -0,0 +1,127 @@
...
+                                      const T& max1,
+                                      const T& min2,
+                                      const T& max2 )
+{
+  AXOM_STATIC_ASSERT( std::is_floating_point< T >::value );
```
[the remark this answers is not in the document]
gzagaris: I agree -- using floating point quantities is not required here.

src/axom/primal/operators/detail/intersect_bounding_box_impl.hpp
```diff
@@ -0,0 +1,127 @@
...
+ * \brief Helper routine for AABB / AABB intersection test.
+ *
+ * \param [in] min1 the 1st AABB min coordinate along a direction.
+ * \param [in] max1 the 1st AABB max coordinate along a drection.
+ * \param [in] min2 the 2nd AABB min coordinate along a direction.
+ * \param [in] max2 the 2nd AABB max coordinate along a drection.
+ *
+ * \return status true if the AABB's intersect, otherwise false.
+ *
+ * \note This routine is called by the intersect AABB/AABB methods for each
```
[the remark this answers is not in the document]
mclarsen: If there is no fuzzy intersection, It might be worth mentioning that that the AABBs can be scaled which essentially creates a fuzzy intersetion.

src/axom/spin/internal/linear_bvh/TraversalPredicates.hpp
```diff
@@ -97,9 +98,9 @@ class TraversalPredicates
   template < typename RayType >
   AXOM_HOST_DEVICE
   static inline bool rayIntersectsLeftBin(
-      const RayType& r, const vec4_t& s1, const vec4_t& s2,
-      FloatType TOL=numerics::floating_point_limits<FloatType>::epsilon()
-      ) noexcept;
+    const RayType& r, const vec4_t& s1, const vec4_t& s2,
+    FloatType TOL=numerics::floating_point_limits<FloatType>::epsilon()
```
mclarsen, line 102: I think I have mentioned this before, but I think the tolerance should be relative so either the AABB bounds or the total "scene" bounds.

src/axom/primal/operators/detail/intersect_bounding_box_impl.hpp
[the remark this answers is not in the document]
gzagaris: Good point @mclarsen -- Since this method is in primal and can be used outside of the BVH, I think it may make sense for this method to take an optional tolerance

src/axom/spin/internal/linear_bvh/TraversalPredicates.hpp
gzagaris: Yes, those are default arguments. At the call site we can pass a more appropriate tolerance. We are currently not doing that, but I think all the machinery is there for us to do so. I'll create a Github issue for this.

gzagaris: See #221

src/axom/primal/operators/detail/intersect_bounding_box_impl.hpp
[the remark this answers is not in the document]
bmhan12: @kennyweiss , @mclarsen  - Good to know! I added a doxygen comment to use scaling for fuzzy intersections.

PR #222, opened 2020-04-14
BVH Improvements: Memory Management

### LINKED ISSUE #134: Allow BVH to use a supplied (pool) allocator for all temporaries

gzagaris:
# Summary

This PR adds support for specifying a user-supplied allocator to use with the BVH.

This avoids relying to Unified Memory when running with `CUDA_EXEC`, which can have nondeterministic behavior in terms of memory motion. Moreover, this PR allows the host code to supply a pool allocator to use with the BVH, if one is available to mitigate the costs for allocating device memory.

Resolves #134


gzagaris:
Here are some preliminary profiles using this feature -- The sections in blue correspond to kernel ranges inside the BVH.

### Using UM
Unfortunately, using just UM with a host application can result in undesirable memory motion, even if we only touch the data on the GPU. The CUDA runtime will move the data as needed and we generally have no control of that:

particles-14K-UM

We could also try using UM and applying a memory advice (see https://umpire.readthedocs.io/en/develop/cookbook/pool_advice.html) to mitigate the motion, but, there are still no guarantees.

### Using Device Memory

Using Device memory explicitly removes all memory motion. However, the internal allocations
for temporaries use by BVH incur a high-cost, hindering performance:

particles-14K-Device

### Using a Memory Pool

By far, the best strategy is to utilize a device memory pool, which, a host-code should have. This avoids all memory motion, as well as, minimize the cost for memory allocations:

Particles-14K-MemoryPool

gzagaris:
This is ready for review. It builds on top of PR #218. The intent is to change the destination branch to develop and merge this to develop after PR #218 gets merged.

RELEASE-NOTES.md
```diff
@@ -10,6 +10,11 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 ## [Unreleased] - Release date yyyy-mm-dd
 
 ### Added
+- Added the ability to specify a user-supplied allocator ID to use with the
```
rhornung67, line 13: To be clear, this is an Umpire allocator ID, correct? If so, it should be stated explicitly here. If not Umpire, what does the allocator ID refer to?

host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.1_nvcc_xlf.cmake
```diff
@@ -108,6 +108,8 @@ set(CUDA_SEPARABLE_COMPILATION ON CACHE BOOL "")
 
 set(AXOM_CUDA_ARCH "sm_70" CACHE PATH "")
 
+set(AXOM_USE_CUB ON CACHE BOOL "")
```
rhornung67, line 111: Is there a case where Axom uses CUB independent of RAJA?

src/axom/spin/BVH.hpp
```diff
@@ -140,6 +140,7 @@ class BVH
    *
    * \param [in] boxes buffer consisting of bounding boxes for each entity.
    * \param [in] numItems the total number of items to store in the BVH.
+   * \param [in] allocatorID specify the ID of the allocator to use. Optional.
```
rhornung67, line 143: Similar comment to above: Umpire ID, correct?

src/axom/spin/BVH.hpp
```diff
@@ -157,6 +158,10 @@ class BVH
    *    double zmax = boxes[ offset+5 ];
    *  \endcode
    *
+   * \note If an allocatorID is not specified, the code will use the default
```
rhornung67, line 161: Umpire allocator ID?

src/axom/spin/BVH.hpp
```diff
@@ -165,13 +170,21 @@ class BVH
    * \pre boxes != nullptr
    * \pre numItems > 0
    */
-  BVH( const FloatType* boxes, IndexType numItems );
+  BVH( const FloatType* boxes, IndexType numItems,
```
rhornung67, line 173: OK. So I see this here. Is there a reason we don't want to just have this just expose the Umpire allocator stuff. Why are we wrapping it?

src/axom/spin/internal/linear_bvh/BVHData.hpp
```diff
@@ -66,11 +66,11 @@ struct BVHData
     m_leaf_nodes( nullptr )
   {}
 
-  void allocate( int32 size )
+  void allocate( int32 size, int allocID )
   {
     AXOM_PERF_MARK_FUNCTION( "BVHData::allocate" );
-    m_inner_nodes = axom::allocate< Vec< FloatType,4 > >( (size-1)*4 );
-    m_leaf_nodes  = axom::allocate< int32 >( size );
+    m_inner_nodes = axom::allocate< Vec< FloatType,4 > >( (size-1)*4, allocID );
```
rhornung67, line 72: Magic number? A comment here about what '4' refers to would be helpful. Alternatively use a named const variable instead.

RELEASE-NOTES.md
gzagaris: Yes, this is an umpire allocator ID -- I will clarify.

- [ ] Clarify allocator ID.

host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.1_nvcc_xlf.cmake
gzagaris: No, in fact we use the CUB version that is bundled with RAJA. The reason that this option was added is b/c there was a linking issue with earlier versions of CUDA and we needed to turn it off.  If we are not using CUB, the code calls `std::stable_sort()` on the CPU, which would require data motion and currently require using UM for supplied bounding boxes. 

In the future, we would use the`RAJA::sort()` when that is available.

Please, let me know if you have any additional questions/suggestions on this.

src/axom/spin/BVH.hpp
gzagaris: We are not wrapping it, we just expose the integer ID of the umpire allocator instead of exposing umpire data-structures in the API. My preference is not to leak umpire data-structures in the API. 

- Is there a good reason that we should? 

- Are there any limitations with the way it is done currently?

To the best of my knowledge, most host-codes use the integer ID to specify an allocator.

Although this class strictly requires `RAJA` and `Umpire`, Axom can be compiled without `RAJA` and `Umpire`. So, limiting the interface to just basic primitive types make the API cleaner and avoids the otherwise needed `ifdef`.

src/axom/spin/internal/linear_bvh/BVHData.hpp
gzagaris: Yeah, the `Vec< FloatType, 4 >` is the underlying internal data-type used by the BVH. No doubt it is magical. 

I'll try to clean this up some (but, perhaps in a subsequent PR).

src/axom/spin/BVH.hpp
rhornung67: Got it. It was unclear to me what was going on until I looked at this part of the code. I think the way you have done it makes sense. What happens if I build Axom without RAJA and Umpire and try to use this class. Is it unavailable and the application code will not compile?

src/axom/spin/internal/linear_bvh/BVHData.hpp
rhornung67: I think adding a named integer constant and using it everywhere is all that is needed.

host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.1_nvcc_xlf.cmake
rhornung67: RAJA also allows external versions of CUB and camp (i.e., not the RAJA submodules). MARBL builds this way. We should also make sure that that works. It is described here:
https://raja.readthedocs.io/en/master/getting_started.html

src/axom/spin/BVH.hpp
gzagaris: Correct -- This class is only included if `Umpire` and `RAJA` are available. So, if the host code is using this class, it won't compile if the Axom is not built with `RAJA` and `Umpire`. If they are pointing to an axom installation they would get an error be able to find the include.

In retrospect, we do have the following in the header:
https://github.com/LLNL/axom/blob/fb2bb2a1f4e37e8a965e8799bca7896b64084e3a/src/axom/spin/BVH.hpp#L19-L21

So, we may want to install the header regardless so that they get a more meaningful error message when they try to use the class and haven't compiled Axom correctly. Thoughts?

We can also make it work without `RAJA` and `Umpire` as well. To do so, it would require us having our own abstraction layer to do `scans` and `reductions` so that the code compiles without `RAJA`.

src/axom/spin/internal/linear_bvh/BVHData.hpp
gzagaris: I'll fix it here. In other parts of the code, I have defined a template alias `vec4_t`, I think doing that in a more global scope inside the BVH internals would help.

- [ ] cleanup

host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.1_nvcc_xlf.cmake
gzagaris: Since this is orthogonal to this PR, I'll add a Github issue. I think it's just a matter of specifying a different CUB dir.

- [ ] Add Github issue.

kennyweiss: > There was a linking issue with earlier versions of CUDA

@gzagaris -- is this still an issue, or has it been resolved?  

If it has been resolved in our currently tested versions of Axom, I suggest enabling ``AXOM_USE_CUB`` to ``ON`` by default when using ``cuda`` and ``raja``?

kennyweiss: If axom want to use our own ``cub``, it will need to be sent to ``raja`` via spack and added as ``CUB_DIR`` to our host-config.  @rhornung67 -- do you know if ``raja``'s spack is set up to support that? 

It's an important use case for using a device-enabled axom, so we should try to set up our CI plan to test this functionality.

kennyweiss, line 111: Also -- our ``CMake`` configuration options are all of the form ``AXOM_ENABLE_*``, while their corresponding C++ defines are of the form ``AXOM_USE_*``.  

Please update this for consistency.

src/axom/spin/internal/linear_bvh/BVH_impl.hpp
```diff
@@ -497,43 +488,49 @@ void BVH< NDIMS, ExecSpace, FloatType >::findPoints( IndexType* offsets,
   };
 
   // STEP 3: get counts
-  int total_count = bvh_get_counts< NDIMS,ExecSpace >(
-    leftPredicate, rightPredicate, inner_nodes, leaf_nodes,
-    numPts, counts, x, y, z );
+  int total_count = 0;
+  AXOM_PERF_MARK_SECTION( "PASS[1]:count_traversal",
```
kennyweiss, line 492: > "PASS[1]:count_traversal"
"PASS[2]:fill_traversal"

Suggestion: Remove ``PASS[1]`` and ``PASS[2]`` from the annotations -- while this is useful for your current analysis, it might not be as useful to users / future use cases.

host-configs/rzansel-blueos_3_ppc64le_ib_p9-xl@nvcc.cmake
```diff
@@ -112,6 +112,8 @@ set(CUDA_SEPARABLE_COMPILATION ON CACHE BOOL "")
 
 set(AXOM_CUDA_ARCH "sm_70" CACHE PATH "")
 
+set(AXOM_USE_CUB ON CACHE BOOL "")
```
kennyweiss, line 115: I don't see a data change in the TPL set.

Did you build a set of TPLs or just modify the existing host-config?

If it is the latter, can you please confirm that:
(a) the TPLs build with your new change
(b) our blueos host-configs continue to pass the CI tests.

host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.1_nvcc_xlf.cmake
gzagaris: > > There was a linking issue with earlier versions of CUDA
> 
> @gzagaris -- is this still an issue, or has it been resolved?
> 
> If it has been resolved in our currently tested versions of Axom, I suggest enabling `AXOM_USE_CUB` to `ON` by default when using `cuda` and `raja`?

It is not an issue -- I have already enabled it in this PR.

host-configs/rzansel-blueos_3_ppc64le_ib_p9-xl@nvcc.cmake
gzagaris: See 64c1721de2f143bdcc05fd115588677717c6457a and 12c3a07b4a24957a463db255908ece401af58ae1

src/axom/spin/internal/linear_bvh/BVH_impl.hpp
gzagaris: I don't really expect users to dig in at this level of granularity.

host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.1_nvcc_xlf.cmake
gzagaris: See issue #224

src/axom/spin/BVH.hpp
gzagaris: See #238 .

RELEASE-NOTES.md
gzagaris: Fixed.

src/axom/spin/BVH.hpp
gzagaris: Fixed.

src/axom/spin/BVH.hpp
gzagaris: Fixed.

src/axom/spin/internal/linear_bvh/BVHData.hpp
gzagaris: Done.

host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.1_nvcc_xlf.cmake
gzagaris: Fixed -- Also updated the `AXOM_USE_MPI3` variable accordingly.

gzagaris:
@rhornung67 -- I have addressed your comments. Please, let me know if you have any other suggestions.

If anyone else has additional comments/suggestions, please, let me know.

PR #226, opened 2020-04-22
Feature/taylor/shroud update

ltaylor16:
# Generate with Shroud-0.11.0

- Earlier this year the devtools was updated to Shroud-0.11.0.  That update was required to get things to work properly with Spack.
- Update YAML files to use new Shroud syntax.

# Design review (Shroud changes)

1. Preserves case name of C++ identifiers in the C wrapper. Note that Fortran is case insensitive.
2. Better support for namespaces. This is demonstrated in the slic interface where enum changed from `level_error` to `message_error`.  i.e. *enum_variable* to *namespace_variable*.  This is intended to mirror C++ usage. 
3. Improved support for Fortran generics to allow multiple variables to be part of the generic interface.


src/axom/slic/interface/slic_shroud.yaml
```diff
@@ -15,11 +15,10 @@ namespace: axom slic
...
   C_line_length: 1000
   F_module_name_library_template: axom_{library_lower}
-#  F_module_per_class: False
   # Change the default template for Fortran functions to include the library name.
-  F_name_impl_template: "{library_lower}_{class_prefix}{underscore_name}{function_suffix}"
-  # XXX - This will eventually become the default in Shroud
-  C_enum_member_template: "{C_prefix}{enum_member_name}"
+  F_name_impl_template: "{library_lower}_{F_name_scope}{underscore_name}{function_suffix}"
+#  C_enum_member_template: "{C_prefix}{enum_member_name}"
+#  F_enum_member_template: "{enum_name}_{enum_member_name}"
```
gzagaris, line 21: Should these comments be removed?

src/axom/slic/interface/c_fortran/wrapfslic.f
```diff
@@ -20,12 +20,12 @@ module axom_slic
...
-    integer(C_INT), parameter :: level_warning = 1
-    integer(C_INT), parameter :: level_info = 2
-    integer(C_INT), parameter :: level_debug = 3
-    integer(C_INT), parameter :: level_num_levels = 4
+    !  enum axom::slic::message::Level
+    integer(C_INT), parameter :: message_error = 0
+    integer(C_INT), parameter :: message_warning = 1
+    integer(C_INT), parameter :: message_info = 2
+    integer(C_INT), parameter :: message_debug = 3
+    integer(C_INT), parameter :: message_num_levels = 4
```
gzagaris, line 28: Is there a way to make these `slic_message_error` etc. That way, the C interface is mirrored more accurately?

ltaylor16: This does mirror C++ usage.  Here's an example from slic

    addStreamToMsgLevel(wefStream,  message::Error);

In Fortran `message::Error` becomes `message_error`.  If it was a class enum, then the class name would be included in the Fortran name similar to how the class name would be needed in C++.  Shroud is just "flattening" the namespace.

src/axom/slic/interface/slic_shroud.yaml
ltaylor16: Yea.  I tend to leave note-to-self comments but these lines are no longer needed.  `F_name_scope` has the flattened Fortran identifier to better support namespaces.

src/axom/slic/interface/slic_shroud.yaml
```diff
@@ -33,6 +32,8 @@ format:
 declarations:
 - decl: namespace message
   # MessageLevel.h
+  options:
+    flatten_namespace: True
```
ltaylor16, line 36: @gzagaris The `flatten_namespace` option here prevents the message namespace from becoming a module on its own.  It preserves more of the old behavior but ends up renaming the enums.  It might be worth exploring removing this option, it's the only place it's used in Axom.

src/axom/slic/interface/c_fortran/wrapfslic.f
gzagaris: I think in your C++ usage example, assumes `using namespace slic;`. Without this, you would need to have:
```
slic::addStreamToMsgLevel(wefStream, slic::message::Error)
```

Moreover, what prompted my comment was that in the auto-generate C file, we have this:
```
//  axom::slic::message::Level
enum SLIC_Level	enum SLIC_message_Level
{	{
  SLIC_Error,	  SLIC_message_Error,
  SLIC_Warning,	  SLIC_message_Warning,
  SLIC_Info,	  SLIC_message_Info,
  SLIC_Debug,	  SLIC_message_Debug,
  SLIC_Num_Levels	  SLIC_message_Num_Levels
};	};
```

So, I am wondering if the Fortran and C usage could be made more consistent, or if there is a better way to make C++/C/Fortran usage overall more consistent.

src/axom/slic/interface/slic_shroud.yaml
gzagaris: @ltaylor16 -- if that's what you recommend as best practice, I think we should remove it.

src/axom/slic/interface/c_fortran/wrapfslic.f
ltaylor16: The Fortran names are in a module so it's similar to `using namespace slic;`

    use slic
    call add_stream_to_msg_level(wefStream, message_error)

There is no way to access the names without using the module.  There have been suggestions to allow fully qualified names in Fortran similar to C++i.e. `slic%message_error`.  But if that happens it's years away.  In the case of name conflicts between modules, there are two current solutions.  `use slic, only : message_error` or `use slic, my_error => message_error`.

The C names are global, so they are always the fully flattened name with the prefix `SLIC_`.

src/axom/sidre/spio/interface/spio_shroud.yaml
```diff
@@ -27,25 +27,21 @@ typemap:
 - type: axom::sidre::DataStore
   fields:
     base: shadow
-    cxx_header: axom/sidre/core/DataStore.hpp
-#    cxx_type: DataStore
+    wrap_header: axom/sidre/interface/c_fortran/wrapDataStore.h
```
kennyweiss, line 30: Is this telling it what file to put the wrap header into (i.e. overriding the default). ?

src/axom/sidre/interface/sidre_shroud.yaml
```diff
@@ -659,9 +688,11 @@ declarations:
...
   options:
     F_string_len_trim: false
-  format:
-    C_code:  return name != NULL;
-    F_code:  '{F_result} = name .ne. " "'
+  splicer:
+    c:
+    - "return name != NULL;"
+    f:
+    - 'SHT_rv = name .ne. " "'
```
kennyweiss, line 695: @ltaylor16 -- could you please clarify this binding?

Does this completely replace the function call with the given splicer code, or is this added to the function call?

src/axom/slic/interface/c_fortran/wrapfslic.f
gzagaris: Thanks for the explanation. Yes the potential conflicts was the main thing that I was worried about. Sounds like we have workarounds, which is good.

 Using a fully qualified name would be nice and avoid these issues when it happens.

src/axom/sidre/interface/sidre_shroud.yaml
```diff
@@ -76,9 +77,10 @@ declarations:
 #  - decl: bool hasBuffer( IndexType idx ) const
   - decl:  Buffer * getBuffer( IndexType idx )
     fortran_generic:
-      idx:
-      -  int32_t
-      -  int64_t
+    - decl: (int32_t idx)
```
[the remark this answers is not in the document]
ltaylor16: In C, when an argument is int64_t and you pass in a 0 the compiler politely promotes it to the correct type. Since Fortran defaults to pass-by-reference this is not possible. Instead Shroud creates two function, one which accepts int32_t and one that accepts int64_t and allows them to be called by a generic name which will accept either argument. This allows the function to be called with a `0` instead of `0_C_INT64_t`.
Earlier version of Shroud only allowed one argument to be changed in the generic interface. Now multiple arguments can be generic.

src/axom/sidre/spio/interface/spio_shroud.yaml
ltaylor16: No, this is telling it where the generated wrapped headers are located and used with wrapped classes since they generate a struct for each class (the shadow type).  The intent of this file is to allow one wrapped library to be used by another wrapped library.  It's not being used yet but I think there is a use case.  Currently spio_shroud.yaml duplicates some of the information in the file since it uses sidre wrapped types.  But eventually there should be some sort of include facility in Shroud.

src/axom/sidre/interface/sidre_shroud.yaml
ltaylor16: It will completely replace the generated code.

src/axom/slic/interface/slic_shroud.yaml
ltaylor16: I tested removing the `flatten_namespace` option and it creates a new module `slic_message_mod` with parameters for the enums named `error`, `warning`, etc.  I think this is worse since the names `error` and `warning` are very generic. The C api did not change. Shroud only put them in a different file (one file per namespace).
This code was written before we assumed C++11.  Today it would be better to use an `enum class` to scope the enums instead of a namespace. But that's another issue.
I will add the changed names in the Fortran API to  `RELEASE-NOTES.md`.

gzagaris: So are we removing this or not? This would require the Fortran application to have `use slic_message_mod`, which I think would be annoying. I think that what we would want is for the application to be able to simply do `use slic_mod` and get everything, right?

ltaylor16: Keep the `flatten_namespace`.  The main issue is it gives better enums: `message_error` instead of just `error`.  It's possible to put `use slic_message_mod` inside module slic_mod so users could still just `use slic_mod` to get everything so that is less of an issue.

gzagaris: Got it -- Thanks.

src/axom/slic/interface/slic_shroud.yaml
ltaylor16: removed

ltaylor16:
> Please add a ticket to reincorporate `make generate` into our CI plans.

Added issue Test generate target #227

ltaylor16:
> As far as I can tell, this has no effect on Fortran users of our shroud bindings, only on C users. Is that right?

The slic message enums changed names as discussed above.  I added this to `RELEASE-NOTES.md`.

PR #229, opened 2020-04-24
Refactor MapCollection, add ListCollection

nselliott:
# Summary

- This PR is a refactoring
  - Refactors MapCollection, adds virtual base class ItemCollection and new class ListCollection
  - Adds a new way for Groups to hold their child items, where items don't have names and are held in a manner similar to Conduit lists.

The MapCollection API is refactored here and uses as the basis of virtual base class ItemCollection, with MapCollection and ListCollection becoming derived classes.

A Group now holds its child Groups and Views in ItemCollections, which can be constructed as either a MapCollection and ListCollection.  This is determined at construction and does not change during the life of a Group.

By default Groups use MapCollections and in this mode there is no change to functionality.  MapCollections require every Group and View to have a unique non-empty string name.

Usage of ListCollection can be enabled via an optional boolean.  Groups and Views held in a ListCollection can have an empty string for a name.  Non-empty strings are allowed but not used for anything.  The methods that try to access items by name return null pointers.  The expected use case would be to use the iterators to iterate over all items without needing to know any names.

ListCollection may not be a fully accurate name at this point because it's not really a list until there is an iterator that adheres to list ordering semantics.  I think this should be done as a next step.  The current iterator still uses the same indexing scheme as the MapCollection.

This could be merged, but it could also wait for some more features, such as an STL-compliant list iterator, and implementation of loading Conduit lists into a Group using ListCollection.  Currently Conduit lists are ignored and can't be loaded into Sidre.

This also needs Shroud interfaces.


agcapps:
Thanks @nselliott .  

One question: as the PR stands now, it looks like it is possible to put unnamed Views and unnamed Groups in the same Group.  Do we have any reason to allow this?  Should we disallow it?

nselliott:
@agcapps That is intentionally allowed.  I don't see a reason to disallow it.  The Groups and Views will be iterated over separately.

gzagaris:
@nselliott -- I was thinking more about the access differences between `MapCollection` and `ListCollection`. 

Instead of having the access-by-name methods virtual in the base class, it might be better/cleaner design if those are only implemented in the `MapCollection`, since access-by-name makes only sense with a `MapCollection`. This would require the caller to downcast to a `MapCollection` if they need to use access by name and remove the need for returning null pointers for these methods when the `ItemCollection` is a `ListCollection`.

nselliott:
@gzagaris That is a valid point, as nobody wants to create a class with a bunch of methods that are guaranteed never to do anything useful, but the primary user of these classes is `sidre::Group`.  Do we want to do an `if` check and a downcast inside `Group` every time someone calls `group->getView("foo")` or multiple downcasts if someone calls `group->getGroup("path/to/grp")`?

If I did make the change you suggest, all of the name-access methods in `Group`, which is the more customer-focused API, would still be there and would still be returning nulls for any `Group` that is constructed to use `ListCollection`.

gzagaris:
> @gzagaris That is a valid point, as nobody wants to create a class with a bunch of methods that are guaranteed never to do anything useful, but the primary user of these classes is `sidre::Group`. Do we want to do an `if` check and a downcast inside `Group` every time someone calls `group->getView("foo")` or multiple downcasts if someone calls `group->getGroup("path/to/grp")`?
> 
> If I did make the change you suggest, all of the name-access methods in `Group`, which is the more customer-focused API, would still be there and would still be returning nulls for any `Group` that is constructed to use `ListCollection`.

@nselliott -- This may be more involved than I thought. 

I do not have a crystal clear, overarching picture of Sidre's design and implementation, but, based on what you said, I think it will be better to avoid :
- Downcasting inside `Group`, and,
- Percolating the problem to the `Group` class

Can `Group` have both `ListCollection` and `MapCollection` or one or the other?

Would it be possible to mirror this class hierarchy for the `Group` class as well, for example, could we have a `Group` class with no access methods and have two sub-classes `GroupListCollection` and `GroupMapCollection` (probably come up with better names)? 

This would require the caller to check the type and down-cast accordingly. This is likely to cause turbulence  in user code, so, I am not entirely sure it is the right approach.

Perhaps, this refactoring can be done incrementally, i.e., add the downcasts inside `Group` for now and `SLIC_ERROR` when an invalid access operation is encountered and in the future refactor and refine the design of the `Group` class etc.

nselliott:
> 
> @nselliott -- This may be more involved than I thought.
> 
> I do not have a crystal clear, overarching picture of Sidre's design and implementation, but, based on what you said, I think it will be better to avoid :
> 
>     * Downcasting inside `Group`, and,
> 
>     * Percolating the problem to the `Group` class
> 
> 
> Can `Group` have both `ListCollection` and `MapCollection` or one or the other?

A single instance uses one or the other, based on a boolean in the constructor.

> 
> Would it be possible to mirror this class hierarchy for the `Group` class as well, for example, could we have a `Group` class with no access methods and have two sub-classes `GroupListCollection` and `GroupMapCollection` (probably come up with better names)?
> 
> This would require the caller to check the type and down-cast accordingly. This is likely to cause turbulence in user code, so, I am not entirely sure it is the right approach.

I think that would be an unnecessary complication of the design and API.  We would have to maintain two implementations of `Group` whose implementation differences would be largely farmed off to the Collection classes.  As it is now in this PR, users can create a `Group` in either mode with just a boolean.  If they create one in the `ListCollection` mode, it is documented that the name-access methods are not useful.  I do think I should add a run-time check method on the mode, because a code might not know the properties of a `Group` it encounters, especially one passed in from somewhere else.  Checking booleans is much easier than dealing with type-checking and casts.

> 
> Perhaps, this refactoring can be done incrementally, i.e., add the downcasts inside `Group` for now and `SLIC_ERROR` when an invalid access operation is encountered and in the future refactor and refine the design of the `Group` class etc.

The design principle that has been used in Sidre is to not use errors when something invalid is passed into an accessor method, but to give a warning and return a null pointer.  So if someone calls `getView("foo")` and there is no `View` by that name, the code doesn't die, and the user is left to decide what to do about the null pointer.  The behavior when using `ListCollection` is consistent with this principle, it's just more extreme in that all names are invalid or irrelevant.

gzagaris:
> > Would it be possible to mirror this class hierarchy for the `Group` class as well, for example, could we have a `Group` class with no access methods and have two sub-classes `GroupListCollection` and `GroupMapCollection` (probably come up with better names)?
> > This would require the caller to check the type and down-cast accordingly. This is likely to cause turbulence in user code, so, I am not entirely sure it is the right approach.
> 
> I think that would be an unnecessary complication of the design and API. We would have to maintain two implementations of `Group` whose implementation differences would be largely farmed off to the Collection classes. As it is now in this PR, users can create a `Group` in either mode with just a boolean. If they create one in the `ListCollection` mode, it is documented that the name-access methods are not useful. I do think I should add a run-time check method on the mode, because a code might not know the properties of a `Group` it encounters, especially one passed in from somewhere else. Checking booleans is much easier than dealing with type-checking and casts.

Sounds good -- It may be useful to have methods `isListCollection()` and `isMapCollection()` for a `Group` as well, if that doesn't exist already.

> 
> > Perhaps, this refactoring can be done incrementally, i.e., add the downcasts inside `Group` for now and `SLIC_ERROR` when an invalid access operation is encountered and in the future refactor and refine the design of the `Group` class etc.
> 
> The design principle that has been used in Sidre is to not use errors when something invalid is passed into an accessor method, but to give a warning and return a null pointer. So if someone calls `getView("foo")` and there is no `View` by that name, the code doesn't die, and the user is left to decide what to do about the null pointer. The behavior when using `ListCollection` is consistent with this principle, it's just more extreme in that all names are invalid or irrelevant.

Yes, I understand. I am worried about potential users who are not familiar with the internals and erroneously use the class. We've done our due diligence with the documentation, but, I would suggest we have a `SLIC_WARNING_IF` as well (?). This might help mitigate some of the API clunkiness.

cyrush:
Q: Trying to understand does the list collection preserve the insertion order yet?  I think so via the groups existing getView or getGroup by index interface?  

The Group interface doesn't rely on or expose the Collection objects -- so I don't think folks using Group API will be impacted by these changes. It preserves the Sidre API - where you can ask questions to probe aspects of the group, and get defaults back (like nullptr) when you ask for something that doesn't exist.

We probably want to add a method to let a user know if the Group is a list or supports named children.

Also not sure if this should be a construction time option, but not sure of a better approach?

nselliott:
> 
> 
> Q: Trying to understand does the list collection preserve the insertion order yet? I think so via the groups existing getView or getGroup by index interface?

It does unless you start removing items and then add more items.

> 
> The Group interface doesn't rely on or expose the Collection objects -- so I don't think folks using Group API will be impacted by these changes. It preserves the Sidre API - where you can ask questions to probe aspects of the group, and get defaults back (like nullptr) when you ask for something that doesn't exist.
> 
> We probably want to add a method to let a user know if the Group is a list or supports named children.

I will add this.

src/axom/sidre/core/Group.cpp
```diff
@@ -1669,17 +1720,30 @@ void Group::loadExternalData(const hid_t& h5_id)
...
+             bool is_list)
   : m_name(name)
   , m_index(InvalidIndex)
   , m_parent(nullptr)
   , m_datastore(datastore)
-  , m_view_coll(new ViewCollection())
-  , m_group_coll(new GroupCollection())
+  , m_is_list(is_list)
+  , m_view_coll(0)
+  , m_group_coll(0)
```
ltaylor16, line 1731: Should these be initialized with `nullptr`?

ltaylor16:
Is there any performance penalty for making the Collection classes virtual?

src/axom/sidre/core/Group.hpp
```diff
@@ -1156,6 +1156,21 @@ class Group
    */
   bool isEquivalentTo(const Group* other, bool checkName = true) const;
 
+  /*!
+   * \brief Return true if this Group holds items in map format.
+   */
+  bool usingMap()
```
gzagaris, line 1162: Thanks for adding those methods, @nselliott 

1. Those should probably be `const`
2. It may be more consistent withe the rest of the API if they are named as: `isUsingMap()` and `isUsingList()`

cyrush:
Looks good to me!

PR #231, opened 2020-04-25
Fix dispatch when using asynchronous CUDA execution policies with the BVH

### LINKED ISSUE #230: BVH: Using Asynchronous CUDA execution policy defaults to CPU sort
Sorting of the supplied AABBs when constructing the BVH uses a templated function, ``custom_sort`` which is specialized by our Axom execution policies.

For a CUDA execution policy the template specialization is as follows:
https://github.com/LLNL/axom/blob/40b369cd3140393db98e02ecc3712efb5c25594e/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp#L406-L408

However, this only handles the case of using synchronous CUDA execution policies. The custom sort needs to take a second template in this case that specifies whether it is SYNCHRONOUS or ASYNC.

gzagaris:
# Summary

This commit fixes an issue with the template specialization of the `custom_sort()` routine, used internally by the BVH to sort the AABBs, which would dispatch to the CPU sort(), when a CUDA async execution policy is used, i.e., `axom::CUDA_EXEC< BLOCKSIZE, axom::ASYNC >`. Now, the code will correctly dispatch to the GPU sort when a CUDA execution policy is specified irrespective of whether it is synchronous or asynchronous.

Resolves #230

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -51,6 +51,10 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   ex. `level_error` is now `message_error`.
 
 ### Fixed
+- Fixed issue where the BVH would dispatch to the CPU sort() routine when the
+  specified execution policy was CUDA_EXEC async. Now, when the execution policy
+  is CUDA_EXEC the code would correctly dispatch to the GPU sort, using CUB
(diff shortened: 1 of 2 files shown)


gzagaris:
Folks -- this is PR is ready to merge. At your convenience, please take a look.

PR #233, opened 2020-04-27
Fix linking issue with nvToolsExt

gzagaris:
# Summary

Fix issue wherein linking nvToolsExt failed due to not
being able to find the nvToolsExt. This is now handled 
in BLT by passing the appropriate link flags, when the
BLT's registered cuda library is created.

### DIFF
--- a/host-configs/docker/docker-linux-ubuntu16.04-x86_64-gcc@8.1.0.cmake
+++ b/host-configs/docker/docker-linux-ubuntu16.04-x86_64-gcc@8.1.0.cmake
@@ -14,7 +14,7 @@
 # Compiler Spec: gcc@8.1.0
 ##################################
 
-# CMake executable path: /home/axom/axom_tpls/gcc-8.1.0/cmake-3.9.6/bin/cmake
+# CMake executable path: /home/axom/axom_tpls/gcc-8.1.0/cmake-3.10.1/bin/cmake
 
 ##############
 # Compilers
@@ -45,7 +45,7 @@ set(CONDUIT_DIR "${TPL_ROOT}/conduit-master" CACHE PATH "")
 set(MFEM_DIR "${TPL_ROOT}/mfem-4.0" CACHE PATH "")
 
 # hdf5 from uberenv
-set(HDF5_DIR "${TPL_ROOT}/hdf5-1.8.19" CACHE PATH "")
+set(HDF5_DIR "${TPL_ROOT}/hdf5-1.8.21" CACHE PATH "")
 
 # scr not built by uberenv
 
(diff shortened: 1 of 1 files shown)


host-configs/rzansel-blueos_3_ppc64le_ib_p9-xl@nvcc.cmake
```diff
@@ -96,7 +96,7 @@ set(CMAKE_CXX_COMPILER_ID "XL" CACHE PATH "All of BlueOS compilers report clang
 
 set(BLT_FORTRAN_FLAGS "-WF,-C!  -qxlf2003=polymorphic" CACHE PATH "Converts C-style comments to Fortran style in preprocessed files")
 
-set(BLT_EXE_LINKER_FLAGS "-Wl,-rpath,/usr/tce/packages/xl/xl-2019.08.20/lib" CACHE PATH "Adds a missing rpath for libraries associated with the fortran compiler")
+set(BLT_EXE_LINKER_FLAGS "-Wl,-rpath,/usr/tce/packages/xl/xl-2019.08.20/lib -L/usr/tce/packages/cuda/cuda-10.1.168/lib64" CACHE PATH "Adds a missing rpath for libraries associated with the fortran compiler")
```
gzagaris, line 100: Although this works, I think it would be best if BLT does that in `blt_setup_cuda_target()`, e.g., by adding a `LINK_DIRECTORIES` target property to the specified target. Setting `BLT_EXE_LINKER_FLAGS` is handy but, it seems like a large hammer.

I'll look more into BLT and talk to a few folks -- Let me know what you think.

host-configs/rzansel-blueos_3_ppc64le_ib_p9-xl@nvcc.cmake
```diff
@@ -96,7 +96,7 @@ set(CMAKE_CXX_COMPILER_ID "XL" CACHE PATH "All of BlueOS compilers report clang
...
```
gzagaris, line 99: Although this works, I think it would be best if BLT does that in blt_setup_cuda_target(), e.g., by adding a LINK_DIRECTORIES target property to the specified target. Setting BLT_EXE_LINKER_FLAGS is handy but, it seems like a large hammer.

I'll look more into BLT and talk to a few folks -- Let me know what you think.

kennyweiss, line 99: This change will be lost the next time we update our TPLs.  Please incorporate it into axom's spack package.

gzagaris: Thanks @kennyweiss. 

I didn't do that yet here b/c I think this should be handled by BLT's `blt_setup_axom_target()` (See my earlier comment). If we end up hard-coding this in the host-config, I will modify the spack package accordingly.

gzagaris:
The following PR in BLT https://github.com/LLNL/blt/pull/359 handles this more properly.

host-configs/rzansel-blueos_3_ppc64le_ib_p9-xl@nvcc.cmake
gzagaris: See https://github.com/LLNL/blt/pull/359

gzagaris:
I will update BLT in this PR, one the updates are merged in BLT.

gzagaris:
This is ready to go. At your convenience, please take another look at this.

PR #235, opened 2020-04-30
Add arch to raja and umpire, fix cuda path in axom

white238:
Fixing the issue shown in #234 ... host-configs coming eventually.

### DIFF
--- a/host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.1_nvcc_xlf.cmake
+++ b/host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.1_nvcc_xlf.cmake
@@ -29,7 +29,7 @@ set(CMAKE_Fortran_COMPILER "/usr/tce/packages/xl/xl-2019.12.23/bin/xlf2003" CACH
 #------------------------------------------------------------------------------
 
 # Root directory for generated TPLs
-set(TPL_ROOT "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2020_04_28_20_17_54/clang-8.0.1_nvcc_xlf" CACHE PATH "")
+set(TPL_ROOT "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2020_04_29_19_35_53/clang-8.0.1_nvcc_xlf" CACHE PATH "")
 
 set(CONDUIT_DIR "${TPL_ROOT}/conduit-0.5.1" CACHE PATH "")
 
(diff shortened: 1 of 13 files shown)


scripts/uberenv/packages/axom/package.py
```diff
@@ -393,8 +393,13 @@ def hostconfig(self, spec, prefix):
                 cfg.write("#------------------{}\n\n".format("-"*60))
 
                 cfg.write(cmake_cache_option("ENABLE_CUDA", True))
-                cfg.write(cmake_cache_entry("CUDA_TOOLKIT_ROOT_DIR", "/usr/tce/packages/cuda/cuda-10.1.168"))
-                cfg.write(cmake_cache_entry("CMAKE_CUDA_COMPILER", "${CUDA_TOOLKIT_ROOT_DIR}/bin/nvcc"))
+
+                cudatoolkitdir = spec['cuda'].prefix
```
kennyweiss, line 397: :+1:

white238:
this is ready to go!

host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@8.0.1_nvcc_xlf.cmake
```diff
@@ -100,7 +100,7 @@ set(BLT_EXE_LINKER_FLAGS "-Wl,-rpath,/usr/tce/packages/xl/xl-2019.12.23/lib" CAC
 
 set(ENABLE_CUDA ON CACHE BOOL "")
 
-set(CUDA_TOOLKIT_ROOT_DIR "/usr/tce/packages/cuda/cuda-10.1.168" CACHE PATH "")
+set(CUDA_TOOLKIT_ROOT_DIR "/usr/tce/packages/cuda/cuda-10.1.243" CACHE PATH "")
```
gzagaris, line 103: It may be good to set `AXOM_USE_CUB=ON` on this host-config, so that we have one host-config that tests that functionality as well.

white238: im not rebuilding host-configs again... feel free to do that after this is merged.

gzagaris: Sure, we can do later -- it doesn't require a rebuild, we just set it in the host-config and update the axom's spack package when it generates the host-config.

mclarsen:
Does Umpire need to know about the arch? I thought it was only making cuda API calls and not compiling kernels.

davidbeckingsale:
We build a tiny bit of device code if constant memory is turned on (it's off by default)

gzagaris:
> Does Umpire need to know about the arch? I thought it was only making cuda API calls and not compiling kernels.

That's a good question -- In the generated PTX I saw umpire introducing code for the  `sm30` architecture. I am not sure exactly where it is coming from. Perhaps, it is only needed when building Umpire with constant device memory, which we do.

gzagaris:
> > Does Umpire need to know about the arch? I thought it was only making cuda API calls and not compiling kernels.
> 
> That's a good question -- In the generated PTX I saw umpire introducing code for the `sm30` architecture. I am not sure exactly where it is coming from. Perhaps, it is only needed when building Umpire with constant device memory, which we do.

(just saw David's response)

davidbeckingsale:
Also, note that Umpire doesn't do _anything_ with CUDA flags or architectures, and will respect whatever is set in `CMAKE_CUDA_FLAGS`.

PR #247, opened 2020-05-09
Adds sidre test to verify proper importing of exported conduit data

kennyweiss:
# Summary

- This PR adds a Sidre test to verify that we can properly roundtrip to/from Conduit's native layout.
- Specifically, I wanted to verify that everything works properly when we have a zero-sized array View
- This came up while debugging an issue in a user code
- Sidre appears to have been working correctly: I did not have to change anything in Sidre.

### DIFF
--- a/src/axom/core/Macros.hpp
+++ b/src/axom/core/Macros.hpp
@@ -28,9 +28,11 @@
 #if defined(__CUDACC__)
 #define AXOM_DEVICE __device__
 #define AXOM_HOST_DEVICE __host__ __device__
+#define AXOM_HOST __host__
 #else
 #define AXOM_DEVICE
 #define AXOM_HOST_DEVICE
+#define AXOM_HOST
 #endif
 
 /*!
--- a/src/axom/sidre/tests/sidre_native_layout.cpp
+++ b/src/axom/sidre/tests/sidre_native_layout.cpp
@@ -7,6 +7,8 @@
 
 #include "axom/sidre/core/sidre.hpp"
 
+#include "conduit_blueprint.hpp"
+
 #include <vector>
 #include <string>
 #include <iostream>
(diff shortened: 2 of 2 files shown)


kennyweiss:
I added a second test that checks that we can export to conduit and verify the `mcarray` blueprint schema for zero-sized multi-component arrays.

src/axom/sidre/tests/sidre_native_layout.cpp
```diff
@@ -264,6 +266,254 @@ TEST(sidre_native_layout,native_layout_with_scalars)
...
+    axom::sidre::DataStore ds1, ds2;
+    conduit::Node node1, node2;
+
+    int* conduit_data = nullptr;
+
+    std::string grpName = "some_group";
+    std::string viewName = "array_view";
+
+    // Simple lambda to return the expected value at an index
+    auto expValue = [=](int i) {
```
gzagaris, line 289: We define `AXOM_HOST_LAMBDA` as well, which could be used here instead (and in other places below)

kennyweiss: Thanks @gzagaris -- I'll update it.

src/axom/core/Macros.hpp
```diff
@@ -28,9 +28,11 @@
 #if defined(__CUDACC__)
 #define AXOM_DEVICE __device__
 #define AXOM_HOST_DEVICE __host__ __device__
+#define AXOM_HOST __host__
```
gzagaris, line 31: This is redundant, since not having a decorator is effectively a `HOST` functions.

kennyweiss: `AXOM_HOST` doesn't otherwise appear to be defined.  
See the definition of `AXOM_HOST_LAMBDA` below.

Are you suggesting removing the `__host__`  part?
Or removing `AXOM_HOST` from the CUDA part of `AXOM_HOST_LAMBDA` ?

kennyweiss: Here is the code I was referring to (line 53):

https://github.com/LLNL/axom/blob/889cf6c8ca31ac5b2f6d9e3a63c065c1b754cb32/src/axom/core/Macros.hpp#L50-L58

gzagaris: I see (who did this? 😄 ) -- So, my preference would be to remove `AXOM_HOST` from the `AXOM_HOST_LAMBDA` definition. But, honestly either way is fine.

kennyweiss: Thanks @gzagaris.

I'm going to keep this as is for this PR since I've already tested in with our host-configs on LC and being more explicit in this case can't hurt. 

We can always remove it in the future if that is a better approach.

PR #249, opened 2020-05-13
Triangle Intersection Algorithm using BVH

bmhan12:
This PR adds a RAJA implementation of the triangle - triangle intersection algorithm in `mesh_tester.cpp` that uses a Bounding Volume Hierarchy spatial index.

Also, a ` AXOM_SUPPRESS_HD_WARN` macro was added to suppress host device warnings.

### DIFF
--- a/src/axom/core/Macros.hpp
+++ b/src/axom/core/Macros.hpp
@@ -35,6 +35,22 @@
 #define AXOM_HOST
 #endif
 
+/*
+ * \def AXOM_SUPPRESS_HD_WARN
+ *
+ * \brief Macro used to silence __host__ __device__ compiler warnings
+ *  when calling __host__ function from __host__ __device function.
+ */
+#if defined(__CUDACC__)
+  #ifdef WIN32
+  #define AXOM_SUPPRESS_HD_WARN __pragma(nv_exec_check_disable)
+  #else
+  #define AXOM_SUPPRESS_HD_WARN _Pragma("nv_exec_check_disable")
+  #endif
+#else
+#define AXOM_SUPPRESS_HD_WARN
+#endif
+
 /*!
  * \def AXOM_LAMBDA
  *
--- a/src/axom/core/Macros.hpp
+++ b/src/axom/core/Macros.hpp
@@ -36,14 +36,22 @@
 #endif
 
 /*
+ * \def AXOM_STRINGIFY
+ *
+ * \brief Helper Macro to specify strings inside pragmas.
+ */
+#define AXOM_STRINGIFY(x) AXOM_DO_STRINGIFY(x)
+#define AXOM_DO_STRINGIFY(x) #x
+
+/*
  * \def AXOM_PRAGMA
  *
  * \brief Macro used to specify pragma directive
  */
 #ifdef _WIN32
-#define AXOM_PRAGMA(__supplied_pragma) __pragma(__supplied_pragma)
+#define AXOM_PRAGMA(x) __pragma(x)
 #else
-#define AXOM_PRAGMA(__supplied_pragma) _Pragma(#__supplied_pragma)
+#define AXOM_PRAGMA(x) _Pragma(AXOM_STRINGIFY(x))
 #endif
 
 /*
--- a/src/axom/core/Macros.hpp
+++ b/src/axom/core/Macros.hpp
@@ -36,17 +36,24 @@
 #endif
 
 /*
+ * \def AXOM_PRAGMA
+ *
+ * \brief Macro used to specify pragma directive
+ */
+#ifdef _WIN32
+#define AXOM_PRAGMA(__supplied_pragma) __pragma(__supplied_pragma)
+#else
+#define AXOM_PRAGMA(__supplied_pragma) _Pragma(#__supplied_pragma)
+#endif
+
+/*
  * \def AXOM_SUPPRESS_HD_WARN
  *
  * \brief Macro used to silence __host__ __device__ compiler warnings
- *  when calling __host__ function from __host__ __device function.
+ *  when calling __host__ function from __host__ __device__ function.
  */
 #if defined(__CUDACC__)
-  #ifdef WIN32
-  #define AXOM_SUPPRESS_HD_WARN __pragma(nv_exec_check_disable)
-  #else
-  #define AXOM_SUPPRESS_HD_WARN _Pragma("nv_exec_check_disable")
-  #endif
+#define AXOM_SUPPRESS_HD_WARN AXOM_PRAGMA(nv_exec_check_disable)
 #else
 #define AXOM_SUPPRESS_HD_WARN
 #endif
--- a/src/axom/core/execution/execution_space.hpp
+++ b/src/axom/core/execution/execution_space.hpp
@@ -83,6 +83,7 @@ struct execution_space
 
   static constexpr bool async() noexcept { return false; };
   static constexpr bool valid() noexcept { return false; };
+  static constexpr bool onDevice() noexcept { return false; };
   static constexpr char* name() noexcept { return (char*)"[UNDEFINED]"; };
   static int allocatorID() noexcept { return axom::INVALID_ALLOCATOR_ID; };
 };
--- a/src/axom/core/execution/internal/cuda_exec.hpp
+++ b/src/axom/core/execution/internal/cuda_exec.hpp
@@ -79,6 +79,7 @@ struct execution_space< CUDA_EXEC< BLOCK_SIZE, ASYNC > >
 
   static constexpr bool async() noexcept { return true; };
   static constexpr bool valid() noexcept { return true; };
+  static constexpr bool onDevice() noexcept { return true; };
   static constexpr char* name() noexcept {
     return (char*)"[CUDA_EXEC] (async)";
   };
--- a/src/axom/core/execution/internal/cuda_exec.hpp
+++ b/src/axom/core/execution/internal/cuda_exec.hpp
@@ -55,6 +55,7 @@ struct execution_space< CUDA_EXEC< BLOCK_SIZE, SYNCHRONOUS > >
 
   static constexpr bool async() noexcept { return false; };
   static constexpr bool valid() noexcept { return true; };
+  static constexpr bool onDevice() noexcept { return true; };
   static constexpr char* name() noexcept { return (char*)"[CUDA_EXEC]"; };
   static int allocatorID() noexcept
   { return axom::getUmpireResourceAllocatorID(umpire::resource::Unified); };
--- a/src/axom/primal/geometry/Triangle.hpp
+++ b/src/axom/primal/geometry/Triangle.hpp
@@ -59,7 +59,7 @@ public:
    * \brief Default constructor. Creates a degenerate triangle.
    */
   AXOM_HOST_DEVICE
-  Triangle() { }
+  Triangle();
 
   /*!
    * \brief Custom Constructor. Creates a triangle from the 3 points A,B,C.
@@ -312,6 +312,12 @@ namespace axom
 namespace primal
 {
 
+template < typename T, int NDIMS >
+Triangle< T,NDIMS >::Triangle()
+   : m_points{ PointType(), PointType(), PointType() }
+{}
+
+//------------------------------------------------------------------------------
 template < typename T, int NDIMS >
 Triangle< T,NDIMS >::Triangle( const PointType& A,
                                const PointType& B,
--- a/src/axom/primal/operators/compute_bounding_box.hpp
+++ b/src/axom/primal/operators/compute_bounding_box.hpp
@@ -113,7 +113,7 @@ AXOM_HOST_DEVICE
 BoundingBox< T, NDIMS > compute_bounding_box( const Triangle< T, NDIMS > &tri)
 {
   BoundingBox< T, NDIMS > res(tri[0]);
-  for (int i = 1; i < NDIMS; i++)
+  for (int i = 1; i < 3; i++)
   {
     res.addPoint(tri[i]);
   }
--- a/src/axom/primal/operators/compute_bounding_box.hpp
+++ b/src/axom/primal/operators/compute_bounding_box.hpp
@@ -16,6 +16,7 @@
 #include "axom/core/numerics/Matrix.hpp" // for Matrix
 #include "axom/core/numerics/eigen_solve.hpp" // for eigen_solve
 #include "axom/primal/geometry/Point.hpp"
+#include "axom/primal/geometry/Triangle.hpp"
 #include "axom/primal/geometry/Vector.hpp"
 #include "axom/primal/geometry/OrientedBoundingBox.hpp"
 
@@ -97,6 +98,22 @@ BoundingBox< T, NDIMS > merge_boxes( const BoundingBox< T, NDIMS >
 {
   BoundingBox< T, NDIMS > res(l);
   res.addBox(r);
+  return res;
+}
+
+/*!
+ * \brief Creates a bounding box around a Triangle
+ *
+ * \param [in] tri The Triangle
+ */
+template < typename T, int NDIMS >
+BoundingBox< T, NDIMS > compute_bounding_box( const Triangle< T, NDIMS > &tri)
+{
+  BoundingBox< T, NDIMS > res;
+  for (int i = 0; i < NDIMS; i++)
+  {
+    res.addPoint(tri[i]);
+  }
   return res;
 }
 
--- a/src/axom/primal/operators/compute_bounding_box.hpp
+++ b/src/axom/primal/operators/compute_bounding_box.hpp
@@ -14,6 +14,7 @@
 
 #include "axom/primal/geometry/NumericArray.hpp" // for numeric arrays
 #include "axom/core/numerics/Matrix.hpp" // for Matrix
+#include "axom/core/Macros.hpp" // for AXOM_HOST__DEVICE
 #include "axom/core/numerics/eigen_solve.hpp" // for eigen_solve
 #include "axom/primal/geometry/Point.hpp"
 #include "axom/primal/geometry/Triangle.hpp"
@@ -92,6 +93,7 @@ OrientedBoundingBox< T, NDIMS > merge_boxes( const OrientedBoundingBox< T, NDIMS
  * \param [in] l left bb
  * \param [in] r right bb
  */
+
 template < typename T, int NDIMS >
 BoundingBox< T, NDIMS > merge_boxes( const BoundingBox< T, NDIMS >
                                      &l, const BoundingBox< T, NDIMS > &r)
@@ -107,10 +109,11 @@ BoundingBox< T, NDIMS > merge_boxes( const BoundingBox< T, NDIMS >
  * \param [in] tri The Triangle
  */
 template < typename T, int NDIMS >
+AXOM_HOST_DEVICE
 BoundingBox< T, NDIMS > compute_bounding_box( const Triangle< T, NDIMS > &tri)
 {
-  BoundingBox< T, NDIMS > res;
-  for (int i = 0; i < NDIMS; i++)
+  BoundingBox< T, NDIMS > res(tri[0]);
+  for (int i = 1; i < NDIMS; i++)
   {
     res.addPoint(tri[i]);
   }
--- a/src/axom/quest/MeshTester.hpp
+++ b/src/axom/quest/MeshTester.hpp
@@ -6,10 +6,29 @@
 #ifndef MESH_TESTER_HPP_
 #define MESH_TESTER_HPP_
 
+#include "axom/core.hpp"
+#include "axom/primal.hpp"
+#include "axom/spin.hpp"
+#include "axom/mint.hpp"
+
+// C++ includes
+#include <cmath>
+#include <algorithm>
+#include <vector>
+#include <unordered_map>
+#include <functional> // for std::hash
+
 // Axom includes
 #include "axom/config.hpp"
 #include "axom/mint/mesh/UnstructuredMesh.hpp"
 #include "axom/slic/interface/slic.hpp"
+
+// BVH includes
+#if defined(AXOM_USE_RAJA)
+  #include "RAJA/RAJA.hpp"
+  #include "axom/spin/BVH.hpp"
+  #include "axom/mint/execution/internal/structured_exec.hpp"
+#endif
 
 // C/C++ includes
 #include <utility>
@@ -25,6 +44,15 @@ namespace axom
 namespace quest
 {
 
+using UMesh = mint::UnstructuredMesh< mint::SINGLE_SHAPE >;
+using Triangle3 = primal::Triangle<double, 3>;
+
+using Point3 = primal::Point<double, 3>;
+using SpatialBoundingBox = primal::BoundingBox<double, 3>;
+using UniformGrid3 = spin::UniformGrid<int, 3>;
+using Vector3 = primal::Vector<double, 3>;
+using Segment3 = primal::Segment<double, 3>;
+
 /*! Enumeration indicating mesh watertightness */
 enum class WatertightStatus : signed char
 {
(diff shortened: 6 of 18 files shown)


bmhan12:
@gzagaris I am encountering a problem where my bvh triangle intersection query will sometimes return less than the expected number of intersections. This only happens when I use a CUDA policy in RAJA; the OpenMP and sequential policies return the same result. 

I did a `diff` of the member variable contents between queries using a CUDA policy and sequential policy at different stages of `BVH.build()`. After the `propagate_aabbs()` call in `build_radix_tree()`, I noticed there were some AABBs in the `m_inner_aabbs` member of the RadixTree object that had `0.0` values in the `min` coordinate when run with a CUDA policy but not with a sequential policy:
```
$ diff cuda_output.txt seq_output.txt            
1c1
< IN RadixTree: m_inner_aabbs[ 0 ] is x-range: [0, 100.857]; y-range: [12.9616, 57.0383]; z-range: [0, 37.432]
---
> IN RadixTree: m_inner_aabbs[ 0 ] is x-range: [2.12674, 100.857]; y-range: [12.9616, 57.0383]; z-range: [7.5006, 37.432]
84384c84384
< IN RadixTree: m_inner_aabbs[ 84383 ] is x-range: [0, 99.137]; y-range: [12.9616, 57.0383]; z-range: [0, 37.432]
---
> IN RadixTree: m_inner_aabbs[ 84383 ] is x-range: [8.02208, 99.137]; y-range: [12.9616, 57.0383]; z-range: [21.7348, 37.432]
96806c96806
< IN RadixTree: m_inner_aabbs[ 96805 ] is x-range: [47.5546, 99.137]; y-range: [14.1554, 44.4068]; z-range: [0, 37.432]
---
> IN RadixTree: m_inner_aabbs[ 96805 ] is x-range: [47.5546, 99.137]; y-range: [14.1554, 44.4068]; z-range: [22.164, 37.432]
103328c103328
< IN RadixTree: m_inner_aabbs[ 103327 ] is x-range: [80.0539, 99.137]; y-range: [23.5458, 35.0012]; z-range: [0, 30.6476]
---
> IN RadixTree: m_inner_aabbs[ 103327 ] is x-range: [80.0539, 99.137]; y-range: [23.5458, 35.0012]; z-range: [22.2783, 30.6476]
104136c104136
< IN RadixTree: m_inner_aabbs[ 104135 ] is x-range: [80.0539, 99.137]; y-range: [23.749, 35.0012]; z-range: [0, 30.6476]
---
> IN RadixTree: m_inner_aabbs[ 104135 ] is x-range: [80.0539, 99.137]; y-range: [23.749, 35.0012]; z-range: [25.0012, 30.6476]
```
Could this be affecting the number of candidates?

gzagaris:
> @gzagaris I am encountering a problem where my bvh triangle intersection query will sometimes return less than the expected number of intersections. This only happens when I use a CUDA policy in RAJA; the OpenMP and sequential policies return the same result.
> 
> I did a `diff` of the member variable contents between queries using a CUDA policy and sequential policy at different stages of `BVH.build()`. After the `propagate_aabbs()` call in `build_radix_tree()`, I noticed there were some AABBs in the `m_inner_aabbs` member of the RadixTree object that had `0.0` values in the `min` coordinate when run with a CUDA policy but not with a sequential policy:
> 
> ```
> $ diff cuda_output.txt seq_output.txt            
> 1c1
> < IN RadixTree: m_inner_aabbs[ 0 ] is x-range: [0, 100.857]; y-range: [12.9616, 57.0383]; z-range: [0, 37.432]
> ---
> > IN RadixTree: m_inner_aabbs[ 0 ] is x-range: [2.12674, 100.857]; y-range: [12.9616, 57.0383]; z-range: [7.5006, 37.432]
> 84384c84384
> < IN RadixTree: m_inner_aabbs[ 84383 ] is x-range: [0, 99.137]; y-range: [12.9616, 57.0383]; z-range: [0, 37.432]
> ---
> > IN RadixTree: m_inner_aabbs[ 84383 ] is x-range: [8.02208, 99.137]; y-range: [12.9616, 57.0383]; z-range: [21.7348, 37.432]
> 96806c96806
> < IN RadixTree: m_inner_aabbs[ 96805 ] is x-range: [47.5546, 99.137]; y-range: [14.1554, 44.4068]; z-range: [0, 37.432]
> ---
> > IN RadixTree: m_inner_aabbs[ 96805 ] is x-range: [47.5546, 99.137]; y-range: [14.1554, 44.4068]; z-range: [22.164, 37.432]
> 103328c103328
> < IN RadixTree: m_inner_aabbs[ 103327 ] is x-range: [80.0539, 99.137]; y-range: [23.5458, 35.0012]; z-range: [0, 30.6476]
> ---
> > IN RadixTree: m_inner_aabbs[ 103327 ] is x-range: [80.0539, 99.137]; y-range: [23.5458, 35.0012]; z-range: [22.2783, 30.6476]
> 104136c104136
> < IN RadixTree: m_inner_aabbs[ 104135 ] is x-range: [80.0539, 99.137]; y-range: [23.749, 35.0012]; z-range: [0, 30.6476]
> ---
> > IN RadixTree: m_inner_aabbs[ 104135 ] is x-range: [80.0539, 99.137]; y-range: [23.749, 35.0012]; z-range: [25.0012, 30.6476]
> ```
> 
> Could this be affecting the number of candidates?

Thanks @bmhan12 -- This looks like a bug. I would expect that we generate the same BVH on both the GPU and CPU.

By default, `AXOM_ENABLE_CUB` is set to `ON` on rzansel in which case, CUB's sort is used to  sort the bounding boxes. On the CPU (sequentially or with OpenMP), we are using `stable_sort()`. I believe CUB's sort is also a "stable_sort()", but they don't explicitly state anything about this on their documentation.

I'll look more closely at this in the afternoon.

rhornung67:
@gzagaris could this be rooted in data being initialized differently on host vs. device?

gzagaris:
> @gzagaris could this be rooted in data being initialized differently on host vs. device?

Yes, if the input is different, that would do it. I am assuming that is not the case, but, that would be an easy thing to check and rule out.

src/tools/mesh_tester.cpp
```diff
@@ -132,12 +132,14 @@ void Input::parse(int argc, char** argv, CLI::App& app)
   app.add_option("-r,--resolution", resolution,
                  "Resolution of uniform grid. \n"
                  "Set to 1 to run the naive algorithm (without a spatial index). \n"
-                 "Set to less than 1 to use the spatial index with a resolution \n"
-                 "of the cube root of the number of triangles.")
+                 "Set to 2 to use a bounding volume hierarchy spatial index.\n"
```
kennyweiss, line 135: @bmhan12 -- in terms of user interface, I think the choice between the naive `O(n^2)` algorithm and the spatial index should be a separate input parameter, maybe `-m,--method == {naive, spatial_index}`  (or something better?)

bmhan12: A separate `method` parameter makes a lot of sense. Overloading the `resolution` parameter as-is makes things unclear.

src/axom/core/Macros.hpp
```diff
@@ -35,6 +35,22 @@
...
 
+/*
+ * \def AXOM_SUPPRESS_HD_WARN
+ *
+ * \brief Macro used to silence __host__ __device__ compiler warnings
+ *  when calling __host__ function from __host__ __device function.
+ */
+#if defined(__CUDACC__)
+  #ifdef WIN32
+  #define AXOM_SUPPRESS_HD_WARN __pragma(nv_exec_check_disable)
```
gzagaris, line 46: I suggest we split this into two separate macros: (a) `AXOM_PRAGMA` and (b) `AXOM_SUPPRESS_HD_WARN`. That way, we can reuse the `AXOM_PRAGMA` in other places as needed.

Something like this:
```
#ifdef WIN32
#define AXOM_PRAGMA(__supplied_pragma) __pragma(__supplied_pragma)
#else 
#define AXOM_PRAGMA(__supplied_pragma) _Pragma("__supplied_pragma")
#endif

#if defined(__CUDACC__)
#define AXOM_SUPPRESS_HD_WARN AXOM_PRAGMA(nv_exec_check_disable)
#else 
#define AXOM_SUPPRESS_HD_WARN AXOM_PRAGMA
#endif 
```

**Note:** There is a way to handle the `"` inside a macro, I think you just add a`#` in front of it or something along those lines

src/axom/quest/MeshTester.hpp
```diff
@@ -33,12 +61,196 @@ enum class WatertightStatus : signed char
...
+  FloatType* zmax = axom::allocate< FloatType >( ncells );
+
+  // Each access-aligned bounding box represented by 2 (x,y,z) points
+  FloatType* aabbs = axom::allocate< FloatType >( ncells * stride );
+
+  // Finding degenerate indices and bounding boxes from Triangles
+  Triangle3 t1 = Triangle3();
+  Triangle3 t2 = Triangle3();
+
+  for (int i=0 ; i < ncells ; i++)
```
gzagaris, line 154: You may use `mint::for_all_cells< exec, xargs::coords >` as we do in the `tests/spin_bvh.cpp` to compute the AABBs in parallel.

src/axom/quest/MeshTester.hpp
```diff
@@ -33,12 +61,196 @@ enum class WatertightStatus : signed char
...
+  {
+    t1=getMeshTriangle(i, surface_mesh);
+    tris[i] = t1;
+
+    if (t1.degenerate())
+    {
+      degenerateIndices.push_back(i);
```
gzagaris, line 161: Instead of using a vector here and calling `push_back()`, which is not parallel friendly, I suggest we allocate an integer array of size `ncells` to mark degenerate cells with a `1`' or `0`. In fact, I think there is code somewhere in this file that creates such an as a cell-centered field variable and that's what we use to visualize the degenerate cells.

src/axom/quest/MeshTester.hpp
```diff
@@ -33,12 +61,196 @@ enum class WatertightStatus : signed char
...
+
+  int* intersection_pairs =
+    axom::allocate <int> (totalCandidates * 2);
+
+  using ATOMIC_POL =
+          typename axom::execution_space< ExecSpace >::atomic_policy;
+  int* counter = axom::allocate <int> (1);
+  counter[0] = 0;
+
+  for_all< ExecSpace >( ncells, AXOM_LAMBDA (IndexType i)
```
kennyweiss, line 215: I think this loop over `ncells` is the main source of the poor performance on the GPU! It is likely to be very load imbalanced.

Is there a way to iterate over the returned candidates, rather than over the indices of the triangles in the original mesh?
Many of the input triangles will have no candidates, and some will have a lot.

My suggestion for the above would be to have a separate pass that expands the `candidates` array and the `offsets` array into two arrays of size `# candidates`, where each entry corresponds to the first and second triangle index of a candidate pair, respectively.  

Optional: You could also potentially use this as an opportunity to filter out the candidates where `i <= candidate_index`.

src/axom/quest/MeshTester.hpp
```diff
@@ -33,12 +61,196 @@ enum class WatertightStatus : signed char
...
+          intersection_pairs[idx + 1] = candidate_index;
+        }
+      }
+    }
+  } );
+
+  // Initialize pairs of clashes
+  for (int i = 0 ; i < counter[0] ; i += 2)
+  {
+    intersections.push_back(std::make_pair(intersection_pairs[i],
```
kennyweiss, line 235: I think I'm missing something -- when does the memory associated with `intersection_pairs` get moved from device to host (for the `raja-cuda` runs)?

src/axom/quest/MeshTester.hpp
```diff
@@ -33,12 +61,196 @@ enum class WatertightStatus : signed char
...
+  // Each access-aligned bounding box represented by 2 (x,y,z) points
+  FloatType* aabbs = axom::allocate< FloatType >( ncells * stride );
+
+  // Finding degenerate indices and bounding boxes from Triangles
+  Triangle3 t1 = Triangle3();
+  Triangle3 t2 = Triangle3();
+
+  for (int i=0 ; i < ncells ; i++)
+  {
+    t1=getMeshTriangle(i, surface_mesh);
```
kennyweiss, line 156: This is on the host, right?

src/axom/quest/MeshTester.hpp
```diff
@@ -33,12 +61,196 @@ enum class WatertightStatus : signed char
...
+    if (t1.degenerate())
+    {
+      degenerateIndices.push_back(i);
+    }
+
+    SpatialBoundingBox triBB = compute_bounds(t1);
+
+    const IndexType offset = i * stride;
+
+    xmin[i] = aabbs[ offset ]     = triBB.getMin()[0];
```
kennyweiss, line 168:
```cpp
     xmin[i] = aabbs[ offset ]     = triBB.getMin()[0]; 
```

Are these arrays on the host as well?

kennyweiss: I don't see explicit memory copies.  Is this function using unified memory? That might be another source of inefficiency for the GPU version.  

If so, I don't think it would take too much effort with axom's memory manager (using umpire) to use explicit memory transfers on a few arrays.

If not, I am not sure how the results are getting transferred.  Perhaps the bounding box bug is somehow related to this?

src/axom/quest/MeshTester.hpp
kennyweiss: This should be combined with the previous suggestion to use a kernel to check for degeneracies and compute the bounding boxes.

src/axom/quest/MeshTester.hpp
gzagaris: This code is definitely relying on Unified memory. 

Unfortunately, using unified memory will still have memory motion even if the data is not accessed on the host, it depends on the driver/runtime, so we cannot really rely on it. 

You may be able to create a unified memory allocator with hints, which may offer some benefits.

My suggestion is to:

1. Add annotations (we now have annotation macros) and look a the profile with `nvvp`

2. Use a device memory pool and use explicit memory copies back to the host (this would have tremendous impact in performance)

3. Pass the pool allocator when constructing the BVH so that the BVH will use the pool allocator internally as well.

kennyweiss: Thanks @gzagaris -- found it -- https://github.com/LLNL/axom/blob/0ed1af3682b6fdf71bc08c4788f94aa0d54993f1/src/axom/core/execution/internal/cuda_exec.hpp#L60

src/axom/quest/MeshTester.hpp
```diff
@@ -33,12 +61,196 @@ enum class WatertightStatus : signed char
...
+    ymin[i] = aabbs[ offset + 1 ] = triBB.getMin()[1];
+    zmin[i] = aabbs[ offset + 2 ] = triBB.getMin()[2];
+
+    xmax[i] = aabbs[ offset + 3 ] = triBB.getMax()[0];
```
kennyweiss, line 172:
```cpp
  xmax[i] = aabbs[ offset + 3 ] = triBB.getMax()[0]; 
```
Any chance these double assignments are the cause of the AABB initialization bug?
(I'm not sure how or why that would happen though).

src/axom/quest/MeshTester.hpp
```diff
@@ -33,12 +61,196 @@ enum class WatertightStatus : signed char
...
+    ymax[i] = aabbs[ offset + 4 ] = triBB.getMax()[1];
+    zmax[i] = aabbs[ offset + 5 ] = triBB.getMax()[2];
+  }
+
+  // Construct BVH
+  axom::spin::BVH< NDIMS, ExecSpace, FloatType > bvh( aabbs, ncells );
+  bvh.setScaleFactor( 1.0 ); // no scaling
```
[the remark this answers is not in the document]
mclarsen: I would go as far as to say that the scale factor should be enforced to be > 1.

gzagaris:
I believe the default scale factor is set to 1.001. Some of the unit tests disable scaling b/c the tests are tailored made for an axis-aligned uniform mesh. But, I agree we should remove this line from here.

src/tools/mesh_tester.cpp
bmhan12: I added a `-m,--method == {"naive", "uniform", "bvh"}` parameter to switch between the algorithms.

src/axom/primal/geometry/Triangle.hpp
```diff
@@ -59,7 +59,7 @@ class Triangle
    * \brief Default constructor. Creates a degenerate triangle.
    */
   AXOM_HOST_DEVICE
-  Triangle() { }
+  Triangle();
```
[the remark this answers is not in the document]
bmhan12: I kept the constructor as-is. I was getting the nvcc compiler error warnings when specifying the `default` constructor.

src/axom/quest/MeshTester.hpp
bmhan12: > My suggestion is to:
> 
> 
> Add annotations (we now have annotation macros) and look a the profile with nvvp
> 
> 
> Use a device memory pool and use explicit memory copies back to the host (this would have tremendous impact in performance)
> 
> 
> Pass the pool allocator when constructing the BVH so that the BVH will use the pool allocator internally as well.

Thanks for the suggestion!
Here was the profile of the `findTriMeshIntersectionsBVH` function using Unified memory:

nvvp mesh tester no pool

I added a device pool to reduce the memory motion:
 
nvvp mesh tester device pool

bmhan12:
@gzagaris @kennyweiss I made further changes to address the PR comments. Please give it a look when you have a chance.

src/axom/quest/MeshTester.hpp
gzagaris: Outstanding!

src/axom/primal/operators/compute_bounding_box.hpp
```diff
@@ -100,6 +103,23 @@ BoundingBox< T, NDIMS > merge_boxes( const BoundingBox< T, NDIMS >
...
+ * \brief Creates a bounding box around a Triangle
+ *
+ * \param [in] tri The Triangle
+ */
+template < typename T, int NDIMS >
+AXOM_HOST_DEVICE
+BoundingBox< T, NDIMS > compute_bounding_box( const Triangle< T, NDIMS > &tri)
+{
+  BoundingBox< T, NDIMS > res(tri[0]);
+  for (int i = 1; i < NDIMS; i++)
```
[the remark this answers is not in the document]
bmhan12: Good  catch!

src/tools/mesh_tester.cpp
```diff
@@ -111,30 +113,44 @@ struct Input
...
-                 "Set to 1 to run the naive algorithm (without a spatial index). \n"
-                 "Set to less than 1 to use the spatial index with a resolution \n"
-                 "of the cube root of the number of triangles.")
+                 "With \'-m uniform\', set resolution of uniform grid. \n"
+                 "Set to less than 1 to use the uniform grid spatial index\n"
+                 "with a resolution of the cube root of the number of\n"
+                 "triangles.")
   ->capture_default_str();
 
   app.add_option("-p, --policy", policy,
```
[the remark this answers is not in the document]
bmhan12: I added the map `{{"seq",0}, {"raja_seq",1}, {"raja_omp",2}, {"raja_cuda",3}}` so the policy can be set using either the string key or integer value. I tried just string values at first, but they aren't valid with the switch-case statements.

src/axom/quest/MeshTester.hpp
```diff
@@ -33,12 +58,296 @@ enum class WatertightStatus : signed char
...
+  // Copy counts data back to host
+  int * host_counts = axom::allocate< int >( ncells, 
+    getUmpireResourceAllocatorID(umpire::resource::Host) );
+  axom::copy(host_counts, counts, ncells * sizeof(int));
+
+  int totalCandidates = 0;
+  for (int i = 0 ; i < ncells ; i++ )
+  {
+    totalCandidates += host_counts[i];
+  }
```
gzagaris, line 242: - Could you compute the total number of candidates in parallel using RAJA reduction types, i.e., `RAJA::ReduceSum` in this case?

- Would this alleviate the need to copy the counts back on the host?

src/axom/quest/MeshTester.hpp
```diff
@@ -33,12 +58,296 @@ enum class WatertightStatus : signed char
...
+      } // END for all cells nodes
+
+      if (tri.degenerate())
+      {
+        degenerate[cellIdx] = 1;
+      }
+      else
+      {
+        degenerate[cellIdx] = 0;
+      }
```
gzagaris, line 188: style suggestion: ``degenerate[ cellIdx ] = ( tri.degenerate() ? 1 : 0 );``

src/axom/quest/MeshTester.hpp
```diff
@@ -33,12 +58,296 @@ enum class WatertightStatus : signed char
...
+            << " in execution Space: "
+            << axom::execution_space< ExecSpace >::name());
+
+  constexpr size_t POOL_SIZE   = (1024 * 1024 * 1024) + 1; 
+  umpire::ResourceManager& rm  = umpire::ResourceManager::getInstance();
+
+  // Use device pool for CUDA policy, host pool otherwise
+  std::string exec_name (axom::execution_space< ExecSpace >::name());
+  umpire::Allocator allocator =
+    (exec_name.find("[CUDA_EXEC]") != std::string::npos ?
```
gzagaris, line 125: Hmm...This can be made cleaner if we'd expand `execution_space` traits class and add another `constexpr bool` method. 

Specifically, I suggest we change the definition in `axom/core/execution_space.hpp`: https://github.com/LLNL/axom/blob/bac157c8834ac6b2e027701461ad5cfb50f0e9ab/src/axom/core/execution/execution_space.hpp#L75-L88

To something like this:
```
template < typename ExecSpace >
struct execution_space
{
  using loop_policy   = void;

  using reduce_policy = void;
  using atomic_policy = void;
  using sync_policy   = void;

  static constexpr bool async() noexcept { return false; };
  static constexpr bool valid() noexcept { return false; };
```
  **``static constexpr bool onDevice() noexcept {return false; };``**
```
  static constexpr char* name() noexcept { return (char*)"[UNDEFINED]"; };
  static int allocatorID() noexcept { return axom::INVALID_ALLOCATOR_ID; };
};
```
Subsequently, each template specialization of the ``execution_space`` traits class implements the ``onDevice`` method accordingly.

This would allow us to call ``axom::execution_space< ExecSpace >::onDevice()`` instead of relying on the name, which is something that would be generally useful as well.

src/axom/core/Macros.hpp
```diff
@@ -35,6 +35,29 @@
...
 
+/*
+ * \def AXOM_PRAGMA
+ *
+ * \brief Macro used to specify pragma directive
+ */
+#ifdef _WIN32
+#define AXOM_PRAGMA(__supplied_pragma) __pragma(__supplied_pragma)
+#else
+#define AXOM_PRAGMA(__supplied_pragma) _Pragma(#__supplied_pragma)
```
gzagaris, line 46: Does the `#` work here correctly to handle the quotes? I recall I had issues in the past with something similar...

gzagaris: Here is what RAJA does:
```
#define RAJA_STRINGIFY(x) RAJA_DO_STRINGIFY(x)
#define RAJA_DO_STRINGIFY(x) #x
#ifdef _WIN32
#define RAJA_PRAGMA(x) __pragma(x)
#else
#define RAJA_PRAGMA(x) _Pragma(RAJA_STRINGIFY(x))
#endif
```

The `STRINGIFY` macro is weird, but, it handles the quotes correctly....

We should follow that same pattern here.

src/axom/quest/MeshTester.hpp
bmhan12: Yes! I totally missed this, thanks for the suggestion.

src/axom/quest/MeshTester.hpp
bmhan12: Added!

src/axom/core/Macros.hpp
bmhan12: Added!

src/axom/core/Macros.hpp
```diff
@@ -35,6 +35,37 @@
 #define AXOM_HOST
 #endif
 
+/*
+ * \def AXOM_STRINGIFY
+ *
+ * \brief Helper Macro to specify strings inside pragmas.
+ */
+#define AXOM_STRINGIFY(x) AXOM_DO_STRINGIFY(x)
+#define AXOM_DO_STRINGIFY(x) #x
```
gzagaris, line 44: Nice!

src/axom/core/execution/internal/cuda_exec.hpp
```diff
@@ -55,6 +55,7 @@ struct execution_space< CUDA_EXEC< BLOCK_SIZE, SYNCHRONOUS > >
 
   static constexpr bool async() noexcept { return false; };
   static constexpr bool valid() noexcept { return true; };
+  static constexpr bool onDevice() noexcept { return true; };
```
gzagaris, line 58: Note, the `cuda_exec.hpp` file consists of two `execution_space` specializations: 

1. ``axom::execution_space< CUDA_EXEC< BLOCK_SIZE, SYNCHRONOUS > >`` , and 
2. ``axom::execution_space< CUDA_EXEC< BLOCK_SIZE, ASYNC > >``

It looks like you only implemented the ``onDevice()`` method for ``CUDA_EXEC< BLOCK_SIZE, SYNCHRONOUS >``. 

Please add it for both. 

In addition, please extend our unit test accordingly in ``core/tests/core_execution_space.cpp``, which is designed to catch this kind of inconsistencies

src/axom/core/execution/execution_space.hpp
```diff
@@ -83,6 +83,7 @@ struct execution_space
 
   static constexpr bool async() noexcept { return false; };
   static constexpr bool valid() noexcept { return false; };
+  static constexpr bool onDevice() noexcept { return false; };
```
gzagaris, line 86: Thank you so much for adding this!

src/axom/quest/MeshTester.hpp
```diff
@@ -33,12 +58,284 @@ enum class WatertightStatus : signed char
...
+            << axom::execution_space< ExecSpace >::name());
+
+  constexpr size_t POOL_SIZE   = (1024 * 1024 * 1024) + 1; 
+  umpire::ResourceManager& rm  = umpire::ResourceManager::getInstance();
+
+  // Use device pool for CUDA policy, host pool otherwise
+  umpire::Allocator allocator =
+    (axom::execution_space< ExecSpace >::onDevice() ? 
+     rm.getAllocator(umpire::resource::Device) :
+     rm.getAllocator(axom::execution_space< ExecSpace >::allocatorID()));
```
gzagaris, line 126: This looks much better now! Thanks!

src/axom/quest/MeshTester.hpp
```diff
@@ -33,12 +58,284 @@ enum class WatertightStatus : signed char
...
+  axom::deallocate(xmin);
+  axom::deallocate(ymin);
+  axom::deallocate(zmin);
+
+  axom::deallocate(xmax);
+  axom::deallocate(ymax);
+  axom::deallocate(zmax);
+
+  int* intersection_pairs =
+    axom::allocate <int> (totalCandidates * 2);
```
gzagaris, line 248: Does this work? I thought you had to do ``totalCandidates.get()`` on the. RAJA reduction type? (but that may have changed)

PR #251, opened 2020-05-20
Refresh SCR integration

adammoody:
# Summary

- This PR updates SPIO and example programs for the SCR API.  It is not yet complete, but I would like to hand this over for further review and final polishing by SPIO developers.
- It updates IOManager.cpp to:
  - Remove SCR_Start/Complete calls which have been elevated up to the example application programs.
  - Add SCR_Route_file calls before opening/creating individual HDF5 data files.
  - Remove mkdir operations when using SCR, since SCR creates directories as needed.
  - It's likely that either additional SCR calls are needed or perhaps they could be moved higher up in the callpath so that fewer are required.
- It adds example programs:
  - IOSCRCkpt - example use of checkpoint / restart within an application
  - IOSCROutput - example use of SCR for pure output dataset (no checkpoint)
  - The IOSCRRead/Write examples should be updated similarly or dropped.
- For applications using SCR with SPIO:
  - As the examples above show, the applications should make Start/Complete calls as needed.  SPIO should then internally call SCR_Route_file calls for its component files.
  - All processes in MPI_COMM_WORLD must call the SCR collective calls (Init/Start/Complete/Finalize), but it is acceptable to pass a subcommunicator to SPIO.  In fact, it should work if the processes in MPI_COMM_WORLD are divided into groups where each group passes its own communicator to SPIO.
  - If the application wants to cache checkpoints in node-local storage, they must be sure to set num_files to the number of processes in the communicator, so that each process writes/reads its own file.  When using node-local cache, two different processes may not share access to a file.
  - It is acceptable to allow shared access to files, but then the user should configure SCR to use bypass mode so that files are read/written on the parallel file system instead of node local cache.
  - One may also write output sets using non-shared mode but read checkpoints back on restart with shared access.  This is useful for applications that may restart with a different number of processes than were used to write the checkpoint.  One can set SCR_GLOBAL_RESTART=1 to support this mode.


adammoody:
@nselliott , here is the PR we were talking about.  I believe this should compile with the SCR you might already be using.  However, it occurs to me now that you'll probably need to update your SCR to get it to run correctly.  When you get to that point, let me know and I can help get you a fresh install to work with.

nselliott:
@adammoody Thank you.  To the axom team, this is based on some discussions with @adammoody where he has given some suggestions that will get our code better aligned with recommended SCR usage and make use of more recent SCR features.  It is not ready and I have some revisions I need to make as well.

rhornung67:
Thanks @adammoody. I thought you only do ML stuff these days? 😄

adammoody:
> 
> 
> Thanks @adammoody. I thought you only do ML stuff these days? 😄

I relapse to old habits at times :-)

adammoody:
@nselliott , it looked to me like all ranks try to read the root file from getProtocol:
https://github.com/LLNL/axom/blob/78a8b77ae2f82ea7fb86d11bbc7773b5055fbc7e/src/axom/sidre/spio/IOManager.cpp#L399

Though in getNumFilesFromRoot and getNumGroupsFromRoot, only rank 0 reads the root file and bcasts the data to others.
https://github.com/LLNL/axom/blob/78a8b77ae2f82ea7fb86d11bbc7773b5055fbc7e/src/axom/sidre/spio/IOManager.cpp#L878
https://github.com/LLNL/axom/blob/78a8b77ae2f82ea7fb86d11bbc7773b5055fbc7e/src/axom/sidre/spio/IOManager.cpp#L902

To help detect other cases where multiple ranks attempt to read from a file, you could configure SCR to store files in node local storage like /dev/shm and then run a test using at least two compute nodes.  The target file would only exist in /dev/shm on one node, so procs on other nodes should get an error when they try to open the file.

adammoody:
@nselliott , it's looking like we can support shared file writes in SCR after all.  I've got a pending PR on SCR where it's working for me so far under limited testing.  To allow shared files during write, SCR has to be configured to bypass the cache and write files directly on the parallel file system.  That means one doesn't get the full performance and features one gets when storing data in cache, but it does work.  I don't think you even need to special case this within SPIO -- it'll be an SCR configuration the application developer/user can choose.

- If the user wants to use the SCR cache (e.g., for improved speed and background transfer), they should use file per-process mode in SPIO, and SCR_Route_file will direct SPIO to write to node-local storage.
- If the user wants to use shared files in SPIO to consolidate data into fewer files, they can configure SCR to use bypass mode, in which case SCR_Route_file directs SPIO to write files directly on the parallel file system)

Just pushed a new commit to this PR while testing it out:
https://github.com/LLNL/axom/pull/251/commits/76d738c219888cfb043e800d72feeb859173e50e

nselliott:
@adammoody So is that a choice that can only be made during configuration of the SCR installation?

adammoody:
It's a runtime configuration under SCR.  It's possible to use both types within the same application run: where some output sets write to the cache with file per process and some write shared files direct to parallel file system.

adammoody:
@nselliott , there are a few things I could use your help on:

Looking at the places I added SCR_Route_file calls, are there more places those should be added?  It looked like there might be other I/O paths that I didn't account for.

Should the SCR_Route_file calls be moved up or down in the software stack to integrate more cleanly?  Maybe fewer calls are needed if we put them in the right spot.  They need to be close enough that we have the final actual file path/name, so I put them very close to the open calls.

Would you like to update the IOSCRRead/Write examples or just go with the IOSCRCkpt/Output instead?  To keep them, we need to add a couple SCR calls to each.

nselliott:
@adammoody Sorry I haven't had the opportunity to address your question yet.  I think we can reduce the SCR_Route_file calls, but I need to examine everything in detail.

nselliott:
I merged from upstream develop because that brings in the fix that makes sure that only rank 0 writes and reads the root file.

src/axom/sidre/spio/IOManager.cpp
```diff
@@ -84,7 +84,15 @@ IOManager::IOManager(MPI_Comm comm,
 {
   MPI_Comm_size(comm, &m_comm_size);
   MPI_Comm_rank(comm, &m_my_rank);
-#ifndef AXOM_USE_SCR
+#ifdef AXOM_USE_SCR
+  int world_size;
+  MPI_Comm_size(MPI_COMM_WORLD, &world_size);
+  if (m_use_scr && m_comm_size != world_size)
```
nselliott, line 90: The text of this warning isn't correct.  What is the behavior we should warn about if m_comm_size does not equal the size of MPI_COMM_WORLD?

gzagaris: I am not sure I understand this restriction. Can we not use SCR on a sub-communicator?

adammoody: Actually, I think we can now probably drop this check and its message.  I think I added this before I moved the Start_checkpoint call.  Since I moved the Start_checkpoint call up to the application, it should now work even when the user creates IOManager with a communicator that is smaller than comm world.

adammoody:
> 
> 
> I merged from upstream develop because that brings in the fix that makes sure that only rank 0 writes and reads the root file.

Great!  That should let the user have the option of restarting from node-local storage then.

src/axom/sidre/spio/IOManager.cpp
adammoody: Most SCR functions are an implied collective across all procs in MPI_COMM_WORLD.  We dup comm_world during SCR_Init.  All procs in the job have to call start/complete, but it's acceptable for a subset of procs to do the actual I/O.  The ranks that read/write data must call SCR_Route_file for each of their files.  I changed the examples so that it's the responsibility of the application to make the collective calls, and then SPIO only has to call Route_file, which is local to the calling process (not a collective).

src/axom/sidre/spio/IOManager.cpp
```diff
@@ -173,26 +181,26 @@ void IOManager::write(sidre::Group* datagroup, int num_files,
     m_baton = new IOBaton(m_mpi_comm, num_files, m_comm_size);
   }
 
-  std::string root_string = file_string;
-#ifdef AXOM_USE_SCR
-  if (m_use_scr)
-  {
-    SCR_Start_checkpoint();
+  if (m_use_scr && num_files != m_comm_size) {
+    SLIC_WARNING("SCR requires a file per process");
```
adammoody, line 185: This is no longer technically required either.  As part of this work, I modified SCR to allow for shared files.  If the user wants multiple processes to share access to a file, they must configure SCR to read/write those datasets on the parallel file system where all procs can see them.  However, if they are using file-per-process, then they can configure SCR to store files in node-local storage instead.  So the user can gain performance when using a file-per-process, but they can still use SCR for shared files.

adammoody:
During a recent SCR team meeting in which I was talking about SPIO, we thought it would be nice to offer a second option in SPIO that calls the SCR collective functions for the user.  That would require the user to pass in a communicator that includes all ranks in comm world.  That would be a simpler interface for those applications that can use it.  I wanted to start with this non-collective approach, since it's more general.  We can revisit the collective option later.

src/axom/sidre/spio/IOManager.cpp
nselliott: I'm going to leave this for now, as all of the SCR implementation here had the expectation to only work in file-per-process mode, so there isn't an immediate need to change this.  If we get everything to a state that we're happy with, we can think about loosening this restriction.

gzagaris:
> During a recent SCR team meeting in which I was talking about SPIO, we thought it would be nice to offer a second option in SPIO that calls the SCR collective functions for the user. That would require the user to pass in a communicator that includes all ranks in comm world. That would be a simpler interface for those applications that can use it. I wanted to start with this non-collective approach, since it's more general. We can revisit the collective option later.

@adammoody -- if the supplied communicator must always consist of all the ranks in comm world, does it need to be supplied? Could we just dup `MPI_COMM_WORLD` as you mentioned previously? 

I am also a bit fuzzy on why the collective calls require the supplied communicator to consist all ranks in comm world. 

- Is this a design artifact in the present implementation?
- Or am I missing something more fundamental that would prevent doing that?

adammoody:
> 
> 
> > During a recent SCR team meeting in which I was talking about SPIO, we thought it would be nice to offer a second option in SPIO that calls the SCR collective functions for the user. That would require the user to pass in a communicator that includes all ranks in comm world. That would be a simpler interface for those applications that can use it. I wanted to start with this non-collective approach, since it's more general. We can revisit the collective option later.
> 
> @adammoody -- if the supplied communicator must always consist of all the ranks in comm world, does it need to be supplied? Could we just dup `MPI_COMM_WORLD` as you mentioned previously?
> 
> I am also a bit fuzzy on why the collective calls require the supplied communicator to consist all ranks in comm world.
> 
>     * Is this a design artifact in the present implementation?
> 
>     * Or am I missing something more fundamental that would prevent doing that?

Good point.  Yes, you could just drop the communicator argument from SPIO for that case, since it's required to effectively be MPI_COMM_WORLD in the SCR calls.

As far as the SCR requirement for comm world, our main challenge is ensuring that we get a consistent cut for the checkpoint across all processes in the job.  The easiest way to ensure that is to have a synchronous collective across all procs.

We are thinking about how we might add support for communicators, but it's been lower priority than adding other functionality so far.  In practice, it will take some significant developer time, because that assumption is scattered throughout the implementation.

For now, we're trying to gather a picture of the various use cases.  We've heard about four different cases so far, and they each impose different requirements on what they'd need in the SCR API.  I'm guessing there are more we haven't yet seen.  I'm hoping we can build a more complete picture as we integrate with more apps.

gzagaris:
> > > During a recent SCR team meeting in which I was talking about SPIO, we thought it would be nice to offer a second option in SPIO that calls the SCR collective functions for the user. That would require the user to pass in a communicator that includes all ranks in comm world. That would be a simpler interface for those applications that can use it. I wanted to start with this non-collective approach, since it's more general. We can revisit the collective option later.
> > 
> > 
> > @adammoody -- if the supplied communicator must always consist of all the ranks in comm world, does it need to be supplied? Could we just dup `MPI_COMM_WORLD` as you mentioned previously?
> > I am also a bit fuzzy on why the collective calls require the supplied communicator to consist all ranks in comm world.
> > ```
> > * Is this a design artifact in the present implementation?
> > 
> > * Or am I missing something more fundamental that would prevent doing that?
> > ```
> 
> Good point. Yes, you could just drop the communicator argument from SPIO for that case, since it's required to effectively be MPI_COMM_WORLD in the SCR calls.
> 
> As far as the SCR requirement for comm world, our main challenge is ensuring that we get a consistent cut for the checkpoint across all processes in the job. The easiest way to ensure that is to have a synchronous collective across all procs.
> 
> We are thinking about how we might add support for communicators, but it's been lower priority than adding other functionality so far. In practice, it will take some significant developer time, because that assumption is scattered throughout the implementation.
> 
> For now, we're trying to gather a picture of the various use cases. We've heard about four different cases so far, and they each impose different requirements on what they'd need in the SCR API. I'm guessing there are more we haven't yet seen. I'm hoping we can build a more complete picture as we integrate with more apps.

Thanks for the explanation. Makes sense.

adammoody:
@nselliott , I know it's been a while since any of us updated this PR.

I think the IOSCRRead/Write examples may still need to be fixed.

Did you get a chance to finish up checking the various SCR_Route_file calls?

Are there other tasks we need to do before it would be merged?

The SCR team has additional ideas for integrating with SPIO, but I'd like to get this much added in before we add more on.

nselliott:
@adammoody The commits I made on June 4 addressed the SCR_Route_file calls.  I moved some things around in a way that looks satisfactory to me.  I don't think there's anything more needing changes in the library code for this PR.  For the examples I need to go over them one more time to make a decision on what to do.  The new examples you wrote should be the primary examples, since they do things in a way that resembles how SCR checkpoints are supposed to operate.  The old examples don't do very much at all, so the decision would be between removing them or keeping them as a very basic check.

I should have time tomorrow to review these things and refresh my memory since the last time I looked at this.

nselliott:
@adammoody I looked over things today and I'm ready to approve the PR.  I'm in favor of keeping the old examples because they can at least be used to verify the basic utility of I/O of a single dataset using SCR's file routing.

adammoody:
Great!  And thanks for your time and help with this, @nselliott !

kennyweiss:
@adammoody  and @nselliott -- I'm a bit late to this PR and am reviewing it now, but had a few quick questions at the outset regarding the SCR version and axom's testing of SCR.

Specifically:
1. Which versions of SCR does/should Axom support? 
    * Does this update depend on a specific version of SCR?
    * Which version was used for testing?

2. Do we currently have any axom host-configs that test SCR ? 
    * A quick grep for `SCR_DIR` seems to indicate that the answer is "No".
    * I am pretty sure that we used to test SCR in at least one host-config
    * Should we add it to any / all LLNL host-configs?
    * Should we add it to any of the azure / docker images
    * Should this be done in this PR or in a follow-up PR?

nselliott:
@kennyweiss I tested with v. 2.0.0.  I think even the current code before this PR won't work with earlier releases.

There used to be a host-config with SCR, but I think it got removed with nobody noticing.  It was linking with an installation outside of our standard TPL package builds.  I would be in favor of adding SCR to the regular TPL suite and all of the host-configs (unless there are some configurations not compatible with SCR), but probably that should be a separate task after this PR.

kennyweiss:
Thanks @nselliott -- I vote for adding SCR to all possible LLNL host-configs, and possibly to some docker images. I agree that we can do this in a separate PR.

I'll create a ticket to continue the discussion.

src/axom/sidre/examples/spio/IOSCRCkpt.cpp
```diff
@@ -0,0 +1,202 @@
...
+using namespace axom::utilities;
+
+/**************************************************************************
+ * Subroutine:  main
+ * Purpose   :
+ *************************************************************************/
+
+int main(int argc, char* argv[])
+{
+  MPI_Init(&argc, &argv);
```
kennyweiss, line 32: @adammoody -- can you please refactor this example to have the following structure, with `main` as a skeleton and all the actual SCR work implemented in functions?

```cpp
int main()
{
  // init MPI and SCR

  // parse command line args

  // The data store should exist before attempting to restart
  DataStore ds = new Datastore(...);

  // attempt to load restart; all SCR calls for loading restart here
  bool hasRestart = loadRestart(file_base, num_ckpts, ds); 
   if( !hasRestart )
   {
       initializeDataStore(ds);   // if not restart, initialize groups and views as desired
   }

   //  do some work
   forall checkpoints 
   {
        // update state in existing datastore
        //  i.e. don't create views or groups here

       // attempt to create checkpoint; all SCR calls for dumping checkpoint here
        dumpCheckpoint(ds);
   }

   // finalize SCR and MPI
}
```

src/axom/sidre/examples/spio/IOSCRCkpt.cpp
```diff
@@ -0,0 +1,202 @@
...
+      // Each SCR output set should be given a name.
+      // This name should be user-friendly, because
+      // end users may need to type it at a command line.
+      // It should also encode enough information that
+      // one can construct the full path to the file on the
+      // parallel file system.
+      std::string ckptname = "time." + std::to_string(t);
+      SCR_Start_output(ckptname.c_str(), SCR_FLAG_CHECKPOINT);
+
+      DataStore* ds = new DataStore();
```
kennyweiss, line 145: To better match how we'd expect users to use this in practice, the datastore should already exist outside this loop.
It should not be created each cycle.

src/axom/sidre/examples/spio/IOSCRCkpt.cpp
kennyweiss: @adammoody -- this is the main change that I'm requesting for this PR.  The others were mainly suggestions about possible improvements.

adammoody:
> 
> 
> Thanks @nselliott -- I vote for adding SCR to all possible LLNL host-configs, and possibly to some docker images. I agree that we can do this in a separate PR.
> 
> I'll create a ticket to continue the discussion.
> 
> EDIT: See #310

Most of my work was done using a development version of SCR.  The SCR team is making a push to do fresh integration in a number of applications, and we are adding functionality and APIs calls to SCR as we go.  We're probably a month or two away from completing that exercise.  The plan is to then stamp a new SCR release to capture all of that.  However, if you'd like a release sooner than that, we could probably stamp a temporary version you could use.  Just let me know.

adammoody:
@kennyweiss , thanks for the review.  I can take a look at those changes next week.

kennyweiss:
@adammoody -- I think any version of SCR that will work with this PR and can be built using our spack scripts (i.e. `<axom>/scripts/uberenv/uberenv.py` ) should be ok for us.

If you'd prefer, we can also hold off on #310 until you have a release that you'd like us to use.

adammoody:
I'll take a look next week to see where we stand on a useable SCR version for SPIO and get back to you.

src/axom/sidre/spio/IOManager.cpp
```diff
@@ -173,33 +173,33 @@ void IOManager::write(sidre::Group* datagroup, int num_files,
     m_baton = new IOBaton(m_mpi_comm, num_files, m_comm_size);
   }
 
-  std::string root_string = file_string;
-#ifdef AXOM_USE_SCR
-  if (m_use_scr)
-  {
-    SCR_Start_checkpoint();
```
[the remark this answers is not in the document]
nselliott: They don't, but `IOSCRWrite` needs `SCR_Start_output` and `SCR_Complete_output`, so I will add.

src/axom/sidre/spio/IOManager.cpp
```diff
@@ -209,18 +209,38 @@ void IOManager::write(sidre::Group* datagroup, int num_files,
...
+        utilities::filesystem::getDirName(dir_name, hdf5_name);
+        if (!dir_name.empty())
+        {
+          utilities::filesystem::makeDirsForPath(dir_name);
+        }
+        h5_file_id = conduit::relay::io::hdf5_create_file(hdf5_name);
+      } else {
+#ifdef AXOM_USE_SCR
+        // register original path, and get new path from SCR
+        char scr_name[SCR_MAX_FILENAME];
```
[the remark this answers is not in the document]
nselliott: I'll go ahead with this change.

nselliott:
Sorry, my last commit broke something.  Looking into it.

src/axom/sidre/examples/spio/IOSCRCkpt.cpp
```diff
@@ -0,0 +1,202 @@
...
+/**************************************************************************
+ *************************************************************************/
+
+#include "mpi.h"
+
+#include "axom/core.hpp"
+#include "axom/slic.hpp"
+#include "axom/sidre.hpp"
+
+#ifdef AXOM_USE_SCR
```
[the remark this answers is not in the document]
gzagaris: My vote would be to assume that SCR is available on this file and add the compile-time error check above.

nselliott:
@gzagaris It requires SCR v. 2.0.0, but that was also true before this PR.  It sounds like we're going to wait for the next SCR release before merging this, though.

gzagaris:
> @gzagaris It requires SCR v. 2.0.0, but that was also true before this PR. It sounds like we're going to wait for the next SCR release before merging this, though.

Sounds good @nselliott -- I think adding a line in the `RELEASE_NOTES`, pointing to the appropriate SCR version would be helpful before merging this.

adammoody:
I should have time to get back to this in the next day or two.  I see @nselliott has already addressed some of the recommendations.  Thanks!

src/axom/sidre/spio/IOManager.cpp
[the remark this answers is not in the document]
adammoody: Thanks @nselliott .  Good to resolve this item now?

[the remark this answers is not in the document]
nselliott: @adammoody Yes. this made for some cleaner code in several of the calling locations.

src/axom/sidre/spio/IOManager.cpp
adammoody: @kennyweiss , from the SCR side, we can now support shared file access for both checkpoint (write) and restart (read).  If the user wants to use shared files (e.g., ``num_files < m_comm_size``), SCR must be configured to direct files to the parallel file system so that all procs can see all files.  If the user is running with file-per-process, they then have the option to configure SCR use node-local storage like ramdisk or an ssd, which gives a speed boost and enables things like background transfers.

Our plan right now is to have SCR_Route_file point at the parallel file system by default.  To use cache, the user will need to take some extra steps to configure SCR to point at cache *and* they will have to be sure they are using file-per-process in SPIO.

I have some text about that here:  https://github.com/LLNL/axom/pull/251#issuecomment-631817064

So it sounds like we should perhaps change this to an error, or perhaps it could be removed?  Which way do you guys want to go with it?

kennyweiss: Thanks @adammoody.

My first take is that we should keep this as an error for now, and create a ticket to rethink/rework this in the future.
Thoughts: @nselliott  or others?

src/axom/sidre/examples/spio/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ set(example_sources
 if (SCR_FOUND)
     list(APPEND example_sources IOSCRRead.cpp)
     list(APPEND example_sources IOSCRWrite.cpp)
+    list(APPEND example_sources IOSCRCkpt.cpp)
```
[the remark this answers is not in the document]
adammoody: I've renamed these examples to add underscores

src/axom/sidre/examples/spio/IOSCRCkpt.cpp
adammoody: I've updated the IO_SCR_Ckpt example for this.  I haven't created functions for the datastore, but I reduced each of those to a few lines each so that it's simpler.  Is that reduced datastore code simple enough as it is, or would you still like to have functions here?

There are a few spots where it comes up:
- reading values from the datastore into local variables after reading a restart
- initializing the datastore if there was no restart to read
- updating datastore values before writing a checkpoint

src/axom/sidre/examples/spio/IOSCRCkpt.cpp
adammoody: I moved the datastore create/delete outside of the loop

adammoody:
@kennyweiss and @nselliott , I updated the IO_SCR_Ckpt and IO_SCR_Output examples.

The checkpoint example does *not* have functions to fill in and update the DataStore.  Though, I did simplify the datastore structure, so that it's just a few lines of code now.  I can still add in functions there, but I left them out in the first pass since it seemed like having functions wouldn't hide much code at this point.  If you'd still like functions in those spots, just let me know.

The IO_SCR_Read/Write examples have not been updated yet.  Those examples don't add much over what's currently in Ckpt/Output.  I think updating IO_SCR_Write would eventually boil down to the same code in IO_SCR_Output, and IO_SCR_Read would match what's in loadRestart within IO_SCR_Ckpt.  I could update those, but I didn't want to bother if you thought we should drop them.

Alternatively, from the SCR side we have wondered about an optional, even deeper level of SCR integration within SPIO to further simplify things for users.  This would also abstract the Start_output/Complete_output calls for the user, which gets us back to the direction @nselliott had initially coded up.  However, to do that, the user will need SPIO to infer some SCR parameters on their behalf, like the string to use for the SCR dataset name and the SCR_FLAG_CHECKPOINT/OUTPUT flags.  Or SPIO will need to allow the user to set those values through new arguments.  If we do that, those Read/Write examples could be used to demonstrate that alternate interface.  In short, this would amount to SPIO wrapping the logic in the loadRestart and dumpCheckpoint / dumpOutput functions in the Ckpt/Output examples.

adammoody:
@nselliott , we have a number of places calling SCR_Route_file in IOManager.  Now that you added the getSCRPath function, I was wondering if we could convert more of those over to call getSCRPath instead.  I can do that, but I haven't checked in detail to see whether you left those in place for some other reason.

adammoody:
To be complete, we should also probably update these examples to set valid=1/0 appropriately depending on the success/failure of the IOManager.write() and read() operations.

adammoody:
Just realized we have some redundant ```use_scr``` settings in the read path.
https://github.com/adammoody/axom/blob/develop/src/axom/sidre/spio/IOManager.hpp

To use SCR, one enables it the IOManger constructor:
```
  IOManager(MPI_Comm com, bool use_scr = false);
```
But then we have another parameter that the user must set in the IOManager.read() call:
```
  void read(sidre::Group* group,
            const std::string& root_file,
            bool preserve_contents = false,
            bool use_scr = false);
```
Can we can drop the ```use_scr``` parameter from the read call, or is there a good reason to have this in two spots?

src/axom/sidre/spio/IOManager.cpp
nselliott: I agree with @kennyweiss, this should be an error for now.  Let's get the file-per-process mode in a good state and then add more flexibility in the future.  Since we don't have an alternate code path for when this check fails, it's better to be an error than a warning.

nselliott:
> 
> 
> @nselliott , we have a number of places calling SCR_Route_file in IOManager. Now that you added the getSCRPath function, I was wondering if we could convert more of those over to call getSCRPath instead. I can do that, but I haven't checked in detail to see whether you left those in place for some other reason.

I count four remaining calls to SCR_Route_file that don't go through getSCRPath.  I don't remember if I had a good reason not to use getSCRPath in each of those places.  Let me give it a once-over to either convert them or decide that they should stay.

nselliott:
@adammoody for the use_scr argument in the read method, I think you're correct.  This gives too many unnecessary user options, which could be confusing.  As it is, there's the option to construct an instance of IOManager to use SCR, but then to call read() in a way that doesn't use SCR, but this option only exists on the read side and not the write side.  It would be better if an SCR-enabled instance of IOManager always uses SCR.  If an application wants to also do non-SCR I/O, it should create another instance of IOManager for that purpose.

I'll go ahead and clean this up.

nselliott:
I got all SCR_Route_file calls factored into getSCRPath, and I removed that use_scr flag from the read method.  This added some new files to the PR because the method signature change caused changes to the shroud-generated Fortran interface code.

adammoody:
@nselliott , if IOManager hits a read/write error, does it signal that back to the caller?  Those calls both return void.  Do they throw an exception?

If no errors come back, that's ok, but if they do, I wanted to update the examples to respond to that to set valid=0 that gets passed to the SCR Complete calls.

nselliott:
@adammoody They don't.  That's probably something that should be improved, but outside the scope of this PR.

src/axom/sidre/spio/IOManager.cpp
adammoody: I changed this from a warning to an error

adammoody:
@nselliott , I also pushed some commits that refactor the code to try to clean up a bit.  Let me know if you want to back any of these out.

I moved the ``if (m_use_scr)`` check into the getSCRPath call so that we can just always call it.  It only changes the path if built with scr, and if m_use_scr is true.  I thought it makes for more readable code this way.

I moved some of the getSCRPath calls closer to where the point where the file is actually opened.

With those, the readWithSCR function simplified back to the same code as the read, so I dropped the readWithSCR.

Can you please take a look at all of these changes and let me know if you see problems?  Thanks!

nselliott:
@adammoody, Thanks, these changes do make things look better.  I like that you were able to get rid of the private readWithSCR.  One thing I'm not fully sure about is that some of the getSCRPath calls you added might be unnecessary because they might be passing in paths that were already returned from a previous getSCRPath call, so it's passing an SCR-routed path back into SCR_Route_file.  My understanding is that SCR_Route_file just leaves the path unchanged in this case, so it won't be a bug but might be extra calls.  Nonetheless, I need to look at it in more depth to see if my concern is correct or not, so no need to revert anything at this point.

adammoody:
> 
> 
> @adammoody, Thanks, these changes do make things look better. I like that you were able to get rid of the private readWithSCR. One thing I'm not fully sure about is that some of the getSCRPath calls you added might be unnecessary because they might be passing in paths that were already returned from a previous getSCRPath call, so it's passing an SCR-routed path back into SCR_Route_file. My understanding is that SCR_Route_file just leaves the path unchanged in this case, so it won't be a bug but might be extra calls. Nonetheless, I need to look at it in more depth to see if my concern is correct or not, so no need to revert anything at this point.

Yes, you're correct in that it would be a bug if I am passing a path to SCR twice, so let's fix those up if you find any.  In the case where SCR is using node-local storage, there will be a difference.  The first call will use a path that points to the parallel file system (good), but the second call would point to node local storage (bad).  It's valid to call SCR_Route_file for the same underlying file multiple times, but all of those calls must use the path to where the file should be on the parallel file system.  Thanks!

nselliott:
@adammoody After looking in more detail, the getSCRPath calls look correct and won't cause double calls on SCR_Route_file.  I'm going to add a documentation comment explaining getSCRPath and also fix the file conflicts that github is reporting.

adammoody:
> 
> 
> @adammoody After looking in more detail, the getSCRPath calls look correct and won't cause double calls on SCR_Route_file. I'm going to add a documentation comment explaining getSCRPath and also fix the file conflicts that github is reporting.

Thanks for checking.  Your point also makes me realize it would be a nice feature if SCR could detect and report an error if someone gives SCR_Route_file the same path that SCR had returned.  When bypassing cache, SCR won't know the difference, but I should be able to flag the case where SCR changes the path to node local storage.  I'll add that to the TODO list for SCR.

nselliott:
@adammoody We left this PR hanging, and I want to see if we can get it completed and merged. Do you see anything still outstanding from what you intended to do here? I just merged from axom develop to get things up to date, and everything still works. The axom project wants to get SCR more integrated into its CI testing, and we can use the example programs you added here, but that would be a new task after this is merged.

@kennyweiss you had some change requests here. Has Adam done enough to address the things you asked for?

adammoody:
Thanks, @nselliott !  Yes, we all collectively put some solid effort on this, and it'd be good to finish it off to get it merged before the underlying code base deviates too much.

In skimming the above thread, I note that I talk a lot about different configurations users might be interested in.  I'll write those up.  It'd be useful if you can then give those a shot to get a feel for it and so I can explain the finer details.  Would you be willing to give those a shot?

nselliott:
@adammoody Yes, I can look at that.

src/axom/CMakeLists.txt
```diff
@@ -76,7 +76,8 @@ endif()
 
 # If Axom is the main project, set FOLDER property for top-level check targets
 if ("${PROJECT_SOURCE_DIR}" STREQUAL "${CMAKE_SOURCE_DIR}")
-    set(_code_check_targets "docs" "doxygen_docs"
+    #set(_code_check_targets "docs" "doxygen_docs"
+    set(_code_check_targets "docs"
```
adammoody, line 80: @nselliott , I forget whether this was intentional or if it got in by mistake.

src/axom/sidre/spio/IOManager.hpp
```diff
@@ -325,11 +322,24 @@ class IOManager
...
+  /*!
+   * \brief If needed, get a file path created by SCR.
+   *
+   * When using this class with the SCR library, SCR must create a file path
+   * to a storage location that it controls, where the actual I/O operations
+   * will occur.  This private method invokes this SCR operation and returns
+   * a string containing the SCR-controlled path.
+   *
+   * When SCR is not being used, either because the IOManager instance was
+   * constructed with the SCR flag set to fales, or because Sidre was not
```
adammoody, line 334:
```suggestion
   * constructed with the SCR flag set to false, or because Sidre was not
```

src/axom/sidre/spio/IOManager.cpp
```diff
@@ -167,28 +167,17 @@ void IOManager::write(sidre::Group* datagroup,
     m_baton = new IOBaton(m_mpi_comm, num_files, m_comm_size);
   }
 
+  SLIC_ERROR_IF(m_use_scr && num_files != m_comm_size,
+                "SCR requires a file per process");
+
```
adammoody, line 172: It is now possible to read/write shared files with SCR.  To do that, the user needs to configure SCR to read/write files to a global file system, i.e., use cache bypass mode, which is the default.  Under this mode, users don't benefit from caching, but the SCR API is still valid.  We added that feature to improve consistency and portability.

I remember we discussed this before in this context, and left it as possible future work.  Just figured I'd repeat it now to keep that fresh.

adammoody:
@nselliott , I noticed there are some file changes at the bottom that may have come in from a merge.  I'm not sure if those are intended to be part of this PR or not.  But otherwise, this looks good to me.

src/axom/CMakeLists.txt
nselliott: Hmmm, this must be a mistake.  Unrelated to anything we did here.  I'll clean it up.

src/axom/sidre/spio/IOManager.cpp
nselliott: Yes, I think this should still be left for future work.  We can change this requirement in IOManager at any time.

nselliott:
> 
> 
> @nselliott , I noticed there are some file changes at the bottom that may have come in from a merge. I'm not sure if those are intended to be part of this PR or not. But otherwise, this looks good to me.

Those are things I added.  We use the shroud tool to auto-generate wrappers for a Fortran interface, so when this PR changed the API of IOManager, I needed to update those files.

src/axom/sidre/spio/interface/c_fortran/wrapIOManager.cpp
```diff
@@ -18,86 +18,89 @@ extern "C" {
 // splicer begin class.IOManager.C_definitions
 // splicer end class.IOManager.C_definitions
```
nselliott, line 20: Somehow clangformat handled the `*` differently in this file than in any other file.  I don't know why this is.

rhornung67 (approved):

src/cmake/thirdparty/FindSCR.cmake
```diff
@@ -27,13 +27,30 @@ find_path( SCR_INCLUDE_DIRS scr.h
            NO_CMAKE_SYSTEM_PATH)
 
 find_library( SCR_LIBRARY NAMES scr libscr
```
kennyweiss, line 29: These can use `blt_find_libraries`:  https://llnl-blt.readthedocs.io/en/develop/api/utility.html#blt-find-libraries

It essentially has the same implementation but with some improved error handling:
https://github.com/LLNL/blt/blob/aea5fbf046e122bd72888dad0a7f97a07b9ff08d/cmake/BLTMacros.cmake#L1084-L1098

kennyweiss: It looks like the following can replace all of these calls
```cmake
blt_find_libraries(
        FOUND_LIBS SCR_LIBRARIES
        NAMES      scr jvtree dtcmp
        REQUIRED   TRUE
        PATHS      ${SCR_DIR}/lib64 ${SCR_DIR}/lib )
include(FindPackageHandleStandardArgs)
```

kennyweiss (approved):
Other than some minor suggestions, this looks really good.

Thanks for seeing this through @nselliott  and Adam.

src/cmake/thirdparty/FindSCR.cmake
```diff
@@ -27,13 +27,30 @@ find_path( SCR_INCLUDE_DIRS scr.h
...
+              PATHS ${SCR_DIR}/lib64
+              NO_DEFAULT_PATH
+              NO_CMAKE_ENVIRONMENT_PATH
+              NO_CMAKE_PATH
+              NO_SYSTEM_ENVIRONMENT_PATH
+              NO_CMAKE_SYSTEM_PATH)
+
+find_library( KVTREE_LIBRARY NAMES kvtree
+              PATHS ${SCR_DIR}/lib64
```
white238, line 38: Per the last paragraph on this file: 

https://cmake.org/cmake/help/latest/command/find_library.html

This should be `lib` not `lib64` .  CMake toggles this automatically for you based on the system you are on and some automatically set global variables

adammoody:
Hooray!  Thanks everyone for helping to find and fix things along the way on this.

PR #254, opened 2020-05-29
Feature/nselliott/item collection

nselliott:
# Summary

- This PR is a feature
  - Adds the ability to import conduit Nodes that contain lists into Sidre Groups using the new ListCollection.
  - Makes necessary fixes so that Groups using ListCollection can do I/O using save/load calls.


src/axom/sidre/core/Group.cpp
```diff
@@ -2111,9 +2131,33 @@ void Group::importFrom(conduit::Node& node,
...
+      if (!create_list && n_group.has_child("groups"))
+      {
+        if (n_group["groups"].dtype().is_list())
+        {
+          create_list = true;
+        }
+      }
       std::string group_name = groups_itr.name();
-      Group* group = createGroup(group_name);
+      if (!m_is_list)
```
ltaylor16, line 2152: I prefer to avoid negative logic if possible. So my preference is to reverse the clauses so the test is `if (m_is_list)`.  More of a style issue.

PR #259, opened 2020-06-17
Basic Inlet

white238:
This PR is a good starting point for the basic functionality of Inlet.  I would like to start playing around with it in Serac.  I have added doxygen comments and a stub for the Sphinx docs.


src/axom/inlet/Table.hpp
```diff
@@ -0,0 +1,229 @@
...
+   */
+  bool required();
+private:
+  axom::sidre::Group* baseFieldAdd(const std::string& name,
+                                   const std::string& description);
+
+  std::string m_name;
+  std::shared_ptr<Reader> m_reader;
+  // Inlet's Root Sidre Group
+  axom::sidre::Group* m_sidreRootGroup;
```
kennyweiss, line 221: Do you need to store the root group?  You can always get it from the current group using something like
```cpp
m_sidreGroup->getDataStore()->getRootGroup();
```
(I didn't look it up, so it might not be 100% syntactically accurate).

white238:
I noticed that the legacy README.md's were repurposed as the main page's for each component in doxygen.  This was causing some weird rendering in Github pages.  I renamed them so they wont cause that conflict.  Also there was some duplication in the component's title that was bleeding over to the root doxygen page.

src/axom/inlet/Table.cpp
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  return shared_from_this();
+}
+
+bool Table::required()
+{
+  SLIC_ASSERT_MSG(m_sidreGroup != nullptr, "Inlet's Table specific Sidre Datastore Group not set");
+
+  if (!m_sidreGroup->hasView("required"))
```
kennyweiss, line 169: We discussed this, but I think this would be a great use case for `sidre::Attributes`, which can have a default value of `false` (or 0, or whatever)

src/axom/inlet/Table.cpp
```diff
@@ -0,0 +1,193 @@
...
+  if (sidreGroup == nullptr)
+  {
+    //TODO: better idea?
+    return std::shared_ptr<Field>(nullptr);
+  }
+
+  bool value;
+  if(m_reader->getBool(fullName, value))
+  {
+    if (value)
```
kennyweiss, line 71: Suggestion: Something like:
```cpp
   if(m_reader->getBool(fullName, value)) 
   {
      sidreGroup->createViewScalar("value", value? int8(1) : int8(0) );
   }
      
```

src/axom/inlet/LuaReader.cpp
```diff
@@ -0,0 +1,177 @@
...
+    SLIC_WARNING(fmt::format("Inlet: Given Lua string could not be loaded: {0}",
+                             luaString));
+    m_luaState = nullptr;
+    return false;
+  }
+
+  return true;
+}
+
+// TODO allow alternate delimiter at sidre level
```
kennyweiss, line 81: 👍

src/axom/inlet/LuaReader.cpp
```diff
@@ -0,0 +1,177 @@
...
+#define SCOPE_DELIMITER '/'
+
+bool LuaReader::findVariable(const std::string& id)
```
kennyweiss, line 84: suggestion:  Rename to `hasVariable` ?  
`findVariable` makes me think that it will return something, e.g. a pointer to the found item.

src/axom/inlet/Inlet.cpp
```diff
@@ -0,0 +1,183 @@
...
+//-------------------------------------------------
+
+axom::sidre::View* Inlet::baseGet(const std::string& name)
+{
+  SLIC_ASSERT_MSG(m_sidreRootGroup != nullptr, "Inlet's Sidre Datastore Group not set");
+
+  // All data hangs under the group's name
+  if (!m_sidreRootGroup->hasGroup(name))
+  {
+    return nullptr;
```
kennyweiss, line 67: It might be nice to provide the user with a log of what went wrong.

Perhaps with a way to modify the verbosity when desired?

src/axom/inlet/Inlet.cpp
```diff
@@ -0,0 +1,183 @@
...
+  }
+  axom::sidre::Group* group = m_sidreRootGroup->getGroup(name);
```
kennyweiss, line 69: Would it make sense to delegate these accesses to the `Table` class?

kennyweiss: I'm guessing this will be done in future updates? along with changes to the `Field` class ?

src/axom/mainpage.md
```diff
@@ -19,8 +20,9 @@ Dependencies between components are as follows:
 - Core has no dependencies, and the other modules depend on Core
 - Slic optionally depends on Lumberjack
 - Slam, Primal, Mint, Quest, Spin, and Sidre depend on Slic
+- Inlet depends on Sidre, and Slic
 - Mint optionally depends on Sidre
-- Spin depends on Primal and Slam
+- Spin depends on Primal, and Slam
 - Quest depends on Slam, Primal, Spin, and Mint
 
 The figure below summarizes the dependencies between the modules.  Solid links
```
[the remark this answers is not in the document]
white238: As far as I can tell.  I also tried moving one (lumberjack's) to under the docs/doxygen directory and then it could no longer find it even though I changed the path in the Doxyfile.in.

src/axom/inlet/tests/inlet_LuaReader.cpp
```diff
@@ -0,0 +1,108 @@
...
+{
+  axom::inlet::LuaReader lr;
+  lr.parseString(
+    "foo = { bar = \"this is a test string\"; baz = \"TesT StrInG\" }");
+
+  bool retValue;
+  std::string value;
+
+  value = "";
+  retValue = lr.getString("foo/bar", value);
```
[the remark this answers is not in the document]
white238: its short for LuaReader... i could just do luaReader.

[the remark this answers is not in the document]
white238: Logged an issue for later. #278

src/axom/inlet/tests/inlet_LuaReader.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+TEST(inlet_LuaReader_getBool, getInsideBools)
+{
+  axom::inlet::LuaReader lr;
+  lr.parseString("foo = { bar = false; baz = true }");
+
+  bool value, retValue;
+
+  value = true;
+  retValue = lr.getBool("foo/bar", value);
```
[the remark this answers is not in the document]
white238: Adding/defining and getting values are both strongly typed.  I thought that was best.

src/axom/inlet/LuaReader.cpp
white238: This does more than just confirm that it is there.  It moves the Lua state to where the variable is so it can then be accessed/used.

src/axom/inlet/Inlet.cpp
white238: True, but what if the value is optional?  Should I add a hasValue maybe to allow them to check before hand?

src/axom/inlet/Inlet.cpp
white238: The idea was that the "get" functions would be used later in the program and the "add" functions would be used when you were defining what should/can be in the input deck.  You're right an example would clear this up.

src/axom/inlet/Table.cpp
white238: done.

src/axom/inlet/Field.hpp
```diff
@@ -0,0 +1,94 @@
...
+   *****************************************************************************
+   * \brief Set the required status of this Field.
+   *
+   * Set whether this Field is required, or not, to be in the input deck.
+   * The default behavior is to not be required.
+   *
+   * \param [in] isRequired Boolean value of whether Field is required
+   *****************************************************************************
+   */
+  std::shared_ptr<Field> required(bool isRequired);
```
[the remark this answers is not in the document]
white238: Ah yes. I forgot to go back and do those.  Will do them in a bit!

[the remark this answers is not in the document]
white238: done.

PR #263, opened 2020-06-19
Fix issue when no fortran is defined

white238:
I added "no fortran" spec so this doesn't happen again

This issue was brought up by a Spack user. This is probably the second or third time not testing against a spec that didn't have fortran has come up.


white238:
> Thanks @white238.
> 
> I'm assuming you're planning to add a host-config for this when it builds, right?

It just succeeded my small test. I will build just this spec and add its specific host-config

white238:
I'm running into a weird error where my compiler specs are bleeding together. I am trying something to see if it goes away.

white238:
This is ready to go.  Just so everyone knows... spack sees gcc@8.1.0_no_fortran as a replacement/upgrade of gcc@8.1.0.  So the former overrides the later.  I have talked to the spack team about it and its a known design problem/flaw/feature.

PR #269, opened 2020-06-25
BUGFIX: Using proper variables in ImplicitGrid Constructor

Parqua:
# Summary

This PR is a bugfix for ImplicitGrid in Spin. It ensures that the constructor for an ImplicitGrid from raw doubles and integers utilizes the integer argument for resolution of the grid. Previously, it would always take the values 0 for the resolution in all dimensions, due to pulling from an internal data structure. It does so by allocating a Point using the supplied values, then using that for further initialization, thus maintaining the data types used for the rest of the code. 

Also included is an expansion of the existing unit test for the initialization of ImplicitGrid, ensuring that values are being properly used.

### DIFF
--- a/src/axom/spin/ImplicitGrid.hpp
+++ b/src/axom/spin/ImplicitGrid.hpp
@@ -127,20 +127,13 @@ public:
    
     GridCell res;
     if(gridRes != nullptr){
-        if(NDIMS == 2){
-            res = GridCell::make_point(gridRes[0], gridRes[1]);
-        }
-        else if(NDIMS == 3){
-            res = GridCell::make_point(gridRes[0], gridRes[1], gridRes[2]);
-        }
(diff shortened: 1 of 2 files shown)


src/axom/spin/ImplicitGrid.hpp
```diff
@@ -124,12 +124,16 @@ class ImplicitGrid
 
     // Set up the grid resolution from the gridRes array
     //   if NULL, GridCell parameter to initialize should also be NULL
-    const GridCell* pRes =
-      (gridRes != nullptr) ? &m_gridRes : nullptr;
+   
+    GridCell res;
+    if(gridRes != nullptr){
+            res = GridCell(gridRes, NDIMS);
```
white238, line 130: Please run "make style" on this.. you indentation seems off

PR #272, opened 2020-06-30
Fixes some compiler warnings on LLNL clusters

kennyweiss:
# Summary

- This PR fixes some compiler warnings after a recent upgrade to several `host-configs` on LLNL's `toss3` clusters.

## Notes

- Our clang host-configs still have the following warnings due to what appears to be an MPI build on the default gcc compiler:
`/usr/bin/ld: warning: libgfortran.so.3, needed by /usr/tce/packages/mvapich2/mvapich2-2.3-clang-10.0.0/lib/libmpi.so, may conflict with libgfortran.so.5`
- There are still numerous warnings about incorrect/invalide host/device decoration. I believe @bmhan12 has a forthcoming PR to address this.  (Possibly #249 ?)

### DIFF
--- a/src/axom/spin/internal/linear_bvh/vec.hpp
+++ b/src/axom/spin/internal/linear_bvh/vec.hpp
@@ -38,8 +38,8 @@ public:
   //
   //  Use only default contructors to keep this a POD type
   //
-  inline AXOM_HOST_DEVICE Vec<T,S>() = default;
-  inline AXOM_HOST_DEVICE Vec<T,S>(const Vec<T,S>& v) = default;
+  inline Vec<T,S>() = default;
+  inline Vec<T,S>(const Vec<T,S>& v) = default;
 
   inline AXOM_HOST_DEVICE bool operator==(const Vec<T,S> &other) const
   {
--- a/src/axom/spin/internal/linear_bvh/vec.hpp
+++ b/src/axom/spin/internal/linear_bvh/vec.hpp
@@ -36,8 +36,10 @@ public:
   T m_data[S];
 
   //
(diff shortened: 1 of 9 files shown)


src/axom/spin/internal/linear_bvh/vec.hpp
```diff
@@ -36,8 +36,10 @@ class Vec
   T m_data[S];
 
   //
-  //  No contructors so this is a POD type
+  //  Use only default contructors to keep this a POD type
```
kennyweiss, line 39: @gzagaris -- I verified that `Vec` is still a POD type with `default` constructors using 
```cpp
   static_assert(std::is_pod< Vec >::value, "Vec<T,S> should be a POD");
```
but didn't keep this static_assert in the code.

See: https://stackoverflow.com/questions/7169658/is-this-struct-pod-in-c11

src/tools/mesh_tester.cpp
```diff
@@ -599,32 +599,33 @@ int main( int argc, char** argv )
...
-     {
-     case seq:
-      collisions = naiveIntersectionAlgorithm(surface_mesh, degenerate, params.intersectionThreshold);
-  #ifdef AXOM_USE_RAJA
-     case raja_seq:
+      switch (params.policy)
+      {
+      case seq:
+        collisions = naiveIntersectionAlgorithm(surface_mesh, degenerate, params.intersectionThreshold);
+        break;
```
kennyweiss, line 606: @bmhan12 -- I added a missing `break` statement here.

bmhan12: Nice catch!

PR #286, opened 2020-07-07
Fix for adding Inlet Fields via Table

aeshapar:
# Summary

- This PR is a bugfix
- It does the following:
  - Modifies Table.cpp to utilize the full path of a Field when adding it to the Inlet

### DIFF
--- a/src/axom/inlet/tests/inlet_Inlet.cpp
+++ b/src/axom/inlet/tests/inlet_Inlet.cpp
@@ -398,6 +398,60 @@ TEST(inlet_Inlet_basic, getNestedStrings)
   EXPECT_EQ(value, "1000");
 }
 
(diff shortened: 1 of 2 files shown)


src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -398,6 +398,60 @@ TEST(inlet_Inlet_basic, getNestedStrings)
...
+  auto inlet = createBasicInlet(&ds, testString);
+
+  std::shared_ptr<axom::inlet::Field> currField;
+
+  // Check for existing fields
+  auto table = inlet->addTable("foo", "A table called foo");
+  table->required(true);
+
+  currField = table->addString("bar", "bar's description");
+  // currField = inlet->addString("foo/bar", "bar's description");
```
white238, line 420: Please remove commented out code

Issue #288, opened 2020-07-07
Windows TPL build issues


agcapps:
After pulling Axom from Github, the TPL build fails with a message like "package axom has no feature hdf5."

agcapps:
TPLs are built on Windows using vcpkg.  The CONTROL file was specifying conduit, mfem, and hdf5 as default features and build-dependencies.  I removed the build dependencies and added Feature paragraphs (see https://vcpkg.readthedocs.io/en/latest/maintainers/control-files/#feature-paragraphs) for mfem and conduit.  I left hdf5 completely out of axom's CONTROL file, since conduit already requires hdf5 as a build dependency.

agcapps:
TPLs and src build must match word size.  The TPL build produces a 64-bit installation by default; the config-build script for axom generates a 32-bit build by default.  I changed the axom default to 64-bits to match the TPLs.

agcapps:
Fixed in branch #289 .

PR #289, opened 2020-07-08
Fix feature list and dependencies for package axom.

agcapps:
# Summary

- This PR is a bugfix: see https://github.com/LLNL/axom/issues/288
- It allows Windows TPL and src to build in a freshly-pulled repository.
  This PR does the following:
  - Changes the vcpkg CONTROL file for the axom package so build dependencies and features are properly specified
  - Sets the default for the source build to 64-bits, to match the TPLs


agcapps:
> Thanks @agcapps.
> 
> Does this imply that, for Windows, Axom will depend on hdf5 when conduit does?

Yes, because the vcpkg CONTROL file for the conduit package lists hdf5 as a build dependency.

PR #293, opened 2020-07-10
Fix bug related to obtuse bug regarding the lua stack index

white238:
Fix a seg fault in LuaReader that was caused by some combination of existant/non-existant reading of multi-nested tables
Added regression test that seg faulted prior to fix

... This was very annoying to find and I am still not entirely sure I understand what triggers this behavior but this does make it go away.

### DIFF
--- a/src/axom/inlet/LuaReader.cpp
+++ b/src/axom/inlet/LuaReader.cpp
@@ -108,6 +108,8 @@ bool LuaReader::findVariable(const std::string& id)
   std::vector<std::string> tokens;
   axom::utilities::string::split(tokens, temp_id, SCOPE_DELIMITER);
 
+  // Clear the lua stack because we always call with fully qualified names
+  lua_settop(m_luaState, 0);
   for (std::string token : tokens)
   {
     if(atGlobalScope)
--- a/src/axom/inlet/tests/inlet_Inlet.cpp
+++ b/src/axom/inlet/tests/inlet_Inlet.cpp
@@ -460,8 +460,8 @@ TEST(inlet_Inlet, mixLevelTables)
   auto luaReader = std::make_shared<axom::inlet::LuaReader>();
   luaReader->parseString(
     "thermal_solver={"
-    "   u0 = { type = \"function\", func = \"BoundaryTemperature\"},"
-    "   kappa = { type = \"constant\", constant = 0.5},"
+    "   u0 = { type = 'function', func = 'BoundaryTemperature'},"
+    "   kappa = { type = 'constant', constant = 0.5},"
     "   solver = {"
     "     rel_tol = 1.e-6,"
     "     abs_tol = 1.e-12,"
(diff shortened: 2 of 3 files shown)


src/axom/inlet/tests/inlet_LuaReader.cpp
```diff
@@ -88,6 +88,33 @@ TEST(inlet_LuaReader_getString, getInsideStrings)
   EXPECT_EQ(value, "TesT StrInG");
 }
 
+
+TEST(inlet_LuaReader, mixLevelTables)
```
white238, line 92: I tried to create a really dumbed down version to show at the LuaReader level but it didn't work. It's a good test anyways.

src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -451,6 +451,105 @@ TEST(inlet_Inlet_basic, getNestedValuesAddedUsingTable)
...
+TEST(inlet_Inlet, mixLevelTables)
+{
+  // regression test for the lua stack being at the wrong
+  // level with nested tables with non-nested values
+
+  axom::sidre::DataStore dataStore;
+  auto luaReader = std::make_shared<axom::inlet::LuaReader>();
+  luaReader->parseString(
+    "thermal_solver={"
+    "   u0 = { type = \"function\", func = \"BoundaryTemperature\"},"
```
kennyweiss, line 463: Suggestion: You could use the single quote (`'`) as an internal string delimiter in lua.
e.g.
```lua
     "   u0 = { type = 'function', func = 'BoundaryTemperature'}," 
```

src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -451,6 +451,105 @@ TEST(inlet_Inlet_basic, getNestedValuesAddedUsingTable)
...
+    "   }"
+    "}");
+
+  auto inlet = std::make_shared<axom::inlet::Inlet>(luaReader, dataStore.getRoot());
+
+  //
+  // Define input deck schema
+  //
+
+  inlet->addString("thermal_solver/mesh/filename", "filename");
```
kennyweiss, line 481: `thermal_solver/mesh` doesn't appear to be in the above lua string.  

If that's part of what you're testing for, it might help to add a comment.

src/axom/inlet/LuaReader.cpp
```diff
@@ -108,6 +108,8 @@ bool LuaReader::findVariable(const std::string& id)
   std::vector<std::string> tokens;
   axom::utilities::string::split(tokens, temp_id, SCOPE_DELIMITER);
 
+  // Clear the lua stack because we always call with fully qualified names
+  lua_settop(m_luaState, 0);
```
kennyweiss, line 112: Was this the fix?

src/axom/inlet/tests/inlet_Inlet.cpp
white238: Done.

src/axom/inlet/tests/inlet_Inlet.cpp
white238: Done.

src/axom/inlet/LuaReader.cpp
white238: Yes.  I still don't quite understand how exactly the lua stack index works but this resets it back to the global level.  Which is fine for this function because the names are fully qualified at this point.

white238: https://www.lua.org/pil/24.2.html

I'm still reading up on how the stack works but here is a pretty good explanation if anyone cares

PR #298, opened 2020-07-11
Bugfix for 64 bit indexing in slam, spin and quest

### LINKED ISSUE #295: Enable InOutOctree to run with 64 bit indexes
The `InOutOctree` has a max depth of 32 levels due to its dependence on `axom::IndexType` for the (integer) coordinates of each octree block.

@agcapps reported a problem that a user encountered in the quest in/out query where 32 levels was insufficient to separate the mesh's inside from its outside.

To resolve this, axom's octree class should either:
* Support 64-bit indexes when `axom::IndexType == int64`
* Add an extra configuration option to allow 64-bit indexes, regardless of the value of `axom::IndexType`

(or more likely, it should support both of these cases, with a default `CoordType` of `axom::IndexType`)

### LINKED ISSUE #296: Compiler errors when using AXOM_USE_64BIT_INDEXTYPE
While debugging #295, I attempted to configure axom with ``AXOM_USE_64BIT_INDEXTYPE = ON`` and encountered several compiler errors.

These should be fixed, and possibly enabled in one of our CI configurations.

kennyweiss:
# Summary

- This PR is a bugfix
- Resolves #296 -- Axom can once again be configured with `AXOM_USE_64BIT_INDEXTYPE`
- Resolves #295 -- Quest's `InOutOctree` can use up to 64 levels when axom is configured with `AXOM_USE_64BIT_INDEXTYPE`

A future PR will consider extentions to the `Octree` and `InOutOctree` classes to allow more than 32 levels of resolution, regardless of the value of `axom::IndexType`. This might take the form of an extra template parameter that defaults to `axom::IndexType`.

### TODO:
- [x] Add a CI configuration that uses 64 bit indexing
- [x] Look into warning about octree coloring from user-supplied mesh for #295

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,6 +34,13 @@ strategy:
       COMPILER: 'g++'
       TEST_TARGET: 'linux_gcc8'
       HOST_CONFIG: 'docker-linux-ubuntu16.04-x86_64-gcc@8.1.0'
+    linux_gcc8_64bit:
+      VM_ImageName: 'ubuntu-16.04'
+      Compiler_ImageName: 'axom/tpls:gcc-8'
+      CMAKE_EXTRA_FLAGS: '-DAXOM_USE_64BIT_INDEXTYPE:BOOL=ON'
+      COMPILER: 'g++'
+      TEST_TARGET: 'linux_gcc8_64bit'
+      HOST_CONFIG: 'docker-linux-ubuntu16.04-x86_64-gcc@8.1.0'
     linux_clang4:
        VM_ImageName: 'ubuntu-16.04'
        Compiler_ImageName: 'axom/tpls:clang-4'
--- a/src/axom/quest/InOutOctree.hpp
+++ b/src/axom/quest/InOutOctree.hpp
@@ -2315,8 +2315,8 @@ public:
   using SpaceTriangle = typename InOutOctreeType::SpaceTriangle;
 
   using LeafVertMap = slam::Map<slam::Set<VertexIndex>, VertexIndex>;
-  using LeafIntMap = slam::Map<slam::Set<int>, int>;
-  using LeafGridPtMap = slam::Map<slam::Set<int>, GridPt>;
+  using LeafIntMap = slam::Map<slam::Set<axom::IndexType>, axom::IndexType>;
+  using LeafGridPtMap = slam::Map<slam::Set<axom::IndexType>, GridPt>;
 
   using DebugMesh = mint::UnstructuredMesh< mint::MIXED_SHAPE>;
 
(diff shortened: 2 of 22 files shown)


src/axom/spin/OctreeLevel.hpp
```diff
@@ -158,7 +158,7 @@ class OctreeLevel
    */
   CoordType maxCoord() const
   {
-    return (1<< m_level) -1;
+    return (CoordType(1)<< m_level) -CoordType(1);
```
kennyweiss, line 161: This one (here and elsewhere) was a bit dubious.

For 64-bit shift operations on numeric literals, we need to ensure that we are using a 64-bit register by casting to the appropriate int type.

src/axom/quest/InOutOctree.hpp
```diff
@@ -2634,10 +2634,10 @@ class InOutOctreeMeshDumper
 
 
     // Add the fields to the mint mesh
-    VertexIndex* vertID = addIntField(debugMesh, "vertID" );
-    VertexIndex* lLevel = addIntField(debugMesh, "level" );
+    auto* vertID = addIntField(debugMesh, "vertID" );
```
gzagaris, line 2637: I think it would be more clear to use `axom::IndexType*` instead of `auto*`

azure-pipelines.yml
```diff
@@ -34,6 +34,13 @@ strategy:
       COMPILER: 'g++'
       TEST_TARGET: 'linux_gcc8'
       HOST_CONFIG: 'docker-linux-ubuntu16.04-x86_64-gcc@8.1.0'
+    linux_gcc8_64bit:
+      VM_ImageName: 'ubuntu-16.04'
+      Compiler_ImageName: 'axom/tpls:gcc-8'
+      CMAKE_EXTRA_FLAGS: '-DAXOM_USE_64BIT_INDEXTYPE:BOOL=ON'
```
gzagaris, line 40: Thanks for turning this ON. We need to be testing this.

src/axom/spin/SparseOctreeLevel.hpp
```diff
@@ -96,7 +96,7 @@ struct BroodRepresentationTraits<CoordType, DIM, BroodDataType,
                                  primal::Point<CoordType,DIM> >
 {
   using GridPt = primal::Point<CoordType,DIM>;
-  using PointRepresenationType = GridPt;
+  using PointRepresentationType = GridPt;
```
agcapps, line 99: Thanks for the spelling correction.

src/axom/quest/InOutOctree.hpp
kennyweiss: I used `auto` since:
* We don't particularly care about the type as long as it is correct
* In this case in particular (`InOutOctree`), I wasn't sure if I wanted to keep them tied to ``axom::IndexType`` or to make them user settable.  I.e. the user would be able to set a parameter to control if the octree is using 64 bit or 32 bit indexing, regardless of the configuration of `axom::IndexType`.  (see my PR summary comment)
   * So, using `auto` seemed like it would introduce the least amount of change.

I agree that it is a bit inconsistent since I had to use the explicit types for the arrays, so I'll update the code to be more consistent.

gzagaris: Thanks @kennyweiss -- I think not using `auto` in this case will make the code cleaner and more consistent.

kennyweiss:
I reran the user-supplied mesh against this branch, but couldn't reproduce the warning about adjacent octree blocks (described in #295)

kennyweiss:
Passed all LC host-configs with 32-bit and 64-bit index types on toss3 and blueos.

PR #302, opened 2020-07-15
Add verify function to Inlet

aeshapar:
# Summary

- This PR is a feature
- It does the following:
  - Adds a verify function to Inlet

### DIFF
--- a/src/axom/inlet/Inlet.cpp
+++ b/src/axom/inlet/Inlet.cpp
@@ -190,29 +190,29 @@ void Inlet::writeDoc() {
 }
 
 bool Inlet::verify() {
-  return verifyRecursive(m_sidreRootGroup);
+  bool verifySuccess = true;
+  verifyRecursive(m_sidreRootGroup, verifySuccess);
+  return verifySuccess;
 }
 
-bool Inlet::verifyRecursive(axom::sidre::Group* sidreGroup) {
+void Inlet::verifyRecursive(axom::sidre::Group* sidreGroup, bool& verifySuccess) {
   SLIC_ASSERT_MSG(sidreGroup, "Root was nullptr");
 
-  axom::sidre::IndexType viewIndex = sidreGroup->getFirstValidViewIndex();
-  while (axom::sidre::indexIsValid(viewIndex)) {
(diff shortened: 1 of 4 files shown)


src/axom/inlet/Inlet.cpp
```diff
@@ -189,5 +189,31 @@ void Inlet::writeDoc() {
...
+}
+
+bool Inlet::verifyRecursive(axom::sidre::Group* sidreGroup) {
+  SLIC_ASSERT_MSG(sidreGroup, "Root was nullptr");
+
+  axom::sidre::IndexType viewIndex = sidreGroup->getFirstValidViewIndex();
+  while (axom::sidre::indexIsValid(viewIndex)) {
+    if (sidreGroup->getView(viewIndex)->getName() == "required"
+        && !sidreGroup->getView(viewIndex)->isScalar()) {
+      return false;
```
white238, line 203: Two small improvements:

* keep a boolean in Inlet whether this failed and start it at the verify() level then return that value.  This way we can verify the whole tree, instead of stopping at the first one.
* print a warning what failed each time with something like:
   
```
 std::string msg = fmt::format("Inlet: {0}: Required field was not specified in Input Deck", fullSidrePathName) // there is a couple ways to get this name
 SLIC_WARNING(msg);
```

src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -669,6 +669,25 @@ TEST(inlet_Inlet, mixLevelTables)
...
+  ds.getRoot()->getGroup("NewTable")->createViewString("required", "string");
+  EXPECT_FALSE(inlet->verify());
+
+  ds.getRoot()->getView("NewTable/required")->rename("renamedView");
+  currField = inlet->addString("NewTable/str");
+  currField->required(true);
+  currField = inlet->addInt("NewTable/int");
+  EXPECT_TRUE(inlet->verify());
+
+  ds.getRoot()->getGroup("NewTable/int")->createView("required");
```
white238, line 687: Do this through inlet for a non-existent value and mark it required instead of directly through the datastore to ensure that Inlet is broken/working.

white238:
#308 pulls in all this work.

PR #303, opened 2020-07-16
Replace hard-coded string in root file with a user-provided argument.

nselliott:
# Summary

- This PR is a refactoring
  - Adds an optional "tree_pattern" string argument to IOManager::write(), to replace a hard-coded string placed in the root file.
  - Partially addresses issue #264 
  - The "tree_pattern" string is used by the Blueprint index as a path to the location of domains in the data files.  The hard-coded string effectively enforces an assumption of a single domain per rank.  Setting it by argument allows applications to provide a path according to their data layouts.


nselliott:
Closing this, as the specific issue here has been addressed or made obsolete through the PR mentioned here and other changes in conduit for the blueprint index.  If a need for more changes is discovered, there should be a new issue posted.

PR #306, opened 2020-07-20
1D axom::array

bmhan12:
This PR:

- Moves and renames the current axom::Array to axom::MCArray
- Moves IteratorBase from slam to core
- Implements a one dimensional axom::Array based on std::vector


src/axom/core/memory_management.hpp
```diff
@@ -191,7 +193,7 @@ inline void deallocate( T*& pointer ) noexcept
 
 //------------------------------------------------------------------------------
 template < typename T >
-inline T* reallocate( T* pointer, std::size_t n ) noexcept
+inline T* reallocate( T* pointer, std::size_t n, int allocID ) noexcept
```
gzagaris, line 196: Hmm...I am not sure what it means to reallocate a pointer using a different allocator (?) 

When we reallocate a pointer, umpire can give us the allocator ID that was used to reallocate the pointer. Does `reallocate` makes sense on a different allocator ID?

@davidbeckingsale , @rhornung67 -- Any thoughts on that?

src/axom/sidre/core/Array.hpp
```diff
@@ -60,7 +60,7 @@ constexpr axom::IndexType ZERO = 0;
  * \see View
  */
 template< typename T >
-class Array : public axom::Array<T>
+class Array : public axom::mint::Array<T>
```
gzagaris, line 63: I believe we discussed moving the old Array to `axom::core::MCArray` or something like that as a first step. I don't think we want Sidre to depend on Mint b/c that creates a circular dependency. Would that be feasible in this PR ?

src/axom/core/Array.hpp
```diff
@@ -80,68 +113,153 @@ class Array
...
+   */ 
+  Array& operator=( const Array& other ) 
+  { 
+
+    if ( this != &other ) 
+    { 
+      m_resize_ratio = other.m_resize_ratio; 
+      m_is_external = false;
+      initialize(other.size(), other.capacity()); 
+      std::memcpy(m_data, other.data(),
```
gzagaris, line 222: Should this be `axom::copy()`, which also handles different memory spaces, that way if the users wants to shoot themselves in the foot, that can do it correctly....

src/axom/core/Array.hpp
```diff
@@ -757,12 +1203,41 @@ inline void Array< T >::dynamicRealloc( IndexType new_num_tuples )
...
+//------------------------------------------------------------------------------
+template< typename T >
+std::ostream& operator<<( std::ostream& os, const Array< T >& arr )
+{
+  arr.print(os);
+  return os;
+}
+
+template < typename T >
+bool operator==(const Array< T >& lhs, const Array< T >& rhs)
```
gzagaris, line 1223: How is this supposed to work on GPUs? More interestingly, what if `lhs` and `rhs` are in different memory spaces?

NOTE:
--------
1. For a given Array object, we can get the allocator ID and from that we could subsequently determine the execution space (we probably need some more traits mojo for that, but, it is feasible)
2. We could in theory bring move data around such that  both `lhs` and `rhs` are in the both memory space, but, I would strongly advise against that, just from a performance standpoint.

Perhaps, the thing to do is to throw an error if the two arrays are in a different memory space and just not allow for that operation? Thoughts?

src/axom/core/IteratorBase.hpp
```diff
@@ -9,15 +9,11 @@
  * \brief Contains iterator base classes
  */
 
-#ifndef SLAM_ITERBASE_H_
-#define SLAM_ITERBASE_H_
-
-#include "axom/slam/Set.hpp"
+#ifndef ITERBASE_H_
+#define ITERBASE_H_
```
gzagaris, line 13: Please prefix the header include guards with `AXOM_`, i.e., `AXOM_ITERBASE_HPP_`

src/axom/core/Macros.hpp
```diff
@@ -197,8 +197,8 @@
  * \endcode
  */
 #define DISABLE_COPY_AND_ASSIGNMENT(className)                                 \
-  className( const className & ) = delete;                                     \
-  className& operator=(const className&) = delete
+  className( className & ) = delete;                                     \
+  className& operator=( className& ) = delete
```
gzagaris, line 201: What was the issue with `const` there? Just curious?

src/axom/core/Array.hpp
```diff
@@ -550,121 +827,291 @@ Array< T >::~Array()
...
+  other.m_is_external = temp_is_external;
+
 }
 
 //------------------------------------------------------------------------------
 template< typename T >
-inline void Array< T >::initialize( IndexType num_tuples,
-                                    IndexType num_components,
+inline std::ostream& Array< T >::print(std::ostream& os) const
+{
```
gzagaris, line 1079: Make sure we copy the array on the host? 

If the Array is on the GPU and this gets called, perhaps we could throw a warning:
```
"You are printing an Array whose contents are on the GPU. I hope you know what you're doing!"
```

src/axom/core/memory_management.hpp
bmhan12: For the case of a null pointer, `axom::reallocate()` will use the allocator associated with the given allocator ID (or default allocator if not provided) to allocate the data.
I was trying to make the usage similar to [umpire:: reallocate() ](https://github.com/LLNL/Umpire/blob/817779995749109b09334bed5a4b7a9ce1bbe166/src/umpire/ResourceManager.hpp#L227).

src/axom/core/Macros.hpp
bmhan12: Ah, this is a mistake on my part. `const` is fine here.  I meant to remove const from DISABLE_MOVE_AND_ASSIGNMENT to match the syntax on[ cppreference](https://en.cppreference.com/w/cpp/language/move_assignment).

src/axom/core/memory_management.hpp
gzagaris: I see -- You are absolutely right.

Perhaps adding a `\note` in the doxygen documentation indicating that the allocator ID is only used when a null pointer is supplied and otherwise ignored would be helpful.

src/axom/sidre/core/Array.hpp
kennyweiss: @gzagaris -- this is a good catch. We absolutely do not want `sidre` to depend on `mint`.

We had originally discussed a two-pronged approach:
* First: Moving the current `axom::Array` int `mint` to minimize the disruption for this initial change
* Later: Creating some form of multi-component array (`MCArray`)  for sidre and mint that is implemented in terms of Brian's new `axom::Array` class

I don't think we considered that the current `sidre::Array` depends on the current `axom::Array`.
Since it does, it sounds like the best short-term approach would be to follow your suggestion.
I.e. 
* move `mint::Array` back to `core` and rename it `MCArray`
* update `sidre:Array` to derive from `MCArray`

@bmhan12 -- Does that sound reasonable?

bmhan12: Yes, I think this makes sense, and I'll have this change put in this PR.

src/axom/core/Array.hpp
agcapps: I think `operator==()` only makes sense if `lhs` and `rhs` are in the same memory space.  It makes sense to me to either return false or throw an error if they're in different memory spaces.

src/axom/core/Array.hpp
```diff
@@ -550,121 +830,303 @@ Array< T >::~Array()
...
 template< typename T >
-inline void Array< T >::resize( IndexType new_num_tuples )
+inline std::ostream& Array< T >::print(std::ostream& os) const
 {
-  assert( new_num_tuples >= 0 );
 
-  if ( new_num_tuples > m_capacity )
+  #if defined(AXOM_USE_UMPIRE)
+    if (m_allocator_id == 
+          axom::getUmpireResourceAllocatorID( umpire::resource::Device ) ||
```
gzagaris, line 1086: I don't think this check is sufficient here. 

An application may have created a custom allocator, e.g., a pool allocator, on the Device that has a different ``allocator_id``, but is on the Device.

 We should check what else is available through the Umpire API. Specifically, given an ``allocator_id``, is there a way to query the Umpire Resource type associated with a given allocator?

src/axom/core/Array.hpp
```diff
@@ -550,121 +830,303 @@ Array< T >::~Array()
   m_data = nullptr;
 }
 
+//------------------------------------------------------------------------------
+template< typename T >
+inline void Array< T >::fill( const T& value )
+{
+  for (IndexType i = 0 ; i < m_num_elements ; i++)
```
gzagaris, line 837: A few things for ``Array::fill``:

- It should be also templated on the desired execution space. The second template argument should be optional and default to ``axom::SEQ_EXEC``
-  Then, implement this using ``axom::for_all()``
- It would be good to add some sanity check to ensure the specified execution space makes sense for the underlying allocator_id (i.e., memory space) for the Array (we may need some more infrastructure to do this).

src/axom/core/Array.hpp
```diff
@@ -550,121 +830,303 @@ Array< T >::~Array()
...
-  if ( new_size > m_capacity )
+//------------------------------------------------------------------------------
+template< typename T >
+inline void Array< T >::set( const T* elements, IndexType n, IndexType pos )
+{
+  assert( elements != nullptr );
+  assert( pos >= 0 );
+  assert( pos + n <= m_num_elements );
+
+  for ( IndexType i = 0 ; i < n ; ++i )
```
gzagaris, line 865: Ditto. As with the ``fill()`` we should probably implement this with an ``axom::for_all`` kernel launch and add a second, optional template argument to specify the execution space.

src/axom/core/Array.hpp
```diff
@@ -57,9 +93,9 @@ constexpr IndexType ZERO = 0;
  *  * <b> External Storage </b> <br />
  *
```
agcapps, line 94: Here or in MCArray, please compare and contrast the two classes.  Why two?  How are they different, and what are they used for?  These two classes mayor may not be temporary, but as long as we have two classes as close as these we need documentation on why they're different.

src/axom/core/Array.hpp
```diff
@@ -79,115 +121,169 @@ class Array
...
 
-  /*!
-   * \brief Accessor, returns a reference to the given component of the
-   *  specified tuple.
+  /*! 
+   * \brief Copy assignment operator for Array 
+   * 
+   * \note If you use the copy assignment operator on an argument Array 
+   *  with data from an external data buffer, the copy-assigned Array 
+   *  will have a deep copy of the data and own the data copy.
```
joshessman-llnl, line 217: I'm not a huge fan of the ambiguity here (in terms of whether copying a non-owning Array would result in the LHS being an owning or non-owning Array)  - in general it's preferable for containers (especially user-facing ones) to be either owning or non-owning.  If it would be useful to have a separate type with view semantics that could be implemented in a follow-up, but is having this non-owning capability a hard requirement?

joshessman-llnl: This will be addressed in a follow-up

agcapps (dismissed):
Thanks!

src/axom/core/Array.hpp
```diff
@@ -11,28 +11,69 @@
...
+// Forward declare the templated classes and operator function(s)
+template <typename T>
+class Array;
+
+/// \name Overloaded Array Operator(s)
+/// @{
+
+/*! 
+ * \brief Overloaded output stream operator. Outputs the Array in to the
+ *  given output stream.
```
kennyweiss, line 30: Typo:

```suggestion
 * \brief Overloaded output stream operator. Outputs the Array to the
 *  given output stream.
```

src/axom/core/Array.hpp
```diff
@@ -11,29 +11,54 @@
...
  *
- * \brief Provides a generic multi-component array container.
+ * \brief Provides a generic array container.
  *
- *  The Array class provides a generic multi-component array container with
- *  dynamic re-allocation and insertion. Each element in the array is a tuple
- *  consisting of 1 or more components, which are stored contiguously.
+ *  The Array class provides a generic  array container with
+ *  dynamic re-allocation and insertion. Each element in the array is 
+ *  stored contiguously.
```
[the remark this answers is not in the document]
kennyweiss: @joshessman-llnl -- please adapt the above as you see fit

src/axom/core/tests/CMakeLists.txt
```diff
@@ -60,6 +60,16 @@ blt_add_executable( NAME       core_serial_tests
...
+# Separate Array tests due to invalid device error with axom::for_all tests
+blt_add_executable( NAME       core_array_test
+                    SOURCES    core_array.cpp
+                    OUTPUT_DIR ${TEST_OUTPUT_DIRECTORY}
+                    DEPENDS_ON ${utils_tests_depends}
+                    FOLDER     axom/core/tests )
+
+axom_add_test( NAME    core_array
+               COMMAND core_array_test )
+
```
kennyweiss, line 72: Thanks for the comment.
Is this still an issue?

If so, please create a ticket so we can address this in the future.

src/axom/core/memory_management.hpp
```diff
@@ -220,7 +226,14 @@ inline T* reallocate(T* pointer, std::size_t n) noexcept
 
   // Umpire 2.1.0 and above handles reallocate(0) natively
```
kennyweiss, line 227: Should we remove logic relating to versions of Umpire earlier than 2.1.2 ? (They're currently on version 5.0.1)

Perhaps we should create an issue to remind us to discuss this and decide if we still need this.

src/axom/core/examples/core_containers.cpp
```diff
@@ -17,7 +17,7 @@
   */
 
 // Axom includes
```
kennyweiss, line 19: It seems like this example file should be updated to the new `axom::Array`, where possible.

This should likely happen in a future PR after the next round of updates to `axom::Array`.

src/axom/core/CMakeLists.txt
```diff
@@ -38,10 +38,12 @@ set(core_headers
     numerics/polynomial_solvers.hpp
 
     ## core
-    Array.hpp
     Macros.hpp
     Map.hpp
     Path.hpp
+    Array.hpp
+    MCArray.hpp
+    IteratorBase.hpp
```
kennyweiss, line 46: This should probably remain sorted alphabetically

src/axom/core/tests/core_array.cpp
```diff
@@ -0,0 +1,1103 @@
...
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include "axom/core/Array.hpp"             /* for axom::Array */
+#include "axom/core/memory_management.hpp" /* for alloc() and free() */
+
+#include "gtest/gtest.h" /* for TEST and EXPECT_* macros */
+
+// C/C++ includes
+#include <algorithm> /* for std::fill_n */
```
kennyweiss, line 12: Suggestion: We can remove the comments on these includes

src/axom/core/tests/core_array.cpp
```diff
@@ -0,0 +1,1103 @@
...
+  axom::Array<T> v_two_copy(v_two);
+
+  EXPECT_EQ(v, v_copy);
+  EXPECT_EQ(v_two, v_two_copy);
+
+  /* Swap */
+  v.swap(v_two);
+
+  EXPECT_EQ(v_two, v_copy);
+  EXPECT_EQ(v, v_two_copy);
```
kennyweiss, line 675: Perhaps this should instead loop through the array and check that the values have actually been swapped?

Alternatively, it would likely be sufficent to show that
```cpp
EXPECT_NE(v, v_two);
EXPECT_NE(v, v_copy);
```

src/axom/core/tests/core_array.cpp
```diff
@@ -0,0 +1,1103 @@
...
+
+  for(IndexType capacity = 2; capacity < 512; capacity *= 2)
+  {
+    Array<int> v_int(ZERO, capacity);
+    internal::check_storage(v_int);
+
+    Array<double> v_double(ZERO, capacity);
+    internal::check_storage(v_double);
+  }
+}
```
kennyweiss, line 770: Suggestion for a future refactor: Perhaps the tests in this file should be set up to use a templated parametrized test fixture, as in https://github.com/LLNL/axom/blob/develop/src/axom/quest/tests/quest_inout_interface.cpp

kennyweiss (approved):
Nice work @bmhan12 !

Please make sure to update the `RELEASE-NOTES`.
In particular, please mention that:
* `axom::Array` is now modeled after `std::vector`
* the previous `axom::Array` was renamed to `axom::MCArray` 
* we're expecting some follow-up changes to both `axom::Array` and `axom::MCArray`
* `sidre::Array` now derives from `axom::MCArray`
* `slam::IteratorBase` was moved to `axom::IteratorBase`

src/axom/core/MCArray.hpp
```diff
@@ -0,0 +1,768 @@
+// Copyright (c) 2017-2020, Lawrence Livermore National Security, LLC and
```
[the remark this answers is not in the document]
bmhan12: Yes, `git mv` was used. In a copy of this branch, I've tried rebasing with the delete/checkout/git mv you suggested to be sure, but I end up with the same result, github not recognizing `MCArray` used to be Array and new Array has old Array's history.

src/axom/core/tests/core_mcarray.cpp
```diff
@@ -0,0 +1,1063 @@
+// Copyright (c) 2017-2020, Lawrence Livermore National Security, LLC and
```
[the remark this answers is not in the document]
bmhan12: Same issue as above. Oddly enough, `Blame `still shows old changes made when this was `core_array.hpp` and you can redirect to `core_array.hpp` history from there, but not directly through the `History` of `core_mcarray.hpp`

src/axom/core/Array.hpp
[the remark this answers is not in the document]
joshessman-llnl: Made a few tweaks but this has been added - these are useful clarifications!

src/axom/core/memory_management.hpp
joshessman-llnl: Logged as #618

white238 (approved):

joshessman-llnl (approved):

PR #308, opened 2020-07-21
Add default values, ranges, valid values, and a verify function for the input deck

aeshapar:
# Summary

- This PR is a feature
- It does the following:
  - Allows specifying default values and ranges for Fields in the input deck.
  - Provides a verification function for the input deck.


src/axom/inlet/Field.cpp
```diff
@@ -64,6 +64,126 @@ bool Field::required()
   return (bool)intValue;
 }
 
+std::shared_ptr<Field> Field::addDefaultString(std::string value) {
```
white238, line 67: I think it would be cleaner to change these to 4 overloaded functions like defaultValue(std::string value) then check if m_type matches just like in your assert but throw a warning if its wrong.

src/axom/inlet/Field.cpp
```diff
@@ -64,6 +64,126 @@ bool Field::required()
...
+  SLIC_ASSERT_MSG(m_type == FieldType::STRING, "Default value type did not match STRING");
+
+  if (m_sidreGroup->hasView("defaultValue"))
+  {
+    std::string msg = fmt::format("Inlet Field has already defined default value: {0}",
+                                  m_sidreGroup->getName());
+    SLIC_WARNING(msg);
```
white238, line 74: Just a thought.  Maybe we should be setting a global error flag at the base of the given root sidre group level anytime there is a warning thrown at the in defining the schema.  It could then be checked in the verify function.

src/axom/inlet/Field.hpp
```diff
@@ -23,6 +23,8 @@ namespace axom
 namespace inlet
 {
 
+enum class FieldType {BOOL, INT, DOUBLE, STRING, UNSPECIFIED};
```
white238, line 26: These are already defined inside of sidre.  I use them in the add functions.  We shouldn't be duplicating definitions.

src/axom/inlet/examples/document_generation.cpp
```diff
@@ -80,15 +82,19 @@ void defineSchema(std::shared_ptr<Inlet> inlet)
...
+  currField->addDoubleRange(0.5, 100.7);
   
   currField = table->addDouble("abs_tol", "description for solver abs tol");
   currField->required(true);  
 
   currField = table->addInt("print_level", "description for solver print level");
-  currField->required(true);  
+  currField->required(true); 
+  int arr[] = {1, 3, 5, 7};
+  currField->addDiscreteIntRange(arr, 4);
```
white238, line 93: This needs a more descriptive name. validValues()? open to ideas...

aeshapar: That was for adding a set of valid values, which I stored in an sidre array

src/axom/inlet/Field.cpp
```diff
@@ -64,11 +64,14 @@ bool Field::required()
   return (bool)intValue;
 }
 
-std::shared_ptr<Field> Field::addDefaultString(std::string value) {
-  SLIC_ASSERT_MSG(m_type == FieldType::STRING, "Default value type did not match STRING");
+std::shared_ptr<Field> Field::addDefaultValue(const char* value) {
```
white238, line 67: I'd lean towards not giving them this functionality. But if you want it you need to check for nullptr before converting to string because you will get a runtime error.

src/axom/inlet/Field.cpp
```diff
@@ -120,8 +121,8 @@ std::shared_ptr<Field> Field::addDefaultInt(int value) {
   return shared_from_this();
 }
 
-std::shared_ptr<Field> Field::addDefaultDouble(double value) {
-  SLIC_ASSERT_MSG(m_type == FieldType::DOUBLE, "Default value type did not match DOUBLE");
+std::shared_ptr<Field> Field::addDefaultValue(double value) {
```
white238, line 127: Remove add so its not confused with the other add values.

src/axom/inlet/Field.cpp
```diff
@@ -106,7 +106,7 @@ std::shared_ptr<Field> Field::addDefaultValue(bool value) {
 }
 
 std::shared_ptr<Field> Field::addDefaultValue(int value) {
-  SLIC_WARNING_IF(m_type ==  axom::sidre::DataTypeId::INT_ID, 
+  SLIC_WARNING_IF(m_type !=  axom::sidre::DataTypeId::INT_ID, 
                     "Default value type did not match BOOL");
```
white238, line 110: This should be int not BOOL

src/axom/inlet/examples/document_generation.cpp
```diff
@@ -80,15 +82,19 @@ void defineSchema(std::shared_ptr<Inlet> inlet)
...
   currField->required(false);
+  currField->range(0.5, 100.7);
   
   currField = table->addDouble("abs_tol", "description for solver abs tol");
   currField->required(true);  
 
   currField = table->addInt("print_level", "description for solver print level");
-  currField->required(true);  
+  currField->required(true); 
+  int arr[] = {1, 3, 5, 7};
```
white238, line 92: This would be a better user experience as a std::vector then convert it internally for sidre's storage.

src/axom/inlet/examples/example1_expected_documentation.rst
```diff
@@ -118,7 +118,7 @@ Description: This is the solver sub-table in the thermal_solver table
    * - rel_tol
      - description for solver rel tol
      - 
-     - 
+     - 0.500000 to 100.700000
```
white238, line 121: The title of this column needs to be more general... its not just "Range" anymore. "Valid Values/Range"

src/axom/inlet/Field.cpp
```diff
@@ -43,7 +21,10 @@ std::shared_ptr<Field> Field::required(bool isRequired)
   {
     std::string msg = fmt::format("Inlet Field has already defined required value: {0}",
                                   m_sidreGroup->getName());
-    issueWarning(msg, m_sidreRootGroup);
+    SLIC_WARNING(msg);
+    if (!m_sidreRootGroup->hasView("warningFlag")) {
```
white238, line 25: Thanks@aesha for moving this back so we still get the file/line number functionality.  Can you move these three lines into an inline private function so its more of a middle ground between what is now here and what you had before?  Just the setting the warningFlag logic.

white238: Maybe this first function could be a new internal inlet helper function file. inlet_utils.hpp?  There was a couple other things that could go into a file like that but it escapes me atm.

src/axom/inlet/inlet_utils.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+#ifndef INLET_UTILS_HPP
+#define INLET_UTILS_HPP
+
+namespace axom
+{
+namespace inlet
+{
+
+void setWarningFlag(axom::sidre::Group* root);
```
white238, line 11: Document this function please.. just a quick "This function is used to mark if anything went wrong during the defining phase of inlet so verify() will properly fail."

src/axom/inlet/inlet_utils.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+}
+}
+
+#endif
```
white238, line 16: newline

src/axom/inlet/inlet_utils.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+
+void setWarningFlag(axom::sidre::Group* root) {
+  if (!root->hasView("warningFlag")) {
+    root->createViewScalar("warningFlag", 1);
+  }
+}
+
+}
+}
```
white238, line 16: newline

src/axom/inlet/Field.cpp
```diff
@@ -58,12 +60,168 @@ bool Field::required()
...
+    }
+    if (!m_sidreGroup->hasView("value")) {
+      m_sidreGroup->createViewScalar("value", (int8)value);
+    }
+  }
+  return shared_from_this();
+}
+
+std::shared_ptr<Field> Field::defaultValue(int value) {
+  if (m_type != axom::sidre::DataTypeId::INT_ID) {
```
kennyweiss, line 121: Will this cause problems if the field is a double, but the user sets the default value using an int.
E.g.
```cpp
   addDouble("Table1/Table2/foo")->defaultValue(0.); //  OK
   addDouble("Table1/Table2/foo")->defaultValue(0);  //  WARNING
```

I think you'd only want to warn the user if it cannot be cast to the desired type.

kennyweiss: Please add an example like this to the unit tests.
Or create a ticket if it does not currently work as expected.

src/axom/inlet/Field.cpp
```diff
@@ -58,12 +60,168 @@ bool Field::required()
...
+    int tuple[2];
+    tuple[0] = startVal;
+    tuple[1] = endVal;
+    auto view = m_sidreGroup->createViewAndAllocate("range", axom::sidre::INT_ID, 2);
+    view->getBuffer()->copyBytesIntoBuffer(tuple, 2*sizeof(int));
+  }
+  return shared_from_this();
+}
+
+std::shared_ptr<Field> Field::validValues(std::vector<int> set) {
```
kennyweiss, line 205: I didn't get there yet, but is there a check that the default value is within the specified range?

src/axom/inlet/Field.cpp
```diff
@@ -58,12 +60,168 @@ bool Field::required()
...
+  if (m_type != axom::sidre::DataTypeId::INT_ID) {
+    SLIC_WARNING("Field value type did not match INT");
+    setWarningFlag(m_sidreRootGroup);
+  }
+
+  if (m_sidreGroup->hasView("range") || m_sidreGroup->hasView("validValues")) {
```
kennyweiss, line 211: This seems to prefer the first specified range (or default).
A different approach would be to overwrite an existing value, i.e. to use the last specified range.

kennyweiss: Note: There might be cases where one would want to allow overwriting a range.
(although, I don't have any on hand at the moment and this can always be improved in the future).

src/axom/inlet/Field.hpp
```diff
@@ -87,9 +94,121 @@ class Field : public std::enable_shared_from_this<Field>
...
+   * \brief Set the range of this Field.
+   *
+   * Set the continuous range for the Field in the input deck.
+   *
+   * \param [in] set An integer vector containing the set of allowed values
+   *
+   * \return Shared pointer to this Field instance
+   *****************************************************************************
+  */
+  std::shared_ptr<Field> validValues(std::vector<int> set);
```
kennyweiss, line 204: should this be a `std::set` instead of a ``std::vector``?

src/axom/inlet/Field.hpp
```diff
@@ -87,9 +94,121 @@ class Field : public std::enable_shared_from_this<Field>
...
+
```
kennyweiss, line 205: Note: We'll definitely need a case where there is a set of valid strings for a field.

src/axom/inlet/Inlet.cpp
```diff
@@ -184,10 +190,77 @@ void Inlet::registerDocWriter(std::shared_ptr<DocWriter> writer) {
...
+        }
+      } else {
+        verifySuccess = false;
+      }
+    } else if (sidreGroup->hasView("validValues")) {
+      int val = sidreGroup->getView("value")->getScalar();
+      int* range = sidreGroup->getView("validValues")->getArray();
+      size_t size = sidreGroup->getView("validValues")->getBuffer()->getNumElements();
+      // Checks if the value specified is one of the valid values
+      bool found = false;
```
kennyweiss, line 245: Possible optimization: If you sort the array, you can use ``std::binary_search``
(although, it is not clear that this would matter in practice)

src/axom/inlet/Inlet.cpp
```diff
@@ -184,10 +190,77 @@ void Inlet::registerDocWriter(std::shared_ptr<DocWriter> writer) {
...
+        double val = sidreGroup->getView("value")->getScalar();
+        if (!(range[0] <= val && val <= range[1])) {
+          verifySuccess = false;
+        }
+      } else {
+        verifySuccess = false;
+      }
+    } else if (sidreGroup->hasView("validValues")) {
+      int val = sidreGroup->getView("value")->getScalar();
+      int* range = sidreGroup->getView("validValues")->getArray();
```
kennyweiss, line 242: Note: You can use ``sidre::Array`` instead.
See: https://github.com/LLNL/axom/blob/c3a64b372d25e53976b3ba8676a25acc49a9a6cd/src/axom/sidre/core/Array.hpp#L93

I think this would look something like:
```cpp
sidre::Array<int> vals = sidreGroup->getView("validValues");
for(axom::IndexType i=0; i< vals.size(); ++i)
{
   if (vals[i] == val) {
          found = true;
          break;
}
```

src/axom/inlet/SphinxDocWriter.hpp
```diff
@@ -145,14 +145,27 @@ class SphinxDocWriter : public DocWriter {
...
     }
 
     std::string tableName;
     std::string description;
     std::vector<std::vector<std::string>> rstTable;
   };
 
+  /*!
+    *******************************************************************************
+    * \brief Stores all field info into m_rstTables
```
kennyweiss, line 159: Could you please improve the documentation for this function?
After reading this a few times, it is not entirely clear to me what this does.

Perhaps the function should be renamed to something like `extractFieldMetadata`?

kennyweiss: Also, the docs should not refer to implementation details like `m_rstTables`.

src/axom/inlet/SphinxDocWriter.cpp
```diff
@@ -133,5 +114,65 @@ void SphinxDocWriter::writeAllTables() {
...
+
+  fieldAttributes[0] = sidreGroup->getName();
+  if (sidreGroup->hasView("description")) {
+    fieldAttributes[1] = std::string(sidreGroup->getView("description")->getString());
+  } 
+
+  if (sidreGroup->hasView("defaultValue")) {
+    axom::sidre::TypeID type = sidreGroup->getView("defaultValue")->getTypeID();
+    if (type == axom::sidre::TypeID::INT8_ID) {
+      int8 val = sidreGroup->getView("defaultValue")->getData();
```
kennyweiss, line 129: Can this logic be cleaned up using a function that extracts the values and returns a sting?

E.g.  perhaps add the following to `inlet_utils`
```cpp
std::string getValueAsString(View* view)
{
   axom::sidre::TypeID type = view->getTypeID(); 
    if (type == axom::sidre::TypeID::INT8_ID) {
      int8 val = view->getData();
      return std::to_string(val);
    } else if (type == axom::sidre::TypeID::INT_ID) {
      int val = view->getData();
      return std::to_string(val);
    } else if (type == axom::sidre::DOUBLE_ID) {
      double val = view->getData();
      return std::to_string(val);
    } else {
      return view->getString();
    } 
}
```

You could then call it here as:
```cpp
   if (sidreGroup->hasView("defaultValue")) 
   { 
      fieldAttributes[2] = getValueAsString( sidreGroup->getView("defaultValue") );
   }
```

Similar calls throughout inlet (or at least this file) could clean up the code a lot.

src/axom/inlet/inlet_utils.hpp
```diff
@@ -0,0 +1,26 @@
...
+/*!
+  *****************************************************************************
+  * \brief This function is used to mark if anything went wrong during the 
+  * defining phase of inlet so verify() will properly fail.
+  *
+  * \param [in] root Pointer to the Sidre Root Group where the warning flag 
+  * will be set.
+  *****************************************************************************
+  */
+void setWarningFlag(axom::sidre::Group* root);
```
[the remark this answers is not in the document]
white238: I'm not sure why that would be helpful. The point of the function is to indicate something went wrong in the defining the input deck part of the code.  Why would you want someone to override that a warning already happened?

src/axom/inlet/SphinxDocWriter.cpp
white238: That's a very good idea, but note that INT8_ID is actually a boolean value in Inlet.  So that case would return "true" or "false"

src/axom/inlet/Field.cpp
aeshapar: Yep, Inlet::verify() takes care of that

src/axom/inlet/Field.hpp
aeshapar: I used a vector since I was converting set into an array to store it in sidre. Is there a way to store an std::set in a sidre View?

src/axom/inlet/Field.hpp
```diff
@@ -28,8 +28,8 @@ namespace inlet
  * \class Field
  *
  * \brief Provides functions to help define how individual field variables in an
- *        input deck are expected to behave.  It also holds the Sidre Group to the
- *        individual field.
+ *        input deck are expected to behave.  It also holds the Sidre Group to 
+ * the individual field.
```
white238, line 32: This is not formatted properly.

src/axom/inlet/Field.cpp
white238: Not sure that is true.  It will check if the value is not in the input deck and there is a default value (because it will set value to the default value at that point i think) but if you have a value in the input deck AND specify a default value, it doesn't specifically check the default value.

The case that needs to be checked is this (and add a test for it please):

// input deck = "foo=5"

inlet()->addInt("foo")->defaultValue(6)->validValues({5}); // should throw a warning

src/axom/inlet/Field.hpp
white238: You should be copying the values out of the given container and into sidre. I would worry that the vector/set gets destroyed then you are accessing invalid memory.

src/axom/inlet/Inlet.cpp
```diff
@@ -184,10 +190,77 @@ void Inlet::registerDocWriter(std::shared_ptr<DocWriter> writer) {
...
+    verifySuccess = false;
+  }
+
+  if (sidreGroup->hasView("required")) {
+    int8 required = sidreGroup->getView("required")->getData();
+    if (required && !sidreGroup->hasView("value")) {
+      std::string msg = fmt::format("Inlet: {0}: Required field was not specified in Input Deck", 
+                                    sidreGroup->getPathName());
+      SLIC_WARNING(msg);
+    setWarningFlag(m_sidreRootGroup);
```
white238, line 216: Indented wrong.

src/axom/inlet/SphinxDocWriter.hpp
aeshapar: I mentioned m_rstTables since the function is private--does that change anything?

white238: m_rstTables doesn't tell you anything about how or why you are storing the data. It also doesn't say what you are doing with the information.

white238: Something like "This extracts information about the Field stored in the given Sidre Group. This information is stored internally by the class and then written out by writeAllTables"

src/axom/inlet/Field.cpp
aeshapar: Sorry, ^ that's right I misread the intial question. I'll add a check specifically for the default value.

src/axom/inlet/Field.hpp
aeshapar: That's what it's doing right now; I copied the vector contents into a sidre buffer.

src/axom/inlet/inlet_utils.hpp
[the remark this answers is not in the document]
kennyweiss: Thanks @white238  -- sounds like this is not needed.

src/axom/inlet/Field.cpp
kennyweiss: Let's hold off on this.  
After speaking to @white238 offline, this sounds like a possible feature for the future if/when this case arises.

src/axom/inlet/Inlet.cpp
kennyweiss: Probably a good idea to hold off on this until the need arises

src/axom/inlet/Field.cpp
```diff
@@ -58,12 +60,168 @@ bool Field::required()
...
+std::shared_ptr<Field> Field::defaultValue(const std::string& value) {
+  if (m_type != axom::sidre::DataTypeId::CHAR8_STR_ID) {
+    SLIC_WARNING("Field value type did not match STRING");
+    setWarningFlag(m_sidreRootGroup);
+  }
+  if (m_sidreGroup->hasView("defaultValue")) {
+    std::string msg = fmt::format("Inlet Field has already defined default value: {0}",
+                                  m_sidreGroup->getName());
+    SLIC_WARNING(msg);
+    setWarningFlag(m_sidreRootGroup);
```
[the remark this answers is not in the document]
kennyweiss: This can be done in a future PR

src/axom/inlet/Field.hpp
kennyweiss: This isn't super critical -- using `std::set` would remove duplicates, but the behavior and usage would otherwise be pretty similar.

src/axom/inlet/Inlet.cpp
kennyweiss: On second thought, this is probably overkill for this use-case and can be skipped for now.

src/axom/inlet/Field.cpp
white238: #312 so we don't forget.

src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -669,6 +670,213 @@ TEST(inlet_Inlet, mixLevelTables)
...
+  EXPECT_TRUE(sidreGroup->hasView("Table1/float1/defaultValue"));
+  doubleVal = sidreGroup->getView("Table1/float1/defaultValue")->getScalar();
+  EXPECT_EQ(doubleVal, 3.14);
+  
+  EXPECT_TRUE(sidreGroup->hasView("Table1/Table2/Table4/str1/defaultValue"));
+  strVal = sidreGroup->getView("Table1/Table2/Table4/str1/defaultValue")->getString();
+  EXPECT_EQ(strVal, "default for old string"); 
+}
+
+TEST(inlet_Field, defaultValuesDocsDisabled) {
```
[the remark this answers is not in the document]
aeshapar: I forgot to disable docs and add the additional assertions for that case. Thanks for catching it!

src/axom/inlet/Field.hpp
```diff
@@ -210,7 +210,7 @@ class Field : public std::enable_shared_from_this<Field>
    * \return Shared pointer to this Field instance
    *****************************************************************************
   */
-  std::shared_ptr<Field> validValues(std::vector<std::string> set);
+  std::shared_ptr<Field> validStringValues(std::vector<std::string> set);
```
white238, line 213: I know i told you to do this.. but the overloaded parameter is enough info.  you can drop the "String" from the parameter name.

src/axom/inlet/Inlet.cpp
```diff
@@ -255,27 +253,30 @@ bool Inlet::verifyValue(axom::sidre::Group* sidreGroup) {
       double val = sidreGroup->getView("value")->getScalar();;
       return checkRange(sidreGroup, val);
     }
+  } else if (sidreGroup->hasGroup("validStringValues")) {
```
white238, line 256: Can you add a specific check to make sure that a Field does not have both a validStringValues and validValues?  That is a case that shouldn't happen but we want to guard against.

aeshapar: I added a warning for this in the validValues functions

white238:
@aeshapar dont worry about styling your code, @kennyweiss is going to have a large PR soon that fixes the style on the whole code base and this will fix it for you.

src/axom/inlet/Field.cpp
```diff
@@ -56,14 +58,210 @@ bool Field::required()
...
+  }
+
+  if (m_sidreGroup->hasView("range") || m_sidreGroup->hasView("validValues")) {
+    std::string msg = fmt::format("Inlet Field has already defined range: {0}",
+                                   m_sidreGroup->getPathName());
+    SLIC_WARNING(msg);
+    setWarningFlag(m_sidreRootGroup);
+  } else {
+    double tuple[2];
```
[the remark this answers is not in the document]
aeshapar: I agree that is cleaner. Thanks!

src/axom/inlet/Field.cpp
```diff
@@ -56,14 +58,189 @@ bool Field::required()
...
+  }
+  setDefaultValue(value);
+  return shared_from_this();
+}
+
+template<typename T>
+void Field::setRange(T startVal, T endVal) {
+  if (m_sidreGroup->hasView("range") || m_sidreGroup->hasView("validValues")
+      || m_sidreGroup->hasView("validStringValues")) {
+    std::string msg = fmt::format("Inlet Field has already defined range: {0}",
```
white238, line 160: This warning message should be broken into three separate if's to clarify to the user what they need to fix:

```
if (m_sidreGroup->hasView("range")) {
    std::string msg = fmt::format("Inlet Field has already defined range: {0}", m_sidreGroup->getPathName());
    SLIC_WARNING(msg);
    setWarningFlag(m_sidreRootGroup);
}
else if(m_sidreGroup->hasView("validValues") || m_sidreGroup->hasView("validStringValues")){
    std::string msg = fmt::format("Cannot set range on Inlet Field after setting valid value: {0}", m_sidreGroup->getPathName());
    SLIC_WARNING(msg);
    setWarningFlag(m_sidreRootGroup);
}
```

src/axom/inlet/Field.cpp
```diff
@@ -56,14 +58,189 @@ bool Field::required()
...
+  return shared_from_this();
+}
+
+std::shared_ptr<Field> Field::range(double startVal, double endVal) {
+  switch(m_type) {
+  case axom::sidre::DataTypeId::DOUBLE_ID:
+    this->setRange(startVal, static_cast<double>(endVal));  
+    break;
+  default: // incompatible type
+    SLIC_WARNING("Field value type did not match INT or DOUBLE");
```
white238, line 194: Remove "Int or"

src/axom/inlet/Field.cpp
```diff
@@ -56,14 +58,189 @@ bool Field::required()
...
+  return shared_from_this();
+}
+
+std::shared_ptr<Field> Field::validValues(std::vector<int> set) {
+  if (m_type != axom::sidre::DataTypeId::INT_ID) {
+    SLIC_WARNING("Field value type did not match INT");
+    setWarningFlag(m_sidreRootGroup);
+  }
+
+  if (m_sidreGroup->hasView("range") || m_sidreGroup->hasView("validValues")
```
white238, line 207: Fix this message like previous code snippet (but switch the warnings to match this case) to clarify that they are setting valid values when range was set or that valid values was already set.  It's okay to bunch validValues and validStringValues warning messages together generically as "valid values" in the warning message.

src/axom/inlet/Field.cpp
```diff
@@ -56,14 +58,189 @@ bool Field::required()
...
+      || m_sidreGroup->hasView("validStringValues")) {
+    std::string msg = fmt::format("Inlet Field has already defined range: {0}",
+                                   m_sidreGroup->getPathName());
+    SLIC_WARNING(msg);
+    setWarningFlag(m_sidreRootGroup);
+  } else {
+    auto view = m_sidreGroup->createViewAndAllocate("validValues", 
+                                                    axom::sidre::INT_ID, set.size());
+    view->getBuffer()->copyBytesIntoBuffer(&set[0], set.size()*sizeof(int));
```
white238, line 216: This is a very useful function I didn't know about!

src/axom/inlet/Field.hpp
```diff
@@ -87,9 +95,173 @@ class Field : public std::enable_shared_from_this<Field>
...
+  /*!
+   *****************************************************************************
+   * \brief Set the valid values for this Field.
+   *
+   * \param [in] set An vector containing the set of allowed integer values
+   *
+   * \return Shared pointer to this Field instance
+   *****************************************************************************
+  */
+  std::shared_ptr<Field> validValues(std::vector<int> set);
```
white238, line 204: Change "std::vector<int> set" to "const std::vector<int>& set" so that it indicates that it isn't changing the vector but more importantly it doesn't copy the vector when this is called.

src/axom/inlet/Inlet.cpp
```diff
@@ -184,10 +191,124 @@ void Inlet::registerDocWriter(std::shared_ptr<DocWriter> writer) {
...
+  while (axom::sidre::indexIsValid(groupIndex)) {
+    if (sidreGroup->getGroupName(groupIndex) != "validStringValues") {
+      verifyRecursive(sidreGroup->getGroup(groupIndex), verifySuccess);
+    }
+    groupIndex = sidreGroup->getNextValidGroupIndex(groupIndex);
+  }
+}
+
+bool Inlet::verifyValue(axom::sidre::Group* sidreGroup) {
+  auto type = sidreGroup->getView("value")->getTypeID();
```
white238, line 244: Move the "auto type = ..." line closer to where it's used (after line 248)

src/axom/inlet/Inlet.hpp
```diff
@@ -275,16 +277,128 @@ class Inlet : public SchemaCreator
...
+  bool verify(); 
+
   // TODO add update value functions
 private:
+  /*!
+   *****************************************************************************
+   * \brief Verifies the contents of the sidreGroup according to Inlet 
+   * requirements.
+   *
+   * This recrusively checks the correctness of each Field and Table in the sidre
```
white238, line 300: Typo in recursively...

I would move most of this description to the verify() function since that is what the users will read.  This description can just be something simple like "recursive internal implementation of verify()"

src/axom/inlet/SphinxDocWriter.cpp
```diff
@@ -36,50 +36,35 @@ void SphinxDocWriter::writeDocumentation() {
   m_oss << ".. |uncheck|    unicode:: U+2610 .. UNCHECKED BOX" << std::endl;
   m_oss << ".. |check|      unicode:: U+2611 .. CHECKED BOX" << std::endl;
   writeDocumentationHelper(m_sidreRootGroup);
-  m_rstTables.push_back(m_currentTable);
+  // m_rstTables.push_back(m_currentTable);
```
white238, line 39: remove commented out lines. since i assume this isnt needed anymore

src/axom/inlet/SphinxDocWriter.cpp
```diff
@@ -36,50 +36,35 @@ void SphinxDocWriter::writeDocumentation() {
...
   writeAllTables();
   m_outFile.open(m_fileName);
   m_outFile << m_oss.str();
   m_outFile.close();
 }
 
 void SphinxDocWriter::writeDocumentationHelper(axom::sidre::Group* sidreGroup) {
-  SLIC_ASSERT_MSG(sidreGroup, "Root was nullptr");
+  SLIC_WARNING_IF(!sidreGroup, "Root was nullptr");
```
white238, line 47: Clarify this message: Inlet's Sidre root group was not defined while writing documentation

src/axom/inlet/SphinxDocWriter.cpp
```diff
@@ -121,16 +106,110 @@ void SphinxDocWriter::writeTable(const std::string& title,
...
+    double* range = view->getArray();
+    oss << range[0] << " to " << range[1];
+  }
+  return oss.str();
+}
+
+std::string SphinxDocWriter::getValidValuesAsString(axom::sidre::View* view) {
+  // auto
+  SLIC_WARNING_IF(view->getTypeID() != axom::sidre::INT_ID,
+                  "discrete range is only valid for integers");
```
white238, line 155: I am not sure what "// auto" is so either clarify it or remove it (my guess is its a left over from copy/paste)

Shouldn't we be supporting valid values of doubles?  Is there a precision problem you are worried about? Also the placement of this warning makes this warning a bit odd sounding if the users ever hits it.  Hopefully they wouldn't though.

"discrete range"  to "valid values"

src/axom/inlet/Field.hpp
aeshapar: I had originally attempted this, but it results in an error because copyBytes doesn't take a const pointer

white238:
LGTM

PR #315, opened 2020-07-29
Fix Parallel BVH construction on GPUs due to incoherent L1 cache

### LINKED ISSUE #307: BVH construction bug due to incoherent L1 cache on GPUs
@mclarsen made a painful discovery(see https://github.com/LLNL/devil_ray/pull/46)

Building the BVH hierarchy by propagating the AABBs bottom up can be problematic on GPUs and lead to some data corruption in the BVH bounding boxes b/c the L1 cache is not coherent across all threads on GPU. Although, this doesn't happen often, it can happen, and with higher probability for larger problems.

This problem is inherent in the following code:
https://github.com/LLNL/axom/blob/7a0c16d4b3fa702dd994e1b46cd43d1945717167/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp#L610-L682

One solution, currently employed in DevilRay, is to disable the use of the L1 cache for that file(compilation unit), with the following compiler flags: ``-Xptxas -dlcm=cg"``

However, for header-only code that is compiled as part of another application (i.e., a different compilation unit), the responsibility to specify these compiler flags is propagated up to the host application which is not desirable.

Potential Solutions:
--------------------

1. Use ``__threadfence()`` calls after each store and before each load operation to ensure the memory is consistent (cross fingers)
2. Explicitly use uncached loads/stores -- it seems that there is some high-level facility that can emit the desired PTX code. See(https://github.com/NVlabs/fermat/blob/14ed50a34c02823096b3932cd9545e0edde1dc1a/contrib/cub/thread/thread_store.cuh)

gzagaris:
# Summary

Adds a call to ``__threadfence_system()`` in the ``propagate_aabbs()`` method which does a bottom-up  parallel construction of the BVH. The memory fence ensures that when a thread computes and stores a parent to global memory, the *write*  is observed by all threads in the system.

This fix currently only applies to NVIDIA GPUs where the bug was observed due to the incoherent L1 cache on NVIDIA GPUs.

Resolves #307

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -62,6 +62,10 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   ex. `level_error` is now `message_error`.
 
 ### Fixed
+- Fixed issue in the parallel construction of the BVH on GPUs, due to incoherent
+  L1 cache that could result in some data corruption in the BVH. The code now
+  calls ``__threadfence_system()`` after the parent is computed and stored back
+  to global memory to ensure that the *write*  is visible to all threads. 
 - Fixed issue in Mint that would cause the clang@9.0.0 compiler to segfault. The
   `mint_cell_types.cpp` test was causing a segfault in the compiler. The main
   issue triggering this compiler bug was the use of `constexpr` when defining the
--- a/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp
+++ b/src/axom/spin/internal/linear_bvh/build_radix_tree.hpp
@@ -606,6 +606,22 @@ static void array_memset(T* array, const int32 size, const T val)
   } );
 }
 
+/*!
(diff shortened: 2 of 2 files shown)


gzagaris:
> Thanks @gzagaris!
> 
> Have you verified that this fixes the issue in #249 ?

@bmhan12 is going to confirm. Based on the observations in #249, I believe this will fix the issue.

gzagaris:
> They only thing I might change is adding `SPIN` in front of `BVH_THREAD_FENCE_SYSTEM`. Otherwise, lgtm.

Thanks @mclarsen -- Note, I am `undefing` the macro at after the end of the `propagate_aabbs` function, but, I like your suggestion.

gzagaris:
> LGTM! I've verified this fixes the issue I was seeing in #249. Thanks @gzagaris !

Good to hear!

kennyweiss:
Thanks for verifying @bmhan12

PR #317, opened 2020-08-03
Add lambda verifiers for Inlet

aeshapar:
# Summary

- This PR is a feature
- It does the following:
  - Adds lambda verifiers for the input deck.

### DIFF
--- a/src/axom/inlet/tests/inlet_Inlet.cpp
+++ b/src/axom/inlet/tests/inlet_Inlet.cpp
@@ -1058,8 +1058,30 @@ TEST(inlet_Inlet_verify, verifyValidStringValues) {
   field = inlet1->addString("NewTable/field5");
   field->validValues(strs)->defaultValue("zyx");
   EXPECT_FALSE(inlet1->verify());
+}
 
-  
+TEST(inlet_verify, verifyLambda) {
+  std::string testString = "field1 = true; field2 = 'abc'; NewTable = { field3 = 'xyz'; field4 = 'yes' }";
+  DataStore ds;
+  auto inlet = createBasicInlet(&ds, testString);
+
+  auto field1 = inlet->addBool("field1");
+  auto field2 = inlet->addString("field2");
+  auto field3 = inlet->addString("NewTable/field3");
+
+  inlet->registerFieldVerifier([&](axom::sidre::Group* group) -> bool {
+    std::string str = group->getView("field2/value")->getString();
+    return (str.size() >= 1 && str[0] == 'a');
+  });
+  EXPECT_TRUE(inlet->verify());
+
+  inlet->registerFieldVerifier([&](axom::sidre::Group* group) -> bool {
+    std::string str = group->getView("NewTable/field3/value")->getString();
+    return (str.size() >= 1 && str[0] == 'a');
+  });
+
+  EXPECT_FALSE(inlet->verify());
+
 }
 
 //------------------------------------------------------------------------------
--- a/src/axom/inlet/Inlet.cpp
+++ b/src/axom/inlet/Inlet.cpp
@@ -199,6 +199,14 @@ void Inlet::writeDoc() {
 bool Inlet::verify() {
   bool verifySuccess = true;
   verifyRecursive(m_sidreRootGroup, verifySuccess);
+  
+  for (auto& lambda : m_verificationLambdas) {
+    if (!lambda(m_sidreRootGroup)) {
+      SLIC_WARNING("registered verification failed");
+      verifySuccess = false;   
+    }  
+  }
+
   return verifySuccess;
 }
 
@@ -328,7 +336,9 @@ bool Inlet::searchValidValues(axom::sidre::Group* sidreGroup, std::string value)
   return false;
 }
 
-
+void Inlet::registerFieldVerifier(std::function<bool(axom::sidre::Group*)> lambda) {
+  m_verificationLambdas.push_back(lambda);
+}
 
 } // end namespace inlet
 } // end namespace axom
--- a/src/axom/inlet/Inlet.hpp
+++ b/src/axom/inlet/Inlet.hpp
@@ -286,7 +286,8 @@ public:
    * This recursively checks the correctness of each Field and Table in the Sidre
    * Group: ensuring that required Fields are specified, each Field's value 
    * and default value are within the specified range or are equal to a valid 
-   * value, and types are consistent.
+   * value, and types are consistent. Also ensures that the registered verification
+   * lambdas hold true.
    * 
    * \return true if contents are correct and false if not.
    *
@@ -294,6 +295,12 @@ public:
    */
   bool verify(); 
 
+   /*!
+   *****************************************************************************
+   * \brief Requires the lambda-specified verification to pass when verify() is 
+   * called.
+   *****************************************************************************
+   */
   void registerFieldVerifier(std::function<bool(axom::sidre::Group*)> lambda);
 
   // TODO add update value functions
--- a/src/axom/inlet/Inlet.hpp
+++ b/src/axom/inlet/Inlet.hpp
@@ -18,6 +18,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <functional>
 
 #include "axom/inlet/SchemaCreator.hpp"
 #include "axom/inlet/Table.hpp"
@@ -293,6 +294,8 @@ public:
    */
   bool verify(); 
 
+  void registerFieldVerifier(std::function<bool(axom::sidre::Group*)> lambda);
+
   // TODO add update value functions
 private:
   /*!
@@ -413,6 +416,7 @@ private:
 
   std::shared_ptr<DocWriter> m_docWriter;
   bool m_docEnabled = false;
+  std::vector<std::function<bool(axom::sidre::Group*)>> m_verificationLambdas;
 };
 
 } // end namespace inlet


src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -1058,8 +1058,30 @@ TEST(inlet_Inlet_verify, verifyValidStringValues) {
...
+}
+
+TEST(inlet_verify, verifyLambda) {
+  std::string testString = "field1 = true; field2 = 'abc'; NewTable = { field3 = 'xyz'; field4 = 'yes' }";
+  DataStore ds;
+  auto inlet = createBasicInlet(&ds, testString);
+
+  auto field1 = inlet->addBool("field1");
+  auto field2 = inlet->addString("field2");
+  auto field3 = inlet->addString("NewTable/field3");
```
joshessman-llnl, line 1070: Instead of having the user implement the `getView` + cast to appropriate type, possibly they could optionally specify a verifier when they add a field, e.g.
```cpp
auto field2 = inlet->addString("field2", [](const std:string& field) { return str.size() >= 1 && str[0] == 'a'; });
```
with signature `addString(name, std::function<bool(string)> verifier, desc)`

...and then wrap up the boilerplate getView + cast when you register the lambda, since you'd know both the name of the view and the type you'd need to cast it to.

white238: an alternative is to follow the ongoing convention of chaining functions like so:

`auto field2 = inlet->addString("field2")->addVerifier([](const std:string& str) { return str.size() >= 1 && str[0] == 'a'; });`

That's assuming the only thing users would want to verify is the value.  Which might be true. 

I would like to avoid users having to do sidre calls but maybe that's wishful thinking.

@kennyweiss  thoughts?

kennyweiss: I agree with @white238 -- 
* The `Field` class should have an `addVerifier()` function that takes a `std::function` (which could also be a `lambda`).
* The `field` has sufficient context to know the path to the associated group, so that does not need to be passed in
* The sidre abstraction should definitely not be exposed to the user -- that is an implementation detail that they shouldn't have to be concerned with.

Does the verifier need any additional parameters, in general?

kennyweiss: The way `cli11` does this makes sense to me: https://github.com/CLIUtils/CLI11#validators

They have a function called `check()` that takes a `Validator` object, and there are several built-in validators.

For `Inlet`, it would make sense to start with a lambda without parameters. If more structure is needed, perhaps this can be extended to an abstract base class that can be used to validate the fields.

Bonus points to consider for the future: It would be **really** cool if you could define the validators in a scripting language like lua.  (I'm not entirely sure how that would work though...)

aeshapar: I agree that the Sidre details should ideally not be revealed to users. If lambda verifiers would only be needed for checking the values of Fields and no other properties, I can go ahead and add addVerifier(). However, if we would like to allow users to verify multiple Field properties, the above approaches wouldn't allow that flexibility. Is providing lambda verifiers for only Field values sufficient?

kennyweiss: Fair enough.  

In these cases, it almost always makes sense to brainstorm some expected use cases and program for those, rather than something generic that will work in all possible cases.

I'd imagine that most validators would be on a single field. Perhaps the more general ones can be validators on an entire `Table`?

white238: I think there is probably multiple levels of verifiers we want to add...

* Specific Field
* Specific Table
* a global one that works on all Fields
* same but on Tables
* Then one that exposes the sidre root group and the users is expected to know what they are doing.

kennyweiss:
@white238 -- It seems like it would be easier for a user to define and maintain verifiers that are defined on the objects (e.g. field and table).

I'd suggest adding the verifiers as they are needed, starting with the per-field and per-table verifiers, since their use case is clear.

src/axom/inlet/Field.hpp
```diff
@@ -257,6 +258,17 @@ class Field : public std::enable_shared_from_this<Field>
    *****************************************************************************
   */
   std::shared_ptr<Field> validValues(const std::initializer_list<double>& set);
+
+  std::shared_ptr<Field> registerVerifier(std::function<bool()> lambda);
+  bool verify();
```
white238, line 263: If i read this correctly this function (verify) is not meant to be called by the user but is called by the Inlet::verify function.  It needs to be documented and that explicitly called out.

src/axom/inlet/Field.cpp
```diff
@@ -285,5 +285,23 @@ std::shared_ptr<Field> Field::validValues(const std::initializer_list<double>& s
   return validValues(std::vector<double>(set));
 }
 
+std::shared_ptr<Field> Field::registerVerifier(std::function<bool()> lambda) {
+  verifier = lambda;
```
white238, line 289: Since only one is allowed, we should throw a warning if there is already one.

src/axom/inlet/Field.cpp
```diff
@@ -285,5 +285,23 @@ std::shared_ptr<Field> Field::validValues(const std::initializer_list<double>& s
...
+  if (verifier && !verifier()) {
+    SLIC_WARNING(fmt::format("Field {0} failed verification", 
+                 m_sidreGroup->getPathName()));
+    return false;
+  }
+  return true;
+}
+
+std::string Field::getName() {
+  return m_sidreGroup->getName();
```
white238, line 303: This exposes a need for another helper function.  This name isn't enough but the getPathName is probably too much.

https://axom.readthedocs.io/en/develop/doxygen/html/classaxom_1_1sidre_1_1Group.html#a7a4cb38dfb777394b1e60d66ec5158bb

Can you create a function called something like getFullName in the new helper function files that takes a Field or a Table and returns the sidre path name minus the root groups name?

So basically if Inlet's root group is "something/inlet" and the path name of this sidre group/view is "somethign/inlet/sometable/somefield" it would just return "sometable/somefield".

src/axom/inlet/Table.cpp
```diff
@@ -193,5 +192,47 @@ bool Table::required()
...
+
+bool Table::hasField(const std::string& fieldName) {
+  return m_fieldChildren.find(getFullName(m_name, fieldName)) != m_fieldChildren.end();
+}
+
+bool Table::hasTable(const std::string& tableName) {
+  return m_tableChildren.find(getFullName(m_name, tableName)) != m_tableChildren.end();
+}
+
+std::string Table::getPathName() {
```
white238, line 233: getPath, Table::getPathName, and getFullName are a bit confusingly named.

'path' doesn't make sense for Inlet as it is a convention of Sidre.  For Inlet, to make things easy on ourselves we should only report the fully qualified variable name without the Sidre prefix.

For example, the user may have an input deck like:

```
foo.bar = 3
```

but if they choose to store their inlet tree under the sidre group "serac/inlet", something that is not exposed to the user on the Serac level, throwing a warning like the following might confuse the user that something else has gone wrong:

```
Warning: Inlet variable "serac/inlet/foo/bar" has an invalid number.
```

Can we simplify this to two name conventions? getFullSidrePath and getName?

src/axom/inlet/Table.hpp
```diff
@@ -16,6 +16,8 @@
 
 #include <memory>
 #include <string>
+#include <functional>
+#include <unordered_set>
```
joshessman-llnl, line 20: It looks like only `unordered_map` is used in this header and not `unordered_set`

white238:
So those comments ended up in a not great spot. If they confuse you I will recomment in the places i made them.

src/axom/inlet/Table.cpp
```diff
@@ -193,5 +246,95 @@ bool Table::required()
...
+
+bool Table::hasField(const std::string& fieldName) {
+  return m_fieldChildren.find(getFullName(m_name, fieldName)) != m_fieldChildren.end();
+}
+
+bool Table::hasTable(const std::string& tableName) {
+  return m_tableChildren.find(getFullName(m_name, tableName)) != m_tableChildren.end();
+}
+
+std::shared_ptr<Table> Table::getTable(std::string name) {
```
white238, line 287: I feel like getTable and getField should be functions of the Inlet class.

src/axom/inlet/Table.cpp
```diff
@@ -74,15 +99,24 @@ std::shared_ptr<Field> Table::addBool(const std::string& name,
...
 }
 
 std::shared_ptr<Field> Table::addDouble(const std::string& name,
                                         const std::string& description)
 {
   std::string fullName = getFullName(m_name, name);
-  axom::sidre::Group* sidreGroup = baseFieldAdd(fullName, description);
+  size_t found =  name.find_last_of("/");
+
+  auto currTable = shared_from_this();
```
white238, line 114: Some of this can be moved into baseFieldAdd because its common logic between all add Field functions.

src/axom/inlet/Table.cpp
```diff
@@ -193,5 +246,95 @@ bool Table::required()
...
+  m_verifier = lambda;
+  return shared_from_this();
+ } 
+
+ bool Table::verify() {
+   bool verified = true;
+   // Verify this Table
+   if (m_verifier && !m_verifier()) {
+     verified = false;
+     SLIC_WARNING(fmt::format("Table {0} failed verification", m_sidreGroup->getPathName()));
```
white238, line 261: Use the Table's name not getPathName here.

src/axom/inlet/Table.cpp
white238: Also change the parameter to "const std:string& name" and please don't alter it during the finding.

src/axom/inlet/inlet_utils.cpp
```diff
@@ -12,5 +12,21 @@ void setWarningFlag(axom::sidre::Group* root) {
   }
 }
 
+std::string getFullName(const std::string& prefix, const std::string& name)
+{
+  if (prefix == "")
```
white238, line 17: Can you update this to a Ternary Operator?  I admit still I'm forcing myself to get used to them.

```
std::string getFullName(const std::string& prefix, const std::string& name)
{
    return (prefix == "") ? name : prefix + "/" + name;
}
```

src/axom/inlet/inlet_utils.hpp
```diff
@@ -10,16 +10,40 @@ namespace inlet
...
+*****************************************************************************
+* \brief This function appends the prefix name to the ending name.
+*
+* \param [in] The prefix string name.
+* \param [in] The ending string name.
+*
+* \return The appended string.
+*****************************************************************************
+*/
+std::string getFullName(const std::string& prefix, const std::string& name);
```
white238, line 33: These are confusingly named.  Maybe

 getFullName -> appendPrefix
 getName -> removePrefix

src/axom/inlet/inlet_utils.cpp
```diff
@@ -12,5 +12,21 @@ void setWarningFlag(axom::sidre::Group* root) {
...
+    return name;
+  }
+  else
+  {
+    return prefix + "/" + name;
+  }
+}
+
+std::string getName(const std::string& prefix, const std::string& pathName) {
+  return pathName.substr(prefix.size()-1);
```
white238, line 28: Throw an error if pathName doesn't start with prefix.

src/axom/inlet/inlet_utils.cpp
```diff
@@ -12,5 +12,17 @@ void setWarningFlag(axom::sidre::Group* root) {
   }
 }
 
+std::string appendPrefix(const std::string& prefix, const std::string& name)
+{
+  return (prefix == "") ? name : prefix + "/" + name;
+}
+
+std::string removePrefix(const std::string& prefix, const std::string& name) {
+  SLIC_ASSERT_MSG(name.find(prefix) != std::string::npos,
```
white238, line 21: https://github.com/LLNL/axom/blob/b156d86fae4c93e51eb247269c2ee70e08f9d246/src/axom/core/utilities/StringUtilities.hpp#L51

I should have linked you this before but please use this as its more accurate and already written.

src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -1058,8 +1058,119 @@ TEST(inlet_Inlet_verify, verifyValidStringValues) {
...
+    std::string str = "";
+    inlet->get("NewTable/field3", str);
+    return (str.size() >= 1 && str[0] == 'x');
+  });
+  EXPECT_TRUE(inlet->verify());
+
+  table1->registerVerifier([&]() -> bool {
+    return table1->hasField("field3");
+  });
+  EXPECT_TRUE(inlet->verify());
```
kennyweiss, line 1114: Could you please clarify something:  
What would have happened if you had `table1->required()`  and it was not present? I.e. which code check would fail and where and when would that happen?

src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -1058,8 +1058,119 @@ TEST(inlet_Inlet_verify, verifyValidStringValues) {
...
+
+  table1->registerVerifier([&]() -> bool {
+    return table1->hasField("field22");
+  });
+  EXPECT_FALSE(inlet->verify());
 
+  std::string testString2 = "thermal_solver={}\n"
```
kennyweiss, line 1121: Should this be a separate test? Or are you testing that you have have two separate `inlet` instances?

src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -1058,8 +1058,119 @@ TEST(inlet_Inlet_verify, verifyValidStringValues) {
...
+  auto inlet2 = createBasicInlet(&ds2, testString2);
+  auto thermalOrder = inlet2->addInt("thermal_solver/order");
+  auto thermalTimeStep = inlet2->addString("thermal_solver/timestepper");
+  auto solidOrder = inlet2->addInt("solid_solver/order");
+  auto soldTimestep = inlet2->addString("solid_solver/timestepper");
+  auto attrib = inlet2->addInt("material/attribute");
+  auto globalTable = inlet2->getGlobalTable();
+  auto material = globalTable->getTable("material");
+
+  globalTable->registerVerifier([&]() -> bool {
```
kennyweiss, line 1141: This is nice!

kennyweiss, line 1141: I haven't gotten there yet, but does inlet support multiple verifiers on the same Field or Table?

kennyweiss: In either case, perhaps there should be a test that attempts to register more than one verifier.

src/axom/inlet/inlet_utils.hpp
```diff
@@ -10,16 +12,40 @@ namespace inlet
...
+*****************************************************************************
+* \brief This function appends the prefix name to the ending name.
+*
+* \param [in] The prefix string name.
+* \param [in] The ending string name.
+*
+* \return The appended string.
+*****************************************************************************
+*/
+std::string appendPrefix(const std::string& prefix, const std::string& name);
```
kennyweiss, line 35: Should this function take an optional third parameter for the separator? With a default of "/" ?

src/axom/inlet/inlet_utils.cpp
```diff
@@ -12,5 +12,17 @@ void setWarningFlag(axom::sidre::Group* root) {
...
+std::string appendPrefix(const std::string& prefix, const std::string& name)
+{
+  return (prefix == "") ? name : prefix + "/" + name;
+}
+
+std::string removePrefix(const std::string& prefix, const std::string& name) {
+  SLIC_ASSERT_MSG(axom::utilities::string::startsWith(name, prefix), 
+                  fmt::format("Provided name {0} does not contain prefix{1}", 
+                  name, prefix));
+  return name.substr(prefix.size()-1);
```
kennyweiss, line 24: One issue with this approach is that it will only fail in debug builds, and happily remove N characters from your string without and indications of trouble in release builds.

A different approach would be to remove the prefix if it is there, and otherwise just return the original string.

kennyweiss: Also, it would probably be a good idea to check that `name` has length at least `prefix.size()`.

src/axom/inlet/Table.hpp
```diff
@@ -240,16 +327,38 @@ class Table : public SchemaCreator, public std::enable_shared_from_this<Table>
...
   std::string m_name;
   std::shared_ptr<Reader> m_reader;
   // Inlet's Root Sidre Group
   axom::sidre::Group* m_sidreRootGroup;
   // This Table's Sidre Group
   axom::sidre::Group* m_sidreGroup;
   bool m_docEnabled;
+  std::unordered_map<std::string, std::shared_ptr<Table>> m_tableChildren;
+  std::unordered_map<std::string, std::shared_ptr<Field>> m_fieldChildren;
+  std::function<bool()> m_verifier;
```
kennyweiss, line 361: Ok -- so there is only on verifier.

src/axom/inlet/Table.cpp
```diff
@@ -193,5 +217,97 @@ bool Table::required()
   return (bool)intValue;
 }
 
+ std::shared_ptr<Table> Table::registerVerifier(std::function<bool()> lambda) {
+  SLIC_WARNING_IF(m_verifier, fmt::format("Verifier for Table {0} already set",
```
kennyweiss, line 221: Ok -- so it is a warning to register a second verifier, but the new one will then overwrite the old one.

src/axom/inlet/tests/inlet_Inlet.cpp
white238: inlet->verify() will fail

src/axom/inlet/tests/inlet_Inlet.cpp
white238: It does not support more than one verifier per Field or Table, it should throw a warning if that is the case and just overwrite what was there.

I agree there should be a test.

src/axom/inlet/inlet_utils.hpp
white238: There needs to be a follow up PR that adds support for changing the default '/'

src/axom/inlet/tests/inlet_Inlet.cpp
aeshapar: It would probably be better in a separate test; I have moved it

src/axom/inlet/Field.hpp
```diff
@@ -306,6 +332,7 @@ class Field : public std::enable_shared_from_this<Field>
   axom::sidre::Group* m_sidreRootGroup = nullptr;
   axom::sidre::DataTypeId m_type = axom::sidre::DataTypeId::NO_TYPE_ID;
   bool m_docEnabled = false;
+  std::function<bool()> verifier;
```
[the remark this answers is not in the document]
aeshapar: Whenever the member verifier is not initialized, verify returns true (since that means it passed all "verification").

src/axom/inlet/Table.hpp
```diff
@@ -240,16 +347,62 @@ class Table : public SchemaCreator, public std::enable_shared_from_this<Table>
...
+   *****************************************************************************
+   * \brief Retrieves the matching Table.
+   * 
+   * \param [in] The string indicating the target name of the Table to be searched for.
+   * 
+   * \return The Table matching the target name. If no such Table is found,
+   * a nullptr is returned.
+   *****************************************************************************
+   */
+  std::shared_ptr<Table> getTablePtr(const std::string& tableName);
```
white238, line 382: Can you add to the documentation that this is the internal implementation of getTable and getField.  I was trying to think of a more descriptive name but I'm at a loss.

src/axom/inlet/Table.hpp
```diff
@@ -231,6 +235,109 @@ class Table : public SchemaCreator, public std::enable_shared_from_this<Table>
...
+
+  /*!
+   *****************************************************************************
+   * \brief Return whether a Feld with the given name is present in this Table's
+   *  subtree.
+   *
+   * \return Boolean value of whether this Table's subtree contains this Field.
+   *****************************************************************************
+   */
+  bool hasField(const std::string& fieldName);
```
white238, line 283: Why is there a hasField vs hasChildField (same for Table)? what is a child vs a subtree?

white238: Also "Feld"

PR #318, opened 2020-08-03
Feature/essman/bring in sidredc

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Adds assertions to the imported `SidreDataCollection` to catch invalid states (e.g. attempting to add a field when a mesh has not been configured)
  - Adds an implementation of `SidreDataCollection::verifyMeshBlueprint` to check that the internal datastore satisfies the Mesh Blueprint
  - Adds a set of tests for the imported class that focus on ensuring that common operations do not invalidate the mesh blueprint, including save/reload
  - Adds some introductory documentation on using the imported class 

This PR is a WIP and ideally would not be merged until the MFEM PR to remove the class (forthcoming) is approved.   
Some remaining questions include:

- Whether `SidreDataCollection::verifyMeshBlueprint` should be a public method (currently this is only used for testing the SDC class)
- Whether the MFEM examples (ex9, ex9p, etc) using `SidreDataCollection` should be added to the Sidre examples (perhaps a Sidre version of a simpler MFEM example would be better), as breaking the circular dependency requires removing them from MFEM


cyrush:
could you already get the conduit trees you wanted with toNativeLayout?

src/axom/sidre/core/Group.cpp
```diff
@@ -2348,6 +2350,23 @@ bool Group::importConduitTreeExternal(conduit::Node &node,
...
+ *
+ * Exports this Group into a Conduit node tree.
+ * This exports a generic conduit tree, not one with sidre conventions.
+ *
+ *************************************************************************
+ */
+
+conduit::Node Group::exportConduitTree() 
+{
+  conduit::Node result;
```
cyrush, line 2364: FYI: Returning with this style will invoke lots of copies (including copies of external data) - so the style we typically use is to pass in a ref to a Node.

joshessman-llnl:
Looks like the `NativeLayout` function was what I wanted, good catch!

src/axom/sidre/datacollection/SidreDataCollection.hpp
```diff
@@ -0,0 +1,528 @@
...
+    Therefore, field registration is left to the user of SidreDataCollection and
+    there are no methods that automatically register GridFunction%s using just
+    the content of the Sidre DataStore. Such capabilities can be implemented in
+    a derived class, adding any desired object management routines.
+
+    @warning This class is still _experimental_, meaning that in future
+    releases, it may not be backward compatible, and the output files generated
+    by the current version may become unreadable.
+*/
+class SidreDataCollection : public mfem::DataCollection
```
gzagaris, line 157: A couple of questions/suggestions for your consideration:

- Since this class in now in Axom, would it be better to call it `MFEMDataCollection`, since it can only be used with MFEM?

- Do you envision adding more DataCollection classes in the future or just this one? If not does it need to be in its own directory?

src/axom/sidre/datacollection/SidreDataCollection.hpp
```diff
@@ -0,0 +1,528 @@
...
+{
+public:
+   typedef mfem::NamedFieldsMap< mfem::Array<int> > AttributeFieldMap;
```
gzagaris, line 160: minor style: in axom we tend to prefer the _alias_ syntax instead of the `typedef`:
```
using AtributeFieldMap = mfem::NameFieldMap< mfem::Array<int> >;
```

src/axom/sidre/datacollection/SidreDataCollection.hpp
cyrush: @gzagaris  good point -- MFEMDataCollection or some change from SidreDataCollection would cause less confusion.
what about: `MFEMSidreDataCollection` , since there is already `mfem::DataCollection` in mfem.

src/axom/sidre/datacollection/SidreDataCollection.hpp
joshessman-llnl: Good catch, I went ahead and made some other C++11 updates as well (e.g. `NULL` to `nullptr`.)

joshessman-llnl:
The Sidre parallel IO in the imported class requires that MFEM is built with MPI support, which is currently [unconditionally disabled by Axom](https://github.com/LLNL/axom/blob/develop/scripts/uberenv/packages/axom/package.py#L127).  Is there any obstacle to having MFEM's MPI variant enabled when Axom's is, i.e.:
```py
depends_on("mfem~mpi~hypre~metis~zlib", when="+mfem~mpi")
depends_on("mfem+mpi~hypre~metis~zlib", when="+mfem+mpi")
```

cyrush:
> The Sidre parallel IO in the imported class requires that MFEM is built with MPI support, which is currently [unconditionally disabled by Axom](https://github.com/LLNL/axom/blob/develop/scripts/uberenv/packages/axom/package.py#L127). Is there any obstacle (possibly within Quest) to having MFEM's MPI variant enabled when Axom's is, i.e.:
> 
> ```python
> depends_on("mfem~mpi~hypre~metis~zlib", when="+mfem~mpi")
> depends_on("mfem+mpi~hypre~metis~zlib", when="+mfem+mpi")
> ```

This will introduce a link dependency of MPI on anything that uses MFEM. So I think we should work through the details of what that implies.

white238:
> > The Sidre parallel IO in the imported class requires that MFEM is built with MPI support, which is currently [unconditionally disabled by Axom](https://github.com/LLNL/axom/blob/develop/scripts/uberenv/packages/axom/package.py#L127). Is there any obstacle (possibly within Quest) to having MFEM's MPI variant enabled when Axom's is, i.e.:
> > ```python
> > depends_on("mfem~mpi~hypre~metis~zlib", when="+mfem~mpi")
> > depends_on("mfem+mpi~hypre~metis~zlib", when="+mfem+mpi")
> > ```
> 
> This will introduce a link dependency of MPI on anything that uses MFEM. So I think we should work through the details of what that implies.

Trying this build now...

src/axom/sidre/datacollection/MFEMSidreDataCollection.hpp
```diff
@@ -0,0 +1,558 @@
...
+
+/** @brief Data collection with Sidre routines following the Conduit mesh
+    blueprint specification. */
+/** MFEMSidreDataCollection provides an HDF5-based file format for visualization
+   or
+    restart capability.  This functionality is aimed primarily at customers of
+    LLNL's axom project that run problems at extreme scales.
+
+    For more information, see:
+    - Sidre component of LLNL's axom project (to be open-sourced),
```
white238, line 31: It has been open-sourced! man would we be in trouble otherwise... (yes i realize its a copy/paste.. i have to amuse myself)

gzagaris: I'd say this comment should probably go away since this is now *part* of sidre... 😄

src/axom/sidre/CMakeLists.txt
```diff
@@ -95,6 +95,11 @@ if (ENABLE_FORTRAN)
     endif()
 endif()
 
+if (MFEM_FOUND AND ENABLE_MFEM_SIDRE_DATACOLLECTION)
```
white238, line 98: Can you add this option to ``src/cmake/AxomOptions.cmake`` defaulted to OFF?

white238 (approved):

rhornung67 (approved):

src/axom/sidre/CMakeLists.txt
```diff
@@ -105,6 +110,7 @@ set(sidre_depends
 
 blt_list_append(TO sidre_depends ELEMENTS hdf5 IF HDF5_FOUND)
 blt_list_append(TO sidre_depends ELEMENTS sparsehash IF SPARSEHASH_FOUND)
+blt_list_append(TO sidre_depends ELEMENTS mfem IF MFEM_FOUND)
```
kennyweiss, line 113: This can be potentially problematic: I think `sidre` should only depend on `mfem` when `MFEM_FOUND AND AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION`

src/axom/sidre/tests/mfem_sidre_datacollection.cpp
```diff
@@ -0,0 +1,119 @@
...
+}
+
+TEST(sidre_datacollection, dc_save)
+{
+    // 1D mesh divided into 10 segments
+    mfem::Mesh mesh(10);
+    MFEMSidreDataCollection sdc(COLL_NAME, &mesh);
+    
+    // The data produced isn't important for this, so just throw it in /tmp
+    sdc.SetPrefixPath("/tmp/dc_save_test");
```
kennyweiss, line 90: Will this work on Windows?

src/axom/sidre/docs/sphinx/mfem_sidre_datacollection.rst
```diff
@@ -0,0 +1,74 @@
...
+   - ``conduit_hdf5``
+
+.. Note::
+   The ``AXOM_USE_HDF5`` build option must be enabled to use the HDF5-based formats.
+
+A typical visualization example might look like the following:
+
+.. code-block::
+
+   mfem::Mesh mesh;
```
kennyweiss, line 42: Would it be possible to move this code snippet to a `sidre` example, that way we are sure that it continues to work?

src/axom/sidre/core/MFEMSidreDataCollection.hpp
```diff
@@ -0,0 +1,556 @@
+// Copyright (c) 2017-2020, Lawrence Livermore National Security, LLC and
```
kennyweiss, line 1: @joshessman-llnl  -- are there any notable diffs between this and the original mfem versions?

src/axom/sidre/tests/mfem_sidre_datacollection.cpp
```diff
@@ -0,0 +1,119 @@
...
+    sdc.Save();
+
+    EXPECT_TRUE(sdc.verifyMeshBlueprint());
+}
+
+TEST(sidre_datacollection, dc_reload)
```
kennyweiss, line 96: @joshessman-llnl -- should we add some MPI tests for `MFEMSidreDataCollection`?

kennyweiss (changes requested):
Thanks @joshessman-llnl !

Could you please:

- [ ] Verify that this works on Windows. If you do not have access to Windows, please let me know and I can build/run the code there.
- [x] Run `make style` (if you have not already done this)
- [x] Move the code snippet from the user docs to a sidre example. Perhaps this can be done in a follow up PR with a simplified mfem `ex9`?
- [ ] Update the `RELEASE-NOTES`.  In the update, please be sure to mention the changes to the dependencies, and mention that this code was adapted from mfem's `SidreDataCollection` class

kennyweiss:
@joshessman-llnl  wrote:
>  Some remaining questions include:
    Whether SidreDataCollection::verifyMeshBlueprint should be a public method (currently this is only used for testing the SDC class)

Please keep this function as part of the public API for now.  It can be helpful in diagnosing errors.

In the future, we can remove all (most of) the logic related to explicitly generating the blueprint index in favor of conduit's native blueprint index generation functionality.

src/axom/sidre/core/MFEMSidreDataCollection.hpp
joshessman-llnl: Other than the format changes and the `VerifyMeshBluePrint` implementation, I don't believe so.

joshessman-llnl:
> Verify that this works on Windows. If you do not have access to Windows, please let me know and I can build/run the code there.

I don't have a MSVC setup, so I would appreciate your help with running the tests on Windows.

As for https://github.com/LLNL/axom/pull/318#discussion_r493190254, the CI was running the tests for this class up until yesterday when I added the guard option, so it must have been succeeding.  I'm not sure what that line would have done on Windows, though, so it may be preferable to leave the prefix off entirely and just use the working directory.

src/axom/sidre/examples/mfem_sidre_datacoll.cpp
```diff
@@ -0,0 +1,64 @@
...
+  for(int i = 0; i < n_iter; i++)
+  {
+    // Calculate the next iteration of the solution field...
+    // For simplicity, every element in the field is set to the current time
+    soln = dt * i;
+
+    // then save it after updating the time information...
+    dc.SetCycle(i);
+    dc.SetTime(dt * i);
+    dc.Save("mfem_sidre_datacoll_ex", "sidre_hdf5");
```
kennyweiss, line 62: I wasn't able to load these files into Visit on Windows. I suspect it is due to `hdf5-1.8` in VisIt vs. `hdf5-1.10` in Axom.

kennyweiss: @joshessman-llnl  -- were you able to load these `*.root` files into VisIt on LC systems? 

I got the following error:
>VisIt could not read from the file ".../build-axom/mfem_sidre_datacoll_ex_*.root database".
  .
  The generated error message was:
  .
 There was an error opening .../build-axom/mfem_sidre_datacoll_ex_*.root database. It may be an invalid file.
VisIt tried using the following file format readers to open the file: Blueprint, PDB, Silo

It might be due to a missing MPI.... I am checking now.

src/axom/sidre/CMakeLists.txt
```diff
@@ -105,6 +110,7 @@ set(sidre_depends
 
 blt_list_append(TO sidre_depends ELEMENTS hdf5 IF HDF5_FOUND)
 blt_list_append(TO sidre_depends ELEMENTS sparsehash IF SPARSEHASH_FOUND)
+blt_list_append(TO sidre_depends ELEMENTS mfem IF MFEM_FOUND AND AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION)
```
kennyweiss, line 113: > MFEM_FOUND AND AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION

@white238  -- I thought these compound expressions don't work in this macro.
I am checking that this is doing the right thing, and will post what I find out.

kennyweiss (commented):
Working through some issues....

src/axom/sidre/CMakeLists.txt
kennyweiss: I think this line has to change...

I just configured with:
```
>./config-build.py -hc host-configs/rzgenie-toss_3_x86_64_ib-gcc@8.1.0.cmake -DAXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION=OFF
```
and mfem was included in sidre's dependencies.  I.e. `blt_print_target_properties(TARGET mfem)` included:
```
-- [sidre property] BLT_SIDRE_DEPENDS_ON: core;conduit::conduit;slic;hdf5;sparsehash;mfem;fmt;mpi
                                                                                     ~~~~
```

I'll fix this.

kennyweiss: Update: Fixed.

kennyweiss:
@joshessman-llnl, @cyrush, @white238:  Where did we land on our MFEM build requiring MPI?  (see discussion above).

I am not sure if the `MFEMSidreDataStore` will function correctly without this.

white238:
> @joshessman-llnl, @cyrush, @white238: Where did we land on our MFEM build requiring MPI? (see discussion above).
> 
> I am not sure if the `MFEMSidreDataStore` will function correctly without this.

https://github.com/LLNL/axom/pull/319/commits/d531a7b5990ac10a54f0b197cb4d431bfd11f25f

This commit on my previously all encompasing PR (which most has been moved into an already merged PR) shows which spec did not build with mfem+mpi.  It was all name mangling linking problems prior to ever getting to mfem if i remember correctly.

I can go back and attempt this specific thing again very soon.

src/axom/sidre/tests/mfem_sidre_datacollection.cpp
kennyweiss: Apparently, it does.

kennyweiss:
@joshessman-llnl wrote:
> ... the CI was running the tests for this class up until yesterday when I added the guard option, so it must have been succeeding. I'm not sure what that line would have done on Windows,

It is worth noting that our CI on Windows currently (a) disables SIDRE (and all TPLS) and (b) only attempts to compile this code; it does not run the unit tests.

src/axom/sidre/examples/sidre_mfem_datacollection.cpp
```diff
@@ -4,7 +4,7 @@
 // SPDX-License-Identifier: (BSD-3-Clause)
 
 /**
```
kennyweiss, line 6: I think this file will also need

```cpp
#include "axom/config.hpp"

#ifdef AXOM_USE_MPI
#include "mpi.h"
#endif 
```

kennyweiss (approved):
Thanks @joshessman-llnl!

Please add an issue to track the remaining tasks before this feature can be enabled by default.

PR #320, opened 2020-08-06
Feature/taylor/shroud 0.12

ltaylor16:
# Update to Shroud-0.12.2

- This PR is a update
- It does the following:
  - Fortran derived-type constructors are now generic functions named afer the derived type.
    `datastore_new` is now `SidreDataStore`
    `iomanager_new` is now `IOManager`
  - Use `nullptr` instead of `NULL` (C++11).
  - Remove `return` from `void` functions. This helps clang static analysis.
  - Remove unneeded casts when assigning to `void *`.
  - Preserve `const` of pointer arguments in C wrapper.
  - Mangle names of Fortran helper functions with C_prefix. This fixes a conflict when two Shroud generated wrappers are used in the same subprogram. Ideally these helpers would be `PRIVATE`, but gfortran does not allow `PRIVATE` and `BIND(C)` together.

Shroud-0.12.2 will be built in devtools but the public versions have not been updated yet.  For now a public version of shroud can be used with `./config-build.py ...  -DSHROUD_EXECUTABLE=/usr/apps/shroud/bin/shroud-0.12.2`


src/axom/sidre/interface/c_fortran/wrapDataStore.cpp
```diff
@@ -26,7 +26,7 @@ SIDRE_DataStore* SIDRE_DataStore_new(SIDRE_DataStore* SHC_rv)
   // splicer begin class.DataStore.method.new
   axom::sidre::DataStore* SHCXX_rv = new axom::sidre::DataStore();
   SHC_rv->addr = static_cast<void*>(SHCXX_rv);
-  SHC_rv->idtor = 0;
+  SHC_rv->idtor = 1;
```
kennyweiss, line 29: What's `SCH_rv->idtor`?  

Does it make sense for this to have change from 0 to 1?

src/axom/sidre/interface/c_fortran/wrapfsidre.F
```diff
@@ -1625,80 +1625,80 @@ function c_group_create_group_bufferify(self, path, Lpath, &
...
                 bind(C, name="SIDRE_Group_move_group")
             use iso_c_binding, only : C_PTR
-            import :: SHROUD_group_capsule
+            import :: SIDRE_SHROUD_group_capsule
             implicit none
-            type(SHROUD_group_capsule), intent(IN) :: self
-            type(SHROUD_group_capsule), intent(IN) :: grp
-            type(SHROUD_group_capsule), intent(OUT) :: SHT_crv
+            type(SIDRE_SHROUD_group_capsule), intent(IN) :: self
+            type(SIDRE_SHROUD_group_capsule), intent(INOUT) :: grp
```
kennyweiss, line 1673: Was this expected to change?

src/axom/sidre/interface/c_fortran/wrapfsidre.F
```diff
@@ -1865,277 +1865,277 @@ function c_view_get_owning_group(self, SHT_crv) &
...
 
         subroutine c_view_attach_buffer_only(self, buff) &
                 bind(C, name="SIDRE_View_attach_buffer_only")
-            import :: SHROUD_buffer_capsule, SHROUD_view_capsule
+            import :: SIDRE_SHROUD_buffer_capsule, SIDRE_SHROUD_view_capsule
             implicit none
-            type(SHROUD_view_capsule), intent(IN) :: self
-            type(SHROUD_buffer_capsule), intent(IN) :: buff
+            type(SIDRE_SHROUD_view_capsule), intent(IN) :: self
+            type(SIDRE_SHROUD_buffer_capsule), intent(INOUT) :: buff
```
kennyweiss, line 2089: Same question for the next few functions... were these expected to change form `IN` to `INOUT` ?

src/axom/sidre/tests/sidre_allocatable_F.f
```diff
@@ -35,7 +35,7 @@ subroutine external_allocatable_int
 
     call set_case_name("external_allocatable_int")
 
-    ds = datastore_new()
+    ds = SidreDataStore()
```
kennyweiss, line 38: 👍

white238:
If this gets merged before I split my PR. I can handle building it with the devtools

white238:
> If this gets merged before I split my PR. I can handle building it with the devtools

nm.. there are some comments by @kennyweiss  that need to be addressed by Lee.

kennyweiss:
@ltaylor16 -- please merge develop and update the shroud interfaces.  I added a function to the quest `signed_distance` API in https://github.com/LLNL/axom/pull/323

src/axom/sidre/interface/c_fortran/wrapDataStore.cpp
ltaylor16: idtor is the index of the destructor. A bug was fixed with https://github.com/LLNL/shroud/commit/6a740252ad3ddf3d9f105ad5830fbaf843a88d72#diff-03302db14fba2604ab029652dfc22804 which changed the owner to "caller" for constructors. Before, when it was 0, the destructor was not being called properly.
The way most people use the Fortran wrapper, the constructor is not called directly from Fortran. Instead it is called from C++ then the address of the object is stuffed into the Fortran derived type where owner is set to "library" i.e. it will not be destroyed from Fortran.

src/axom/sidre/interface/c_fortran/wrapfsidre.F
ltaylor16: Yes. When a derived type is passed to Fortran with intent(OUT) the compiler will clear all of the members since the routine is declaring that it will assign them (intent(OUT)).  By changing to intent(INOUT), the compiler will not clear the derived type and will not complain if the subroutine assigns to individual members of the struct.  intent(IN) says that no members will change.

src/axom/sidre/interface/c_fortran/wrapfsidre.F
ltaylor16: Yes. See above.

src/axom/sidre/interface/c_fortran/wrapDataStore.cpp
kennyweiss: Thanks @ltaylor16 ! Just checking to make sure there were no inadvertent changes.

PR #328, opened 2020-08-24
Link hdf5 when using Sidre with Mint

gzagaris:
# Summary

In some instances the code was failing to compile b/c
the compiler was not able to resolve the hdf5 header include
in one of the Sidre header. To handle this more robustly, hdf5
needs to be specified explicitely for mint when using Sidre.
Fixed that.

### DIFF
--- a/src/axom/mint/CMakeLists.txt
+++ b/src/axom/mint/CMakeLists.txt
@@ -117,9 +117,10 @@ set( mint_dependencies
      slic
      )
 
-blt_list_append( TO mint_dependencies
-                 ELEMENTS sidre conduit::conduit hdf5
-                 IF AXOM_MINT_USE_SIDRE )
+if (AXOM_MINT_USE_SIDRE)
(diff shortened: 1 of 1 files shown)


src/axom/mint/CMakeLists.txt
```diff
@@ -118,7 +118,7 @@ set( mint_dependencies
      )
 
 blt_list_append( TO mint_dependencies
-                 ELEMENTS sidre conduit::conduit
+                 ELEMENTS sidre conduit::conduit hdf5
```
white238, line 121: (looking at sidre's cmake files) It looks like hdf5 is not a hard requirement of sidre.  Can you should probably guard it here as well.

sidres line:

```
blt_list_append(TO sidre_depends ELEMENTS hdf5 IF HDF5_FOUND)
```

gzagaris: That's a good point -- I'll make that fix

PR #330, opened 2020-08-26
Inlet arrays for basic types

aeshapar:
This PR is a feature. It does the following:
- Expands LuaReader's functionalities to process basic Lua arrays 
- Provides Inlet Arrays for basic types (bool, int, double, string)
- Adds Sol as a TPL for Lua and C++ binding


src/axom/inlet/LuaReader.cpp
```diff
@@ -42,18 +38,12 @@ bool LuaReader::parseFile(const std::string& filePath)
...
-      lua_pcall(m_luaState, 0, 0, 0))
-  {
-    SLIC_WARNING(fmt::format(
-                   "Inlet: Given Lua input deck could not be loaded: {0}",
-                   filePath));
-    m_luaState = nullptr;
-    return false;
+  auto script = m_lua.script_file(filePath);
+  if (!script.valid()) {
+    SLIC_WARNING(fmt::format("Inlet: Given Lua input deck does not exist: {0}",
```
white238, line 43: is this warning message correct? what if it is invalid lua?

src/axom/inlet/LuaReader.cpp
```diff
@@ -64,114 +54,105 @@ bool LuaReader::parseString(const std::string& luaString)
...
-    if(lua_isnil(m_luaState, -1))
-    {
-      // variable not found
-      return false;
-    }
-    atGlobalScope = false;
-  }
-
-  return true;
+bool LuaReader::getInt(const std::string& id, int& value)
```
white238, line 76: Maybe its time to revisit my getInt getBool getDouble getString functions... @kennyweiss  thoughts?

src/axom/inlet/LuaReader.cpp
```diff
@@ -64,114 +54,105 @@ bool LuaReader::parseString(const std::string& luaString)
...
+    if (t[tokens[i]].valid()) {
+      t = t[tokens[i]];
+    } else {
+      return false;
+    }
+  }
+
+  auto it = t.cbegin();
+  while (it != t.cend()) {
+    if ((*it).first.get_type() == sol::type::number
```
white238, line 150: Add a comment clarifying what this if is doing.. "get only numbered items in table" or something.

src/axom/inlet/Table.hpp
```diff
@@ -190,8 +202,8 @@ class Table : public SchemaCreator, public std::enable_shared_from_this<Table>
    *****************************************************************************
    */
   std::shared_ptr<Field> addInt(const std::string& name,
-                                const std::string& description="");
-
+                                const std::string& description="",
+                                bool forArray = false, int num = 0);
```
white238, line 206: Please document the new parameters.

src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -1236,6 +1235,100 @@ TEST(inlet_verify, verifyTableLambda3) {
...
+  inlet->getGlobalTable()->addDoubleArray("luaArrays/arr4");
+  group = inlet->getTable("luaArrays/arr4/_inlet_array")->sidreGroup();
+
+  idx = group->getGroup("12");
+  EXPECT_TRUE(idx);
+  double doubleVal = idx->getView("value")->getScalar();
+  EXPECT_EQ(doubleVal, 2.4);
+
+}
+
```
white238, line 1331: There needs to be more tests of each function added.  There are several new functions that are not called at all in any tests.  Also an example of how the users would use the new functionality is needed.

src/axom/inlet/Table.cpp
```diff
@@ -92,73 +151,65 @@ std::shared_ptr<Field> Table::addField(axom::sidre::Group* sidreGroup,
...
     sidreGroup->createViewScalar("value", value? int8(1) : int8(0) );
   }          
   return addField(sidreGroup, axom::sidre::DataTypeId::INT8_ID, fullName, name);         
 }
 
 std::shared_ptr<Field> Table::addDouble(const std::string& name,
-                                        const std::string& description)
-{
+                                        const std::string& description,
+                                        bool forArray, double num) {
```
white238, line 172: I'm confused by the new parameters.  Can you explain why they are needed? Are you expecting users to ignore them? If not, maybe move this to an internal function the users wont see.

src/axom/inlet/tests/inlet_Inlet.cpp
aeshapar: Agreed on the example. As for the tests, all of the new functions seem to be included in the tests. Could you please point out which function(s) I missed?

white238: Oh i see.. you are doing the addArray's in this test.. then the getMaps above...

src/axom/inlet/Table.cpp
aeshapar: These are needed for when the addArray functions add the individual elements as fields. I would expect users to ignore them.

white238: Then we should move to an internal function that the users dont see.

src/axom/inlet/Table.hpp
```diff
@@ -138,6 +138,102 @@ class Table : public SchemaCreator, public std::enable_shared_from_this<Table>
...
+   *****************************************************************************
+   */
+  std::shared_ptr<Table> addStringArray(const std::string& name,
+                                        const std::string& description="");
+  
+   /*!
+   *****************************************************************************
+   * \brief Get a boolean array represented as an unordered map from the input deck
+   *
+   * \param [in] name Name of the array
```
white238, line 197: These parameters documentations are wrong for all the get*Array functions

src/axom/inlet/examples/arrays.cpp
```diff
@@ -0,0 +1,28 @@
+#include <iostream>
+#include <unordered_map>
+
+#include "axom/inlet.hpp"
+
+void example() {
```
white238, line 6: There is no need for the example function.  Move it to main().  Also return 0 to avoid a warning at the end of main.

Add some comments since this is an example for the users.  For example, "Initialize Inlet" "Parse example input file" "Add verifier that array was found" (also check some of the values inside).

src/axom/inlet/Table.cpp
```diff
@@ -51,8 +50,100 @@ std::shared_ptr<Table> Table::addTable(const std::string& name,
...
+      table->addStringHelper(std::to_string(p.first), "", true, p.second);
+    }
+  } else {
+    SLIC_WARNING(fmt::format("String array {0} not found.", fullName));
+  }
+  return table;
+}
+
+bool Table::getBoolArray(std::unordered_map<int,bool>& map) {
+  size_t found = m_name.find("/_inlet_array");
```
white238, line 114: pretty sure that '/' is not supposed to be there.

white238: Is find specific enough here? is m_name a fully qualified name that may have one embedded in the full path? would endswith (there is axom::utilities::string::endsWith()) be better?

src/CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ if (“${PROJECT_SOURCE_DIR}” STREQUAL “${CMAKE_SOURCE_DIR}”)
     # Set some default BLT options before loading BLT only if not included in
     # another project
     if (NOT BLT_CXX_STD)
-        set(BLT_CXX_STD "c++11" CACHE STRING "")
+        set(BLT_CXX_STD "c++14" CACHE STRING "")
```
kennyweiss, line 50: Is this a strict requirement for inlet / sol ? Or a nice-to-have feature ?

I think this change should be done is a separate PR where it can be tested/validated separately.

kennyweiss: Also, not all of our users are on `C++14` yet, so we should confirm that this will not be a breaking update.

(I am not expecting problems, but we've been burned in the past, and do not want to unnecessarily introduce problems that will prevent our users from upgrading)

white238: According to their docs (which they were nice enough to build for us), it should work with C++11 so we can safely revert this change.  This happened early when we weren't sure. 

https://sol2.readthedocs.io/en/v2.20.6/compilation.html

white238: Though @gzagaris 's PR will soon require C++14 due to raja being updated.  At some point we are going to have to bite that bullet.

src/axom/inlet/LuaReader.hpp
```diff
@@ -15,6 +15,7 @@
 #define INLET_LUAMAP_HPP
 
 #include "axom/inlet/Reader.hpp"
+#include <sol/sol.hpp>
```
kennyweiss, line 18: Is there a missing edit to `src/axom/inlet/CMakeLists.txt` that adds `sol` as a dependency to `inlet`?
(I realize it is header only, but it should still be added).

src/axom/inlet/LuaReader.cpp
kennyweiss: Could be -- definitely in a separate PR though.
on the other hand, the current way makes it explicit for the user.

src/axom/inlet/LuaReader.cpp
```diff
@@ -64,114 +54,107 @@ bool LuaReader::parseString(const std::string& luaString)
...
+    value = m_lua[id];
+    return true;
+  }
+  
+  if (!m_lua[tokens[0]].valid()) {
     return false;
   }
-  value = (int)lua_tonumber(m_luaState, -1);
-  return true;
+  sol::table t = m_lua[tokens[0]];
```
kennyweiss, line 115: Should these be by-value or by-reference?

src/axom/inlet/LuaReader.cpp
```diff
@@ -64,114 +54,107 @@ bool LuaReader::parseString(const std::string& luaString)
...
+  if (t[tokens.back()].valid()) {
+    value = t[tokens.back()];
+    return true;
+  }
+  
+  return false;
 }
 
+template <typename T>
+bool LuaReader::getMap(const std::string& id, std::unordered_map<int, T>& values, sol::type type) {
```
kennyweiss, line 132: Do we always want to extract the data to an explicit map?
Are there potential issues when `T` is not a native type (i.e. int, string, double,bool) ?

kennyweiss (changes requested):
Looks like a nice addition @aeshapar !

I've added some comments.
Mainly, I'm concerned about the change from `C++11` to `C++14` -- this should be tested separately, and possibly deferred to the future if it is not strictly necessary.

Please also:
* document the addition of `sol` in the appropriate places 
* add a note about its license.  @white238 can point out where
* add a line to the RELEASE-NOTES about the new built-in TPL

src/axom/inlet/LuaReader.cpp
```diff
@@ -11,28 +11,24 @@
...
 #include "axom/inlet/LuaReader.hpp"
 
 #include "axom/core/utilities/FileUtilities.hpp"
 #include "axom/core/utilities/StringUtilities.hpp"
+#include "axom/inlet/inlet_utils.hpp"
 
 #include "fmt/fmt.hpp"
 #include "axom/slic.hpp"
 
+#define SOL_ALL_SAFETIES_ON 1
```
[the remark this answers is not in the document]
white238: We do have specific changes to the sol.hpp.  its hard to see them in this PR because the file is too long.  but if you view the sol.hpp file and search for axom you can see them there.

src/CMakeLists.txt
aeshapar: C++ 14 was a requirement for sol

gzagaris: That's correct,  C++14 is required (though no strongly enforced) with newer RAJA. Consequently, everyone is going to be using C++14. All our compilers already support it. We should run with this.

src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -1236,6 +1235,131 @@ TEST(inlet_verify, verifyTableLambda3) {
...
 
+TEST(lua_Arrays, basicArraysInLua) {
+  std::string testString = "luaArray = { [1] = 4, [2] = 5, [3] = 6 , [4] = true, [8] = false, [12] = 2.4, [33] = 'hello', [200] = 'bye' }";
+  LuaReader lr;
+  lr.parseString(testString);
+
+  std::unordered_map<int,int> ints;
+  bool found = lr.getIntMap("luaArray", ints);
+  EXPECT_TRUE(found);
+  std::unordered_map<int,int> expectedInts{{1,4},{2,5},{3,6},{12,2}};
```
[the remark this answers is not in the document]
aeshapar: Yes, since sol doesn't distinguish between doubles and ints, they're all just grouped under sol::type::number

src/axom/inlet/LuaReader.cpp
aeshapar: As of now, we don't provide an interface for extracting non-native types.

src/axom/inlet/LuaReader.cpp
aeshapar: Since we're repeatedly changing t as we search deeper in the table, by-value

src/axom/inlet/CMakeLists.txt
```diff
@@ -41,6 +41,7 @@ blt_list_append(TO inlet_sources ELEMENTS LuaReader.cpp IF LUA_FOUND)
 set(inlet_depends core slic sidre fmt)
 
 blt_list_append(TO inlet_depends ELEMENTS lua IF LUA_FOUND)
+blt_list_append(TO inlet_depends ELEMENTS sol IF SOL_FOUND)
```
white238, line 44: Can you change these two lines to just:

```
blt_list_append(TO inlet_depends ELEMENTS sol lua IF SOL_FOUND)
```

src/axom/inlet/LuaReader.cpp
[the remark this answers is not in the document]
white238: Can you move this line into sol.hpp with the other defines we had to do?

src/axom/inlet/LuaReader.hpp
```diff
@@ -139,23 +134,19 @@ class LuaReader : public Reader
    */
   bool getString(const std::string& id, std::string& value);
 
+  bool getIntMap(const std::string& id, std::unordered_map<int, int>& values);
```
white238, line 137: These need to be documented.

src/axom/inlet/Reader.hpp
```diff
@@ -132,6 +133,11 @@ class Reader
    *****************************************************************************
    */
   virtual bool getString(const std::string& id, std::string& value) = 0;
+
+  virtual bool getIntMap(const std::string& id, std::unordered_map<int, int>& values) = 0;
```
white238, line 137: Same documentation here.

src/axom/inlet/examples/arrays.cpp
```diff
@@ -0,0 +1,36 @@
...
+    if (!vals->getStringArray(map)) {
+      std::cout << "Error: Array not found\n";
+    }
+    return map.size() == 3;
+  });
+
+  // We expect verfication to pass since values array has 3 elements
+  inlet->verify() ? std::cout << "Verification passed\n"
+                  : std::cout << "Verification failed\n";
+  return 0;
```
white238, line 35: Can you print out the found values here?

src/axom/inlet/examples/arrays.cpp
```diff
@@ -0,0 +1,41 @@
...
+  auto inlet = std::make_shared<axom::inlet::Inlet>(lr, ds.getRoot());
+
+  // Register the verifier, which will verify the array values
+  auto vals = inlet->getGlobalTable()->addStringArray("values");
+  vals->registerVerifier([&]() -> bool {
+    std::unordered_map<int,std::string> map;
+    if (!vals->getStringArray(map)) {
+      std::cout << "Error: Array not found\n";
+    } else {
+      std::cout << "Map Contents:\n";
```
white238, line 29: Little strange from a user-perspective to do this in the verify function.  Maybe change this to check that the given values are either "start", "stop", or "pause" to show a semi-real life example.  Then after the verify() call, just loop through the values and print them out like you are now.

white238 (approved):

RELEASE-NOTES.md
```diff
@@ -55,6 +55,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   candidate BVH bins that intersect each bounding box.
 - Added an `axom-config.cmake` file to axom's installation to streamline incorporating axom
   into user applications. See `<axom-install>/examples/axom` for example usages.
+- Added a new built-in TPL, Sol, for a fast, simple C++ and Lua Binding 
+  (automatically enabled when LUA_DIR is found)
```
kennyweiss, line 59: Please add: 
* Which version of SOL we are using
* that SOL requires `C++14`
* a link to SOL.

src/axom/inlet/LuaReader.hpp
```diff
@@ -38,15 +39,9 @@ namespace inlet
...
-   *****************************************************************************
-   * \brief Destructor for the LuaReader.
-   *
-   * This performs any cleanup work the Reader needs to do before going
-   * away.
-   *****************************************************************************
-   */
-  ~LuaReader();
+  LuaReader() {
+    m_lua.open_libraries(sol::lib::base);
```
kennyweiss, line 43: Does this need to be guarded, e.g. with an error message if it cannot open the library, or does SOL/lua already handle that for you?

src/axom/inlet/LuaReader.hpp
```diff
@@ -139,23 +134,74 @@ class LuaReader : public Reader
...
+private:
+  // Expect this to be called for only Inlet-supported types.
+  template<typename T>
+  bool getValue(const std::string& id, T& value);
 
-  lua_State* m_luaState;
+  // Expect this to be called for only Inlet-supported types.
+  template <typename T>
+  bool getMap(const std::string& id, std::unordered_map<int, T>& values, sol::type type);
+  sol::state m_lua;
```
kennyweiss, line 204: Suggestion (feel free to ignore):
For readability, I like to separate private methods from private variables with an extra access modifier, e.g.

```cpp
private:
  // Expect this to be called for only Inlet-supported types.
  template<typename T>
  bool getValue(const std::string& id, T& value);

  // Expect this to be called for only Inlet-supported types.
  template <typename T>
  bool getMap(const std::string& id, std::unordered_map<int, T>& values, sol::type type);

private:
  sol::state m_lua;
```

src/axom/inlet/Table.cpp
```diff
@@ -109,73 +189,66 @@ std::shared_ptr<Field> Table::addField(axom::sidre::Group* sidreGroup,
...
-                                      const std::string& description)
-{
+std::shared_ptr<Field> Table::addBoolHelper(const std::string& name,
+                                            const std::string& description, 
+                                            bool forArray, bool num) {
   std::string fullName = appendPrefix(m_name, name);
   axom::sidre::Group* sidreGroup = createSidreGroup(fullName, description);
-  if(sidreGroup == nullptr)
-  {
+  if (sidreGroup == nullptr) {
```
kennyweiss, line 197: Please run `make style` on your branch before merging.

src/axom/inlet/examples/CMakeLists.txt
```diff
@@ -11,6 +11,11 @@
 #------------------------------------------------------------------------------
 
 if (LUA_FOUND)
```
kennyweiss, line 13: Should this be guarded on `if(SOL_FOUND)`?  It uses `LuaReader`.

kennyweiss: Please build and test this branch against a `C++11`  configuration that has LUA, but not SOL

kennyweiss (changes requested):
Thanks @aeshapar  -- looks like we're almost there.

I had a few suggestions to harden this PR, and to ensure that we don't run into problems related to `C++11` vs. `C++14`.

Please also add a note about SOL and its copyright to the bottom of  `README.md`

If you modified SOL, we should indicate how it was modified. 
@white238  -- perhaps we should also add a patch file?

src/axom/inlet/examples/CMakeLists.txt
white238: SOL is builtin in so she cant... unless she deletes the header from our repo which seems drastic.

src/CMakeLists.txt
kennyweiss: Thanks @gzagaris.  Per our team meeting on 9/14, we will hold off on requiring `C++14` for axom for a few months, but can certainly optionally build with `C++14`

src/axom/inlet/examples/CMakeLists.txt
kennyweiss: Got it, so  `SOL_FOUND` and `LUA_FOUND` should always have the same value?

src/axom/inlet/LuaReader.hpp
aeshapar: It's not needed for our usage of Lua, so I've removed it.

src/axom/inlet/examples/CMakeLists.txt
white238: Not necessarily... for example if you give Axom a LUA_DIR, LUA_FOUND will be true but if you are C++11, SOL_FOUND will be false.  I guarded all the places that were causing problems though.

src/axom/inlet/LuaReader.hpp
white238: @aeshapar can you do this or just put a line break between the member functions and variables?

white238 (approved):

RELEASE-NOTES.md
```diff
@@ -55,8 +55,9 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   candidate BVH bins that intersect each bounding box.
 - Added an `axom-config.cmake` file to axom's installation to streamline incorporating axom
   into user applications. See `<axom-install>/examples/axom` for example usages.
-- Added a new built-in TPL, Sol, for a fast, simple C++ and Lua Binding 
-  (automatically enabled when LUA_DIR is found)
+- Added a new built-in TPL, Sol (https://github.com/ThePhD/sol2), for fast and 
+  simple C++ and Lua Binding (automatically enabled when LUA_DIR is found). 
+  The version of Sol used in this release is v2.20.6, which requires C++14.
```
kennyweiss, line 60: 👍

src/thirdparty/sol/sol.patch
```diff
@@ -0,0 +1,13 @@
+72a73,84
+> // start axom change
```
kennyweiss, line 2: 👍

kennyweiss (commented):
Thanks @aeshapar -- looks great!

kennyweiss (approved):

PR #334, opened 2020-09-03
Clarify when script fails

white238:
small change to reprint the error at the bottom after permission output.


kennyweiss (approved):
Thanks @white238 !

white238:
We are all mad here...

PR #340, opened 2020-09-10
Fix build error when AXOM_VERSION_EXTRA is not defined

white238:
This is fixes a build error when not in a git repository.

... i just want to use a new Axom in Serac....

### DIFF
--- a/src/axom/core/utilities/About.cpp
+++ b/src/axom/core/utilities/About.cpp
@@ -144,7 +144,10 @@ void about(std::ostream &oss)
 std::string getVersion()
 {
   std::ostringstream oss;
-  oss << AXOM_VERSION_FULL << "-" << AXOM_VERSION_EXTRA;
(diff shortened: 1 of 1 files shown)


src/axom/core/utilities/About.cpp
```diff
@@ -144,7 +144,10 @@ void about(std::ostream &oss)
 std::string getVersion()
 {
   std::ostringstream oss;
-  oss << AXOM_VERSION_FULL << "-" << AXOM_VERSION_EXTRA;
+  oss << AXOM_VERSION_FULL;
+#ifdef AXOM_VERSION_EXTRA
+  oss << "-" << AXOM_VERSION_EXTRA;
+#endif
```
gzagaris, line 150: Good catch! Thanks for fixing this.

gzagaris (approved):

kennyweiss (approved):

PR #341, opened 2020-09-16
Guard hdf5

white238:
A user showed that we were guarding hdf5 properly in our lulesh2 example.  I reproduced this by turning off axom's and conduits hdf5 variant then building.  Because hdf5 is not a target in this state CMake blindly puts -lhdf5 on the link line hoping it is on the system (it actually is on rzgenie, unhelpfully?).  But this fix made it go away.

(All the labels for @kennyweiss )


kennyweiss (approved):
Thanks @white238 !

kennyweiss:
I appreciate the high labels-to-edit ratio!

rhornung67 (approved):

PR #350, opened 2020-09-22
new strategy to re-wire conduit handlers

cyrush:
# Summary

Fodder to address issues with general conduit errors being routed to SLIC

* Adds static methods:
  * `DataStore::wireConduitMessageHandlersToSLIC()`
  * `DataStore::restoreConduitDefaultMessageHandlers()`

(Better names welcome)

From Lee's feedback -- I did this instead of a default arg to the constructor b/c this setting will impact all DataStore instances (static method seems more appropriate)

* Wraps our tests that depended on the old behavior with a rewire/restore calls.

This PR doesn't do anything to address providing these methods in C or Fortran. Not sure if we need those options right now, but I do thing those cases (C or Fortran) might actually be a place were we may need this option.


src/axom/sidre/tests/sidre_attribute.cpp
```diff
@@ -161,6 +161,9 @@ TEST(sidre_attribute,create_attr)
 
 TEST(sidre_attribute,view_attr)
 {
+  // Note: This is the test that relies on re-writing conduit error handlers
```
cyrush, line 164: this probably should say `This test relies`.

cyrush:
I branched off of main, I'll fix.

src/axom/sidre/core/DataStore.cpp
```diff
@@ -129,6 +129,40 @@ DataStore::~DataStore()
   }
 }
 
+/*
+ *************************************************************************
+ *
+ * Rewire Conduit Message Handlers to SLIC.
+ *  ... Not the best function name, but its is descriptive :-)
```
cyrush, line 136: `its is`

src/axom/sidre/core/DataStore.cpp
```diff
@@ -129,6 +129,40 @@ DataStore::~DataStore()
...
+
+/*
+ *************************************************************************
+ *
+ * Restore Conduit Message Handlers to Conduit Defaults.
+ *  ... Not the best function name, but it is descriptive :-) 
+ *
+ *************************************************************************
+ */
+void DataStore::restoreConduitDefaultMessageHandlers()
```
agcapps, line 157: "restore" bothers me a little.  I'd rather "set" or "wire".  (Just a naming issue, probably not important...)

agcapps (approved):
Thanks @cyrush . This looks good.

src/axom/sidre/core/DataStore.cpp
cyrush: I agree `set` or `wire` is probably better, let's see if other folks have strong opinions.

kennyweiss: Since this function name is not into the whole brevity thing, how about `resetConduitMessageHandlersToDefault()` ?
(or maybe just `resetConduitMessageHandlers()` )

src/axom/sidre/core/DataStore.cpp
```diff
@@ -129,6 +129,40 @@ DataStore::~DataStore()
...
 
+/*
+ *************************************************************************
+ *
+ * Rewire Conduit Message Handlers to SLIC.
+ *  ... Not the best function name, but it is descriptive :-) 
+ *
+ *************************************************************************
+ */
+void DataStore::wireConduitMessageHandlersToSLIC()
```
kennyweiss, line 140: I'm ok with `wire`, but as an alternate suggestion to complement `reset...` below, how about just `setConduitMessageHandlersToSLIC()` ?

src/axom/sidre/core/DataStore.hpp
```diff
@@ -371,8 +371,8 @@ class DataStore
    *        this DataStore.
    *
    * If this DataStore contains data for a mesh that adheres to Conduit's
-   * Blueprint format, this method genearates a Blueprint index and stores
-   * it at a specified locaiton within this DataStore.  It uses as input a
+   * Blueprint format, this method generates a Blueprint index and stores
+   * it at a specified location within this DataStore.  It uses as input a
```
kennyweiss, line 375: 👍

kennyweiss (approved):
Thanks @cyrush 

Please add a (strongly worded) note/warning to the `RELEASE-NOTES` indicating this change in default behavior and the workaround to get the old behavior back.

src/axom/sidre/core/DataStore.cpp
white238: Is it possible a user might have set the conduit handlers to something non-default and you are inadvertently removing their custom handler?  Like so:

```
// user set to some custom handler
...
// wireConduitMessageHandlersToSLIC()
// Ascent does its thing
// restoreConduitDefaultMessageHandlers()
...
// user no longer has a custom handler
```

src/axom/sidre/core/DataStore.cpp
cyrush: Leaning towards `set`

src/axom/sidre/core/DataStore.cpp
cyrush: @white238  that is possible -- we could implement a push-pop style, but honestly folks can also do that themselves at the same (coarse) level.

I think `setConduitMessageHandlersToDefault()` is the most clear reflection of the logic. I will go with that and `setConduitMessageHandlersToSLIC()`

rhornung67 (approved):
Looks good to me. I'll defer to you folks to decide on the new method names. Thanks @cyrush

cyrush:
thanks all, I am going to merge now.

I also added a ticket (#353) to later audit why we relied on the re-wired error handler logic for those few cases that i found in this pr .

PR #356, opened 2020-09-27
Inlet cleanup

white238:
This PR consists of the the last work that Aesha did.

* Makes SchemaCreator's functions pure virtual and adds missing functions
* Adds ability to access Inlet's held Reader class
* Adds ability to access Sol's Lua state held by LuaReader

### DIFF
--- a/src/axom/inlet/SchemaCreator.hpp
+++ b/src/axom/inlet/SchemaCreator.hpp
@@ -58,8 +58,8 @@ public:
    * \return Shared pointer to the created Table
    *****************************************************************************
    */
-  std::shared_ptr<Table> addTable(const std::string& name,
-                                  const std::string& description);
+  virtual std::shared_ptr<Table> addTable(const std::string& name,
+                                  const std::string& description) = 0;
 
   /*!
    *****************************************************************************
@@ -76,8 +76,8 @@ public:
    * \return Shared pointer to the created Field
    *****************************************************************************
    */
-  std::shared_ptr<Field> addBool(const std::string& name,
-                                 const std::string& description);
+  virtual std::shared_ptr<Field> addBool(const std::string& name,
+                                 const std::string& description) = 0;
 
   /*!
    *****************************************************************************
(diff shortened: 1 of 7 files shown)


agcapps (approved):
Thanks, @white238 and @aeshapar .

src/axom/inlet/SchemaCreator.hpp
```diff
@@ -130,8 +130,162 @@ class SchemaCreator
...
+   */
+  virtual bool hasField(const std::string& fieldName) = 0;
+
+  /*!
+   *****************************************************************************
+   * \return An unordered map from Field names to the child Field pointers for 
+   * this Table.
+   *****************************************************************************
+   */
+  virtual std::unordered_map<std::string, std::shared_ptr<Field>> getChildFields() = 0;
```
gzagaris, line 160: Should this be returned as a `const` reference or by reference to avoid copying the map?

src/axom/inlet/SchemaCreator.hpp
```diff
@@ -130,8 +130,162 @@ class SchemaCreator
...
+   *****************************************************************************
+   * \brief Add an array of Double Fields to the input deck schema.
+   *
+   * \param [in] name Name of the array
+   * \param [in] description Description of the Field
+   *
+   * \return Shared pointer to the created Field
+   *****************************************************************************
+   */
+  virtual std::shared_ptr<Table> addDoubleArray(const std::string& name,
```
gzagaris, line 230: Have you considered using `templates` for the add & get methods here? Would that make more sense? The add/get methods seem to have the same signature, so I am not entirely sure what is the difference in the implementation -- Perhaps, something to think about as your design evolves.

gzagaris (approved):
Looks good to me -- I did a quick pass and added some general API comments for your consideration.

src/axom/inlet/SchemaCreator.hpp
white238: The idea for not templating the add functions was because we lock down what types that Inlet supports (bool, int, double, and string) for correctness. I do go back and forth on this though...

src/axom/inlet/SchemaCreator.hpp
white238: This is probably true.  I need to do a general once over of what is passed around for efficiency.

src/axom/inlet/SchemaCreator.hpp
gzagaris: You can also lock down on the types using `AXOM_STATIC_ASSERT`....

PR #358, opened 2020-09-28
Fix build errors found with Serac's configuration

white238:
I swear I tested this. I am not sure how the hdf5 ever worked. Serac is the only configuration I know that turns on mpi for hdf5 but I turned it on 25 days ago and there has been multiple TPL builds since then...

### DIFF
--- a/src/thirdparty/tests/CMakeLists.txt
+++ b/src/thirdparty/tests/CMakeLists.txt
@@ -37,11 +37,8 @@ if ( RAJA_FOUND )
   set( raja_smoke_dependencies RAJA gtest )
 
   blt_list_append( TO raja_smoke_dependencies ELEMENTS openmp IF ENABLE_OPENMP)
-
   blt_list_append( TO raja_smoke_dependencies ELEMENTS umpire IF UMPIRE_FOUND)
-
   blt_list_append( TO raja_smoke_dependencies ELEMENTS cuda IF ENABLE_CUDA)
-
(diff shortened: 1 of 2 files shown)


src/thirdparty/tests/CMakeLists.txt
```diff
@@ -59,10 +56,15 @@ endif()
...
+
+    set( hdf5_smoke_dependencies hdf5 gtest )
+
+    blt_list_append( TO hdf5_smoke_dependencies ELEMENTS mpi IF ENABLE_MPI)
+
     blt_add_executable(NAME hdf5_smoke_test
                        SOURCES hdf5_smoke.cpp
                        OUTPUT_DIR ${TEST_OUTPUT_DIRECTORY}
-                       DEPENDS_ON hdf5 gtest
+                       DEPENDS_ON ${hdf5_smoke_dependencies}
```
gzagaris, line 67: I guess you are building with an MPI version of HDF5....

gzagaris (approved):

src/thirdparty/tests/CMakeLists.txt
white238: ... but how did this not break 25 days ago... maybe spack has tests off by default.

gzagaris: Hmm...were you using a different version of CMake perhaps? I have run into an issue where different versions of CMake wouldn't propagate dependencies correctly...

agcapps (approved):
Thumbs up

Issue #321, opened 2020-08-07
Add ability to register a custom abort handler to SLIC


white238:
Currently, SLIC_ERROR() and SLIC_ASSERT() call "processAbort()" which, in turn calls std::abort() or  MPI_Abort() + std::abort() when MPI is enabled. Although this is sufficient for the most part, there are some use cases, where it would be desirable to override this default behavior. For example, for supporting death tests within an MPI process, it is desirable not to call MPI_Abort(). Having the ability to register a custom abort handler to Slic could provide one solution to that.

gzagaris:
I think we have a way to do a `setAbortOnError` to false, etc., which I think would be useful for what you are describing.

white238:
I moved this one from Jira because it came up in Serac.  I already have the setAbortOnError and setAbortOnWarning to false, but it would be nice to be able to register a custom abort to cut down on duplicate code throughout the code base.

PR #359, opened 2020-09-30
Allow user to specify an abort callback for slic abort

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Adds the ability for the user to specify a custom abort function - `axom::utilities::processAbort` remains the default

`std::function` could be used in place of raw function pointers.

An example usage was also added.

### DIFF
--- a/src/axom/slic/core/Logger.cpp
+++ b/src/axom/slic/core/Logger.cpp
@@ -20,7 +20,10 @@ Logger* Logger::s_Logger = nullptr;
 std::map<std::string, Logger*> Logger::s_loggers;
 
 //------------------------------------------------------------------------------
-Logger::Logger() : m_abortOnError(true), m_abortOnWarning(false), m_abortFunction(axom::utilities::processAbort)
+Logger::Logger()
+  : m_abortOnError(true)
+  , m_abortOnWarning(false)
+  , m_abortFunction(axom::utilities::processAbort)
 {
   // by default, all message streams are disabled
   for(int i = 0; i < message::Num_Levels; ++i)
--- a/src/axom/slic/core/Logger.cpp
+++ b/src/axom/slic/core/Logger.cpp
@@ -20,7 +20,7 @@ Logger* Logger::s_Logger = nullptr;
 std::map<std::string, Logger*> Logger::s_loggers;
 
 //------------------------------------------------------------------------------
-Logger::Logger() : m_abortOnError(true), m_abortOnWarning(false)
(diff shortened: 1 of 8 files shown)


src/axom/slic/core/Logger.cpp
```diff
@@ -58,6 +62,18 @@ Logger::~Logger()
   }  // END for all levels
 }
 
+//------------------------------------------------------------------------------
+void Logger::setAbortFunction(void (*abort_func)(void))
```
gzagaris, line 66: I would add an `AbortFunctionPtr` alias instead for readability

gzagaris (approved):
Thanks!

src/axom/slic/core/Logger.cpp
```diff
@@ -58,6 +62,18 @@ Logger::~Logger()
...
+{
+  if(abort_func == nullptr)
+  {
+    std::cerr << "WARNING: supplied abort function is NULL!\n";
```
kennyweiss, line 70: Suggestion: Please clarify this error message to let the user know where the problem is, 
e.g. something like:
> WARNING: slic::Logger::setAbortFunction() -- Supplied abort function was invalid

src/axom/slic/interface/slic.cpp
```diff
@@ -147,6 +147,19 @@ bool isAbortOnWarningsEnabled()
   return (Logger::getActiveLogger()->isAbortOnWarningsEnabled());
 }
 
+//------------------------------------------------------------------------------
+void setAbortFunction(void (*abort_func)(void))
+{
+  if(!isInitialized())
+  {
+    std::cerr << "[ERROR]: slic::initialize() must be called first "
```
kennyweiss, line 155: Suggestion: Remove the word `first`

src/axom/slic/examples/basic/custom_abort_logger.cpp
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2017-2020, Lawrence Livermore National Security, LLC and
```
kennyweiss, line 1: 👍 Thanks for the new example!

kennyweiss (approved):
Thanks @joshessman-llnl 

Please also update the `RELEASE-NOTES`

gzagaris:
I believe this would address #321, correct?

joshessman-llnl:
> 
> 
> I believe this would address #321, correct?

Good catch.  This resolves #321

white238 (approved):

PR #360, opened 2020-09-30
Added std::initializer_list constructors to primal primitives

estebanpauli:
The primal class templates NumericArray, Point, and Vector now have
constructors from initializer lists. As per a conversation with Kenny
and Rich, these are not marked with AXOM_HOST_DEVICE for now.

# Summary

- This PR is a  feature
- It does the following
  - Creates constructors on Vector, Point, and NumericArray to create these classes from instances of std::initializer_list.

### DIFF
--- a/src/axom/primal/tests/primal_numeric_array.cpp
+++ b/src/axom/primal/tests/primal_numeric_array.cpp
@@ -57,6 +57,28 @@ TEST(primal_numeric_array, constructors)
     CoordType expVal = i < halfPt ? valsArr[i] : CoordType();
     EXPECT_EQ(arrHalfVec[i], expVal);
   }
+
+  primal::NumericArray<int, 3> fromInitializerListRightSize = {10, 20, 30};
+  for(int i = 0; i < 3; ++i)
+  {
+    EXPECT_EQ(10 * (i + 1), fromInitializerListRightSize[i]);
+  }
+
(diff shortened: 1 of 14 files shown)


src/axom/primal/tests/primal_numeric_array.cpp
```diff
@@ -57,6 +57,28 @@ TEST(primal_numeric_array, constructors)
     CoordType expVal = i < halfPt ? valsArr[i] : CoordType();
     EXPECT_EQ(arrHalfVec[i], expVal);
   }
+
+  primal::NumericArray<int, 3> fromInitializerListRightSize = {10, 20, 30};
```
kennyweiss, line 61: It might be helpful to also have some examples of the form
```cpp
   primal::NumericArray<int, 3> fromInitializerList{10, 20, 30}; 
```

Or perhaps:
```cpp
   primal::NumericArray<int, 3> fromInitializerList{{10, 20, 30}}; 
```

kennyweiss (approved):
Thanks @estebanpauli !

gzagaris (approved):
Thanks!

rhornung67 (approved):
Looks good. Thanks! 👍

src/axom/primal/tests/primal_numeric_array.cpp
estebanpauli: Good idea. I will add tests that do this.

agcapps (approved):
Thanks @estebanpauli .

PR #362, opened 2020-10-02
Move Inlet access logic to Inlet::Fields

### LINKED ISSUE #361: Inlet: Add get function to Table so you can get nested values without fully qualifying all the time
It would be nice if you didn't have to give a fully qualified path every time.  Currently there is only a get from the Inlet class. Adding Inlet::getTable and Table::get would simplify some deeply nested tables.

For example:

```
inlet->get("nonlinear_solid/solver/nonlinear/rel_tol", nonlin_params.rel_tol);
inlet->get("nonlinear_solid/solver/nonlinear/abs_tol", nonlin_params.abs_tol); 
inlet->get("nonlinear_solid/solver/nonlinear/max_iter", nonlin_params.max_iter);
inlet->get("nonlinear_solid/solver/nonlinear/print_level", nonlin_params.print_level);
```

would change to:

```
auto table = inlet->getTable("nonlinear_solid/solver/nonlinear");
table->get("rel_tol", nonlin_params.rel_tol);
table->get("abs_tol", nonlin_params.abs_tol); 
table->get("max_iter", nonlin_params.max_iter);
table->get("print_level", nonlin_params.print_level);
```

It would also help with passing tables into functions to fill nested structures.

joshessman-llnl:
# Summary

- This PR is a refactoring
- It does the following:
  - Moves the `get` logic from the `Inlet` object to the individual `Fields`s

This allows tables to be retrieved from the Inlet object (and other tables) and have their contained values accessed without needing to reference the "singleton" inlet object.   This will assist in pulling structs out of Inlet, e.g. for some struct `data_struct`, 
conversions could be implemented via a free function `from_inlet(const Inlet::Table& table, data_struct& data)`
and then `from_inlet(inlet.getTable("path/to/data/struct/info", data)`.

This introduces slightly more boilerplate (as an access from an `Inlet` object forwards to a `Table` object which forwards to a `Field` object), but keeps the separates the value get logic from the traversal logic in the Inlet/Table containers.

Resolves #361


white238 (approved):
Thanks @joshessman-llnl !

agcapps (approved):
Thanks: this is a good feature.

PR #364, opened 2020-10-07
Add experimental support for user-defined types in Inlet

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Adds the ability for users to construct instances of their own types from inlet by defining specializations for a `from_inlet` function
  - Adds direct `get` calls to inlet that `throw std::runtime_error` on failure instead of using an out-param
  - Adds a new `operator[]` to `Table` that returns a proxy view to a subobject (Field or Table) with the provided name
  - Adds `contains` to `Table` to check for the existence of a subobject (Field or Table) with the provided name

To see how this might be used, see https://github.com/LLNL/serac/compare/feature/essman/inlet_structs?expand=1.


src/axom/inlet/Field.hpp
```diff
@@ -347,6 +348,18 @@ class Field : public std::enable_shared_from_this<Field>
    */
   bool get(std::string& value);
 
+  template <typename T>
```
kennyweiss, line 351: It seems like it might be confusing to have two functions called `get ` that behave differently.
One is of the form:
```cpp
bool get(PrimitiveType& value)
```
the other is of the form:
```
template<typename T>
T get()
```

Should one of them be renamed?
Would `contains` be better for the form that returns the value as a reference out parameter?

src/axom/inlet/Inlet.hpp
```diff
@@ -194,80 +194,69 @@ class Inlet : public SchemaCreator
...
-  bool get(const std::string& name, double& value)
+ *******************************************************************************
+ * \brief Gets a value of arbitrary type out of the datastore
+ * 
+ * Retrieves a value of user-defined type.
+ * 
+ * \param [in] name The name of the subtable representing the root of the object
+ * \return The retrieved value
+ * \tparam The type to retrieve
+ * \pre Requires a specialization of from_inlet<T>(axom::inlet::Table&)
```
kennyweiss, line 229: Perhaps it would be good to mention which ones are built-in? (or that the native types are built-in?)

src/axom/inlet/Inlet.hpp
```diff
@@ -194,80 +194,69 @@ class Inlet : public SchemaCreator
...
+   * \brief Obtains a proxy view into the datastore.
+   * 
+   * \see Table::operator[]
+   *******************************************************************************
    */
-  bool get(const std::string& name, std::string& value)
-  {
-    return m_globalTable->get(name, value);
-  }
+  Proxy operator[](const std::string& name) { return (*m_globalTable)[name]; }
```
kennyweiss, line 259: Perhaps there is a better user-facing name for the `Proxy` type, e.g. `inlet::FieldView` or `inlet::FieldProxy` (or something better)?

src/axom/inlet/Inlet.hpp
```diff
@@ -465,9 +454,9 @@ class Inlet : public SchemaCreator
    * \return Whether or not the array was found
    *****************************************************************************
    */
-  bool getBoolArray(std::unordered_map<int, bool>& map)
+  bool getArray(std::unordered_map<int, bool>& map)
```
kennyweiss, line 457: Could the value type by templated? E.g.
```cpp
template<typename T>
bool getArray(std::unordered_map<int, T>& map);
```
Presumably, one would also want an array of user-defined structs.

kennyweiss: Perhaps that is for a future PR?

src/axom/inlet/Table.cpp
```diff
@@ -145,25 +144,25 @@ std::shared_ptr<Table> Table::addStringArray(const std::string& name,
...
+bool Table::getArray(std::unordered_map<int, double>& map)
 {
   return axom::utilities::string::endsWith(m_name, "_inlet_array") &&
     m_reader->getDoubleMap(m_name.substr(0, m_name.size() - 12), map);
 }
 
-bool Table::getStringArray(std::unordered_map<int, std::string>& map)
+bool Table::getArray(std::unordered_map<int, std::string>& map)
 {
   return axom::utilities::string::endsWith(m_name, "_inlet_array") &&
```
kennyweiss, line 167: Outside the scope of this PR, but this logic should be centralized into a utility function:
```cpp
  return axom::utilities::string::endsWith(m_name, "_inlet_array") &&
    m_reader->getDoubleMap(m_name.substr(0, m_name.size() - 12), map);
```

src/axom/inlet/Table.cpp
```diff
@@ -295,6 +294,62 @@ std::shared_ptr<Field> Table::addStringHelper(const std::string& name,
...
+
+Proxy Table::operator[](const std::string& name)
+{
+  auto has_table = hasTable(name);
+  auto has_field = hasField(name);
+
+  // Ambiguous case - both a table and field exist with the same name
+  if(has_table && has_field)
+  {
+    std::string msg = fmt::format(
```
kennyweiss, line 315: fwiw, sidre has logic to prevent a child `group` and a `view` within the same group from having the same name. Not sure if that comes into play here.

src/axom/inlet/Table.hpp
```diff
@@ -19,17 +19,184 @@
...
 {
+/*!
+ *******************************************************************************
+ * \class is_lua_primitive
+ *
+ * \brief A type trait for checking if a type is isomorphic to a Lua primitive
+ * \tparam T The type to check
+ *******************************************************************************
+ */
+template <typename T>
```
kennyweiss, line 62: Should these functions be in a `detail` namespace?

src/axom/inlet/Table.hpp
```diff
@@ -19,17 +19,184 @@
...
+ * \class Proxy
+ *
+ * \brief Provides a uniform interface for access and conversion to primitive
+ * and user-defined types
+ *
+ * \see Inlet Field
+ * \see Inlet Table
+ *******************************************************************************
+ */
+class Proxy
```
kennyweiss, line 105: See earlier comment about whether `Proxy` is a good user-facing name for this type.

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -0,0 +1,273 @@
...
+  bool baz;
+};
+
+void from_inlet(axom::inlet::Table& base, Foo& f)
+{
+  f.bar = base["bar"];
+  f.baz = base["baz"];
+}
+
+TEST(inlet_object, simple_struct_by_ref)
```
kennyweiss, line 45: 👍 -- This is nice!

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -0,0 +1,273 @@
...
+  Foo foo;
+  bool found = inlet->get("foo", foo);
+
+  EXPECT_TRUE(found);
+  EXPECT_TRUE(foo.bar);
+  EXPECT_FALSE(foo.baz);
+}
+
+template <>
+Foo from_inlet<Foo>(axom::inlet::Table& base)
```
kennyweiss, line 70: So, it would be up to the user to choose which version to define ?

kennyweiss (approved):
This is really nice @joshessman-llnl !

src/axom/inlet/Field.hpp
kennyweiss: Should have ready your PR description better :)

kennyweiss:
We should discuss the exceptions part and which areas of the API use them.

Is all the exception-based functionality duplicated by call paths that do not have exceptions? In that case, would it make sense to add a config variable / compiler define to enable/disable this API?

src/axom/inlet/Inlet.hpp
joshessman-llnl: I agree that it may not be immediately clear what a `Proxy` is for - though the name is used by the `json` and `sol` libraries for roughly the same thing.  I'd be hesitant to put `Field` in the name since it can also wrap a `Table`. Maybe `SubObjectView`?

src/axom/inlet/Inlet.hpp
joshessman-llnl: I had considered (and attempted) this but the `getArray` functions are part of the `SchemaCreator` interface, and template functions cannot be virtual.  I anticipate that adding arrays of user-defined structs (not currently supported) could require some significant changes, so that will probably be in a separate PR.

src/axom/inlet/tests/inlet_object.cpp
joshessman-llnl: Yes, the template specialization version that returns a T is intended for move-only/non-default-constructible types, though other types could use that version as well.  It is possible to change the by-reference get (soon to be renamed to `get_to`) to deduce at compile time which version was defined for a particular type since it can be implemented in terms of either, so I may look into implementing that.

joshessman-llnl: I have added in some more deduction logic that makes these specializations/overloads significantly more flexible and no longer defers existence checks until link time - they can be checked as part of template instantiation.

src/axom/inlet/Field.hpp
joshessman-llnl: To avoid ambiguity I have renamed the `get(T& val)` to `get_to`

src/axom/inlet/Inlet.hpp
white238: We could remove the SchemaCreator class entirely, I put it there to ease the chaining of nested function calls, but if there is a better way to handle that, I'm perfectly fine with that.  I agree this should be in a separate PR.

src/axom/inlet/Table.hpp
```diff
@@ -19,17 +19,228 @@
...
+/*!
+ *******************************************************************************
+ * \class is_lua_primitive
+ *
+ * \brief A type trait for checking if a type is isomorphic to a Lua primitive
+ * \tparam T The type to check
+ *******************************************************************************
+ */
+template <typename T>
+struct is_lua_primitive
```
white238, line 66: I'm hesistant to have the input language used here.  Looking forward to future where there is going to be multiple input language options, do we see having multiples of these here (is_lua_primitive, is_json_primitive, etc) or should they be at the Reader level (LuaReader::is_primitive)?  Or should this be is_inlet_primitive?

joshessman-llnl: I don't think it's possible to implement type traits as part of an inheritance hierarchy, but for now I think it's reasonable to switch to `is_inlet_primitive` as JSON at least has roughly the same primitives.

src/axom/inlet/Table.hpp
```diff
@@ -19,17 +19,225 @@
...
+ * with the required T operator()(axom::inlet::Table&)
+ * \tparam T The type to check
+ *******************************************************************************
+ */
+template <typename T, typename SFINAE = void>
+struct has_from_inlet_specialization : std::false_type
+{ };
+
+template <typename T>
+struct has_from_inlet_specialization<
```
white238, line 109: Should this be has_FromInlet_specialization?

src/axom/inlet/Field.cpp
```diff
@@ -245,24 +245,20 @@ std::shared_ptr<Field> Field::range(double startVal, double endVal)
...
-  }
   // There is no boolean type in conduit/sidre so we use int8
-  if(valueView->getTypeID() != axom::sidre::INT8_ID)
+  if(valueView == nullptr || valueView->getTypeID() != axom::sidre::INT8_ID)
   {
     std::string msg = fmt::format(
       "[Inlet] Boolean named '{0}' was asked for"
       " but recieved type {1}",
       name(),
       valueView->getTypeID());
```
white238, line 259: this is possibly a nullptr.

src/axom/inlet/Inlet.hpp
white238: We could go with how Sidre handles this by not having this logic at the Inlet class level and making the user call something like `auto table = myInlet->getGlobalTable()` then doing the defining.

src/axom/inlet/Table.cpp
white238: @joshessman-llnl  also pointed out this was double reading from Lua.. this info is already inside of the datastore.  This was missed in the previous PR.

src/axom/inlet/Field.cpp
```diff
@@ -245,12 +245,14 @@ std::shared_ptr<Field> Field::range(double startVal, double endVal)
...
-bool Field::get(bool& value)
+template <>
+bool Field::get<bool>()
 {
   axom::sidre::View* valueView = m_sidreGroup->getView("value");
   if(valueView == nullptr)
   {
-    return false;
+    SLIC_ERROR("[Inlet] Field does not contain a value");
+    setWarningFlag(m_sidreRootGroup);
```
white238, line 255: This warning flag is only checked during `Inlet::verify()` so that all the verifications are run before finally failing.  I don't think it makes sense to mark this.  Just assume the users is going to error out at this point.

src/axom/inlet/Inlet.hpp
```diff
@@ -194,80 +194,45 @@ class Inlet : public SchemaCreator
...
-  bool get(const std::string& name, double& value)
+   *******************************************************************************
+   * \brief Gets a value of arbitrary type out of the datastore
+   * 
+   * Retrieves a value of user-defined type, i.e., not double, int, bool, or string.
+   * 
+   * \param [in] name The name of the subtable representing the root of the object
+   * \return The retrieved value
+   * \tparam The type to retrieve
+   * \pre Requires a specialization of from_inlet<T>(axom::inlet::Table&)
```
white238, line 205: FromInlet?

src/axom/inlet/examples/user_defined_type.cpp
```diff
@@ -0,0 +1,139 @@
...
+      schema.addTable("solver",
+                      "Information about the iterative solver used for Ku = f");
+    LinearSolver::defineSchema(*solver_table);
+  }
+};
+
+template <>
+struct FromInlet<ThermalSolver>
+{
+  // This is also implicitly recursive - will call the from_inlet
```
white238, line 100: FromInlet?

src/axom/inlet/examples/user_defined_type.cpp
```diff
@@ -0,0 +1,139 @@
...
+  // Create a table off the global table for the thermal_solver object
+  // then define its schema
+  auto thermal_solver_table =
+    inlet->addTable("thermal_solver",
+                    "Configuration for a thermal conduction module");
+  ThermalSolver::defineSchema(*thermal_solver_table);
+
+  if(!inlet->verify())
+  {
+    SLIC_INFO("Inlet failed to verify against provided schema");
```
white238, line 134: This should be SLIC_ERROR.

src/axom/inlet/Table.hpp
```diff
@@ -19,17 +19,225 @@
...
+ * \class Proxy
+ *
+ * \brief Provides a uniform interface for access and conversion to primitive
+ * and user-defined types
+ *
+ * \see Inlet Field
+ * \see Inlet Table
+ *******************************************************************************
+ */
+class Proxy
```
white238, line 129: I like to keep one class per file unless there is a reason.  Can you move this into its own file?

src/axom/inlet/Table.hpp
```diff
@@ -476,6 +685,20 @@ class Table : public SchemaCreator, public std::enable_shared_from_this<Table>
...
+   * \brief Return whether a Table or Field with the given name is present in 
+   * this Table's subtree.
+   *
+   * \return Boolean value indicating whether this Table's subtree contains a
+   * Field or Table with the given name.
+   *****************************************************************************
+   */
+  bool contains(const std::string& name)
+  {
+    return hasTable(name) || hasField(name);
```
white238, line 699: Can `hasTable`, `hasField`, and `contains` be collapsed into just `contains`?  I don't think it makes sense to have a field and a table of the same name at the same scope.

src/axom/inlet/Field.cpp
```diff
@@ -272,21 +273,22 @@ bool Field::get(bool& value)
       "[Inlet] Invalid integer value stored in "
       " boolean value named {0}",
       name());
-    SLIC_WARNING(msg);
+    SLIC_ERROR(msg);
```
kennyweiss, line 276: This seems a bit weird to me...  you are setting an error, which will typically abort, but are also setting a warning flag for when the program continues.

Is this the intended behavior?  Is it handling the case where `slic::setAbortOnError(false)` ?

src/axom/inlet/Field.cpp
```diff
@@ -272,21 +273,22 @@ bool Field::get(bool& value)
...
+  return static_cast<bool>(intValue);
 }
 
-bool Field::get(double& value)
+template <>
+double Field::get<double>()
 {
   axom::sidre::View* valueView = m_sidreGroup->getView("value");
+
   if(valueView == nullptr)
```
kennyweiss, line 288: This duplicated error handling logic should be refactored if/when possible.

src/axom/inlet/Inlet.hpp
```diff
@@ -194,80 +194,45 @@ class Inlet : public SchemaCreator
...
+   * \note This function does not indicate failure
```
kennyweiss, line 206: >\note This function does not indicate failure

Is this still true?  Doesn't it call `SLIC_ERROR` on (some) errors?

src/axom/inlet/Table.cpp
```diff
@@ -145,25 +144,25 @@ std::shared_ptr<Table> Table::addStringArray(const std::string& name,
   return table;
 }
 
-bool Table::getBoolArray(std::unordered_map<int, bool>& map)
+bool Table::getArray(std::unordered_map<int, bool>& map)
```
kennyweiss, line 147: 👍

src/axom/inlet/Inlet.hpp
kennyweiss: Please replace all mentions of `from_inlet` if it no longer exists.

src/axom/inlet/Table.hpp
```diff
@@ -19,17 +19,225 @@
...
 {
 namespace inlet
 {
+namespace detail
+{
+/*!
+ *******************************************************************************
+ * \class is_inlet_primitive
+ *
+ * \brief A type trait for checking if a type is isomorphic to a Lua primitive
```
kennyweiss, line 61: I think `Lua` should be removed from here, and "inlet primitive" should be properly defined (here or elsewhere).

src/axom/inlet/Table.hpp
kennyweiss: Not sure... the user might need to have different logic for when something is a `Table` vs. when something is a `Field`.

src/axom/inlet/examples/user_defined_type.cpp
```diff
@@ -0,0 +1,139 @@
...
+};
+
+template <>
+struct FromInlet<ThermalSolver>
+{
+  // This is also implicitly recursive - will call the from_inlet
+  // functions defined for the subobjects
+  ThermalSolver operator()(axom::inlet::Table& base)
+  {
+    return {base["mesh"].get<Mesh>(), base["solver"].get<LinearSolver>()};
```
kennyweiss, line 104: This is really nice!

kennyweiss: Would it be possible to add an array case? Or is that not yet ready?

src/axom/inlet/examples/verification.cpp
```diff
@@ -30,12 +30,13 @@ int main()
...
   // _inlet_workflow_verification_start
   v->registerVerifier([&]() -> bool {
-    int dim;
-    myInlet->get("dimensions", dim);
-    int value;  // field value doesnt matter just that it is present in input file
-    bool x_present = v->hasField("x") && myInlet->get("vector/x", value);
-    bool y_present = v->hasField("y") && myInlet->get("vector/y", value);
-    bool z_present = v->hasField("z") && myInlet->get("vector/z", value);
+    int dim = (*myInlet)["dimensions"];
+    bool x_present = v->hasField("x") &&
```
kennyweiss, line 34: Minor: For a vector with "x", "y" and "z", one might expect the type to be `double` rather than `int`

src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -58,22 +59,17 @@ TEST(inlet_Inlet_basic, getTopLevelBools)
   //
 
   bool value = false;
-  bool found = false;
 
   // Check for existing fields
-  found = inlet->get("foo", value);
-  EXPECT_TRUE(found);
+  value = (*inlet)["foo"];
```
kennyweiss, line 64: Suggestion: It might be good to use both the `operator[]()` and the `get` option in this test, to demonstrate that both are valid.

kennyweiss (approved):
Thanks @joshessman-llnl -- this is looking really nice.

Approving, but please address my concerns/observations below.

src/axom/inlet/Field.cpp
white238: We should probably rename this to be more specific, like `setSchemaWarningFlag` or `setSchemaStepThrewWarningFlag`.

src/axom/inlet/examples/user_defined_type.cpp
joshessman-llnl: This is in-progress on https://github.com/LLNL/axom/tree/feature/essman/inlet_record_arrays but I wanted to keep that functionality as a separate PR since this one is already rather large, in my opinion.

src/axom/inlet/Field.cpp
joshessman-llnl: Per Chris's comment about the intended usage of the warning flag, I've removed it from the `get` functions.  That being said, the original intent was for when the program would continue (since these were originally exceptions), but in most cases where I've replaced exceptions with `SLIC_ERROR`, the program will immediately segfault if SLIC does not terminate on error.

src/axom/inlet/Inlet.hpp
joshessman-llnl: I've added a clarification to this note - failure is not indicated in a way that can be handled by a program as neither a boolean return value or exceptions are used.

src/axom/inlet/examples/verification.cpp
joshessman-llnl: I agree - I think my only changes to this example were to switch to the new accessors but it won't be a problem to change these to doubles.

white238 (approved):

PR #369, opened 2020-10-14
Feature/essman/inlet record arrays

joshessman-llnl:
# Summary

- This PR is a refactoring of the inlet module
- It does the following:
  - Adds support for arrays of user-defined type

A major priority when implementing this was that a schema for a struct could be defined in a way that allows the same `defineSchema` function to work for a single instance of that struct or an array of that struct.  See the `user_defined_type` example for an example usage.

This required a fair number of changes to implement in a reasonable way:
- Adding new `Verifiable` and `VerifiableScalar` interfaces/concepts for the return types of `add*` functions so constraints can be applied to groups of struct fields (i.e., all the instances of that field in an array of structs)
- Removed the `SchemaCreator` interface - schemas must be defined through tables
- New `Table::addGenericArray` function for creating an array of user-defined type - this returns a table to which fields corresponding to struct fields can be added
- Changed the lambda verifier interface to use predicates with the signature `bool pred(Proxy&)` instead of `bool ()` and forcing the user to capture everything
- Instead of having `Inlet::verify` traverse the entire sidre group, each table/field is now completely responsible for (recursively) verifying itself, meaning that `Inlet::verify` now just verifies the global table

Also, it removes some existing duplication, e.g., duplicated logic for adding a scalar or array type, or duplicated reading of the input file for arrays.

Finally, this PR modifies the LuaReader class to support integer-valued keys (specifically, array indexing).

This PR did end up being fairly large but a unfortunately a large number of changes were necessary to cleanly implement arrays of structs.


src/axom/inlet/Table.hpp
```diff
@@ -238,8 +250,8 @@ class Table : public SchemaCreator, public std::enable_shared_from_this<Table>
    * \return Shared pointer to the created Field
    *****************************************************************************
    */
-  std::shared_ptr<Table> addBoolArray(const std::string& name,
-                                      const std::string& description = "");
+  std::shared_ptr<Verifiable> addBoolArray(const std::string& name,
```
kennyweiss, line 253: Ok, so you can still chain calls like `->required(true)`, but you no longer get the field.

I suppose you can no longer return a `Table` or `Field` since an array is `Verifiable` while a `Field` is a `VerifiableScalar` ?

src/axom/inlet/Table.hpp
```diff
@@ -730,10 +875,116 @@ class Table : public SchemaCreator, public std::enable_shared_from_this<Table>
   bool m_docEnabled;
   std::unordered_map<std::string, std::shared_ptr<Table>> m_tableChildren;
   std::unordered_map<std::string, std::shared_ptr<Field>> m_fieldChildren;
-  std::function<bool()> m_verifier;
+  std::function<bool(Proxy&)> m_verifier;
+};
+
+// To-be-defined template specializations
```
kennyweiss, line 881: Cleanup: Possibly for a separate PR, but I wonder if some of these helper routines should be moved to an implementation/detail file.

src/axom/inlet/Table.hpp
```diff
@@ -730,10 +875,116 @@ class Table : public SchemaCreator, public std::enable_shared_from_this<Table>
...
+  */
+  std::shared_ptr<Verifiable> registerVerifier(std::function<bool(Proxy&)> lambda);
+
+private:
+  std::vector<std::shared_ptr<Verifiable>> m_tables;
 };
 
 }  // end namespace inlet
-}  // end namespace axom
+}  // namespace axom
```
kennyweiss, line 988: why did the `end` get removed ?

I think clang-format should add it back... unless there is an issue with the namespaces?

src/axom/inlet/Table.cpp
```diff
@@ -65,110 +65,88 @@ std::shared_ptr<Table> Table::addTable(const std::string& name,
   return currTable;
 }
```
kennyweiss, line 67: Side note: I'm starting to think that Axom should have a proper `Path` class that both `inlet` and `sidre` could use.
It can have utility functions like returning the parent, stem, etc...

Perhaps it can live in `core`?

src/axom/inlet/Verifiable.hpp
```diff
@@ -0,0 +1,313 @@
...
+
+/*!
+ *******************************************************************************
+ * \class VerifiableScalar
+ *
+ * \brief Basic interface for verifiable scalar values of Inlet primitive type,
+ * namely int, double, bool, or std::string
+ *******************************************************************************
+ */
+class VerifiableScalar
```
kennyweiss, line 93: Should `VerifiableScalar` be moved to its own file?

src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -41,10 +42,8 @@ TEST(inlet_Inlet_basic, getTopLevelBools)
   // Define schema
   //
 
-  std::shared_ptr<axom::inlet::Field> currField;
-
   // Check for existing fields
-  currField = inlet->addBool("foo", "foo's description");
+  auto currField = inlet->addBool("foo", "foo's description");
```
kennyweiss, line 46: Minor: This name is a bit misleading now, since `currField` is no longer an `inlet::Field`, it is now an `inlet::VerifiableScalar`, right?

src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -1272,14 +1243,14 @@ TEST(inlet_verify, verifyFieldLambda)
   auto field2 = inlet->addString("field2");
   auto field3 = inlet->addString("NewTable/field3");
 
-  field2->registerVerifier([&]() -> bool {
-    auto str = (*inlet)["field2"].get<std::string>();
+  field2->registerVerifier([](Proxy& field) -> bool {
+    std::string str = field;
```
kennyweiss, line 1247:
```cpp
  field2->registerVerifier([](Proxy& field) -> bool {
    std::string str = field; 
    ...
  });
```
I get why it is implemented this way (i.e. passing in a Proxy to the lambda and having to then extract the actual type), but I'm wondering if this will be confusing for users...

I anticipate having to look up this syntax every time I need to use it.

kennyweiss: Also, should it be ``const Proxy& field`` since we don't want to change the value.
This is likely enforced by the Proxy type itself, but adding `const` might help the user understand the intent.

kennyweiss (approved):
This looks really nice @joshessman-llnl -- I appreciate the effort you put into keeping a consistent interface for scalars and arrays.

src/axom/inlet/tests/inlet_Inlet.cpp
joshessman-llnl: I think for now it is sufficient to have the `Verifiable` signature use a `Table&` and `VerifiableScalar` use a `Field&` - initially I wanted the latter to inherit from the former but that would impose too much on the user.

I completely agree with the const correctness, but because no member functions in inlet are marked const, this would require a fair number of changes.  Const correctness is a primary goal for my next PR.

src/axom/inlet/Table.hpp
joshessman-llnl: It has to do with the `Aggregate` types - the call to `addBoolArray` could either add a single array (returns Table) or, if the calling table is an array of structs, an array for each struct in the array of structs (returns AggregateTable).   I think adding fields to an array of booleans is invalid anyways, but in the general case it does restrict the user a bit.

src/axom/inlet/Verifiable.hpp
```diff
@@ -0,0 +1,313 @@
...
+   *****************************************************************************
+   * \brief Return the required status.
+   *
+   * Return that this object is required, or not, to be in the input file.
+   * The default behavior is to not be required.
+   *
+   * \return Boolean value of whether this object is required
+   *****************************************************************************
+   */
+  virtual bool required() = 0;
```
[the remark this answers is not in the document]
joshessman-llnl: I don't have a problem renaming this (seems like a questionable application of overloading), I was just trying to match the existing interface that `Table` and `Field` provide.

src/axom/inlet/LuaReader.hpp
```diff
@@ -194,6 +194,18 @@ class LuaReader : public Reader
...
+   *****************************************************************************
+   * \brief Get the list of indices for an array
+   *
+   * \param [in]  id    The identifier to the array that will be retrieved
+   * \param [out] map The values of the indices that were retrieved
+   *
+   * \return true if the array was able to be retrieved from the file
+   *****************************************************************************
+   */
+  bool getArrayIndices(const std::string& id, std::vector<int>& indices);
```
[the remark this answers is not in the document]
joshessman-llnl: Good catch (I'm used to functions that return their results) - it should be sufficient to just clear out the vector.

src/axom/inlet/LuaReader.cpp
```diff
@@ -154,7 +199,15 @@ bool LuaReader::getMap(const std::string& id,
   sol::table t = m_lua[tokens[0]];
   for(size_t i = 1; i < tokens.size(); i++)
   {
-    if(t[tokens[i]].valid())
+    // Use the C versions to avoid the exceptions
```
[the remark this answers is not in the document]
joshessman-llnl: The indices differ on one of them, but I was thinking it would be possible to do an iterator-based solution

src/axom/inlet/Table.cpp
white238: Why would it not live in sidre since its used for sidre paths?

kennyweiss: I suppose it could live in Sidre...
I was thinking of a path as just a delimited string -- without implying any sidre class structure.

Also, presumably paths in inlet can be different from those in sidre.

white238: That's true. core it is.

src/axom/inlet/Field.cpp
```diff
@@ -13,7 +14,7 @@ namespace axom
 {
 namespace inlet
 {
-std::shared_ptr<Field> Field::required(bool isRequired)
+std::shared_ptr<VerifiableScalar> Field::required(bool isRequired)
```
white238, line 17: Now that we have changed the `required()` to `isRequired()`, we can make this default to true

src/axom/inlet/LuaReader.cpp
```diff
@@ -102,35 +102,62 @@ bool LuaReader::getStringMap(const std::string& id,
   return getMap(id, values, sol::type::string);
 }
 
-bool LuaReader::getArrayIndices(const std::string& id, std::vector<int>& indices)
+template <typename Iter>
+bool LuaReader::traverseToTable(Iter begin, Iter end, sol::table& table)
```
kennyweiss, line 106: 👍

kennyweiss (approved):
Thanks for the updates @joshessman-llnl -- looks great!

src/axom/inlet/Inlet.hpp
```diff
@@ -418,12 +419,28 @@ class Inlet : public SchemaCreator
...
+   *****************************************************************************
+   * \brief Add an array of Fields to the input deck schema.
+   *
+   * \param [in] name Name of the array
+   * \param [in] description Description of the Field
+   *
+   * \return Shared pointer to the created Field
+   *****************************************************************************
+   */
+  std::shared_ptr<Table> addGenericArray(const std::string& name,
```
white238, line 438: Why is this `Table` not `Verifiable` like the rest of the `add*Array` functions?

src/axom/inlet/Table.hpp
```diff
@@ -277,52 +291,82 @@ class Table : public SchemaCreator, public std::enable_shared_from_this<Table>
...
    *****************************************************************************
    */
-  bool getArray(std::unordered_map<int, int>& map);
+  std::shared_ptr<Table> addGenericArray(const std::string& name,
+                                         const std::string& description = "");
 
   /*!
    *****************************************************************************
-   * \brief Get a double array represented as an unordered map from the input deck
+   * \brief Get an array represented as an unordered map from the input deck
```
white238, line 313: This keeps cropping up from random sources but can you change `input deck` to `input file`

src/axom/inlet/Table.hpp
```diff
@@ -277,52 +291,82 @@ class Table : public SchemaCreator, public std::enable_shared_from_this<Table>
...
+      int* array = view->getArray();
+      for(int i = 0; i < view->getNumElements(); i++)
+      {
+        auto index_label = std::to_string(array[i]);
+        map[array[i]] = getTable(index_label)->get<T>();
+      }
+    }
+    else
+    {
+      SLIC_WARNING("Table does not contain an array");
```
white238, line 365: prefix this message with "[Inlet]" since some of these messages are pretty generic sounding without it.

src/axom/inlet/examples/user_defined_type.cpp
```diff
@@ -6,12 +6,13 @@
 #include "axom/inlet.hpp"
 
 #include <iostream>
```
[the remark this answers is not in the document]
white238: Seconded, either some further validation, or printing values to the screen.

src/axom/inlet/Table.cpp
```diff
@@ -65,110 +65,88 @@ std::shared_ptr<Table> Table::addTable(const std::string& name,
...
+    auto view =
+      table->m_sidreGroup->createViewAndAllocate("_inlet_array_indices",
+                                                 axom::sidre::INT_ID,
+                                                 indices.size());
+    int* raw_array = view->getArray();
+    std::copy(indices.begin(), indices.end(), raw_array);
+    for(const auto idx : indices)
     {
-      table->addStringHelper(std::to_string(p.first), "", true, p.second);
+      table->addTable(std::to_string(idx), description);
```
white238, line 140: Is this the right description?

src/axom/inlet/examples/user_defined_type.cpp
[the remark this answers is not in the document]
white238: Check-in generated documentation?

src/axom/inlet/examples/user_defined_type.cpp
```diff
@@ -76,21 +78,48 @@ struct FromInlet<LinearSolver>
...
+  static void defineSchema(inlet::Table& schema)
   {
     auto mesh_table = schema.addTable("mesh", "Information about the mesh");
     Mesh::defineSchema(*mesh_table);
     auto solver_table =
       schema.addTable("solver",
                       "Information about the iterative solver used for Ku = f");
     LinearSolver::defineSchema(*solver_table);
+    auto bc_table = schema.addGenericArray("bcs", "List of boundary conditions");
+    BoundaryCondition::defineSchema(*bc_table);
```
white238, line 122: I am concerned a user would not understand how this works... more comments are needed like "addGenericArray loops over found indices and creates array of boundary conditions"

Maybe there should a snippet of an example lua in each defineSchema?

white238 (approved):

white238:
Thanks again @joshessman-llnl !

src/axom/inlet/Table.cpp
joshessman-llnl: That description is just the user description for the array of structs being used as the description for each element - it looks like it was previously a blank description, so I'm not sure which would be more helpful.

src/axom/inlet/Inlet.hpp
joshessman-llnl: Arrays of structs within arrays of structs are not yet supported.  I was hoping to do that in a future PR as it would require a new interface that (probably) extends `Verifiable` but also supports adding new fields.

src/axom/inlet/Table.cpp
white238: Is there a way to get a description for each element?

PR #373, opened 2020-10-20
Const correctness in Inlet

joshessman-llnl:
# Summary

- This PR is a refactoring
- It does the following:
  - Adds const to member functions and local variables
  - Adds `true` as the default argument to `Verifiable::required` and `VerifiableScalar::required` (and the implementations of those interfaces)

In the interest of keeping PRs small, other fixes (e.g. removing `shared_ptr` when possible) will be part of future PRs.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -17,6 +17,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Added a `batched` option to quest's InOutOctree containment query example application.
   This uses a kernel to test for containment on an array of points.
   The query uses OpenMP threading, when available.
+- Inlet: Added support for user-defined conversions from Inlet tables to user-defined
+  types, and support for arrays of user-defined types
 
 ### Changed
 - The Sidre Datastore no longer rewires Conduit's error handlers to SLIC by default. 
@@ -25,6 +27,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Inlet: Fixed `SchemaCreator` to an abstract class and added missing functions
 - Inlet: Added ability to access the `Reader` class from `Inlet` and Sol Lua state
   from the `LuaReader` class
+- Inlet: Switched accessor interface to match that of the STL with operator[] and
+  T get<T>()
 
 
 ## [Version 0.4.0] - Release date 2020-09-22
--- a/src/axom/inlet/Table.cpp
+++ b/src/axom/inlet/Table.cpp
@@ -66,7 +66,7 @@ std::shared_ptr<Table> Table::addTable(const std::string& name,
 }
 
 std::vector<std::pair<std::string, std::string>> Table::arrayIndicesWithPaths(
-  const std::string& name)
+  const std::string& name) const
 {
   std::vector<std::pair<std::string, std::string>> result;
   if(!m_sidreGroup->hasView("_inlet_array_indices"))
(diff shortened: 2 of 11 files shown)


src/axom/inlet/Table.cpp
```diff
@@ -44,7 +44,7 @@ std::shared_ptr<Table> Table::addTable(const std::string& name,
     found = currName.find("/");
   }
 
-  std::string currTableName =
+  const auto currTableName =
```
white238, line 47: I kinda like leaving this `std::string` instead of `auto`, helps with readability.

src/axom/inlet/Table.cpp
```diff
@@ -120,6 +120,13 @@ std::shared_ptr<Verifiable> Table::addStringArray(const std::string& name,
 std::shared_ptr<Table> Table::addGenericArray(const std::string& name,
                                               const std::string& description)
 {
+  if(m_sidreGroup->hasView("_inlet_array_indices"))
```
white238, line 123: These two string literals are used in quite a few spots at this point. Can you pull them into a common location?

RELEASE-NOTES.md
```diff
@@ -17,6 +17,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Added a `batched` option to quest's InOutOctree containment query example application.
   This uses a kernel to test for containment on an array of points.
   The query uses OpenMP threading, when available.
+- Inlet: Added support for user-defined conversions from Inlet tables to user-defined
```
white238, line 20: :+1:

white238 (approved):

src/axom/inlet/Table.cpp
```diff
@@ -13,6 +13,9 @@ namespace axom
 {
 namespace inlet
 {
+const std::string Table::INTERNAL_ARRAY_GROUP = "_inlet_array";
```
white238, line 16: Maybe `INTERNAL_ARRAY_GROUP_NAME` or `INTERNAL_ARRAY_GROUP_ID`? maybe drop INTERNAL not sure its needed.

INDEX to INDICES though...

src/axom/inlet/Table.cpp
```diff
@@ -442,15 +452,15 @@ void Table::addPrimitiveArrayHelper<std::string>(Table& table,
   }
 }
 
-Proxy Table::operator[](const std::string& name)
+Proxy Table::operator[](const std::string& name) const
 {
-  auto has_table = hasTable(name);
-  auto has_field = hasField(name);
+  const auto has_table = hasTable(name);
```
white238, line 457: bool

RELEASE-NOTES.md
```diff
@@ -25,6 +27,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Inlet: Fixed `SchemaCreator` to an abstract class and added missing functions
 - Inlet: Added ability to access the `Reader` class from `Inlet` and Sol Lua state
   from the `LuaReader` class
+- Inlet: Switched accessor interface to match that of the STL with operator[] and
+  T get<T>()
```
kennyweiss, line 31: Could you please show what that looked like before and now? (In this PR, not in the RELEASE-NOTES).

kennyweiss (approved):
Thanks @joshessman-llnl !

RELEASE-NOTES.md
joshessman-llnl: Sure, so

```cpp
int value1;
bool success = inlet->get("path/to/value1", value1);
if (success)
{
  // do stuff with value1
}
int value2;
inlet->get("path/to/value2", value2); // has default, no need to check success
```

would become something like

```cpp
auto table = inlet["path/to"];
if (table.contains("value1")
{
    int value1 = table["value1"]; // or int value1 = table["value"].get<int>();
    // do stuff with value1
}

int value2 = table["value2"]; // or inlet["path/to/value2"];
```

kennyweiss: Thanks @joshessman-llnl

PR #377, opened 2020-10-29
Update blt to get fix that prevents a failure when clangformat is autodetected

white238:
Subtle niche error that showed up in a Spack build on WSL


agcapps (approved):
thanks Chris

kennyweiss (approved):

kennyweiss (approved):
Please add a note the RELEASE-NOTES so users know they've got a version of axom that resolves this problem

PR #378, opened 2020-10-30
Quiet warnings about non-virtual destructor

white238:
Small PR.


joshessman-llnl (approved):
Good catch - was this with a higher warning level than the default?

white238:
> Good catch - was this with a higher warning level than the default?

Nope, but they were hidden behind a lot of linker warnings we can't do anything about until I think TOSS4 hits.

kennyweiss (approved):

agcapps (approved):
Thanks, Chris

PR #379, opened 2020-11-02
Add support for mesh/field reconstruction in MFEMSidreDataCollection

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Adds mesh and field reconstruction when a datastore is `Load`ed from

This PR adds some additional information to the output files when running in parallel to aid in the reconstruction of parallel meshes as determining shared edge and face data is nontrivial given just the data prescribed by the Conduit Blueprint.

An example usage of this restart functionality was added as well.  The idea is that in a real integration the conditional logic for a restart vs. nominal run would be restricted to the `MaybeOwner` objects and the state object that contains them.

Parallel mesh reconstruction has also been implemented, but for simplicity it has been moved to a separate branch for a later PR.

## Remaining usage questions

Currently all reconstruction errors are fatal, but I wonder if it would be more useful to just warn on failure and leave the data unreconstructed.

I'm also not sure how this new functionality should behave if a mesh/field already exists (set prior to the `Load`) - right now it will just overwrite, but I'm not familiar enough with potential use cases to know whether it should warn and stop instead.


joshessman-llnl:
To make reviewing this easier I have split the PR in half - the follow-up PR will add support for parallel mesh reconstruction.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -545,13 +571,75 @@ void MFEMSidreDataCollection::createMeshBlueprintAdjacencies(bool hasBP)
...
+          gvertices_data[vi] = pmesh->GroupVertex(gi, vi);
+        }
+      }
 
-      for(int vi = 0; vi < num_gvertices; ++vi)
+      // For all these higher-order elements, we store a flat list of tuples of
+      // vertex indices instead of MFEM's internal edge/face indices for generality
+      // This uses more space but technically the Blueprint does not standardize how
+      // edge/face indices correspond to vertex indices
+      if(dim >= 2)
```
kennyweiss, line 590: @joshessman-llnl -- can you please explain why the original version only needed to copy the shared vertices, but the new version needs to copy shared edges and faces?

Also, would these lists ever need to be updated? E.g. if the mesh changes.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1287,6 +1386,189 @@ std::string MFEMSidreDataCollection::getElementName(mfem::Element::Type elementE
...
+{
+  // mfem::ConduitDataCollection::BlueprintMeshToMesh would be useful here, but
+  // we need all the parameters to construct a ParMesh (the Mesh base subobject
+  // is initialized manually)
+
+  SLIC_ERROR_IF(
+    !verifyMeshBlueprint(),
+    "Cannot reconstruct mesh, data does not satisfy Conduit Blueprint");
+
+  SLIC_ERROR_IF(!m_bp_grp->hasView("coordsets/coords/values/x"),
```
kennyweiss, line 1419: I think this function will need some additional work (perhaps this is what you meant by "experimental").

When reading in a mesh, the coordinate set does not need to be named `coords`.
Similarly, the topology does not need to be named 'mesh'
And it will not always have a "boundary" topology (even if some ranks have a "boudary topology -- we've been burned by that one in the past).

kennyweiss: Perhaps the claim is that this function is only guaranteed/expected to `Load` a blueprint that was `Save`d by this class, `MFEMSidreDataCollection`, in which case some of the coord/topology names can be reasonably hard-coded?

kennyweiss (commented):
This is really nice @joshessman-llnl 

However, I have concerns about the use of `std::unique_ptr`s inside `MFEMSidreDataCollection` and its possible implications for user codes.

I'm not sure if this will require changes to this PR, but at the very least we need to verify that it will not be a problem.

src/axom/sidre/tests/sidre_mfem_datacollection.cpp
```diff
@@ -110,24 +111,86 @@ TEST(sidre_datacollection, dc_reload)
...
+  sdc_writer.Save();
+
+  // No mesh is used here
+  MFEMSidreDataCollection sdc_reader(COLL_NAME);
+
+  #if defined(AXOM_USE_MPI) && defined(MFEM_USE_MPI)
+  sdc_reader.SetComm(MPI_COMM_WORLD);
+  #endif
+
+  sdc_reader.SetPrefixPath("/tmp/dc_reload_test");
```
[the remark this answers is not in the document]
joshessman-llnl: I think we had looked at this in an [earlier PR](https://github.com/LLNL/axom/pull/318#discussion_r493768971), but I do agree that the use of `/tmp` looks strange on Windows.  I think using the default path (the current working directory) would be acceptable here.

src/axom/sidre/examples/sidre_mfem_datacollection_restart.cpp
```diff
@@ -0,0 +1,280 @@
...
+#endif
+
+// MFEM includes - needed to set up simulation
+#include "mfem.hpp"
+
+// Abstraction for an object that may or may not retain ownership of an object
+// via an owning- or non-owning-pointer - this allows for owning and non-owning
+// versions of an object to be accessed uniformly.
+template <typename T>
+class MaybeOwner
```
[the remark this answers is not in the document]
joshessman-llnl: I had initially used this example as a "sandbox" for integrating this into serac, but using something like raw pointers and a set of `bool`s to keep track of ownership is definitely simpler for this minimal example.

src/axom/sidre/docs/sphinx/mfem_sidre_datacollection.rst
```diff
@@ -64,8 +64,7 @@ A typical visualization example might look like the following:
       dc.Save("data_name", "sidre_hdf5");
    }
 
-See the ``sidre_mfem_datacollection_ex`` example for a more thorough example of the above functionality.
-
+See the ``sidre_mfem_datacollection_vis`` example for a more thorough example of the above functionality.
 
 .. Note::
    The ``owns_mesh_data`` option must be set to true when constructing an instance of the class for the
```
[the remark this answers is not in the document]
joshessman-llnl: Yes, this was a "discovery" I had made when starting to implement this functionality - fields also need to be explicitly constructed with a `nullptr` so the sidreDC will do the allocation within the datastore.  I will add this to #355

src/axom/sidre/core/MFEMSidreDataCollection.hpp
```diff
@@ -165,13 +165,9 @@ namespace sidre
...
-    @note MFEMSidreDataCollection does not manage the FiniteElementSpace%s and
-    FiniteElementCollection%s associated with registered GridFunction%s.
-    Therefore, field registration is left to the user of MFEMSidreDataCollection
-       and
-    there are no methods that automatically register GridFunction%s using just
-    the content of the Sidre DataStore. Such capabilities can be implemented in
-    a derived class, adding any desired object management routines.
+    @note MFEMSidreDataCollection will attempt to reconstruct meshes and fields
+    on a restart, and will automatically register them.  This functionality is
+    experimental.
```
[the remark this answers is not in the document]
joshessman-llnl: The main thing that this PR is missing is full support for reloading parallel meshes/fields (which is mostly implemented in a separate branch)

src/axom/sidre/core/MFEMSidreDataCollection.hpp
```diff
@@ -457,6 +456,16 @@ class MFEMSidreDataCollection : public mfem::DataCollection
   // This is stored for convenience.
   Group* m_named_bufs_grp;
 
+  // Used to retain ownership of components of reconstructed Meshes and GridFuncs
+  // Instead of using flags to keep track of ownership between this class
+  // and the mfem::DataCollection subobject, always have the subobject
+  // retain a non-owning pointer and manage memory via unique_ptr in
+  // this class for consistency
+  std::unique_ptr<mfem::Mesh> m_owned_mesh;
```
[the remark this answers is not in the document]
joshessman-llnl: I'd like to hear more about some of these use cases - though from my understanding the existing interface will be preserved in that `mfem::DataCollection::mesh` will remain a non-owning `Mesh*` that references the `unique_ptr` when the mesh is owned by the datastore, and an external mesh otherwise.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
joshessman-llnl: The shared edge/face information is needed to reconstruct the tables used internally by `mfem::ParMesh` (that reconstruction logic will be in a follow-up PR).

I may be incorrect but I do not think this class is tolerant to any kind of refinement after it is constructed - presumably this would require logic to examine the mesh before each `Save()`.  I will document this in #355

src/axom/sidre/core/MFEMSidreDataCollection.cpp
joshessman-llnl: Yes, reconstruction is currently only supported for meshes/fields written by this class.  This requirement will be stricter for parallel meshes as the shared edge/face data is currently "off-label use" of the Conduit blueprint.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -568,9 +656,8 @@ bool MFEMSidreDataCollection::verifyMeshBlueprint()
   bool result = conduit::blueprint::mesh::verify(mesh_node, verify_info);
   // conduit::Node::to_string only available in latest version
   SLIC_WARNING_IF(!result,
-                  "MFEMSidreDataCollection blueprint verification failed: " /*<<
-                                                                               verify_info.to_string()
-                                                                               */                          );
+                  "MFEMSidreDataCollection blueprint verification failed: "
+                  /*<< verify_info.to_string()*/);
```
white238, line 660: Can this be uncommented in the newest conduit that was updated recently?

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1287,6 +1385,189 @@ std::string MFEMSidreDataCollection::getElementName(mfem::Element::Type elementE
   return "unknown";
 }
 
+mfem::Geometry::Type MFEMSidreDataCollection::getElementTypeFromName(
+  const std::string& name)
+{
+  if(name == "point")
```
white238, line 1391: Brackets are needed here due to style guide

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1287,6 +1385,189 @@ std::string MFEMSidreDataCollection::getElementName(mfem::Element::Type elementE
...
+
+  SLIC_ERROR_IF(!m_bp_grp->hasView("coordsets/coords/values/x"),
+                "Cannot reconstruct a mesh without a Cartesian coordinate set");
+
+  View* vertex_view = m_bp_grp->getView("coordsets/coords/values/x");
+  SLIC_ERROR_IF(vertex_view->isExternal(),
+                "Cannot reconstruct a mesh that was built using external data");
+
+  // Assumes that Vertex is a standard layout type with only an array member
+  static_assert(std::is_standard_layout<mfem::Vertex>::value,
```
white238, line 1426: Should we be using SLIC_ASSERT for this?

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -509,13 +510,38 @@ void MFEMSidreDataCollection::createMeshBlueprintAdjacencies(bool hasBP)
     }
   }
 
+  int dim = mesh->SpaceDimension();
```
jamiebramwell, line 513: This should be `pmesh` for consistency.

white238 (approved):

src/axom/sidre/examples/sidre_mfem_datacollection_restart.cpp
```diff
@@ -0,0 +1,214 @@
+// Copyright (c) 2017-2020, Lawrence Livermore National Security, LLC and
```
white238, line 1: Update copyrights please.

src/axom/sidre/examples/sidre_mfem_datacollection_restart.cpp
```diff
@@ -0,0 +1,214 @@
...
+    fecoll = fespace->FEColl();
+  }
+
+  // FEM-related objects needed as part of a simulation
+  // In a real simulation these would be exposed via accessors
+  mfem::Mesh* mesh;
+  const mfem::FiniteElementCollection* fecoll;
+  mfem::FiniteElementSpace* fespace;
+  mfem::GridFunction* soln_field;
+  bool owns_data = false;
```
white238, line 159: m_

src/axom/sidre/core/MFEMSidreDataCollection.cpp
joshessman-llnl: This is a compile-time assertion, my understanding is that SLIC_ASSERTs are for runtime assertions

src/axom/sidre/core/MFEMSidreDataCollection.cpp
joshessman-llnl: As a follow-up, I have added a snippet to the documentation that specifies the required topology/coordset names, and also converted these to variables instead of having the paths hardcoded.  There are still some places that use the hardcoded names, so a follow-up PR will address this.  Having a `Path` class (see #438) will make this easier and more readable.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
white238: yepp your right carry on

kennyweiss (approved):
Thanks @joshessman-llnl !

PR #380, opened 2020-11-03
Split HEADERS from SOURCES on blt_add_executable

white238:
This updates BLT to include the fix i just put in to split HEADERS from SOURCES on blt_add_executable.  This is important for file level dependency checking during the build system run (ie makefiles) as well as IDE folder support.

Surprisingly enough CMake is not smart enough not to compile a header when you flag it for separable compilation, it has logic everywhere else not to compile a header by itself based on file extension.  My guess this is in the old CUDA support and will go away when we switch to the new CMake CUDA support.

### DIFF
--- a/src/axom/core/tests/CMakeLists.txt
+++ b/src/axom/core/tests/CMakeLists.txt
@@ -55,7 +55,8 @@ if (NOT ENABLE_MPI)
 endif()
 
 blt_add_executable( NAME       core_serial_tests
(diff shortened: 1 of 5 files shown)


kennyweiss (approved):

bmhan12 (approved):

rhornung67 (approved):
Looks good. Thanks! 👍

agcapps (approved):
Thanks @white238

PR #381, opened 2020-11-03
Add CUDA includes/libraries when linking against mfem+cuda

joshessman-llnl:
# Summary

- This PR is a bugfix
- It does the following:
  - Adds MFEM's required CUDA libraries/headers as they are not added to `config.mk` as part of the build process

### DIFF
--- a/src/cmake/thirdparty/FindMFEM.cmake
+++ b/src/cmake/thirdparty/FindMFEM.cmake
@@ -110,6 +110,15 @@ else()
 
     list(APPEND MFEM_LIBRARIES ${mfem_tpl_lnk_flags})
 
+    # Check if MFEM was built with CUDA
+    if(mfem_cfg_file_txt MATCHES "MFEM_USE_CUDA += YES")
+        message(STATUS "Adding MFEM CUDA includes/libraries")
+        list(APPEND MFEM_INCLUDE_DIRS ${CUDA_INCLUDE_DIRS})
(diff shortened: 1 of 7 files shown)


src/cmake/thirdparty/FindMFEM.cmake
```diff
@@ -110,6 +110,17 @@ else()
 
     list(APPEND MFEM_LIBRARIES ${mfem_tpl_lnk_flags})
 
+    # Check if MFEM was built with CUDA
+    if(mfem_cfg_file_txt MATCHES "MFEM_USE_CUDA += YES")
+        if(NOT ENABLE_CUDA)
+            message(WARNING "MFEM was built with CUDA but CUDA is not enabled")
+        endif()
+        list(APPEND MFEM_INCLUDE_DIRS ${CUDA_INCLUDE_DIRS})
```
white238, line 118: the cuda registered library is added in a lot of places.  Is it possible it was not in a few places due to the new mfem+cuda functionality?  I would prefer that instead of doubling up this information.

white238: After discussing with @joshessman-llnl , I think this is fine for now but can be improved upon later down the road.

white238 (approved):

kennyweiss (approved):
Thanks @joshessman-llnl 

I'm assuming that you've tested this outside of axom.

Would it make sense to incorporate `mfem+cuda` into at least one of our llnl gpu configurations so that we can also test this within axom? E.g. to avoid regressions.

PR #382, opened 2020-11-03
Bugfix/white238/spackpackage

white238:
The main reason for this was to turn off clang-format explicitly when devtools was not enabled.  This was causing a problem in the version checking regex that will be fixed in BLT soon.

This also unifies a bunch of changes across projects with their own axom/package.py.  This will get upstreamed to spack's repo as well.


kennyweiss (approved):

agcapps (approved):
Many thanks, @white238

rhornung67 (approved):

white238:
@nselliott I think this should fix the SPIO test on all the platforms. It at least fixes rzansel.

PR #384, opened 2020-11-05
Fix Fortran array indexing error

nselliott:
# Summary

- This PR is a bugfix
Fixes array indexing in one test to use Fortran 1:n indexing.


white238 (approved):

agcapps (approved):
Thanks, Noah.

ltaylor16 (commented):
Most Fortran compilers have a bound check flag that would catch this kind of stuff. Maybe BLT could support setting this flag.

PR #391, opened 2020-11-16
Add "dictionary" types to Inlet

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Adds the ability to defined string-keyed (in addition to integer-keyed) associative arrays in an input file

This required more refactoring/additional code than I expected as it doesn't look like Sidre/conduit supports arrays of strings within a single `sidre::View`.  Key storage is still implemented as a contiguous integer array for arrays, but dicts use a group of string-valued `sidre::View`s.

### DIFF
--- a/src/axom/inlet/Inlet.hpp
+++ b/src/axom/inlet/Inlet.hpp
@@ -430,7 +430,7 @@ public:
 
   /*!
    *****************************************************************************
-   * \brief Add an array of Fields to the input file schema.
+   * \brief Add an array of user-defined types to the input file schema.
    *
    * \param [in] name Name of the array
    * \param [in] description Description of the Field
@@ -446,58 +446,82 @@ public:
 
   /*!
    *****************************************************************************
-   * \brief Get a boolean array represented as an unordered map from the input file
+   * \brief Add a dictionary of Boolean Fields to the input file schema.
    *
-   * \param [out] map Unordered map to be populated with array contents
+   * \param [in] name Name of the dict
+   * \param [in] description Description of the Field
    *
-   * \return Whether or not the array was found
+   * \return Reference to the created Field
    *****************************************************************************
    */
-  bool getArray(std::unordered_map<int, bool>& map)
+  Verifiable& addBoolDict(const std::string& name,
+                          const std::string& description = "")
   {
-    return m_globalTable.getArray(map);
+    return m_globalTable.addBoolDict(name, description);
   }
 
   /*!
    *****************************************************************************
-   * \brief Get a int array represented as an unordered map from the input file
+   * \brief Add a dictionary of Integer Fields to the input file schema.
    *
-   * \param [out] map Unordered map to be populated with array contents
+   * \param [in] name Name of the dict
+   * \param [in] description Description of the Field
    *
-   * \return Whether or not the array was found
+   * \return Reference to the created Field
    *****************************************************************************
    */
-  bool getArray(std::unordered_map<int, int>& map)
+  Verifiable& addIntDict(const std::string& name,
+                         const std::string& description = "")
   {
-    return m_globalTable.getArray(map);
+    return m_globalTable.addIntDict(name, description);
   }
 
   /*!
    *****************************************************************************
-   * \brief Get a double array represented as an unordered map from the input file
+   * \brief Add a dictionary of Double Fields to the input file schema.
    *
-   * \param [out] map Unordered map to be populated with array contents
+   * \param [in] name Name of the dict
+   * \param [in] description Description of the Field
    *
-   * \return Whether or not the array was found
+   * \return Reference to the created Field
    *****************************************************************************
    */
-  bool getArray(std::unordered_map<int, double>& map)
+  Verifiable& addDoubleDict(const std::string& name,
+                            const std::string& description = "")
   {
-    return m_globalTable.getArray(map);
+    return m_globalTable.addDoubleDict(name, description);
   }
 
   /*!
    *****************************************************************************
-   * \brief Get a string array represented as an unordered map from the input file
+   * \brief Add a dictionary of String Fields to the input file schema.
    *
-   * \param [out] map Unordered map to be populated with array contents
+   * \param [in] name Name of the dict
+   * \param [in] description Description of the Field
    *
-   * \return Whether or not the array was found
+   * \return Reference to the created Field
    *****************************************************************************
    */
-  bool getArray(std::unordered_map<int, std::string>& map)
+  Verifiable& addStringDict(const std::string& name,
+                            const std::string& description = "")
   {
-    return m_globalTable.getArray(map);
+    return m_globalTable.addStringDict(name, description);
+  }
+
+  /*!
+   *****************************************************************************
+   * \brief Add a dictionary of user-defined types to the input file schema.
+   *
+   * \param [in] name Name of the dict
+   * \param [in] description Description of the Field
+   *
+   * \return Reference to the created Field
+   *****************************************************************************
+   */
+  Table& addGenericDict(const std::string& name,
+                        const std::string& description = "")
+  {
+    return m_globalTable.addGenericDict(name, description);
   }
 
   // TODO add update value functions
--- a/src/axom/inlet/Inlet.hpp
+++ b/src/axom/inlet/Inlet.hpp
@@ -369,9 +369,9 @@ public:
    * \brief Add an array of Boolean Fields to the input file schema.
    *
    * \param [in] name Name of the array
-   * \param [in] description Description of the Field
+   * \param [in] description Description of the array
    *
-   * \return Reference to the created Field
+   * \return Reference to the created array
    *****************************************************************************
    */
   Verifiable& addBoolArray(const std::string& name,
@@ -385,9 +385,9 @@ public:
    * \brief Add an array of Integer Fields to the input file schema.
    *
    * \param [in] name Name of the array
-   * \param [in] description Description of the Field
+   * \param [in] description Description of the array
    *
-   * \return Reference to the created Field
+   * \return Reference to the created array
    *****************************************************************************
    */
   Verifiable& addIntArray(const std::string& name,
@@ -401,9 +401,9 @@ public:
    * \brief Add an array of Double Fields to the input file schema.
    *
    * \param [in] name Name of the array
-   * \param [in] description Description of the Field
+   * \param [in] description Description of the array
    *
-   * \return Reference to the created Field
+   * \return Reference to the created array
    *****************************************************************************
    */
   Verifiable& addDoubleArray(const std::string& name,
@@ -417,9 +417,9 @@ public:
    * \brief Add an array of String Fields to the input file schema.
    *
    * \param [in] name Name of the array
-   * \param [in] description Description of the Field
+   * \param [in] description Description of the array
    *
-   * \return Reference to the created Field
+   * \return Reference to the created array
    *****************************************************************************
    */
   Verifiable& addStringArray(const std::string& name,
@@ -433,9 +433,9 @@ public:
    * \brief Add an array of user-defined types to the input file schema.
    *
    * \param [in] name Name of the array
-   * \param [in] description Description of the Field
+   * \param [in] description Description of the array
    *
-   * \return Reference to the created Field
+   * \return Reference to the created array
    *****************************************************************************
    */
   Table& addGenericArray(const std::string& name,
--- a/src/axom/inlet/Inlet.hpp
+++ b/src/axom/inlet/Inlet.hpp
@@ -454,10 +454,10 @@ public:
    * \return Reference to the created dictionary
    *****************************************************************************
    */
-  Verifiable& addBoolDict(const std::string& name,
-                          const std::string& description = "")
+  Verifiable& addBoolDictionary(const std::string& name,
+                                const std::string& description = "")
   {
-    return m_globalTable.addBoolDict(name, description);
+    return m_globalTable.addBoolDictionary(name, description);
   }
 
   /*!
@@ -470,10 +470,10 @@ public:
    * \return Reference to the created dictionary
    *****************************************************************************
    */
-  Verifiable& addIntDict(const std::string& name,
-                         const std::string& description = "")
+  Verifiable& addIntDictionary(const std::string& name,
+                               const std::string& description = "")
   {
-    return m_globalTable.addIntDict(name, description);
+    return m_globalTable.addIntDictionary(name, description);
   }
 
   /*!
@@ -486,10 +486,10 @@ public:
    * \return Reference to the created dictionary
    *****************************************************************************
    */
-  Verifiable& addDoubleDict(const std::string& name,
-                            const std::string& description = "")
+  Verifiable& addDoubleDictionary(const std::string& name,
+                                  const std::string& description = "")
   {
-    return m_globalTable.addDoubleDict(name, description);
+    return m_globalTable.addDoubleDictionary(name, description);
   }
 
   /*!
@@ -502,10 +502,10 @@ public:
    * \return Reference to the created dictionary
    *****************************************************************************
    */
-  Verifiable& addStringDict(const std::string& name,
-                            const std::string& description = "")
+  Verifiable& addStringDictionary(const std::string& name,
+                                  const std::string& description = "")
   {
-    return m_globalTable.addStringDict(name, description);
+    return m_globalTable.addStringDictionary(name, description);
   }
 
   /*!
@@ -518,10 +518,10 @@ public:
    * \return Reference to the created dictionary
    *****************************************************************************
    */
-  Table& addGenericDict(const std::string& name,
-                        const std::string& description = "")
+  Table& addGenericDictionary(const std::string& name,
+                              const std::string& description = "")
   {
-    return m_globalTable.addGenericDict(name, description);
+    return m_globalTable.addGenericDictionary(name, description);
   }
 
   // TODO add update value functions
--- a/src/axom/inlet/Inlet.hpp
+++ b/src/axom/inlet/Inlet.hpp
@@ -449,9 +449,9 @@ public:
    * \brief Add a dictionary of Boolean Fields to the input file schema.
    *
    * \param [in] name Name of the dict
-   * \param [in] description Description of the Field
+   * \param [in] description Description of the dictionary
    *
-   * \return Reference to the created Field
+   * \return Reference to the created dictionary
    *****************************************************************************
    */
   Verifiable& addBoolDict(const std::string& name,
@@ -465,9 +465,9 @@ public:
    * \brief Add a dictionary of Integer Fields to the input file schema.
    *
    * \param [in] name Name of the dict
-   * \param [in] description Description of the Field
+   * \param [in] description Description of the dictionary
    *
-   * \return Reference to the created Field
+   * \return Reference to the created dictionary
    *****************************************************************************
    */
   Verifiable& addIntDict(const std::string& name,
@@ -481,9 +481,9 @@ public:
    * \brief Add a dictionary of Double Fields to the input file schema.
    *
    * \param [in] name Name of the dict
-   * \param [in] description Description of the Field
+   * \param [in] description Description of the dictionary
    *
-   * \return Reference to the created Field
+   * \return Reference to the created dictionary
    *****************************************************************************
    */
   Verifiable& addDoubleDict(const std::string& name,
@@ -497,9 +497,9 @@ public:
    * \brief Add a dictionary of String Fields to the input file schema.
    *
    * \param [in] name Name of the dict
-   * \param [in] description Description of the Field
+   * \param [in] description Description of the dictionary
    *
-   * \return Reference to the created Field
+   * \return Reference to the created dictionary
    *****************************************************************************
    */
   Verifiable& addStringDict(const std::string& name,
@@ -513,9 +513,9 @@ public:
    * \brief Add a dictionary of user-defined types to the input file schema.
    *
    * \param [in] name Name of the dict
-   * \param [in] description Description of the Field
+   * \param [in] description Description of the dictionary
    *
-   * \return Reference to the created Field
+   * \return Reference to the created dictionary
    *****************************************************************************
    */
   Table& addGenericDict(const std::string& name,
--- a/src/axom/inlet/Reader.hpp
+++ b/src/axom/inlet/Reader.hpp
@@ -150,6 +150,7 @@ public:
    */
   virtual bool getIntMap(const std::string& id,
                          std::unordered_map<int, int>& values) = 0;
+  /// \overload
   virtual bool getIntMap(const std::string& id,
                          std::unordered_map<std::string, int>& values) = 0;
 
@@ -168,6 +169,7 @@ public:
    */
   virtual bool getBoolMap(const std::string& id,
                           std::unordered_map<int, bool>& values) = 0;
+  /// \overload
   virtual bool getBoolMap(const std::string& id,
                           std::unordered_map<std::string, bool>& values) = 0;
 
@@ -186,6 +188,7 @@ public:
    */
   virtual bool getDoubleMap(const std::string& id,
                             std::unordered_map<int, double>& values) = 0;
+  /// \overload
   virtual bool getDoubleMap(const std::string& id,
                             std::unordered_map<std::string, double>& values) = 0;
 
@@ -204,6 +207,7 @@ public:
    */
   virtual bool getStringMap(const std::string& id,
                             std::unordered_map<int, std::string>& values) = 0;
+  /// \overload
   virtual bool getStringMap(const std::string& id,
                             std::unordered_map<std::string, std::string>& values) = 0;
 
@@ -218,6 +222,7 @@ public:
    *****************************************************************************
    */
   virtual bool getIndices(const std::string& id, std::vector<int>& indices) = 0;
+  /// \overload
   virtual bool getIndices(const std::string& id,
                           std::vector<std::string>& indices) = 0;
 };
--- a/src/axom/inlet/Reader.hpp
+++ b/src/axom/inlet/Reader.hpp
@@ -201,16 +201,94 @@ public:
 
   /*!
    *****************************************************************************
+   * \brief Get an string-integer mapping for the given array
+   *
+   * This performs any necessary retrieval and mapping from the given identifier
+   * to what is in the input file.
+   *
+   * \param [in]  id    The identifier to the string that will be retrieved
+   * \param [out] map The values of the ints that were retrieved
+   *
+   * \return true if the dict was able to be retrieved from the file
+   *****************************************************************************
+   */
+  virtual bool getIntDict(const std::string& id,
+                          std::unordered_map<std::string, int>& values) = 0;
+
+  /*!
+   *****************************************************************************
+   * \brief Get an string-bool mapping for the given array
+   *
+   * This performs any necessary retrieval and mapping from the given identifier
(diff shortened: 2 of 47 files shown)


src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -428,6 +426,114 @@ TEST(inlet_object, implicit_conversion_primitives)
...
+
+TEST(inlet_dict, dict_of_struct_containing_dict)
+{
+  std::string testString =
+    "foo = { [\"key3\"] = { arr = { [\"key1\"] = 3 }; }, "
+    "        [\"key4\"] = { arr = { [\"key2\"] = 2 }; } }";
+  DataStore ds;
+  auto inlet = createBasicInlet(&ds, testString);
+
+  auto& dict_table = inlet.getGlobalTable().addGenericDict("foo");
```
white238, line 487: Isn't the `.getGlobalTable()` not needed at this point?

src/axom/inlet/examples/user_defined_type.cpp
```diff
@@ -172,10 +171,10 @@ struct ThermalSolver
  *      -- see above FromInlet<LinearSolver>
  *    },
  *    bcs = {
- *        [1] = {
+ *        ["temperature"] = {
```
white238, line 174: 👍

src/axom/inlet/Table.hpp
```diff
@@ -941,7 +1118,7 @@ class Table : public Verifiable
   std::vector<AggregateField> m_aggregate_fields;
 
   static const std::string ARRAY_GROUP_NAME;
```
white238, line 1120: Should this be changed to `CONTAINER_GROUP_NAME`? It covers both arrays and dicts right?

src/axom/inlet/Table.hpp
```diff
@@ -655,6 +679,37 @@ class Table : public Verifiable
...
+  {
+    T result;
+    // This needs to work transparently for both references to the underlying
+    // internal table and references using the same path as the data file
+    if(axom::utilities::string::endsWith(m_name, ARRAY_GROUP_NAME))
+    {
+      if(!getDict(result))
+      {
+        SLIC_ERROR(
+          "[Inlet] Table does not contain a valid array of requested type");
```
white238, line 702: Add m_name somewhere in this error message and below

src/axom/inlet/Table.hpp
```diff
@@ -371,63 +389,67 @@ class Table : public Verifiable
...
+   */
+  Verifiable& addStringDict(const std::string& name,
+                            const std::string& description = "");
+
+  /*!
+   *****************************************************************************
+   * \brief Add a dictionary of user-defined types to the input file schema.
+   *
+   * \param [in] name Name of the dict
+   * \param [in] description Description of the Field
```
white238, line 446: Description of the dict.. (and above)

src/axom/inlet/Table.hpp
```diff
@@ -921,9 +995,119 @@ class Table : public Verifiable
...
+   */
+  template <typename T>
+  typename std::enable_if<!detail::is_inlet_primitive<T>::value, bool>::type
+  getDict(std::unordered_map<std::string, T>& map) const
+  {
+    for(const auto& indexLabel : containerIndices())
+    {
+      map[indexLabel] = getTable(indexLabel).get<T>();
+    }
+    return true;
```
white238, line 1096: This always returns true. Is there something in the templating that i am missing that returns false?

joshessman-llnl: This is the version for dicts of user-defined type; the version for dicts of primitives can return false (in which case it will just `SLIC_ERROR`.  This function can still fail (if `containerIndices` fails with a `SLIC_ERROR`), but I will add a check to this function for consistency.

src/axom/inlet/Table.hpp
```diff
@@ -921,9 +995,129 @@ class Table : public Verifiable
...
+   * \param [out] map Unordered map to be populated with dictionary contents
+   *
+   * \return Whether or not the dictionary was found
+   *****************************************************************************
+   */
+  template <typename T>
+  typename std::enable_if<detail::is_inlet_primitive<T>::value, bool>::type
+  getDict(std::unordered_map<std::string, T>& map) const
+  {
+    map.clear();
```
white238, line 1069: Add a comment in the docs that this clears the given map

src/axom/inlet/Table.hpp
```diff
@@ -827,13 +889,16 @@ class Table : public Verifiable
    * 
    * \param [inout] table The inlet::Table to add the array to 
    * \param [in] lookupPath The path within the input file to read from
+   * \param[in] isDict Whether to use string keys
```
white238, line 892: \param [in]

src/axom/inlet/Table.hpp
```diff
@@ -560,6 +582,7 @@ class Table : public Verifiable
               typename std::enable_if<detail::is_inlet_primitive<T>::value>::type>
   Verifiable& addPrimitiveArray(const std::string& name,
                                 const std::string& description = "",
+                                const bool isDict = false,
```
white238, line 585: Missing docs for new parameter.

src/axom/inlet/Table.hpp
```diff
@@ -371,63 +389,67 @@ class Table : public Verifiable
...
-   * \brief Get an array represented as an unordered map from the input file
-   * of primitive type
+   * \brief Add a dictionary of Boolean Fields to the input file schema.
    *
-   * \param [out] map Unordered map to be populated with array contents
+   * \param [in] name Name of the dict
+   * \param [in] description Description of the Field
    *
-   * \return Whether or not the array was found
+   * \return Reference to the created Field
```
white238, line 397: `created dictionary`?

src/axom/inlet/Table.cpp
```diff
@@ -72,24 +72,52 @@ Table& Table::addTable(const std::string& name, const std::string& description)
...
-    SLIC_ERROR(fmt::format(
-      "[Inlet] Table '{0}' does not contain an array of user-defined objects",
-      m_name));
+    const auto view = m_sidreGroup->getView(CONTAINER_INDICES_NAME);
+    const int* array = view->getArray();
+    const auto num_elements = view->getNumElements();
+    indices.reserve(num_elements);
+    for(int i = 0; i < num_elements; i++)
+    {
+      indices.push_back(std::to_string(array[i]));
```
white238, line 87: I noticed that we store the array indices as ints (in Sidre) then convert them to strings (out of this function) then convert them back to ints (in get<> function).

Is there no way around this? Is it for keeping the templating clean?

src/axom/inlet/Table.cpp
```diff
@@ -127,26 +155,25 @@ Verifiable& Table::addStringArray(const std::string& name,
...
 {
-  if(m_sidreGroup->hasView(ARRAY_INDICIES_VIEW_NAME))
+  if(isGenericContainer())
   {
-    SLIC_ERROR(
-      fmt::format("[Inlet] Adding array of structs to array of structs {0} is "
-                  "not supported",
-                  m_name));
+    SLIC_ERROR(fmt::format(
+      "[Inlet] Adding array of container to array of structs {0} is "
```
white238, line 161: There is something wrong with this error message.  Is is supposed to be "Adding array of containers"?

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -428,6 +426,114 @@ TEST(inlet_object, implicit_conversion_primitives)
   EXPECT_EQ(arr, expected_arr);
 }
 
+TEST(inlet_dict, basic_dicts)
+{
+  std::string testString =
+    "foo = { [\"key1\"] = 4, [\"key3\"] = 6, [\"key2\"] = 10}";
```
kennyweiss, line 432: Would a string literal be better for defining these types of tests?
See: https://en.cppreference.com/w/cpp/language/string_literal

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -428,6 +426,114 @@ TEST(inlet_object, implicit_conversion_primitives)
...
+  DataStore ds;
+  auto inlet = createBasicInlet(&ds, testString);
```
kennyweiss, line 434: Looking at this PR, I'm realizing that it is not clear to me when to use `auto` and when to use `auto&` within the `inlet` API.

src/axom/inlet/tests/inlet_object.cpp
kennyweiss: Alternatively, you could use single quotes for the internal Lua strings, e.g.
```cpp
     "foo = { ['key1'] = 4, ['key3'] = 6, ['key2'] = 10}"; 
```

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -428,6 +426,114 @@ TEST(inlet_object, implicit_conversion_primitives)
...
+
+  auto& dict_table = inlet.addGenericDict("foo");
+
+  dict_table.addBool("bar", "bar's description");
+  dict_table.addBool("baz", "baz's description");
+  std::unordered_map<std::string, Foo> expected_foos = {{"key1", {true, false}},
+                                                        {"key2", {false, true}}};
+  std::unordered_map<std::string, Foo> foos;
+  foos = inlet["foo"].get<std::unordered_map<std::string, Foo>>();
```
kennyweiss, line 459: Just to clarify, with this new PR, there are separate `get<>` calls to access the integer-indexed entries and the string-indexed entries, right?  (w/ an `unordered_map<int, T>` and an `unordered_map<std::string, T>` template type)

kennyweiss: Follow-up 1: Should there be a test with a dictionary that has both int-valued and string-valued indices?

src/axom/inlet/Table.cpp
```diff
@@ -161,6 +188,60 @@ Table& Table::addGenericArray(const std::string& name,
...
+  std::vector<std::string> indices;
+  const std::string& fullName = appendPrefix(m_name, name);
+  if(m_reader.getIndices(fullName, indices))
+  {
+    auto group = table.m_sidreGroup->createGroup(CONTAINER_INDICES_NAME,
+                                                 /* list_format = */ true);
+    for(const auto idx : indices)
+    {
+      table.addTable(idx, description);
+      group->createViewString("", idx);
```
white238, line 235: Should probably add a comment here (and other similar functions) that says something to the effect of:

Creating groups with indices for their names to be filled by following add functions on returned table.

src/axom/inlet/tests/inlet_object.cpp
kennyweiss: Follow-up 2: Does the struct type for the string-indexed entries have to match the struct for the int-indexed entries?

src/axom/inlet/Reader.hpp
```diff
@@ -198,19 +204,22 @@ class Reader
...
+   * \param [out] indices The values of the indices that were retrieved
    *
-   * \return true if the array was able to be retrieved from the file
+   * \return true if the indices were able to be retrieved from the file
    *****************************************************************************
    */
-  virtual bool getArrayIndices(const std::string& id,
-                               std::vector<int>& indices) = 0;
+  virtual bool getIndices(const std::string& id, std::vector<int>& indices) = 0;
+  virtual bool getIndices(const std::string& id,
```
white238, line 221: Does doxygen automatically recognize this as an overload? and document both of them?

src/axom/inlet/Table.hpp
```diff
@@ -371,63 +389,67 @@ class Table : public Verifiable
...
+   *****************************************************************************
+   * \brief Add a dictionary of user-defined types to the input file schema.
+   *
+   * \param [in] name Name of the dict
+   * \param [in] description Description of the dictionary
+   *
+   * \return Reference to the created dictionary
+   *****************************************************************************
+   */
+  Table& addGenericDict(const std::string& name,
```
kennyweiss, line 451: Should the return type be a `Verifiable`? Or does it have to be a `Table`?

src/axom/inlet/Table.hpp
```diff
@@ -639,18 +663,56 @@ class Table : public Verifiable
     T result;
     // This needs to work transparently for both references to the underlying
     // internal table and references using the same path as the data file
-    if(axom::utilities::string::endsWith(m_name, ARRAY_GROUP_NAME))
+    if(axom::utilities::string::endsWith(m_name, CONTAINER_GROUP_NAME))
```
kennyweiss, line 666:
```cpp
if( axom::utilities::string::endsWith(m_name, CONTAINER_GROUP_NAME) ) {...}
```

Minor: It might be better (w.r.t. readability and maintenance), to make this a utility function.
E..g.
```cpp
if( inlet::is_container_group (m_name) ) { .... }
```

src/axom/inlet/Table.hpp
```diff
@@ -639,18 +663,56 @@ class Table : public Verifiable
     T result;
```
kennyweiss, line 663: I guess this assumes that `T` is default-constucible. 

This is almost definitely a reasonable assumption for inlet, but is an assumption nevertheless. This should probably be added to the doxygen for `get<>`  (assuming it's not already there)

src/axom/inlet/Inlet.hpp
```diff
@@ -430,7 +430,7 @@ class Inlet
 
   /*!
    *****************************************************************************
-   * \brief Add an array of Fields to the input file schema.
+   * \brief Add an array of user-defined types to the input file schema.
    *
    * \param [in] name Name of the array
    * \param [in] description Description of the Field
```
white238, line 436: Description of the array

white238 (approved):

src/axom/inlet/Table.hpp
```diff
@@ -921,9 +995,129 @@ class Table : public Verifiable
...
+  getArray(std::unordered_map<int, T>& map) const
+  {
+    map.clear();
+    if(!axom::utilities::string::endsWith(m_name, CONTAINER_GROUP_NAME))
+    {
+      return false;
+    }
+    for(auto& item : m_fieldChildren)
+    {
+      auto pos = item.first.find_last_of("/");
```
kennyweiss, line 1022: Minor -- It might also be a good idea to have a utility function to find this index.

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -428,6 +426,114 @@ TEST(inlet_object, implicit_conversion_primitives)
...
+  std::unordered_map<std::string, FooWithArray> foos_with_array;
+  foos_with_array =
+    inlet["foo"].get<std::unordered_map<std::string, FooWithArray>>();
+  EXPECT_EQ(foos_with_array, expected_foos);
+}
+
+TEST(inlet_dict, array_of_struct_containing_dict)
+{
+  std::string testString =
+    "foo = { [7] = { arr = { [\"key1\"] = 3 }; }, "
```
kennyweiss, line 522: What would happen if the key has a slash ("/") in it, 
e.g.
```cpp
std::string testString =
    "foo = { [7] = { arr = { ['key1'] = 3 }; },
             [4] = {arr = { ['key1/subkey1'] = 3 }; }, "
```
Is this worth testing?

kennyweiss: More generally,  are there restrictions on the key strings?

src/axom/inlet/Table.hpp
```diff
@@ -921,9 +995,129 @@ class Table : public Verifiable
...
+      auto index = item.first.substr(pos + 1);
```
kennyweiss, line 1077: Here's another case of extracting the index

src/axom/inlet/Table.cpp
kennyweiss: Follow up question -- would the code break if you had a table like the following:
```lua
foo = {
  [ 1 ] = "an int-indexed value",
  ['1'] = "a string-indexed value"
}
```

kennyweiss (commented):
I got about half-way through the PR and will get back to the rest tomorrow.

So far, this looks really nice.
I'd like to see an example table that mixes int-indexed entries with string-indexed entries to make sure that those are supported.

src/axom/inlet/tests/inlet_object.cpp
white238: Would you want to support both index types in a single container? I know it's supported in Lua but I think I find it very confusing from user stand-point.

For example:

```
foo = { ["1"] = "something", [1] = "something else"}
```

src/axom/inlet/Table.cpp
joshessman-llnl: Was supposed to be analogous to the message in its dictionary counterpart - "Adding array of structs to container of structs"

src/axom/inlet/Reader.hpp
joshessman-llnl: I think I need to add a `/// \overload` before the overload - with that it generates the following:
image

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -78,14 +78,13 @@ TEST(inlet_object, simple_array_of_struct_by_value)
...
-  auto& arr_table = inlet.getGlobalTable().addGenericArray("foo");
+  auto& arr_table = inlet.addGenericArray("foo");
 
   arr_table.addBool("bar", "bar's description");
   arr_table.addBool("baz", "baz's description");
   std::unordered_map<int, Foo> expected_foos = {{4, {true, false}},
                                                 {7, {false, true}}};
-  std::unordered_map<int, Foo> foos;
-  EXPECT_TRUE(arr_table.getArray(foos));
+  auto foos = inlet["foo"].get<std::unordered_map<int, Foo>>();
```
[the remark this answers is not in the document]
joshessman-llnl: I would expect the copy to be elided here, though it's technically not guaranteed.  `const auto&` would be valid but only via reference lifetime extension of temporaries, so I would prefer just `auto` (or `const auto`) here.

src/axom/inlet/tests/inlet_object.cpp
joshessman-llnl: I think the single quotes are a better option for consistency with Inlet's other tests - I was not aware that Lua allows for both single- and double-quoted strings like Python.

src/axom/inlet/tests/inlet_object.cpp
joshessman-llnl: References are needed for objects owned/managed by Inlet (child Tables/Fields returned from add* functions).  It's not ideal but because the copy constructors for those objects are explicitly deleted a mistake will always result in a compile-time error instead of at runtime.

src/axom/inlet/tests/inlet_object.cpp
joshessman-llnl: Right now this is not supported - I think it would require something like an `unordered_map<variant<int, std::string>, T>` which is definitely possible and relatively easy to use:
```cpp
    std::unordered_map<std::variant<std::string, int>, std::string> map;
    map[0] = "int-valued";
    map["3"] = "string-valued";
    for (const auto& thing : map)
    {
        std::visit([](const auto& key) {std::cout << key;}, thing.first);
        std::cout << ": " << thing.second << "\n";
    }
// output:
3: string-valued
0: int-valued
```
This would take a little bit of work to implement in C++11/14 so it may belong in a separate PR.

src/axom/inlet/Table.hpp
joshessman-llnl: Right now the `addGeneric*` functions return `Table` so the other `add*` functions are "accessible" - this prohibits an container of structs from being defined within a container of structs (which is detected and reported to the user when attempted).  Removing this limitation would require some modifications/extension of the `Verifiable` interface but is definitely possible.

src/axom/inlet/Table.hpp
joshessman-llnl: In this case the `T` is `unordered_map<int, UserDefinedType>` which is default-constructible even if `UserDefinedType` is not.  This will forward to the `FromInlet<UserDefinedType>` specialization which allows the user to construct the `UserDefinedType` instance however they see fit.

src/axom/inlet/Table.cpp
joshessman-llnl: I had initially left the indices as ints so they could be stored contiguously, but in retrospect I think that was a mistake given that performance isn't a big concern here.  Using identical storage for string and integer keys allowed a lot of duplicated code to be removed.

src/axom/inlet/tests/inlet_object.cpp
joshessman-llnl: This was a good idea - the current behavior is to add a subtable (which is in general inaccessible) to the array table.  I've added an error when the key contains a forward slash.

I tried a couple other potentially problematic keys and discovered that (unsurprisingly) the empty string is not a valid key, so I've added an error for that as well.

If a key that contains a slash is valid, it might be worth looking at user-specified scope delimiters (instead of hardcoded forward-slash) in a future PR.  I'm not sure whether this is an inlet-only change or if the forward-slash is hardcoded in Sidre as well.

src/axom/inlet/tests/inlet_object.cpp
kennyweiss: @white238  re: dictionaries with `"1"` and `1` as keys -- I agree that `inlet` should only support a subset of Lua and that the particular use case that I pointed out (string versions of ints) would be an antipattern that we would not want to support (or at least promote).

My concern was more in the realm of error checking / handling -- i.e. what would happen if a user attempted to do that.  Would the code crash? Would it silently overwrite one of the entries (and which one)? etc...

src/axom/inlet/Table.hpp
kennyweiss: I was asking more from a consistency point of view -- i.e. will the return type make sense to users.

src/axom/inlet/Table.hpp
kennyweiss: Perhaps a doxy `\tparam` comment on `T` would be appropriate?  (Might already be there, it's difficult to navigate through github)

src/axom/inlet/tests/inlet_object.cpp
kennyweiss: Re: sidre -- I think we added some knobs to theoretically allow for it to be user-changeable, but in practice that has not been explored, so it would likely need to fine-tuning / debugging to get that working.

src/axom/inlet/tests/inlet_object.cpp
joshessman-llnl: Given `foo = { ["1"] = "something", [1] = "something else"}`, the answer depends on whether `foo` was added as a dictionary or as an array.

If it was added as a dictionary, the integer-valued entry is ignored completely.  Inlet doesn't warn or error in this case - my understanding is that this was done so other things could be present in a table containing keyed entries:
```lua
foo  = {
    [1] = "something",
    bar = true,
}
```

If it was added as an array, an exception will be thrown by `std::stoi` when the key is converted to an integer.  There was some opposition to exceptions earlier so I've replaced `std::stoi` with the (somewhat returncode-based) libc `strtol` and added a SLIC error message on failure.

src/axom/inlet/Table.hpp
joshessman-llnl: My hope was that users wouldn't have to worry too much about the exact return type of these functions, but that the interface of the return type would reflect valid operations on the result. 

E.g., it wouldn't make sense to add a field to an array of primitives (as primitive arrays are "terminal"), but it would make sense to add a field to a struct (as structs are composite/product types).

src/axom/inlet/Table.hpp
```diff
@@ -1000,45 +1014,46 @@ class Table : public Verifiable
...
-   * \param [out] map Unordered map to be populated with array contents
+   * \param [out] map Unordered map to be populated with container contents, will be
+   * cleared before contents are added
    *
-   * \return Whether or not the array was found
+   * \return Whether or not the container was found
    *****************************************************************************
    */
-  template <typename T>
+  template <typename Key, typename T>
```
kennyweiss, line 1054: Minor: When I saw a `Key` type for a map, I expected its counterpart to be something like `Val`, rather than `T`.

src/axom/inlet/Table.hpp
```diff
@@ -650,68 +693,39 @@ class Table : public Verifiable
...
+  typename std::enable_if<detail::is_inlet_array<T>::value ||
+                            detail::is_inlet_dict<T>::value,
+                          T>::type
+  get() const
   {
     T result;
     // This needs to work transparently for both references to the underlying
     // internal table and references using the same path as the data file
-    if(axom::utilities::string::endsWith(m_name, CONTAINER_GROUP_NAME))
+    if(isContainerGroup(m_name))
```
kennyweiss, line 714: This seems much cleaner to me -- thanks!

src/axom/inlet/Table.hpp
```diff
@@ -650,68 +693,39 @@ class Table : public Verifiable
...
-                      "of requested type",
-                      m_name));
+        SLIC_ERROR(fmt::format(
+          "[Inlet] Table '{0}' does not contain a valid container of "
+          "requested type",
+          m_name));
       }
     }
-    else if(!getTable(CONTAINER_GROUP_NAME).getDict(result))
+    else if(!getTable(detail::CONTAINER_GROUP_NAME).getContainer(result))
```
kennyweiss, line 724: Suggestion: For this one, would something like `getContainerTable(result)` be cleaner?  (Perhaps not though)

src/axom/inlet/Table.hpp
```diff
@@ -394,7 +432,7 @@ class Table : public Verifiable
    * \param [in] name Name of the dict
    * \param [in] description Description of the Field
    *
-   * \return Reference to the created Field
+   * \return Reference to the created dictionary
    *****************************************************************************
    */
   Verifiable& addBoolDict(const std::string& name,
```
kennyweiss, line 438: Suggestion/possibility: Would the more verbose `addBoolDictionary()` be more consistent with the rest of `inlet`?
E.g. the above functions are `get*Array()`, not `get*Arr()`

kennyweiss (approved):
Thanks @joshessman-llnl -- this looks great.

If I understand it correctly, the code does not currently support having integer-indexed and string-indexed entries in the same table, but this might be supported in the future. Is that right?

src/axom/inlet/Table.hpp
joshessman-llnl: I'll go ahead and make sure all these templates are `<Key, Val>` - I think I might use `<K, V>` in a few places as well.

Interestingly enough, the [standard](https://eel.is/c++draft/unord.map.syn#header:%3cunordered_map%3e) uses `<Key, T>` for associative containers.

src/axom/inlet/Table.hpp
joshessman-llnl: Hmm...I think this is currently only used in this one place, but if we end up needing this logic elsewhere I think it would be a good idea to refactor it into a utility function - not sure if `getContainerTable` or `getTableContainer` would be better.

For now I've added an additional comment on this call.

src/axom/inlet/Table.hpp
joshessman-llnl: My initial thinking on the name was that we'd use `dict` as Python does, but as we use `Array` instead of `list` already I don't have a problem changing it to `Dictionary`.

joshessman-llnl:
> 
> If I understand it correctly, the code does not currently support having integer-indexed and string-indexed entries in the same table, but this might be supported in the future. Is that right?

Yes - I've logged this as #400 so we don't forget.

PR #396, opened 2020-11-26
Lua 5.4 compatibility + Static Umpire

white238:
* Backport fix from main sol repo to our older version of sol for Lua 5.4 compatibility
* Umpire was the only library building shared, this fixes it.  (For @gzagaris )

host-configs incoming soonish

### DIFF
--- a/scripts/uberenv/spack_configs/blueos_3_ppc64le_ib/packages.yaml
+++ b/scripts/uberenv/spack_configs/blueos_3_ppc64le_ib/packages.yaml
@@ -116,6 +116,8 @@ packages:
     variants: ~shared
   mfem:
     variants: ~mpi~metis~zlib
+  umpire:
+    variants: ~shared
 
 # Globally lock in versions of Devtools
   cppcheck:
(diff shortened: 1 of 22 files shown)


agcapps (approved):
Happy Thanksgiving.  Thanks for the fix, Chris.

scripts/uberenv/spack_configs/blueos_3_ppc64le_ib/packages.yaml
```diff
@@ -116,6 +116,8 @@ packages:
     variants: ~shared
   mfem:
     variants: ~mpi~metis~zlib
+  umpire:
+    variants: ~shared
```
agcapps, line 120: So, does this mean that when we build TPLs through spack, that Axom requires static compilation for conduit and umpire?  If that's so, why do we do this?

white238: Anything in the packages.yaml is specific to us and the SYS_TYPE in the path.  We do it this way so it is not hardcoded in the shared axom/package.py which goes up to Spack proper.

white238:
This is ready to go.

agcapps (approved):
Still looks good.

kennyweiss (approved):
Thanks @white238

PR #401, opened 2020-12-07
Add support for mixed-key dictionaries in Inlet

### LINKED ISSUE #400: Add mixed-key dictionaries to Inlet
Follow up question -- would the code break if you had a table like the following:
```lua
foo = {
  [ 1 ] = "an int-indexed value",
  ['1'] = "a string-indexed value"
}
```

_Originally posted by @kennyweiss in https://github.com/LLNL/axom/pull/391#discussion_r533911314_

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Adds support for mixed-key dictionaries in Inlet e.g., `[1] = "first", ['second'] = "second"`

About 2/3 of this PR (by LoC) is the implementation of the tagged union type.  This is a stand-in for `std::variant` with some additional Inlet-specific convenience functions.  There is a lot of boilerplate as defaulted constructors/destructors/assignment op/etc are not sufficient for a type- and memory-safe sum type.

## Design review

On 12/7, we reviewed this PR.  We discussed the design ideas:

1. Behavior when an integer key and string key have the same value, e.g., `[1] = "first", ['1'] = "second"`
2. Behavior when the user requests a single-type associative array from a mixed-key dictionary

### 1
It was decided that this should result in an error.  If duplicate keys of this form need to be supported at a later date this can be revisited.

### 2
It was decided that the subset of keys matching the requested type should be returned and the others should be ignored.  Tests demonstrating this were added.

Resolves #400


kennyweiss (approved):
Looks good @joshessman-llnl !

src/axom/inlet/VariantKey.cpp
```diff
@@ -0,0 +1,184 @@
...
+namespace axom
+{
+namespace inlet
+{
+VariantKey& VariantKey::operator=(const VariantKey& other)
+{
+  copyFrom(other);
+  return *this;
+}
+VariantKey& VariantKey::operator=(VariantKey&& other)
```
white238, line 17: You don't need to fix this... but it is an obvious oversight that no newline is required here...

white238 (approved):

PR #407, opened 2020-12-10
Added mfem to gcc8 and clang10 dockers

hennis1:
# Summary

- This PR is for adding the mfem feature to the docker containers.
  - Adds mfem to the docker containers


rhornung67 (approved):
I'm good with this if @white238 is. Thanks @hennis1 👍

white238:
Were you doing just +mfem in this PR or were you trying to update gfortran and CMake here as well? I am okay with leaving CMake at the current version because we test against newer versions on LC.  It may not be worth updating it here...

hennis1:
> Were you doing just +mfem in this PR or were you trying to update gfortran and CMake here as well? I am okay with leaving CMake at the current version because we test against newer versions on LC. It may not be worth updating it here...

gfortran and cmake are in other branches and aren't working. I am keeping them separate. Is there any dependency on mfem for gfortran or cmake?

white238:
You need to update the azure-pipelines.yml to the new docker image names with the time date stamp in them.  It should be in the build logs I think.

white238 (approved):
Looks great! Thanks @hennis1 !

rhornung67 (approved):
Thanks a bunch @hennis1

PR #409, opened 2020-12-15
Refactor Inlet's tests to be language-agnostic

joshessman-llnl:
# Summary

- This PR is a refactoring
- It does the following:
  - Utilizes Google Test's type-parameterized tests to allow for tests of multiple reader types using the same test string

In anticipation of `inlet::Reader` implementations for JSON and YAML Inlet's tests have been refactored to be language-independent.  As Lua's functionality is a superset of JSON's/YAML's, the idea is to keep all tests written in Lua and write a minimal translation tool for the tests whose inputs correspond to valid YAML/JSON.  This mitigates the need to maintain hundreds of tests in different input languages and to make sure that their behavior is identical.

Tests covering features available in Lua but not in JSON/YAML (discontiguous arrays, for example) will remain Lua-specific.

Almost all of this PR is boilerplate changes to use type-parameterized tests - the main exception is tests that needed to be modified to fit the subset of Lua that corresponds to valid YAML.  The original tests were preserved and made to use a fixed (Lua) reader type.


src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -70,49 +77,52 @@ TEST(inlet_object, simple_struct_by_value)
...
+  Inlet inlet = createBasicInlet<TypeParam>(&ds, testString);
 
   auto& arr_table = inlet.addGenericArray("foo");
 
   arr_table.addBool("bar", "bar's description");
   arr_table.addBool("baz", "baz's description");
-  std::unordered_map<int, Foo> expected_foos = {{1, {true, false}},
-                                                {2, {false, true}}};
+  // Lua is 1-indexed
+  const int base_idx =
```
kennyweiss, line 111: Suggestion: Seems like it might be a good idea for the reader to define the `base_idx`

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -154,19 +164,20 @@ struct FromInlet<FooWithArray>
   }
 };
 
-TEST(inlet_object, array_of_struct_containing_array)
+TYPED_TEST(InletObjectTest, array_of_struct_containing_array)
 {
   std::string testString =
-    "foo = { [4] = { arr = { [1] = 3 }; }, "
-    "        [7] = { arr = { [6] = 2 }; } }";
+    "foo = { [0] = { arr = { [0] = 3 }; }, "
```
kennyweiss, line 170: For Lua arrays, should the first index be `1` ?
Or, is the `0` there to support the other readers?

kennyweiss (approved):
Thanks @joshessman-llnl -- this looks great.

Have you tried this with yaml/json files?  Or is this just in anticipation of this working with yaml/json.  (Obviously, this PR does not yet include the yaml/json stuff).

src/axom/inlet/tests/inlet_object.cpp
joshessman-llnl: The other readers will read in arrays implicitly starting at zero, so I've explicitly specified the indices so the Lua reader will produce a matching result (since Lua places no restrictions on indices)

joshessman-llnl:
> 
> Have you tried this with yaml/json files? Or is this just in anticipation of this working with yaml/json. (Obviously, this PR does not yet include the yaml/json stuff).

Yes, I have a mostly-complete YAML reader on a [different branch](https://github.com/LLNL/axom/tree/feature/essman/yaml_reader)

src/axom/inlet/tests/inlet_Inlet.cpp
```diff
@@ -75,11 +83,11 @@ TEST(inlet_Inlet_basic, getTopLevelBools)
   EXPECT_EQ(table.type(), InletType::Nothing);
 }
 
-TEST(inlet_Inlet_basic, getNestedBools)
+TYPED_TEST(BasicInletTest, getNestedBools)
```
white238, line 86: Does this get prefixed with anything?  It is a convention in Axom to prefix all tests with their component and filename for filtering purposes.

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -570,6 +586,104 @@ TEST(inlet_dict, key_with_slash)
 }
 */
 
+// Noncontiguous array tests that are only valid in Lua
```
white238, line 589: 👍

white238 (approved):
Looks great!

PR #411, opened 2020-12-17
Add ability to specify arbitrary function types in input files

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Extends the currently restricted set of possible functions in the input file

The initial Inlet functions PR (#385) hardcoded the set of allowed functions.  To extend this, this PR generates the set of permissible function signatures at compile time and uses that to instantiate the `FunctionWrapper` type.

It also renames the `inlet::FunctionType` enum to `inlet::FunctionTag` (which enumerates the allowed types for function arguments/return values).  It replaces the `inlet::FunctionType` name with a set of type aliases, so `FunctionTag::Vec3D` is the enum used to specify the signature, and `FunctionType::Vec3D` is the vector type (currently `primal::Vector<double, 3>`) that is actually passed to/returned from the function.

Examples are added demonstrating how functions in a Lua input file might be used to define `mfem::Coefficient`s.


src/axom/inlet/tests/inlet_function.cpp
```diff
@@ -179,7 +209,7 @@ TEST(inlet_function, simple_vec3_to_vec3_verify_lambda_fail)
 struct Foo
 {
   bool bar;
-  std::function<axom::primal::Vector3D(axom::primal::Vector3D)> baz;
+  std::function<FunctionType::Vec3D(FunctionType::Vec3D)> baz;
```
kennyweiss, line 212: 👍

src/axom/inlet/tests/inlet_function.cpp
```diff
@@ -54,11 +55,10 @@ TEST(inlet_function, simple_vec3_to_vec3_raw)
   auto inlet = createBasicInlet(&ds, testString);
 
   auto func =
-    inlet.reader().getFunction("foo", FunctionType::Vec3D, {FunctionType::Vec3D});
+    inlet.reader().getFunction("foo", FunctionTag::Vec3D, {FunctionTag::Vec3D});
 
   EXPECT_TRUE(func);
-  auto result =
-    func.call<axom::primal::Vector3D>(axom::primal::Vector3D {1, 2, 3});
+  auto result = func.call<FunctionType::Vec3D>(FunctionType::Vec3D {1, 2, 3});
```
kennyweiss, line 61: Would this function also work if you only supplied 1 or 2 parameters?
e.g.
```cpp
   auto result = func.call<FunctionType::Vec3D>(FunctionType::Vec3D {1});      --> [2,0,0]
   auto result = func.call<FunctionType::Vec3D>(FunctionType::Vec3D {1, 2});   --> [2,4,0]
```

...or, are three parameters always required?

src/axom/inlet/examples/mfem_coefficient.cpp
```diff
@@ -0,0 +1,192 @@
...
+  // We only need the boundary condition sub-table
+  auto& bc_table = inlet.addGenericArray("bcs", "List of boundary conditions");
+  BoundaryCondition::defineSchema(bc_table);
+
+  if(!inlet.verify())
+  {
+    SLIC_ERROR("Inlet failed to verify against provided schema");
+  }
+
+  // Read all the data into a thermal solver object
```
kennyweiss, line 179: Possible copy/paste error: "thermal solver"?

kennyweiss (approved):
Looks great @joshessman-llnl   (although I didn't dig too deep into the template metaprogramming part).

Side question -- are there any `inlet` examples that define a callback function within a lua deck and call that function from within the deck? If not, should there be?

src/axom/inlet/tests/inlet_function.cpp
joshessman-llnl: The `Vector` constructor will value-initialize the remaining components when less than three are provided - I'll add a test for this.

joshessman-llnl:
> 
> Side question -- are there any `inlet` examples that define a callback function within a lua deck and call that function from within the deck? If not, should there be?

I added a lambda verifier to the mfem coefficient example that actually calls the function and checks its result.  We could also project the coefficients onto a GridFunction (thus calling the Lua functions) but my thinking was that demonstrating the use of a verifier on a function is more relevant to an Inlet example.

src/axom/inlet/tests/inlet_function.cpp
kennyweiss: Thanks @joshessman-llnl -- that's what I thought

src/axom/inlet/Function.hpp
```diff
@@ -48,12 +48,25 @@ namespace inlet
...
+/*!
+ *******************************************************************************
+ * \brief The types used to describe function signatures in the input file
+ * 
+ * \note These are aliases intended to improve readability
+ *******************************************************************************
+ */
+struct FunctionType
+{
+  using Vec3D = axom::primal::Vector3D;
```
white238, line 66: Do you want to add Empty|None|whatever you want to call it in this PR or in a follow up?  String would also be helpful

src/axom/inlet/docs/sphinx/advanced_types.rst
```diff
@@ -125,8 +125,8 @@ the following methods:
 
 .. code-block:: C++
 
-  auto coef = inlet["coef"].get<std::function<double(axom::primal::Vector3D)>>();
-  std::function<double(axom::primal::Vector3D)> coef = inlet["coef"];
+  auto coef = inlet["coef"].get<std::function<double(axom::inlet::FunctionType::Vec3D)>>();
```
white238, line 128: Can these lines be referenced from a compiled and tested example so they don't get out of sync with what works?

src/axom/inlet/tests/inlet_function.cpp
```diff
@@ -232,8 +284,8 @@ TEST(inlet_function, simple_vec3_to_vec3_array_of_struct)
   arr_table.addBool("bar", "bar's description");
   arr_table
     .addFunction("baz",
-                 FunctionType::Vec3D,
-                 {FunctionType::Vec3D},
+                 FunctionTag::Vec3D,
```
white238, line 287: Can you add a function that takes and returns a just a double?

```
    inlet.addFunction("foo",
                 FunctionTag::Double,
                 {FunctionTag::Double},
                 "foo's description")
```

white238 (approved):

src/axom/inlet/Function.hpp
joshessman-llnl: I'll save this for a follow up, I anticipate some nuance with the Void type

PR #413, opened 2020-12-21
Add YAML reader to Inlet

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Adds an implementation of the `inlet::Reader` interface for YAML
  - Adds primitive Lua -> YAML translation functionality for testing purposes

I opted for a hand-rolled translation approach to avoid adding dependencies on flex/bison/ANTLR or similar and because the subset of Lua that we care about for testing purposes is fairly minimal.

### DIFF
--- a/src/axom/inlet/YAMLReader.cpp
+++ b/src/axom/inlet/YAMLReader.cpp
@@ -349,7 +349,6 @@ bool YAMLReader::getArray(const std::string& id,
     if(getValue(node, value))
     {
       values[0] = value;
-      return true;
     }
     else
     {
@@ -367,7 +366,7 @@ bool YAMLReader::getArray(const std::string& id,
       const auto& child = itr.next();
       T value;
       // Inlet allows for heterogenous containers, so a failure here is "normal"
-      if(!getValue(child, value))
+      if(getValue(child, value))
       {
         values[index] = value;
       }
--- a/src/axom/inlet/YAMLReader.cpp
+++ b/src/axom/inlet/YAMLReader.cpp
@@ -152,11 +152,12 @@ bool YAMLReader::getValue(const conduit::Node& node, double& value)
   // Match LuaReader functionality - promote from integer
   if(node.dtype().is_number())
   {
-    value = node.to_float64();
+    value = node.to_double();
     return true;
   }
   return false;
 }
+
 bool YAMLReader::getValue(const conduit::Node& node, bool& value)
 {
   // Boolean literals don't appear to be parsed as such - they are strings
@@ -166,10 +167,7 @@ bool YAMLReader::getValue(const conduit::Node& node, bool& value)
     // YAML 1.2 spec, section 10.3.2
     // FIXME: Converting the string to lowercase is not strictly correct, it
     // allows for things like tRue and falsE
-    std::transform(as_str.begin(),
-                   as_str.end(),
-                   as_str.begin(),
-                   [](const unsigned char c) { return std::tolower(c); });
+    utilities::string::toLower(as_str);
     if(as_str == "true")
     {
       value = true;
@@ -275,11 +273,8 @@ bool YAMLReader::getIndices(const std::string& id,
   {
     return false;
   }
-  // FIXME: Update to range-based for loops when Axom begins using Conduit 0.6.0
-  auto itr = node.children();
-  while(itr.has_next())
+  for(const auto& child : node.children())
   {
-    const auto& child = itr.next();
     indices.push_back(child.name());
   }
   return true;
@@ -304,11 +299,8 @@ bool YAMLReader::getDictionary(const std::string& id,
     return false;
   }
 
-  // FIXME: Update to range-based for loops when Axom begins using Conduit 0.6.0
-  auto itr = node.children();
-  while(itr.has_next())
+  for(const auto& child : node.children())
   {
-    const auto& child = itr.next();
     const auto name = child.name();
 
     T value;
@@ -363,11 +355,8 @@ bool YAMLReader::getArray(const std::string& id,
   else
   {
     conduit::index_t index = 0;
-    // FIXME: Update to range-based for loops when Axom begins using Conduit 0.6.0
-    auto itr = node.children();
-    while(itr.has_next())
+    for(const auto& child : node.children())
     {
-      const auto& child = itr.next();
       T value;
       // Inlet allows for heterogenous containers, so a failure here is "normal"
       if(getValue(child, value))
--- a/src/axom/inlet/YAMLReader.cpp
+++ b/src/axom/inlet/YAMLReader.cpp
@@ -35,8 +35,20 @@ bool YAMLReader::parseFile(const std::string& filePath)
                              filePath));
     return false;
   }
-  m_root.load(filePath, "yaml");
-  return true;
+  bool success = true;
+  // Temporarily enable exceptions for Conduit errors so they can be caught and displayed
+  sidre::DataStore::setConduitDefaultMessageHandlers();
+  try
+  {
+    m_root.load(filePath, "yaml");
+  }
+  catch(const conduit::Error& e)
+  {
+    SLIC_WARNING(fmt::format("[Inlet]: Failed to parse YAML:\n{}", e.message()));
+    success = false;
+  }
+  sidre::DataStore::setConduitSLICMessageHandlers();
+  return success;
 }
 
 bool YAMLReader::parseString(const std::string& YAMLString)
@@ -46,8 +58,19 @@ bool YAMLReader::parseString(const std::string& YAMLString)
     SLIC_WARNING("Inlet: Given an empty YAML string to parse.");
     return false;
   }
-  m_root.parse(YAMLString, "yaml");
-  return true;
+  bool success = true;
+  sidre::DataStore::setConduitDefaultMessageHandlers();
+  try
+  {
+    m_root.parse(YAMLString, "yaml");
+  }
+  catch(const conduit::Error& e)
+  {
+    SLIC_WARNING(fmt::format("[Inlet]: Failed to parse YAML:\n{}", e.message()));
+    success = false;
+  }
+  sidre::DataStore::setConduitSLICMessageHandlers();
+  return success;
 }
 
 namespace detail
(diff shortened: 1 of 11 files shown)


estebanpauli:
I have no experience with Inlet, but I'll do my best to review.

src/axom/inlet/YAMLReader.cpp
```diff
@@ -0,0 +1,382 @@
...
+    return false;
+  }
+
+  // FIXME: Update to range-based for loops when Axom begins using Conduit 0.6.0
+  auto itr = node.children();
+  while(itr.has_next())
+  {
+    const auto& child = itr.next();
+    const auto name = child.name();
+    if(!getValue(child, values[name]))
```
estebanpauli, line 311: I had to think about this for a minute to understand what was going on. For me, it would have been simpler like this:
```
T value;
if (getValue(child, value)) {
    values[name] = value;
}
```

src/axom/inlet/YAMLReader.cpp
```diff
@@ -0,0 +1,382 @@
...
+    {
+      // The current interface allows for overlapping types, but we need to
```
estebanpauli, line 313: I'm not sure what this means, but that could just be my limited experience with Inlet.

src/axom/inlet/YAMLReader.cpp
```diff
@@ -0,0 +1,382 @@
...
+template <typename T>
+bool YAMLReader::getArray(const std::string& id,
+                          std::unordered_map<int, T>& values)
+{
+  values.clear();
+  const auto node = detail::traverseNode(m_root, id);
+  // Truly primitive (i.e., not string) types are contiguous so we grab the array pointer
+  if(node.dtype().number_of_elements() > 1)
+  {
+    // The template parameter is not enough to know the type of the conduit array
```
estebanpauli, line 330: Good documentation explaining why you have to do things this way.

src/axom/inlet/YAMLReader.hpp
```diff
@@ -0,0 +1,255 @@
...
+  FunctionVariant getFunction(const std::string& id,
+                              const FunctionType ret_type,
+                              const std::vector<FunctionType>& arg_types);
+
+  /*!
+   *****************************************************************************
+   * \brief The base index for arrays in Lua
+   *****************************************************************************
+   */
+  static const int baseIndex = 0;
```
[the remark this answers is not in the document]
joshessman-llnl: I had missed updating the comment - this is the base index for arrays in YAML that gets used for testing purposes

src/axom/inlet/YAMLReader.cpp
joshessman-llnl: Thanks for the suggestion - I agree that this a lot more intuitive.

src/axom/inlet/YAMLReader.cpp
```diff
@@ -0,0 +1,384 @@
...
+  }
+  else if(!node.dtype().is_list())
+  {
+    // Single-element arrays will be just the element itself
+    // If it's a single element, we know the index is zero
+    T value;
+    if(getValue(node, value))
+    {
+      values[0] = value;
+      return false;
```
white238, line 351: return true?

src/axom/inlet/YAMLReader.cpp
```diff
@@ -0,0 +1,382 @@
...
+YAMLReader::YAMLReader() { }
+bool YAMLReader::parseFile(const std::string& filePath)
+{
+  if(!axom::utilities::filesystem::pathExists(filePath))
+  {
+    SLIC_WARNING(fmt::format("Inlet: Given YAML input file does not exist: {0}",
+                             filePath));
+    return false;
+  }
+  m_root.load(filePath, "yaml");
```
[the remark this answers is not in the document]
joshessman-llnl: This is a good question - failure to parse here will result in a SLIC error (via Conduit's error handler) and subsequently program termination.  Exceptions are (to my knowledge, intentionally) avoided in Axom, so I am not sure that we have the ability to detect/handle a parse error in this context.

src/axom/inlet/YAMLReader.hpp
```diff
@@ -0,0 +1,255 @@
...
+   * \return true if the string was able to be parsed
+   *****************************************************************************
+   */
+  bool parseString(const std::string& YAMLString);
+
+  /*!
+   *****************************************************************************
+   * \brief Return a boolean out of the input file
+   *
+   * This performs any necessary retrieval and mapping from the given identifier
```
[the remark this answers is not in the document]
joshessman-llnl: Similar inheritance hierarchies in Axom duplicate the documentation (e.g. `lumberjack::Communicator`, `mint::Field`), but it looks like Doxygen is able to produce correct documentation for derived classes without the copy-pasted annotations:
image

src/axom/inlet/YAMLReader.cpp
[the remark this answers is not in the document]
estebanpauli: That behavior is inconsistent with the documentation, which states that the function just returns false if the file can't be read. Having the program terminate instead seems quite drastic, especially for something that can be used as a library. If this is reasonably easy to change, I think it should be. Otherwise, I would document this possibility in the API.

The Axom team should chime in here.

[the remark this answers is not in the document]
cyrush: Depending on how Sidre is used, you should be able to catch the conduit exception here, and return false (or otherwise report the details of the error)

src/axom/inlet/YAMLReader.cpp
```diff
@@ -0,0 +1,389 @@
...
+  }
+  return false;
+}
+
+bool YAMLReader::getValue(const conduit::Node& node, double& value)
+{
+  // Match LuaReader functionality - promote from integer
+  if(node.dtype().is_number())
+  {
+    value = node.to_float64();
```
cyrush, line 155: to_double would be better here (even though I know of know system so far where a double isn't 64bit)

cyrush (approved):
Looks good!

src/axom/inlet/tests/inlet_test_utils.hpp
```diff
@@ -11,12 +11,56 @@
...
+ */
+class LuaToYAML
+{
+public:
+  /*!
+   *****************************************************************************
+   * \brief Converts a Lua string to YAML
+   * \param [in] luaString The string to convert
+   * \note This function does not check for syntactic validity.  It is the
+   * responsibility of the callet to pass a valid Lua string.
```
kennyweiss, line 39: typo: `callet` -> `caller`

kennyweiss (approved):
Looks good @joshessman-llnl 

It might be a good idea to provide a few simple test cases that run the `YAMLReader` directly on valid and invalid YAML input (i.e. without implicitly going through the Lua-to-YAML conversion).

At the very least, it can demonstrate to users that (a) the is possible and (b) how to do this.

src/axom/inlet/YAMLReader.cpp
```diff
@@ -0,0 +1,389 @@
...
+    return true;
+  }
+  return false;
+}
+bool YAMLReader::getValue(const conduit::Node& node, bool& value)
```
white238, line 160: New line above this.. have you run this through the style checker?

src/axom/inlet/tests/inlet_test_utils.cpp
```diff
@@ -7,4 +7,158 @@
 #include "axom/slic.hpp"
 
 namespace axom::inlet::detail
```
white238, line 9: One-line namespaces throw a warning on non-C++17 compilers

src/axom/inlet/YAMLReader.cpp
```diff
@@ -0,0 +1,389 @@
...
+{
+  // Boolean literals don't appear to be parsed as such - they are strings
+  if(node.dtype().is_string())
+  {
+    std::string as_str = node.as_string();
+    // YAML 1.2 spec, section 10.3.2
+    // FIXME: Converting the string to lowercase is not strictly correct, it
+    // allows for things like tRue and falsE
+    std::transform(as_str.begin(),
```
white238, line 169: Could you add a to_lower to the string utilities functions in core?  This is useful in more than here and improves readability.

src/axom/inlet/YAMLReader.cpp
```diff
@@ -0,0 +1,389 @@
...
+bool YAMLReader::getIndices(const std::string& id,
+                            std::vector<std::string>& indices)
+{
+  indices.clear();
+  const auto node = detail::traverseNode(m_root, id);
+  if(!node.dtype().is_object())
+  {
+    return false;
+  }
+  // FIXME: Update to range-based for loops when Axom begins using Conduit 0.6.0
```
white238, line 278: conduit 0.6.0 is now here!

joshessman-llnl:
> It might be a good idea to provide a few simple test cases that run the `YAMLReader` directly on valid and invalid YAML input (i.e. without implicitly going through the Lua-to-YAML conversion).

Good idea, I've just taken some of the Inlet Reader tests and "manually" converted the test string from Lua to YAML.

src/axom/inlet/YAMLReader.cpp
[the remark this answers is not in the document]
joshessman-llnl: The exceptions thrown by conduit's default error handler are suppressed by the SLIC error handler - I'm wondering if it would be acceptable to temporarily reset the error handler to the conduit default, catch the exception if it gets thrown, then reset to the SLIC error handler.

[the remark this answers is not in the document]
kennyweiss: We just handled something similar in a different application, but I don't recall the specifics.  
@cyrush can you please help @joshessman-llnl  with the details?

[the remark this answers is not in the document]
kennyweiss: The two functions here are likely sufficient for your needs: 
https://github.com/LLNL/axom/pull/350/files#diff-2908fac76434ec9e7431bfb3b6173f7c87ae79671faec954bf97e62d77c2b98f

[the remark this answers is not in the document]
joshessman-llnl: Yep, it looks like those functions were enough to implement what I had proposed.

white238 (approved):

estebanpauli (approved):

PR #419, opened 2021-01-05
Add a JSON reader to Inlet

joshessman-llnl:
# Summary

- This PR is a feature/refactor
- It does the following:
  - Refactors the `YAMLReader` into a `ConduitReader` with a protocol parameter (i.e., `yaml` for the YAMLReader and `json` for the new JSONReader)
  - Adds a `JSONReader` and extends the Lua translation logic to translate Lua -> JSON


kennyweiss (approved):
Looks great @joshessman-llnl !

Please update the `RELEASE-NOTES`

cyrush (approved):

src/axom/inlet/ConduitReader.cpp
```diff
@@ -26,47 +26,55 @@ namespace axom
 {
 namespace inlet
 {
-YAMLReader::YAMLReader() { }
-bool YAMLReader::parseFile(const std::string& filePath)
+ConduitReader::ConduitReader(const std::string& protocol) : m_protocol(protocol)
+{
+  SLIC_ERROR_IF(
+    (protocol != "yaml") && (protocol != "json"),
+    "Inlet: Only JSON and YAML protocols are supported by ConduitReader");
```
white238, line 33: The mixed casing between the check and the error is concerning.  Can you lower case it before the check? (using the previously mentioned new string utilities function in some other PR I asked for :) )

white238: Or is this hidden behind the yaml/json reader classes from the users?

src/axom/inlet/ConduitReader.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+namespace axom
+{
+namespace inlet
+{
+/*!
+ *******************************************************************************
+ * \class ConduitReader
+ *
+ * \brief A Reader that is able to read variables from a file compatible with
```
white238, line 29: Can you add a message that the user is not to use this class directly but to use the yaml/json classes directly?

white238 (approved):

PR #423, opened 2021-01-12
Incorporate new mpi blueprint features

nselliott:
# Summary

- This PR is a feature addition
  - Incorporates new MPI-enabled blueprint functions for verify() and generate_index().  These take a communicator and work on parallel meshes more robustly.  They both can handle underdecomposed meshes, verify() broadcasts its results so that its boolean return value is the same on all ranks, and generate_index enabled an overloaded DataStore::generateBlueprintIndex that can be an MPI collective call instead requiring if logic to restrict it to one rank.


src/axom/sidre/CMakeLists.txt
```diff
@@ -106,6 +106,7 @@ endif()
 set(sidre_depends
     core
     conduit::conduit
+    conduit::conduit_mpi
```
white238, line 109: This should be guarded with ENABLE_MPI with the following:

```
blt_list_append(TO sidre_depends ELEMENTS conduit::conduit_mpi IF ENABLE_MPI)

File: src/axom/sidre/examples/sidre_generateindex.cpp [hunk 2]
```diff
@@ -0,0 +1,275 @@
...
+}
+#endif
+
+int main(int argc, char** argv)
+{
+  // Deal with unused variables
+  AXOM_DEBUG_VAR(argc);
+  AXOM_DEBUG_VAR(argv);
+#ifdef AXOM_USE_MPI
+  int num_ranks = 1;
```
white238, line 262: Should you initialize a SLIC logger?

rhornung67 (approved):

src/axom/sidre/tests/spio/spio_parallel.hpp
```diff
@@ -997,13 +997,10 @@ TEST(spio_parallel, sidre_simple_blueprint_example)
   MPI_Barrier(MPI_COMM_WORLD);
 
   // rank 0 adds the bp index to the root file
```
agcapps, line 999: Does `writeBlueprintIndexToRootFile()` make sure that rank 0 is the only one that writes?  If the function makes that check, please briefly say so in the comment.  Otherwise, the comment should be removed since the check for rank 0 was removed.

agcapps (commented):
Thanks @nselliott .  Just an issue with what looks like a missed or incomplete comment.

src/axom/sidre/examples/CMakeLists.txt
```diff
@@ -88,6 +89,17 @@ if(ENABLE_MPI)
                            )
         endif()
     endforeach()
+
+    blt_add_test( NAME sidre_createdatastore_parallel
```
kennyweiss, line 93: Should we also test these with a single rank, i.e. `NUM MPI_TASKS 1`?

src/axom/sidre/core/DataStore.cpp
```diff
@@ -618,6 +622,54 @@ bool DataStore::generateBlueprintIndex(const std::string& domain_path,
...
 }
 
+#ifdef AXOM_USE_MPI
+bool DataStore::generateBlueprintIndex(MPI_Comm comm,
+                                       const std::string& domain_path,
+                                       const std::string& mesh_name,
+                                       const std::string& index_path)
+{
+  Group* domain;
+  if (domain_path == "/")
```
kennyweiss, line 632: Should we also handle the empty string (`domain_path == ""`) as the root?
(or perhaps that case is already handled by the `else if` case)

cyrush (approved):

src/axom/sidre/core/DataStore.cpp
```diff
@@ -618,6 +622,54 @@ bool DataStore::generateBlueprintIndex(const std::string& domain_path,
...
+    domain->createNativeLayout(mesh_node);
+  }
+
+  Group* bpindex = getRoot()->hasGroup(index_path)
+    ? getRoot()->getGroup(index_path)
+    : getRoot()->createGroup(index_path);
+
+  bool success = false;
+  conduit::Node info;
+  if(conduit::blueprint::mpi::verify("mesh", mesh_node, info, comm))
```
kennyweiss, line 659: Should we dump the `info` in problem cases?  Perhaps in a `SLIC_DEBUG`?

I suppose the user can do that when this function returns false...

kennyweiss (approved):
Thanks @nselliott -- looks good.

It might be a good idea to add some additional tests for the case where there is only a single MPI rank.

src/axom/sidre/core/DataStore.cpp
nselliott: I think it might be an undocumented feature that "/" represents the root in sidre?  I'm not sure what would happen if an empty string is passed right now.  I could make "" also work for the root for this method, but I'm not sure it will be consistent with everything else in sidre.  Either way, the documentation should be clarified.

src/axom/sidre/examples/CMakeLists.txt
nselliott: Single rank tests for these are created in the foreach loop immediately above this.

src/axom/sidre/core/DataStore.cpp
nselliott: `info` is a local variable that will be destroyed upon return, so the user can't do anything with it.  Using SLIC is a good idea.

src/axom/sidre/examples/CMakeLists.txt
kennyweiss: Thanks @nselliott -- hard to see from github :)

PR #424, opened 2021-01-12
Extend permissible Inlet function argument/return types to include void and string

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Adds `void` and `std::string` as options for return/argument types in input-filed-defined functions

With the addition of two new argument types the number of function signatures increases from 8 to 64.  No additional assembly is generated but this does have an impact on compile time - on my machine, it roughly doubles compilation time for the header.

One option to decrease this would be to use virtual functions, but we would lose the uniform `get<T>` syntax as virtual functions cannot be combined with templates.

We could also use a separate program to generate the "name" of the full template instantiation, and "paste" that into the user-facing header via cmake's `configure_file` or similar.

The snippet below is mostly portable between linux compilers:
```cpp
abi::__cxa_demangle(typeid(BasicFunctionWrapper).name();
```


src/axom/inlet/tests/inlet_function.cpp
```diff
@@ -139,6 +139,84 @@ TEST(inlet_function, simple_double_to_double_through_table)
...
+                    "foo's description");
+
+  auto callable =
+    inlet["foo"].get<std::function<FunctionType::Void(FunctionType::Double)>>();
+  double arg = -6.37;
+  callable(arg);
+
+  inlet.addDouble("bar", "bar's description");
+  double result = inlet["bar"];
+  EXPECT_FLOAT_EQ(result, arg);
```
kennyweiss, line 174: Suggestion: It might be good to check that `inlet["bar"] == 19.9` before `callable(arg)`

src/axom/inlet/tests/inlet_function.cpp
```diff
@@ -139,6 +139,84 @@ TEST(inlet_function, simple_double_to_double_through_table)
...
+}
+
+TEST(inlet_function, simple_string_to_double_through_table)
+{
+  std::string testString =
+    "function foo(s) if s == 'a' then return 9.1 "
```
kennyweiss, line 180: Suggestion: Can you please reformat this lua input to make it easier to read?
E.g. something like following (or better):
```lua
    "function foo(s) "
    "  if     s == 'a' then return 9.1 " 
    "  elseif s == 'b' then return -6.3 "
    "  else            return 66.5 "
    "  end 
    "end"; 
```

src/axom/inlet/tests/inlet_function.cpp
```diff
@@ -139,6 +139,84 @@ TEST(inlet_function, simple_double_to_double_through_table)
...
+
+  EXPECT_FLOAT_EQ(callable("a"), 9.1);
+  EXPECT_FLOAT_EQ(callable("b"), -6.3);
+  EXPECT_FLOAT_EQ(callable("c"), 66.5);
+}
+
+TEST(inlet_function, simple_double_to_string_through_table)
+{
+  std::string testString =
+    "function foo(d) if d == 9.1 then return 'a' "
```
kennyweiss, line 202: Minor: The following appears to work, but to avoid possible issues on some compilers with floating point comparisons, it might be better to have your parameters have integer values (e.g.  `1` and `2` instead of `9.1` and `-6.3`)

kennyweiss (approved):
Thanks @joshessman-llnl  -- looks good.

kennyweiss:
>  No additional assembly is generated but this does have an impact on compile time - on my machine, it roughly doubles compilation time for the header.

@joshessman-llnl -- could you please provide some (estimates of) quantitative numbers here?  
E.g. compile time increasing from 2 to 4 seconds would be different from it increasing from 30 to 60 seconds.

src/axom/inlet/tests/inlet_function.cpp
joshessman-llnl: I agree that this would be useful to check, but currently Inlet will only read something from the input file once (that is, later updates will not affect the Inlet state)  - this is why I had to delay the `addDouble` call until after the function is invoked.

joshessman-llnl:
> 
> 
> > No additional assembly is generated but this does have an impact on compile time - on my machine, it roughly doubles compilation time for the header.
> 
> @joshessman-llnl -- could you please provide some (estimates of) quantitative numbers here?
> E.g. compile time increasing from 2 to 4 seconds would be different from it increasing from 30 to 60 seconds.

On my machine the difference is roughly 0.7s -> 1.4s

joshessman-llnl:
To improve performance I have switched to a lightweight `TypeList` type for storing parameter packs instead of `std::tuple`.  This approach avoids instantiating constructors, destructors, etc for all the instantiated tuple types and also allows for removal of the workaround required to store `void` in a `std::tuple`.

white238:
Looks great @joshessman-llnl ! could you update the new example that was merged just now w/ the new function types?

https://github.com/LLNL/axom/blob/e5131b35a5dd039f677defec5301a4037480956a/src/axom/inlet/examples/lua_library.cpp

white238 (approved):

PR #427, opened 2021-01-13
Apply generic container verifiers to individual elements for consistency

joshessman-llnl:
# Summary

- This PR is a bugfix
- It does the following:
  - Registers verifiers on a generic container to apply to individual elements
  - Sets the "required" flag on a generic container to apply to individual elements

This may end up being a temporary fix - the use of a single table to define the schema for a user-defined type within a container is useful, but raises the question of whether operations are (or should be) performed on all individual elements of the container versus the container as a whole.

That is, when a generic container is defined, there are two useful tables:
- a "forwarding" table that forwards its operations to all the elements (e.g., when `addInt()` is called, an integer field is added to all container elements)
- the table representing the entire container (this might be useful for defining verifiers that work on the entire table, e.g., a lambda that checks the number of elements in the container)

This PR ensures that the return value from `addGenericContainer` is completely the former, but the latter is still useful, so a future PR should make that table accessible.


src/axom/inlet/Table.hpp
```diff
@@ -1124,6 +1124,23 @@ class Table : public Verifiable<Table>
...
+   * 
+   * \pre The calling table must be a generic container, i.e., isGenericContainer()
+   * returns true
+   * 
+   * \pre The function must accept a single argument of type Table&
+   * 
+   *****************************************************************************
+   */
+  template <typename Func>
+  void forEachContainerElement(Func&& func) const;
```
estebanpauli, line 1142: Even though this is private, it seems odd to take the `func` parameter by r-value reference. I suppose it is easy enough to change later if you ever need to hang on to the functor for longer.

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -147,6 +147,82 @@ TYPED_TEST(inlet_object, simple_array_of_struct_verify_reqd)
...
+    [](const axom::inlet::Table& foo) { return foo.contains("bar"); });
+
+  EXPECT_TRUE(inlet.verify());
+}
+
+TYPED_TEST(inlet_object, simple_array_of_struct_verify_lambda_pass)
+{
+  std::string testString =
+    "foo = { [4] = { bar = true;}, "
+    "        [7] = { bar = false;} }";
```
estebanpauli, line 188: This test would be improved if you had "baz" in one of the examples.

estebanpauli (commented):
Other than the couple of small comments I made, this looks reasonable to me. However, my familiarity with this is low.

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -147,6 +147,82 @@ TYPED_TEST(inlet_object, simple_array_of_struct_verify_reqd)
...
 
+TYPED_TEST(inlet_object, simple_array_of_struct_verify_empty)
+{
+  std::string testString = "foo = { }";
+  DataStore ds;
+  Inlet inlet = createBasicInlet<TypeParam>(&ds, testString);
+
+  auto& arr_table = inlet.addGenericArray("foo");
+  // Even though these are required, the input file should still
+  // "verify" as the array is empty
```
kennyweiss, line 158: That seems a bit unexpected.  

(The comment is helpful here, and I suppose I get the logic that `required` is applied to each struct, and this one has none).

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -147,6 +147,82 @@ TYPED_TEST(inlet_object, simple_array_of_struct_verify_reqd)
...
+{
+  std::string testString =
+    "foo = { [4] = { bar = true;}, "
+    "        [7] = { bar = false;} }";
+  DataStore ds;
+  Inlet inlet = createBasicInlet<TypeParam>(&ds, testString);
+
+  auto& arr_table = inlet.addGenericArray("foo");
+
+  arr_table.addBool("bar", "bar's description").required();
```
kennyweiss, line 175: Just checking -- the following two lines are equivalent to each other, right:
```cpp
   arr_table.addBool("bar", "bar's description").required(); 
   arr_table.addBool("bar", "bar's description").required(true); 
```

and similarly, the following two lines are equivalent to each other:
```cpp
   arr_table.addBool("bar", "bar's description").required(false); 
```

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -147,6 +147,82 @@ TYPED_TEST(inlet_object, simple_array_of_struct_verify_reqd)
...
+    "        [7] = { baz = false;} }";
+  DataStore ds;
+  Inlet inlet = createBasicInlet<TypeParam>(&ds, testString);
+
+  auto& arr_table = inlet.addGenericArray("foo");
+
+  arr_table.addBool("bar", "bar's description");
+  arr_table.addBool("baz", "baz's description");
+
+  // Check for mutual exclusivity
```
kennyweiss, line 197: Please clarify this line a bit -- it seems that you'd like either `bar` or `baz` but not both.

(It's clear from staring at the code for a bit, but it would be clearer if the (arbitrary) condition was explicitly stated).

kennyweiss (approved):
Thanks @joshessman-llnl

src/axom/inlet/Table.cpp
```diff
@@ -164,10 +173,6 @@ Table& Table::addGenericContainer(const std::string& name,
     table.m_sidreGroup->createViewScalar(detail::GENERIC_CONTAINER_FLAG,
                                          static_cast<int8>(1));
   }
-  else
```
estebanpauli, line 167: Should the warning stay if the field is required, or is that checked elsewhere?

src/axom/inlet/Table.cpp
```diff
@@ -437,11 +442,8 @@ struct PrimitiveArrayHelper<Key, bool>
   PrimitiveArrayHelper(Table& table, Reader& reader, const std::string& lookupPath)
   {
     std::unordered_map<Key, bool> map;
-    if(!reader.getBoolMap(lookupPath, map))
-    {
-      SLIC_WARNING(
```
estebanpauli, line 442: See previous comment about whether this is still needed for required fields. Looks like there are a few of these. They are either all okay, or all need to be updated. I am not familiar enough with this code to know which one it is.

src/axom/inlet/Table.hpp
joshessman-llnl: In addition, this is actually a [universal reference](https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers), so it wouldn't preclude a `const` functor from being used with this function.

src/axom/inlet/tests/inlet_object.cpp
joshessman-llnl: Yes - in some cases, like https://github.com/LLNL/axom/blob/ec05a4704cc5b3c15043655cf1b5d49a591d1f9f/src/axom/inlet/tests/inlet_object.cpp#L128-L129

I've explicitly specified it one way or another with the intention of improving readability.

src/axom/inlet/Table.cpp
joshessman-llnl: This is a good question - the `required` flag is checked in the verification phase, which is separate from the `add*` functions.  I've added some extra tests to ensure that required arrays cause verification to fail when they are empty.

estebanpauli (approved):

src/axom/inlet/Table.cpp
```diff
@@ -680,27 +675,43 @@ Table& Table::required(bool isRequired)
 
 bool Table::isRequired() const
 {
+  if(isGenericContainer())
+  {
+    bool result = false;
+    forEachContainerElement([&result](Table& table) {
+      if(table.isRequired()) result = true;
```
white238, line 682: Single line if's are not encouraged in Axom due to the bugs that can be introduced due to them.  Please add curly braces.

white238 (approved):
Thanks for jumping on this quickly @joshessman-llnl ! the new test cases look great.

PR #429, opened 2021-01-14
Add conversion of arrays to std::vector<T>

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Adds a method to retrieve `std::vector`s of containers that ignores indices

Inlet currently uses `std::unordered_map<int, T>` as the primary representation for array types, as some `Reader`s support array discontiguous indexing.  In some cases only the contents of the array is useful (i.e., not its indices), so the added method provides a means of obtaining just the values.  Values are returned in ascending order by index.

This approach is more permissive than one that checks for contiguity and errors out accordingly, but since the user is specifically requesting a vector of the contents, I think it is reasonable to not require contiguity as long as the order of the returned vector is guaranteed.

### DIFF
--- a/src/axom/inlet/Table.hpp
+++ b/src/axom/inlet/Table.hpp
@@ -158,12 +158,14 @@ struct is_inlet_primitive_dict<std::unordered_map<VariantKey, T>>
 };
 
 template <typename T>
-struct is_std_vector : std::false_type
+struct is_primitive_std_vector : std::false_type
 { };
 
 template <typename T>
-struct is_std_vector<std::vector<T>> : std::true_type
-{ };
+struct is_primitive_std_vector<std::vector<T>>
+{
+  static constexpr bool value = is_inlet_primitive<T>::value;
+};
 
 /*!
  *******************************************************************************
@@ -768,7 +770,7 @@ public:
   template <typename T>
   typename std::enable_if<
     !detail::is_inlet_primitive<T>::value && !detail::is_inlet_array<T>::value &&
-      !detail::is_inlet_dict<T>::value && !detail::is_std_vector<T>::value,
+      !detail::is_inlet_dict<T>::value && !detail::is_primitive_std_vector<T>::value,
     T>::type
   get(const std::string& name = "") const
   {
@@ -836,7 +838,7 @@ public:
    *******************************************************************************
    */
   template <typename T>
-  typename std::enable_if<detail::is_std_vector<T>::value, T>::type get() const
+  typename std::enable_if<detail::is_primitive_std_vector<T>::value, T>::type get() const
   {
     // Only allow retrieval of std::vectors from integer-keyed containers
     using Key = int;
@@ -849,7 +851,7 @@ public:
     for(const auto& entry : map)
     {
       // FIXME: 1-indexed arrays??
-      SLIC_ERROR_IF(entry.first >= result.size(),
+      SLIC_ERROR_IF(static_cast<std::size_t>(entry.first) >= result.size(),
                     "[Inlet] Cannot convert to vector due to too-large index");
       result[entry.first] = entry.second;
     }
--- a/src/axom/inlet/Table.hpp
+++ b/src/axom/inlet/Table.hpp
@@ -158,6 +158,14 @@ struct is_inlet_primitive_dict<std::unordered_map<VariantKey, T>>
 };
 
 template <typename T>
+struct is_std_vector : std::false_type
+{ };
+
+template <typename T>
+struct is_std_vector<std::vector<T>> : std::true_type
+{ };
+
+template <typename T>
 struct is_primitive_std_vector : std::false_type
 { };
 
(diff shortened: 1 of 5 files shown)


src/axom/inlet/Table.hpp
```diff
@@ -157,6 +157,24 @@ struct is_inlet_primitive_dict<std::unordered_map<VariantKey, T>>
   static constexpr bool value = is_inlet_primitive<T>::value;
 };
 
+template <typename T>
+struct is_std_vector : std::false_type
+{ };
+
+template <typename T>
+struct is_std_vector<std::vector<T>> : std::true_type
```
estebanpauli, line 165: In addition to the type parameter, std::vector has the allocator parameter. Given that Axom is going to be constructing the vector, its probably fine that it is restricted to the default allocator. It would probably be good to note that here, though. It might also be worth just adding the allocator type as a template parameter.

src/axom/inlet/Table.hpp
```diff
@@ -764,6 +782,51 @@ class Table : public Verifiable<Table>
...
+   * 
+   * \return The values in the retrieved container
+   * 
+   * \tparam T The container type, i.e., T = std::vector<V>
+   * 
+   * \note Elements in the returned array will be in ascending order by index,
+   * regardless of index contiguity or base index
+   *******************************************************************************
+   */
+  template <typename T>
```
estebanpauli, line 799: See previous comment about the allocator parameter for vectors.

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -436,6 +436,40 @@ TYPED_TEST(inlet_object, implicit_conversion_primitives)
...
+  Inlet inlet = createBasicInlet<TypeParam>(&ds, testString);
+
+  // Define schema
+  inlet.addIntArray("arr");
+
+  // Attempt both construction and assignment
+  std::vector<int> expected_arr {4, 6, 10};
+  std::vector<int> arr = inlet["arr"];
+  EXPECT_EQ(arr, expected_arr);
+  arr = inlet["arr"];
```
estebanpauli, line 452: Why is this here a second time? Does this test a separate path than the line above?

src/axom/inlet/Table.hpp
kennyweiss: I don't think we'd want to complicate the interface in that way.

I was actually going to suggest that, in the future, this move from `std::vector` to `axom::Array`, since we're trying to eliminate usage of `std::vector` from our codebase.  Note that axom::Array uses umpire allocations instead of std::allocators.

Tag: @bmhan12

src/axom/inlet/tests/inlet_object.cpp
kennyweiss: @estebanpauli -- Despite the syntax, Line 450 actually uses a copy constructor, rather than the assignment operator.

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -639,6 +673,74 @@ TEST(inlet_object_lua, array_from_bracket)
...
+    "        { bar = false; baz = true} }";
+  DataStore ds;
+  Inlet inlet = createBasicInlet<axom::inlet::LuaReader>(&ds, testString);
+
+  auto& arr_table = inlet.addGenericArray("foo");
+
+  arr_table.addBool("bar", "bar's description");
+  arr_table.addBool("baz", "baz's description");
+  std::vector<Foo> expected_foos = {{true, false}, {false, true}};
+  auto foos = inlet["foo"].get<std::vector<Foo>>();
```
kennyweiss, line 740: This is really nice!

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -639,6 +673,74 @@ TEST(inlet_object_lua, array_from_bracket)
...
+
+  // Define schema
+  inlet.addIntArray("arr");
+
+  // Attempt both construction and assignment
+  std::vector<int> expected_arr {4, 6, 10};
+  std::vector<int> arr = inlet["arr"];
+  EXPECT_EQ(arr, expected_arr);
+  arr = inlet["arr"];
+  EXPECT_EQ(arr, expected_arr);
```
kennyweiss, line 690: Minor: For at least one of these tests, it might be good to show the other way of accessing these using a map. Or perhaps to add a comment that the array getter removes the indices.

kennyweiss (approved):
Thanks @joshessman-llnl -- this is a nice contribution!

I added a note below about the possibility of changing this down the road from `std::vector` to `axom::Array`, which mostly be a drop-in replacement for `std::vector`, with some differences related to memory allocation/management/semantics.

Note that `axom::Array` is not yet ready, so my comments are meant to be informative rather than actionable, and the changes will be pretty straightforward.

src/axom/inlet/tests/inlet_object.cpp
joshessman-llnl: Copy construction vs. copy assignment is a slightly different overload resolution context that has caused problems with these implicit conversions in the past, so this test checks both to avoid regressions.

estebanpauli: Sorry, I meant a separate path for the conversion from the result of inlet["arr"] (a Proxy, I believe), to something that the vector can be constructed/assigned from. I guess it is important to test both though, since as you point out, they are not the same.

white238 (approved):

white238:
Thanks @joshessman-llnl !

PR #431, opened 2021-01-19
Update sparsehash headers to v. 2.0.4

nselliott:
# Summary

- This PR is a third-party library update
  - Brings in headers from sparsehash v. 2.0.4.  The code changes come from that library, not from our team.
  - Addresses comment from @rhornung67 in issue #365 but does not address the original user issue.


rhornung67 (approved):
Looks good. Please add an entry for this in the release notes file.

agcapps (approved):
Thanks @nselliott .

kennyweiss:
@nselliott  -- will we still need the following patch -- `sparsehash-2.0.3-no-tr1.patch`  ?
Or should it be removed as part of this PR?

See: https://github.com/LLNL/axom/tree/develop/src/thirdparty/patches

nselliott:
While I'm in the release notes, I'll also add an entry for the recent conduit update, since I neglected to do that at the time.

kennyweiss (approved):

nselliott:
> 
> 
> @nselliott -- will we still need the following patch -- `sparsehash-2.0.3-no-tr1.patch` ?
> Or should it be removed as part of this PR?
> 
> See: https://github.com/LLNL/axom/tree/develop/src/thirdparty/patches

@kennyweiss I think that patch was made obsolete by this commit.  https://github.com/LLNL/axom/commit/053e33c6604cfb0873ba162c9ca08c4ddefd3adb ;  I believe before you added this, we were finding the tr1 header by default even when the standard `<functional>` was present.  Now C++ 11 and later compilers should always find the `<functional>` header and never need to worry about the tr1 stuff.  Does that sound correct @agcapps ?

kennyweiss:
Thanks @nselliott.  If so, I'd suggest removing that patch file.

agcapps:
> 
> 
> > @nselliott -- will we still need the following patch -- `sparsehash-2.0.3-no-tr1.patch` ?
> > Or should it be removed as part of this PR?
> > See: https://github.com/LLNL/axom/tree/develop/src/thirdparty/patches
> 
> @kennyweiss I think that patch was made obsolete by this commit. [053e33c](https://github.com/LLNL/axom/commit/053e33c6604cfb0873ba162c9ca08c4ddefd3adb) ; I believe before you added this, we were finding the tr1 header by default even when the standard `<functional>` was present. Now C++ 11 and later compilers should always find the `<functional>` header and never need to worry about the tr1 stuff. Does that sound correct @agcapps ?

That does sound correct.

PR #433, opened 2021-01-20
Add support to save off a QuadratureFunction to a MFEMSidreDataCollection

rcarson3:
# Summary

- This PR is a feature
- It does the following:
  - Adds the ability to save off `mfem::QuadratureFunction` to MFEMSidreDataCollection
  - Adds necessary details to allow the QuadratureFunction to be reloaded
  - Adds a unit test to ensure QuadratureFunctions can be saved off and then later restored

This PR does require a version of MFEM with at least this commit within it: https://github.com/mfem/mfem/commit/a8ebd745a85770330a691ffd15e78aaf2962b223

# Potential further expansions of this new API

- An additional function could be added to reconstruct an mfem::QuadratureSpace for a given saved off QuadratureFunction. Currently, this would require a user to supply a mfem::mesh object that corresponds to the saved off data. However, it seems that with #379 we should have all the information necessary to do this for the user. 
- An additional function could be added to reconstruct a mfem::QuadratureFunction for a user based on the saved off data.


src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1108,6 +1110,231 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
...
+      return;
+   }
+
+   // Register field_name in the blueprint group.
+   sidre::Group* f = m_bp_grp->getGroup("qfields");
+
+   if (f->hasGroup( field_name ))
+   {
+      // There are two possibilities:
+      // 1. If HasField(field_name) is true - we are overwriting a field that
```
kennyweiss, line 1216: typo in comment: `HasField` -> `HasQField`

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1108,6 +1110,231 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
...
+                         "registered, overwriting the old qfield");
+         }
+#endif
+         DeregisterQField(field_name);
+      }
+   }
+
+   sidre::Group* grp = f->createGroup( field_name );
+
+   // Set the "qspace_order" string using the qf's qspace, overwrite if
```
kennyweiss, line 1237: TODO: Please add a comment here with the schema for a QField.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1108,6 +1110,231 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
...
+   }
+
+   // Delete field_name from the named_buffers group, if allocated.
+   FreeNamedBuffer(field_name);
+}
+
+int MFEMSidreDataCollection::GetQFieldOrder(const std::string &field_name)
+{
+   sidre::Group* grp = m_bp_grp->getGroup("qfields/" + field_name);
+   SLIC_ASSERT_MSG(grp != NULL,
```
kennyweiss, line 1309: Note for the future: We should look into how error checking / reporting is done in this class and make sure we are consistent about runtime check vs. ASSERT (debug-only) checks.

kennyweiss (commented):
Thanks @rcarson3 -- overall, this looks like a nice addition to me.  

I'd like to hear back from @cyrush about how this will work with blueprint and VisIt before approving.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -242,6 +242,7 @@ void MFEMSidreDataCollection::createMeshBlueprintStubs(bool hasBP)
     m_bp_grp->createGroup("coordsets");
     m_bp_grp->createGroup("topologies");
     m_bp_grp->createGroup("fields");
+    m_bp_grp->createGroup("qfields");
   }
```
cyrush, line 246: `qfields` is not a part of the blueprint, so this part of the tree won't be recognized elsewhere. 

(It won't cause any blueprint verify errors, but it won't be checked for, or consumed)

Can we have the quadrature funcs be a field with a specific basis type?

cyrush: or otherwise an extension to a field?

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1108,6 +1110,231 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
...
+   sidre::Group *bp_field_grp = m_bp_grp->getGroup("qfields/" + field_name);
+   sidre::Group *bp_index_field_grp =
+      m_bp_index_grp->createGroup("qfields/" + field_name);
+
+   bp_index_field_grp->createViewString( "path", bp_field_grp->getPathName() );
+   bp_index_field_grp->copyView( bp_field_grp->getView("topology") );
+   bp_index_field_grp->copyView( bp_field_grp->getView("vdim") );
+   bp_index_field_grp->copyView( bp_field_grp->getView("qspace_order") );
+
+   // Fix me?: Should this be qf->GetVDim() instead?
```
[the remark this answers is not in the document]
rcarson3: So, I do have at least one application where I would have vector-valued qfields (more for restart capabilities than anything else) rather than a case where you might have those vector fields saved off as individual components.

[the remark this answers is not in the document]
kennyweiss: Thanks @rcarson3 -- restart functionality is independent of the blueprint index.

From sidre/hdf5's perspective the data looks like an array of doubles, so everything should work as long as you can initialize an `mfem::QuadratureFunction` from this data.

Currently, the blueprint index is only used by VisIt (and possibly ascent ?). So I think the `vdim` is not currently used. Presumably, we'd want this to be used in the future, which is why I wanted feedback from @cyrush.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
rcarson3: So, quadrature functions I, personally, would consider separate from a field variable / basis function. Since, they exist directly at the quadrature points and don't depend on a basis type or field/nodes. For example, if you look at how `mfem` creates them they only require a mesh, the integration order, and vector dimension of the quadrature function in order to be defined. Based on this, do you have an idea for what might be the best to handle such a thing within blueprint?

cyrush: In blueprint speak, a `field` is something that loosely encapsulates values  associated with a mesh topology. That still seems appropriate for a quadrature function, or that a quadrature function could be a new variant of a blueprint field.

One way to think about this, is how would delineate between this new construct in the overall structure: 

https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html 

Many of the same relationships that represent a field would have to be duplicated. Things like index generation would need to be duplicated, vs reused automatically.

rcarson3: Okay, I think I'm starting to get what you're talking about. 

Then as far as the blueprint is concern, it might be possible to have it think of a `QuadratureFunction` as an L2 `GridFunction` with a `Gauss-Legendre` basis type, mfem::Ordering::Type::ByVDIM, and of sufficiently high enough order so its nodes correspond to the quadrature point locations of the `QuadratureFunction`. As long as internally, it doesn't try to project those points to a mesh's nodal locations, since you could have variables in a `QuadratureFunction` that need to remain non-negative. I guess even then you might want to have a flag associated with that field to denote it as a quadrature function.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1108,6 +1110,231 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
   DataCollection::RegisterField(field_name, gf);
 }
 
+// private method
+void MFEMSidreDataCollection::
+addQuadratureFunction(const std::string &field_name, mfem::QuadratureFunction *qf,
+                     const std::string &buffer_name,
+                     axom::sidre::IndexType offset)
+{
+   sidre::Group* grp = m_bp_grp->getGroup("qfields/" + field_name);
```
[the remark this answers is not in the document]
rcarson3: Used your suggestion here and a few other places that had the same line of code.

src/axom/sidre/core/MFEMSidreDataCollection.hpp
```diff
@@ -259,6 +259,44 @@ class MFEMSidreDataCollection : public mfem::DataCollection
...
+                      const std::string &buffer_name,
+                      axom::sidre::IndexType offset);
+   
+   /// Returns the QuadratureSpace order useful for reconstructing a QuadratureFunction
+   int GetQFieldOrder(const std::string &field_name);
+
+   /// Returns the QuadratureSpace VDim useful for reconstructing a QuadratureFunction
+   int GetQFieldVDim(const std::string &field_name);
+
+   /// Returns a pointer to the raw data usefol for reconstructing a QuadratureFunction
```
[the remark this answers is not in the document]
rcarson3: fixed

src/axom/sidre/core/MFEMSidreDataCollection.cpp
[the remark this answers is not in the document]
rcarson3: I swapped it over to `qf->GetVDim()`, but we can revert it back if it's determined that's not the best way to go.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
rcarson3: Fixed

src/axom/sidre/core/MFEMSidreDataCollection.cpp
rcarson3: I took a shot at this, but I have a feeling we'll want to iterate on the description of it a bit more.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1108,6 +1110,231 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
...
+    */
+
+   // Make sure we have the View "values".
+   sidre::View *vv = alloc_view(grp, "values");
+
+   // Describe and apply the "values" View.
+   // If the data store has buffer for field_name (e.g. AllocNamedBuffer was
+   // called, or it was loaded from file), use that buffer.
+   sidre::View *bv = named_buffers_grp()->getView(buffer_name);
+   if (bv)
```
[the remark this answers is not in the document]
rcarson3: Updated this based on your comment. Let me know if the new code is what you had in mind.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
cyrush: Yes -- if we can encode it in some how via a `basis` string name, thats great -- if not we can always add another key/value to specifically identify the QuadratureFunction case.

rcarson3: @cyrush so I brought this up with the MFEM team just to see if they had any thoughts on the topic.

@v-dobrev had this to say:

> In terms of blueprint, a `QuadratureFunction` is another way of defining a field which is different from the way `GridFunctions` are defined. The information to store for a `QuadratureFunction` should be essentially what we save in `QuadratureFunction::Save` — `QuadratureSpace::Save` + `VDim` + data. Note that in the future, `QuadratureSpace::Save` will probably support types other than `Type: default_quadrature`.

@joshessman-llnl then suggested just calling the basis: `QF_Default_[ORDER]` which I like due to it being simple.

Based on their response, let me know what you think of the proposed `basis` string name.

cyrush: that sounds great to me, for folks looking to a reconstruct  QuadFunc, QuadSpac  - if we can have a way to pass that string like we do to mfem for a FiniteElementCollection, that would complete the circuit.

rcarson3: @cyrush I'll work on the necessary PR for that within MFEM and link the PR here. Also, I think if we wanted to properly encode everything within the basis string like its done in the `FiniteElementCollection` the basis string here should be `QF_Default_[ORDER]_[VDIM]`. So, the `QuadratureSpace` would be able to take that name in and just make use of the `QF_Default_[ORDER]` part and supplied mesh to reconstruct itself. The `QuadratureFunction` could take the name in and just use the `[VDIM]` part as long as it's supplied an appropriate `QuadratureSpace`.

Nonetheless, I'll work on updating everything for this PR over from being `qfields` to `fields` for the underlying `blueprint`.

cyrush: Sounds good!

rcarson3:
So, I've updated things now so that `qfields` is now `fields`, and we now encode all of the relevant information to reconstruct things within a basis string that looks like: `QF_Default_[ORDER]_[VDIM]`.

I still need to merge develop into here and update the `reconstructFields` to play nice with QFs. I don't think this should be too difficult.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1108,6 +1108,322 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
   DataCollection::RegisterField(field_name, gf);
 }
 
+// private method
+void MFEMSidreDataCollection::
+addQuadratureFunction(const std::string &field_name, mfem::QuadratureFunction *qf,
```
joshessman-llnl, line 1113: This looks very similar to `addScalarBasedGridFunction` - I wonder if there's an opportunity to combine the two given that both `mfem::GridFunction` and `mfem::QuadratureFunction` inherit from `mfem::Vector`.  It looks like for the most part both of these functions only use the `mfem::Vector` interface.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1108,6 +1108,322 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
...
+   SLIC_ASSERT_MSG(numDofs == 0 || vv->getData() == qf->GetData(),
+                   "View data is different from QuadratureFunction data");
+   SLIC_ASSERT_MSG(vv->getNumElements() == numDofs,
+                   "View size is different from QuadratureFunction size");
+}
+
+// private method
+// Should only be called on mpi rank 0 ( or if serial problem ).
+void MFEMSidreDataCollection::
+RegisterQFieldInBPIndex(const std::string& field_name, mfem::QuadratureFunction *qf)
```
joshessman-llnl, line 1179: This looks pretty similar to `RegisterFieldInBPIndex` - it looks like the only thing needed here from the MFEM object (gridfunc/qfunc) is the vector dimension, so maybe that could replace the second parameter:

```cpp
RegisterFieldInBPIndex(const std::string& field_name, const int num_components)
```

src/axom/sidre/core/MFEMSidreDataCollection.hpp
```diff
@@ -259,6 +259,44 @@ class MFEMSidreDataCollection : public mfem::DataCollection
...
+                      axom::sidre::IndexType offset);
+   
+   /// Returns the QuadratureSpace order useful for reconstructing a QuadratureFunction
+   int GetQFieldOrder(const std::string &field_name);
+
+   /// Returns the QuadratureSpace VDim useful for reconstructing a QuadratureFunction
+   int GetQFieldVDim(const std::string &field_name);
+
+   /// Returns a pointer to the raw data useful for reconstructing a QuadratureFunction
+   double* GetQFieldData(const std::string &field_name);
```
joshessman-llnl, line 298: Is the intent that these will be going away/moved once the reconstruction logic is finished?  I'd be hesitant to add them to the public interface of the class.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1108,6 +1108,322 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
...
+      std::string basis_name = v->getString();
+
+      size_t pos = 0;
+      std::string delimiter = "_";
+
+      // If the DC was reloaded the internal q_field_map was not restored
+      // so we can't just use HasQField here. We should fix this in the  Load
+      // function. 
+      if (basis_name.substr(0, basis_name.find(delimiter)) != "QF")
+      {
```
joshessman-llnl, line 1330: It might be useful to implement something like [mfem::FiniteElementCollection::new](https://mfem.github.io/doxygen/html/classmfem_1_1FiniteElementCollection.html#a15fcfa553d4949eb08f9926ac74d1e80) for this.  It probably belongs in MFEM but it could be temporarily kept here.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1108,6 +1108,322 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
...
+        SLIC_ERROR("field name provided was not for a QF");
+        return -1;
+      }
+      // We know our basis string should be something like:
+      // QF_Default_[ORDER]_[VDIM]
+      pos = basis_name.find(delimiter);
```
joshessman-llnl, line 1336: I added some similar logic as a function in https://github.com/LLNL/axom/pull/473/commits/12d6100cd8c74a3894a0124c4ae97300d4faadd7, if that looks helpful it could possibly be `cherry-pick`ed onto this branch.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1108,6 +1108,322 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
...
+
+      if (basis_name.substr(0, basis_name.find(delimiter)) != "QF")
+      {
+        SLIC_ERROR("field name provided was not for a QF");
+        return nullptr;
+      }
+
+      v = grp->getView("values");
+      void* vptr = v->getVoidPtr();
+      return static_cast<double*>(vptr) + v->getOffset();
```
joshessman-llnl, line 1418: Minor:  This might be a little clearer (and also maybe more type-safe, I think the access through `conduit::Value` is type-checked):
```suggestion
      return v->getData();
```

src/axom/sidre/tests/sidre_mfem_datacollection.cpp
```diff
@@ -127,6 +127,84 @@ TEST(sidre_datacollection, dc_reload)
...
+      (*qv)(2*i+1) = double(Nq - i - 1);
+  }
+
+  sdc_writer.SetPrefixPath("/tmp/dc_qf_test");
+  sdc_writer.SetCycle(5);
+  sdc_writer.SetTime(8.0);
+  sdc_writer.Save();
+
+  MFEMSidreDataCollection sdc_reader(COLL_NAME);
+  sdc_reader.SetPrefixPath("/tmp/dc_qf_test");
```
joshessman-llnl, line 173: Minor: To avoid potential issues on Windows, the new "convention" with these tests is to use the working directory (default prefix path) + the name of the test:
```suggestion
  MFEMSidreDataCollection sdc_reader(testName());
```

src/axom/sidre/tests/sidre_mfem_datacollection.cpp
```diff
@@ -127,6 +127,84 @@ TEST(sidre_datacollection, dc_reload)
...
+  // structure.
+  mfem::QuadratureFunction *qs = new mfem::QuadratureFunction(qspace, nullptr, qs_vdim);
+  mfem::QuadratureFunction *qv = new mfem::QuadratureFunction(qspace, nullptr, qv_vdim);
+
+  int Nq = qs->Size();
+
+  // The mesh and field(s) must be owned by Sidre to properly manage data in case of
+  // a simulated restart (save -> load)
+  bool owns_mesh = true;
+  MFEMSidreDataCollection sdc_writer(COLL_NAME, mesh, owns_mesh);
```
joshessman-llnl, line 153:
```suggestion
  MFEMSidreDataCollection sdc_writer(testName(), mesh, owns_mesh);
```

src/axom/sidre/tests/sidre_mfem_datacollection.cpp
```diff
@@ -127,6 +127,84 @@ TEST(sidre_datacollection, dc_reload)
...
+
+  // The data needs to be instantiated before we save it off
+  for (int i = 0; i < Nq; ++i)
+  {
+      (*qs)(i) = double(i);
+      (*qv)(2*i+0) = double(i);
+      (*qv)(2*i+1) = double(Nq - i - 1);
+  }
+
+  sdc_writer.SetPrefixPath("/tmp/dc_qf_test");
```
joshessman-llnl, line 167:
```suggestion
```

src/axom/sidre/tests/sidre_mfem_datacollection.cpp
```diff
@@ -127,6 +127,84 @@ TEST(sidre_datacollection, dc_reload)
   EXPECT_TRUE(sdc_reader.verifyMeshBlueprint());
 }
 
+TEST(sidre_datacollection, dc_qf_reload)
+{
+  //Set up a small mesh and a couple of grid function on that mesh
+  mfem::Mesh *mesh = new mfem::Mesh(2, 3, mfem::Element::QUADRILATERAL, 0, 2.0, 3.0);
+  mfem::FiniteElementCollection *fec = new mfem::LinearFECollection;
+  mfem::FiniteElementSpace *fespace = new mfem::FiniteElementSpace(mesh, fec);
```
joshessman-llnl, line 135: Minor: Is the manual memory management with `new`/`delete` needed here?

joshessman-llnl (commented):
Looking good so far @rcarson3!
Just a few minor comments on style and possible reuse of existing code.

rhornung67:
@rcarson3 we're planning for an Axom release in about a month. Should this be finished up and merged before that to get in the release?

rcarson3:
@rhornung67 sorry for going silent for the past month. The application code I was working this PR for ended up wanting another feature working for a deadline. So, I've been focusing my attention recently on trying to meet their needs. I can probably work this a bit over a few nights this week nights and week end to address all the various comments and concerns. Although, depending on people's thoughts this PR could require a companion PR within MFEM just to make somethings nicer within here.

joshessman-llnl:
After some discussion with @rcarson3 I plan on taking over this PR and putting the finishing touches on it.

src/axom/core/utilities/StringUtilities.cpp
```diff
@@ -39,6 +39,16 @@ void toUpper(std::string& str)
...
 
+bool checkedConvertToInt(const std::string& number, int& result)
+{
+  // Use the C versions to avoid the exceptions
+  // thrown by std::stoi on conversion failure
+  // FIXME: Switch to std::from_chars when C++17 is available
+  char* ptr;
+  result = strtol(number.c_str(), &ptr, 10);
+  return *ptr == 0;
+}
```
cyrush, line 50: fyi: Conduit already provides both a way to check if a string is a number and a method that does the conversion in its `utils`. 

https://github.com/LLNL/conduit/blob/f9b61cbadd270dd1ef8c7987e63e3f509951c768/src/libs/conduit/conduit_utils.cpp#L956

https://github.com/LLNL/conduit/blob/f9b61cbadd270dd1ef8c7987e63e3f509951c768/src/libs/conduit/conduit_utils.hpp#L443

joshessman-llnl: Thanks for the suggestion - always good to get rid of duplicated functionality!

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1206,13 +1277,100 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
...
+  // <m_bp_grp>/fields/<field_name>/basis        = string that encodes the QF
+  //                                               integration order and vector dimension
+  //                                               "QF_Default_[ORDER]_[VDIM]"
+
+  // Set the "basis" string using the qf's order and vdim, overwrite if
+  // necessary.
+  sidre::View* v = alloc_view(grp, "basis");
+  // In the future, we should be able to have mfem::QuadratureFunction provide us this name.
+  const std::string basis_name =
+    fmt::format("QF_Default_{0}_{1}", qf->GetSpace()->GetOrder(), qf->GetVDim());
```
rcarson3, line 1347:
```suggestion
    fmt::format("QF_Default_{0}_{1}", qf->GetSpace()->GetElementIntRule(0).GetOrder(), qf->GetVDim());
```

This should be all that's needed to run with MFEM v4.2 and get around the newer getter function that was added in between MFEM 4.2 and 4.3.

I've tried compiling this locally on my copy of axom. However, I've been running into some issues with it not picking up my copy of hdf5 :/ after my system updated, which has caused the build to fail earlier on...

cyrush (approved):

white238 (approved):

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1206,13 +1277,103 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
   // Register field_name in the blueprint_index group.
   if(myid == 0)
   {
-    RegisterFieldInBPIndex(field_name, gf);
+    RegisterFieldInBPIndex(field_name, gf->VectorDim());
```
v-dobrev, line 1280: Note that there is a difference between `gf->VectorDim()` and `gf->FESpace()->GetVDim()`. For example, for FE spaces using `mfem::ND_FECollection`, `gf->VectorDim()` will be the same as the `mfem::Mesh::SpaceDimension()` while `gf->FESpace()->GetVDim()` will be 1.

I'm not sure what the intent here is -- I just wanted to point out a potential issue.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -35,6 +36,75 @@ const std::string MFEMSidreDataCollection::s_attribute_suffix =
...
+}
+
+/**
+ * @brief Implements an analogue to mfem::FiniteElementCollection::New
+ * for mfem::QuadratureSpaces - basis currently must be of form QF_Default_[ORDER]_[VDIM]
+ * @param[in] name The name that encodes the QuadratureSpace data
+ * @param[in] mesh The mesh to construct the QuadratureSpace with
+ * @param[out] vdim The vector dimension parsed from the basis name - unlike FEColl/FESpace,
+ * QSpaces do not contain vdim information, but it may be useful for the caller when constructing
+ * the QFunc
```
v-dobrev, line 79: Note that the `vdim` for an `mfem::QuadratureFunction` is stored in the `mfem::QuadratureFunction` itself and not in the `mfem::QuadratureSpace`. This way different `mfem::QuadratureFunction`s with different `vdim` can use the same `mfem::QuadratureSpace`.

This is in contrast to `mfem::(Par)GridFunction` where `vdim` is part of the `mfem::(Par)FiniteElementSpace`.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
joshessman-llnl: This logic is mirrored from
https://github.com/mfem/mfem/blob/96fc6f177ea8f2174676e7dc6d5c9ec1a08a4d60/fem/sidredatacollection.cpp#L988-L991

but given the ambiguity I will add that comment to the call site here.

joshessman-llnl (approved):

src/axom/sidre/core/MFEMSidreDataCollection.cpp
rcarson3: @v-dobrev so for the basis string I go over the logic a bit here: https://github.com/LLNL/axom/pull/433#discussion_r577936473 . So, the string itself encodes information related to the quadrature space and quadrature function vdim. I believe the logic here for this function is that since you're already parsing the basis string to reconstruct the `QuadratureSpace` you ought as well return what the VDim is for the `QF`  to the user in case the user wants it.

v-dobrev: Okay, I just wanted to point out this small difference in handling `vdim` in `mfem::QuadratureFunction` compared to `mfem::(Par)GridFunction`. I guess you already new that. :-)

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1206,13 +1274,106 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
...
 
+void MFEMSidreDataCollection::RegisterQField(const std::string& field_name,
+                                             mfem::QuadratureFunction* qf,
+                                             const std::string& buffer_name,
+                                             axom::sidre::IndexType offset)
+{
+  if(field_name.empty() || buffer_name.empty() || qf == NULL ||
+     qf->GetSpace() == NULL)
+  {
+  #ifdef AXOM_DEBUG
```
kennyweiss, line 1295: This message should be split into a few cases since the current message doesn't adequately describe what happened in all cases.

src/axom/sidre/core/MFEMSidreDataCollection.hpp
```diff
@@ -256,6 +254,37 @@ class MFEMSidreDataCollection : public mfem::DataCollection
...
+    */
+  virtual void RegisterQField(const std::string& field_name,
+                              mfem::QuadratureFunction* qf)
+  {
+    RegisterQField(field_name, qf, field_name, 0);
+  }
+
+  /// Register a mfem::QuadratureFunction in the Sidre DataStore.
+  /** The registration procedure is as follows:
+       - if (@a qf's data is NULL), error out since this means there's no quadrature
```
kennyweiss, line 269: Please improve the docs here. `qf`'s data is `NULL` does not exactly match the check in the code:

```cpp
  if(field_name.empty() || buffer_name.empty() || qf == NULL || qf->GetSpace() == NULL)
```
It seems to me that the situation is that `qf` is null or doesn't have an associated quadrature space.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1206,13 +1274,106 @@ void MFEMSidreDataCollection::RegisterField(const std::string& field_name,
...
   // Register field_name + gf in field_map.
   DataCollection::RegisterField(field_name, gf);
 }
 
+void MFEMSidreDataCollection::RegisterQField(const std::string& field_name,
+                                             mfem::QuadratureFunction* qf,
+                                             const std::string& buffer_name,
+                                             axom::sidre::IndexType offset)
+{
+  if(field_name.empty() || buffer_name.empty() || qf == NULL ||
```
kennyweiss, line 1292: Style: Please convert `NULL` to `nullptr`.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -2056,13 +2221,28 @@ void MFEMSidreDataCollection::reconstructFields()
...
-      // FiniteElementCollection
-      auto basis_name = field_grp->getView("basis")->getString();
-      m_fecolls.emplace_back(mfem::FiniteElementCollection::New(basis_name));
+      int vdim = 1;  // The default
+      // GridFunction vs QuadratureFunction
+      bool is_gridfunc = true;
+      // FiniteElementCollection/QuadratureSpace construction
+      const std::string basis_name = field_grp->getView("basis")->getString();
+      // Check if it's a QuadratureFunction
+      if(basis_name.find("QF") == 0)
```
kennyweiss, line 2230: There's no reason to assume that we could not have a legitimate grid function with the characters `QF` in its name.
Is there a better check we could do?

Is this the only place that attempts to check names to see the type? Perhaps this should be moved to a function?

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -2056,13 +2221,28 @@ void MFEMSidreDataCollection::reconstructFields()
...
+      {
+        // The vdim is being overwritten here with the value in the basis string so we
+        // can correctly construct the QuadratureFunction
+        m_quadspaces.emplace_back(
```
kennyweiss, line 2234: Potential future optimization: Should `m_quadspaces` be a map from string name to `QuadratureSpace`? 
(This would need to be unique for each mesh, but `MFEMSidreDataCollection` is only over a single mesh)

kennyweiss (approved):
This is really nice @rcarson3 and @joshessman-llnl !

RELEASE-NOTES.md
```diff
@@ -46,6 +46,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   delimiter characters
 - Inlet: Added a method to `inlet::Inlet` that retrieves the set of unexpected names
   in the input file
+- Added support for QFunctions to `MFEMSidreDataCollection`
```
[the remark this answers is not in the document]
joshessman-llnl: The comment about this requiring a specific version of MFEM is no longer accurate - this PR uses a workaround (that we can revisit/rip out once Axom bumps its MFEM version)

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -35,6 +36,75 @@ const std::string MFEMSidreDataCollection::s_attribute_suffix =
...
+ * @param[in] mesh The mesh to construct the QuadratureSpace with
+ * @param[out] vdim The vector dimension parsed from the basis name - unlike FEColl/FESpace,
+ * QSpaces do not contain vdim information, but it may be useful for the caller when constructing
+ * the QFunc
+ */
+mfem::QuadratureSpace* NewQuadratureSpace(const std::string& name,
+                                          Mesh* mesh,
+                                          int& vdim)
+{
+  const auto tokens = splitLastNTokens(name, 4, '_');
```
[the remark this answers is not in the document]
joshessman-llnl: There is not, it was invented for the purposes of this PR (after some discussion with the MFEM folks), though once this goes in the logic will be moved to MFEM proper.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
joshessman-llnl: I may be misunderstanding the comment, but this code is checking the basis name - in which case we're only worried about collisions with the strings documented here: https://mfem.github.io/doxygen/html/classmfem_1_1FiniteElementCollection.html#a15fcfa553d4949eb08f9926ac74d1e80 

as opposed to the name assigned to the GridFunction

src/axom/sidre/core/MFEMSidreDataCollection.cpp
joshessman-llnl: m_quadspaces (like its counterpart m_fespaces) is just used for managing ownership - its members are never accessed directly

src/axom/sidre/tests/sidre_mfem_datacollection.cpp
```diff
@@ -237,6 +237,76 @@ TEST(sidre_datacollection, dc_reload_mesh)
...
+
+  MFEMSidreDataCollection sdc_reader(testName());
+#if defined(AXOM_USE_MPI) && defined(MFEM_USE_MPI)
+  sdc_reader.SetComm(MPI_COMM_WORLD);
+#endif
+  sdc_reader.Load(sdc_writer.GetCycle());
+
+  ASSERT_TRUE(sdc_reader.HasQField("qs"));
+  ASSERT_TRUE(sdc_reader.HasQField("qv"));
+  mfem::QuadratureFunction* reader_qs = sdc_reader.GetQField("qs");
```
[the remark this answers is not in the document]
joshessman-llnl: The QuadratureSpace is created/owned by the DataCollection (and stored in m_qspaces).

This approach is analogous to the one I used for reconstruction GridFunctions, in that the DataCollection will construct/own the FEColl and FESpace objects that make up the GridFunction.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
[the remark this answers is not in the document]
joshessman-llnl: I have the current format in the Doxygen annotation for this function as well:

https://github.com/LLNL/axom/blob/614075596fe967d574632b3a0cb1be525ff3edd7/src/axom/sidre/core/MFEMSidreDataCollection.cpp#L73-L74

src/axom/sidre/core/MFEMSidreDataCollection.cpp
joshessman-llnl: This check is analogous to this existing snippet for GridFunctions (with the exception of NULL vs nullptr):

https://github.com/LLNL/axom/blob/614075596fe967d574632b3a0cb1be525ff3edd7/src/axom/sidre/core/MFEMSidreDataCollection.cpp#L1202-L1206

The above case doesn't even provide a message, so I'll go ahead and put some more helpful messages in both places

src/axom/sidre/core/MFEMSidreDataCollection.cpp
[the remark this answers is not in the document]
kennyweiss: Thanks for pointing it out @joshessman-llnl -- I missed it. 
That's perfect.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
kennyweiss: Thanks for clarifying -- you're right, if this is only on the name, and QF is not in `FECollection::Name()`, then this works.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
kennyweiss: I was thinking more in terms of memory savings.
(I haven't looked into the storage costs for `QuadratureSpace`, but `FiniteElementSpace` contains several indexing several arrays)

rcarson3: @kennyweiss I was thinking of doing something along the lines of what you were talking about for FES and QS in a follow-up PR that would be more memory optimization focused. I know for my ECP application that'll be a useful feature to have, since all my data shares the same QS and most of the FES are also shared as well.

joshessman-llnl: Created an issue for this - #533

PR #426, opened 2021-01-13
Expose a wrapped version of Primal's vector class for use in Lua input files

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Replaces vector component expansion (vec -> x,y,z) in Lua functions with a Lua usertype that supports basic vector operations
  - Additionally, resolves a [rule of three](https://en.cppreference.com/w/cpp/language/rule_of_three) warning in Primal due to a user-defined copy ctor but not copy `op=`

PR #434, opened 2021-01-20
Add support for deeply nested containers of structs in Inlet

joshessman-llnl:
# Summary

- This PR is a refactoring/feature
- It does the following:
  - Adds support for deeply nested containers of user-defined types in input files
  - Adds a new example demonstrating the use of deeply nested containers

2/3s of the PR by LoC is the new example and new tests.  The remaining 1/3 is a few changes to the underlying structure of the Table hierarchy and the information contained in the datastore:
* The container indices stored in the datastore are now absolute instead of relative
* References to "child" Tables are stored in a new Table member variable when the nesting is indirect, that is, when a table has nested children corresponding to array elements but is not the internal container group (`_inlet_container`)
* The checked string -> integer conversion function has the signature `bool(string, int&)` instead of `pair<int, bool>(string)` to avoid ambiguity in the return type
* A new `Table::addStruct` method was added as a replacement for `Table::addTable` when defining a nested structure.  A follow-up PR will convert `addTable` to a `private` method and replace it with `addStruct` in the remaining examples/tests to avoid "polluting" this PR with a bunch of identical changes.


src/axom/inlet/Inlet.hpp
```diff
@@ -119,6 +119,23 @@ class Inlet
    */
   Table& addTable(const std::string& name, const std::string& description = "");
 
+  /*!
+   *****************************************************************************
+   * \brief Add a structure to the input file schema.
+   *
+   * Adds a structure/record to the input file schema. Structures can contain
+   * fields and/or substructures.  By default, it is not required unless marked with
```
estebanpauli, line 127: Does it also support lists? I'm not familiar enough with the usage of "field" here to know for sure, but it might be clear from the broader context of the class. I don't know if a change is needed to the documentation, but I just want to make sure lists are supported.

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -0,0 +1,268 @@
...
+{
+  Vector result;
+  // Make sure the elements are accessed in order
+  for(int i = 0; i < 3; i++)
+  {
+    const auto ele = map.find(i);
+    if(ele != map.end())
+    {
+      result[i] = ele->second;
+    }
```
estebanpauli, line 30: Should it be an error if the element is missing? I'm not sure what the expected usage of this is, so I can't tell for sure. Noticing that it's in an example, though, it might not really matter.

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -0,0 +1,268 @@
...
+      result.translate = mapToVector(base["translate"]);
+    }
+    else if(base.contains("rotate"))
+    {
+      result.type = Operator::Type::Rotate;
+      result.rotate = base["rotate"];
+      result.axis = mapToVector(base["axis"]);
+      result.center = mapToVector(base["center"]);
+    }
+    else if(base.contains("x") || base.contains("y") || base.contains("z"))
```
estebanpauli, line 87: FWIW, in my example, "slice" was a nested dictionary:
```
operators:
   translate: [1, 2, 3]
   slice:
      x: 10
```
The idea is that you can slice along an x, y, or z plane. There are also other options available. For example, to tell it which direction should be "up" in the new coordinate system:
```
slice:
   x: 10
   up: [1, 2]
```

If you think this could trip up the implementation, please make sure that the tests cover situations like these.

estebanpauli: Having seen your example below, it looks like maybe this should be `base.contains("slice")`. Is this right, or does `contains()` also go to children?

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -540,6 +540,224 @@ TYPED_TEST(inlet_object, implicit_conversion_primitives)
...
+
+  auto& quux_table = inlet.addGenericArray("quux");
+  auto& foo_table = quux_table.addGenericArray("arr");
+
+  foo_table.addBool("bar", "bar's description");
+  foo_table.addBool("baz", "baz's description");
+
+  // Contiguous indexing for generality
+  std::unordered_map<int, QuuxWithFooArray> expected_quuxs = {
+    {0, {{{0, {true, false}}, {1, {false, true}}}}},
```
estebanpauli, line 581: That's a lot of curly braces :). Some newlines might help readability.

estebanpauli (commented):
Since I'm not familiar with this code base, it's hard for me to judge a lot of this code, but hopefully some of the comments I left are useful.

src/axom/inlet/inlet_utils.cpp
```diff
@@ -119,14 +119,50 @@ std::string removeBeforeDelimiter(const std::string& path, const char delim)
...
+  auto pos = result.find(substr);
+  while(pos != std::string::npos)
+  {
+    result.erase(pos, substr.length());
+    pos = result.find(substr);
+  }
+  return result;
+}
+
+bool checkedConvertToInt(const std::string& number, int& result)
```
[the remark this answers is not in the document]
joshessman-llnl: Good feedback - I had been vacillating between the out-params (which Inlet used to use frequently) and the returned pair when I initially wrote this function.

There are a couple other places where I would have found it convenient to use `std::optional`, too.  I think implementing a limited version of the type could be useful (in `inlet_utils` or something) to start, but it might be a good utility type to pull into `core` as well.  Any thoughts on this @white238 @kennyweiss?

src/axom/inlet/Inlet.hpp
joshessman-llnl: If I'm understanding your question, the intent is that lists of structs would be added via `addGenericArray`, and lists of primitives would be added via `addIntArray`, `addStringArray`, etc.  Maybe `addStructArray`/`addStructDictionary` would be clearer names that match the naming used for this new `addStruct` function?

src/axom/inlet/examples/nested_structs.cpp
joshessman-llnl: In theory, yes, but to avoid cluttering the example with logic that will be added in #429 and #426, I have omitted some verification logic here.

src/axom/inlet/examples/nested_structs.cpp
joshessman-llnl: > "slice" was a nested dictionary

Ah, the spacing must have gotten lost in translation.  I've updated the example accordingly.  The schema adjustments mean that the conditional in `FromInlet<Operator>` can use the more intuitive `base.contains("slice")`.

src/axom/inlet/Inlet.hpp
estebanpauli: My question was meant to be about the documentation more than anything. It says "Structures can contain fields and/or substructures". Since substructures are called out explicitly, it made me wonder if it does not support lists. Sounds like it does, though.

As far as your suggestion goes, I think the more consistent the names are, the better. Given that we have e.g. `addString()/addStringArray()`, having `addStruct()/addStructArray()` would be nice.

src/axom/inlet/inlet_utils.cpp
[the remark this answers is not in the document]
estebanpauli: Yes, I meant output parameters :). I think input parameters are fine.

src/axom/inlet/tests/inlet_object.cpp
kennyweiss: We might be limited by `clang-format` in adding newlines

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -0,0 +1,276 @@
...
+    {
+      result.type = Operator::Type::Slice;
+      // Grab the substructure corresponding to the slice operation
+      // and use it to populate the Operator instance
+      auto slice = base["slice"];
+      if(slice.contains("x"))
+      {
+        result.x = slice["x"];
+      }
+      if(slice.contains("y"))
```
kennyweiss, line 102: Minor: if `y` is not present in the input deck, 
* does/should it get initialized to some default value?
* how does the user know if `y` was present or if the default option was used?
* how/where does the developer add the `*.default()`  and/or `*.required()` tags?

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+struct Shape
+{
+  std::string name;
+  enum class Material
+  {
+    Steel,
+    Wood,
+    Plastic
+  } material;
```
kennyweiss, line 202: Does the `Material` enum and/or `material` instance add anything essential to this example? e.g. w.r.t. supporting this use case?

Tag: @estebanpauli

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -0,0 +1,276 @@
...
+  axom::slic::SimpleLogger logger;
+
+  axom::sidre::DataStore ds;
+  auto reader = std::unique_ptr<inlet::YAMLReader>(new inlet::YAMLReader());
+  reader->parseString(input);
+  inlet::Inlet inlet(std::move(reader), ds.getRoot());
+
+  auto& shapes_table = inlet.addGenericArray("shapes");
+  Shape::defineSchema(shapes_table);
+
```
kennyweiss, line 270: @joshessman-llnl -- it would be instructive to see how one might define a validator on the `Operator` to see if it is a valid translation/rotation/slice.

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -0,0 +1,276 @@
...
+  auto shapes = inlet["shapes"].get<std::unordered_map<int, Shape>>();
```
kennyweiss, line 271: It would also be interesting to see what the generated sphinx docs would look like for this nested structs example. Would it be easy to add a "generate docs" option?

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -0,0 +1,276 @@
...
+    if(material == "steel")
+    {
+      result.material = Shape::Material::Steel;
+    }
+    result.geom = base["geometry"].get<Geometry>();
+    return result;
+  }
+};
+
+const std::string input = R"(
```
kennyweiss, line 238: 👍 for the multiline string literal!

Presumably, clang-format is happy with it?

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -0,0 +1,276 @@
...
+      start_dimensions: 3
+      operators:
+        - translate: [10, 20, 30]
+        - rotate: 45
+          axis: [1, 2, 3]
+          center: [4, 5, 6]
+        - slice:
+            x: 10
+        - translate: [5, 6]
+)";
```
kennyweiss, line 256: @estebanpauli -- do you think it would be a good idea to add a second shape to make sure this all works as expected?

src/axom/inlet/examples/CMakeLists.txt
```diff
@@ -10,22 +10,31 @@
 # Specify list of examples
 #------------------------------------------------------------------------------
 
+set(inlet_examples
+    nested_structs.cpp )
+
 if (SOL_FOUND)
    set(inlet_examples
```
kennyweiss, line 17: Should this be a `list(APPEND ...)` ?  
Or possibly a `blt_list_append`  (https://llnl-blt.readthedocs.io/en/develop/api/utility.html#blt-list-append)

src/axom/inlet/Table.cpp
```diff
@@ -91,39 +112,48 @@ std::vector<VariantKey> Table::containerIndices() const
...
 
 std::vector<std::pair<std::string, std::string>> Table::containerIndicesWithPaths(
   const std::string& name) const
 {
   std::vector<std::pair<std::string, std::string>> result;
-  // Need to go up one level because this is an _inlet_container group
-  const auto pos = m_name.find_last_of("/");
-  const std::string baseName = m_name.substr(0, pos);
-  for(const auto& indexLabel : containerIndices())
+  for(const auto& indexLabel : containerIndices(false))
```
kennyweiss, line 147: Side note: It seems like a lot of this logic would be simplified with `sidre` (or possibly axom `core`) had a `Path` class to extract the stem/root etc...  (and to support different delimiters).

kennyweiss (approved):
Thanks @joshessman-llnl -- this looks great!

Thanks also for supporting the "nested_structs" example.
@estebanpauli -- does this cover all the expected needs for your use case?

src/axom/inlet/examples/nested_structs.cpp
kennyweiss: As far as I can see, this should not affect @joshessman-llnl's solution.

Perhaps the operator could have `slice_x`, `slice_y`, ... `slice_up`, if that would be helpful/clearer.

src/axom/inlet/examples/nested_structs.cpp
estebanpauli: FWIW, in Klee, this is string, not an enum. If this serves to verify that enums work, though, or shows how to use them, I think it helps the example.

src/axom/inlet/examples/nested_structs.cpp
estebanpauli: Sure, that would probably be good.

estebanpauli:
> Thanks @joshessman-llnl -- this looks great!
> 
> Thanks also for supporting the "nested_structs" example.
> @estebanpauli -- does this cover all the expected needs for your use case?

This looks like it would cover what I'm currently seeing as problem. More shortcomings could be found later, but this looks like it should help considerably.

src/axom/inlet/examples/nested_structs.cpp
kennyweiss: @estebanpauli -- I agree. Do you have an example with a second shapes? Or could you please supply one? 
Perhaps a shape without any operators, and a shape with a translation operator?

src/axom/inlet/examples/nested_structs.cpp
joshessman-llnl: In this example enumerations are represented in the input deck as strings, so it would be easy use that approach.

I agree that the enum should be left in, as it may be instructive for future users looking to "deserialize" enumerations with Inlet.

src/axom/inlet/examples/nested_structs.cpp
estebanpauli: Sure. A simple example can serve the purposes here. I don't think it requires anything complicated:
```
  - name: tire
    material: rubber
    geometry:
      format: test_format
      path: path/to/tire.format
      units: cm
      operators:
        - translate: [10, 20]
```

src/axom/inlet/examples/nested_structs.cpp
joshessman-llnl: I think for an example this is indeed minor, but in actual usage it might be helpful to have a nullable utility type (like a limited version of `std::optional`) for these cases.

Regardless of how these slice variables are initialized, I don't think this example has any `default` or `required` fields - I think these would be useful to demonstrate.

src/axom/inlet/examples/nested_structs.cpp
joshessman-llnl: Surprisingly, yes

src/axom/inlet/Inlet.hpp
joshessman-llnl: I think for this PR I can add `addStructArray` et all as aliases for `addGenericArray`.  Since a follow-up PR will be removing `addTable` from the public interface of `inlet::Table`, I think the removal of `addGeneric*` can be made part of that PR.

src/axom/inlet/examples/CMakeLists.txt
joshessman-llnl: I agree that `blt_list_append` would more clearly express intent here.

src/axom/inlet/examples/nested_structs.cpp
joshessman-llnl: Yes, although this appears to have exposed some inadequacies in the SphinxDocWriter logic.  I have logged this as #436

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -65,6 +66,21 @@ struct Operator
...
+      const bool is_slice = table.contains("slice");
+
+      // Can only be one
+      if((is_translate && is_rotate) || (is_translate && is_slice) ||
+         (is_rotate && is_slice))
+      {
+        return false;
+      }
+
+      return true;
```
kennyweiss, line 82: Minor: should this be: 
```cpp
return is_translate || is_rotate || is_slice;
```
Or does this verifier assume that the operator can be empty?

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -151,11 +167,11 @@ struct Geometry
     table.addString("format");
     table.addString("path");
     // A string is used to represent the enumeration
-    table.addString("units");
-    table.addInt("start_dimensions");
-    // addGenericArray is used to represent std::vector<T> where
+    table.addString("units").defaultValue("cm").validValues({"cm", "m"});
```
kennyweiss, line 170: Thanks for improving the example @joshessman-llnl !

kennyweiss (approved):

src/axom/inlet/Inlet.hpp
```diff
@@ -543,6 +578,22 @@ class Inlet
...
+   *****************************************************************************
+   * \brief Add an dictionary of user-defined type to the input file schema.
+   *
+   * \param [in] name Name of the dictionary
+   * \param [in] description Description of the dictionary
+   *
+   * \return Reference to the created dictionary
+   *****************************************************************************
+   */
+  Table& addStructDictionary(const std::string& name,
```
estebanpauli, line 591: This is just to clarify my understanding. What's the difference between a struct and a dictionary? Can a dictionary have arbitrary members, but a struct only specific ones?

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -0,0 +1,343 @@
...
+  static void defineSchema(inlet::Table& table)
+  {
+    // The rotation is in degrees
+    table.addDouble("rotate").range(-180, 180);
+    // Vectors are defined as arrays of doubles
+    table.addDoubleArray("axis");
+    table.addDoubleArray("center");
+    table.addDoubleArray("translate");
+    // The slice operation can have sub-entries, so we represent it as a struct
+    // Note that Inlet does not require a 1-1 correspondence (or any correspondence)
```
estebanpauli, line 63: The note here about not requiring a 1-1 correspondence is useful. Thanks for including it.

estebanpauli (approved):

src/axom/inlet/Inlet.hpp
joshessman-llnl: This is a good question - I've added a snippet on this to the docs, reproduced here:

Although many of Inlet's input file languages do not distinguish between a "dictionary" type
  and a "record" type, Inlet treats them differently for type safety reasons:

  *Dictionaries* use arbitrary strings or integers for their keys, and their values (entries)
  can only be retreived as a homogenous type.  In other words, dictionaries must map to
  ``std::unordered_map<Key, Value>`` for fixed key and value types.

  *Structs* contain a fixed set of named fields, but these fields can be of any type.
  As the name suggests, these map to ``structs`` in C++.

  In short, if the *key values* vary, use a dictionary.  If the *field types* vary, use a struct.

src/axom/inlet/examples/user_defined_type.cpp
```diff
@@ -184,12 +184,12 @@ struct ThermalSolver
   static void defineSchema(inlet::Table& schema)
   {
     // _inlet_userdef_simple_usage_start
-    auto& mesh_table = schema.addTable("mesh", "Information about the mesh");
+    auto& mesh_table = schema.addStruct("mesh", "Information about the mesh");
```
white238, line 187: Calling this `mesh_table` is a tad confusing here due to them not lining up now.

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -0,0 +1,343 @@
...
+    os << "   Rotation operator:\n";
+    os << fmt::format("      with axis: {0}\n", op.axis);
+    os << fmt::format("      with center: {0}\n", op.center);
+    break;
+  case Operator::Type::Slice:
+    os << "   Slice operator:\n";
+    os << fmt::format("      with x-coord: {0}\n", op.x);
+    os << fmt::format("      with y-coord: {0}\n", op.y);
+    os << fmt::format("      with z-coord: {0}\n", op.z);
+    break;
```
white238, line 151: Error out in unknown type?

src/axom/inlet/examples/user_defined_type.cpp
joshessman-llnl: I think keeping the `Table` class name is fine; the intent of the change is to separate the addition of a `Table` (which is specific to the internals of the Inlet hierarchy) and the addition of a struct/record type to the input schema.

I agree that this should probably be renamed though, maybe `mesh_schema`?

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -0,0 +1,343 @@
...
+    slice.addDouble("y");
+    slice.addDouble("z");
+
+    // Verify that exactly one type of operator is defined
+    table.registerVerifier([](const inlet::Table& table) {
+      const bool is_translate = table.contains("translate");
+      const bool is_rotate = table.contains("rotate");
+      const bool is_slice = table.contains("slice");
+
+      // Can only be one
```
white238, line 77: ... missed Highlander reference. PR rejected.

src/axom/inlet/examples/user_defined_type.cpp
white238: `mesh_schema` makes me feel better. which is what is important.

white238 (approved):

src/axom/inlet/tests/inlet_object.cpp
cyrush: > We might be limited by `clang-format` in adding newlines

`clang-format` doesn't like new lines?

kennyweiss: I was suggesting that `clang-format` might remove manually introduced extra lines/whitespace.

(It works at the AST level, not at the line of source code level, so source code whitespaces are not considered in its reformatting)

src/axom/inlet/Inlet.hpp
cyrush: When creating concepts likes this for an interface, even if you document well -- it can be confusing if you attribute new semantics to things people may already have mental models for (Tables, Dictionaries, Structs). 

`if the key values vary, use a dictionary`.  

By vary here you mean the key names aren't validated, where the struct names are? Is that the root of the difference with validating the `names` (or keys) passed when fetching?


An important aspect is if the validation + errors messages for users lead them to understand what is off and make it easy for folks to correct the inputs when things go off the rails, regardless of what constructs you use to do that. 

Hard to understand all that via a PR

src/axom/inlet/tests/inlet_object.cpp
cyrush: well then, today i learned not to get excited about `clang-format`

kennyweiss: My takeaway is that our codebase formatting is significantly more consistent since moving to `clang-format`.

There are a few cases where the whitespaces had semantic utility (e.g. constructors for a 4x4 matrix), and in those limited cases, we've manually tagged the few lines (via ``/* clang-format off */``   ``/* clang-format on */`` )

src/axom/inlet/Inlet.hpp
white238: @cyrush  and i had a conversation about this thread.  We agree there is a confusion between overloaded concepts due to language differences.  We should get these existing PRs in.  Then have a high level discussion to try to minimize the confusion going on.  Then have one PR that addresses our naming conventions.

PR #437, opened 2021-01-26
Use different filenames for serial and parallel test runs of sidre examples

nselliott:
# Summary

- This PR is a bugfix
- It does the following:
  - Adds a "_par" to the filenames used in SPIO output when running the sidre_createdatastore and sidre_generateindex examples in parallel.  This should eliminate the possibility that the serial and parallel runs of these examples might simultaneously write to the same files during regression testing.


kennyweiss (approved):
Thanks @nselliott !

cyrush (approved):

agcapps (approved):
Thanks, @nselliott .

rhornung67 (approved):

PR #446, opened 2021-02-02
Refactor Inlet's DocWriter interface, add support for documenting provided values

joshessman-llnl:
# Summary

- This PR is a refactoring/feature
- It does the following:
  - Refactors the `DocWriter` interface to be more descriptive and renames it to just `Writer`
  - Adds functionality to the SphinxDocWriter that adds a column for the value that was provided in the input file

Currently the `DocWriter` interface only requires that implementers implement a `void writeDocumentation(void)`, which isn't a particularly descriptive signature.  This PR introduces `documentTable(const Table&)` and `finalize()` methods to the interface which implementers use to document a single `Table` and write/flush documentation, respectively.

This allows the base `Inlet` class to manage traversal through the Inlet tree/hierarchy (which is writer-agnostic).

This PR also extends the SphinxDocWriter to document the provided values.  This is reflected in the updated Sphinx page with the sample output.  Note that Inlet does not keep track of whether the provided value was just the default value (i.e., it was not provided by the user), so the generated documentation also does not make this distinction.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -65,6 +65,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - The `MFEMSidreDataCollection` will now reconstruct fields and the mesh when a
   datastore is `Load` ed in
 - Inlet: Exposed primal::Vector in Lua for use in input-file-defined functions
+- Inlet: Cleaned up `Table` interface to eliminate ambiguity and duplicated functionality
 
 ### Fixed
 - Updated to new BLT version that does not fail when ClangFormat returns an empty
@@ -72,6 +73,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   target if version is unknown or wrong.
 - Inlet: Apply lambda verifiers on generic containers to individual elements
   for consistency
+- Inlet: Fixed a bug relating to nested table lookups of primitive arrays and functions
 
 
 ## [Version 0.4.0] - Release date 2020-09-22
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -39,6 +39,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Inlet: Added support for deeply nested containers of structs
 - Inlet: Added support for `void` and strings in Lua-defined functions
 - Inlet: Added `get<std::vector<T>>` for retrieving arrays without index information
+- Inlet: SphinxDocWriter can now output the values provided in an input file
 
 ### Changed
 - The Sidre Datastore no longer rewires Conduit's error handlers to SLIC by default. 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -39,7 +39,6 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Inlet: Added support for deeply nested containers of structs
 - Inlet: Added support for `void` and strings in Lua-defined functions
 - Inlet: Added `get<std::vector<T>>` for retrieving arrays without index information
-- Inlet: SphinxDocWriter can now output the values provided in an input file
 
 ### Changed
 - The Sidre Datastore no longer rewires Conduit's error handlers to SLIC by default. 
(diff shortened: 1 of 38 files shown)


kennyweiss (approved):
Thanks @joshessman-llnl -- this is a nice improvment.

Are the callback functions added to the tables? (Perhaps I missed it).
Ideally, they would list their signature in addition to other metadata.

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -328,9 +328,9 @@ int main(int argc, char** argv)
 
   if(docsEnabled)
   {
-    std::unique_ptr<inlet::SphinxDocWriter> docWriter(
-      new inlet::SphinxDocWriter("nested_structs.rst", inlet.sidreGroup()));
-    inlet.registerDocWriter(std::move(docWriter));
+    std::unique_ptr<inlet::SphinxWriter> writer(
+      new inlet::SphinxWriter("nested_structs.rst", true));
```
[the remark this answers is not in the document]
joshessman-llnl: Also, I think your `outputProvidedValues` is more descriptive than the current `documentProvided`, so I've switched that parameter name.

joshessman-llnl:
> 
> Are the callback functions added to the tables? (Perhaps I missed it).
> Ideally, they would list their signature in addition to other metadata.

Not currently - I'll tag this as an issue.

src/axom/inlet/SphinxWriter.hpp
```diff
@@ -226,13 +207,14 @@ class SphinxDocWriter : public DocWriter
...
 
-  axom::sidre::Group* m_sidreRootGroup;
   std::ofstream m_outFile;
   std::ostringstream m_oss;
   // This is needed to preserve the traversal order of the Inlet::Tables
   std::vector<std::string> m_inletTablePathNames;
   std::unordered_map<std::string, TableData> m_rstTables;
   std::string m_fileName;
+  // Used for the RST tables for fields
+  std::vector<std::string> m_colLabels;
```
kennyweiss, line 217: Minor: Since you already depend on Sidre, you might consider using a `sidre::MapCollection` for this.
The data is stored in an integer-indexed array, but entries can be indexed by strings.

See: https://github.com/LLNL/axom/blob/develop/src/axom/sidre/core/MapCollection.hpp

src/axom/inlet/Writer.hpp
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2017-2021, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/*!
+ *******************************************************************************
+ * \file Writer.hpp
+ *
+ * \brief This file contains the abstract base class definition of Writer.
```
kennyweiss, line 10: I like this abstraction.
I can imagine a use-case where one would want to output each top-level table in a separate file.

src/axom/inlet/SphinxWriter.hpp
joshessman-llnl: I agree that something like this might be useful, but the use of raw pointers as the stored item instead of objects themselves would require introducing manual memory management here (I think) since I want the data to be owned by the container.

I'll have to tweak this logic a bit anyways when I add callback function documentation, so possibly a more elegant solution will present itself.

RELEASE-NOTES.md
```diff
@@ -39,6 +39,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Inlet: Added support for deeply nested containers of structs
 - Inlet: Added support for `void` and strings in Lua-defined functions
 - Inlet: Added `get<std::vector<T>>` for retrieving arrays without index information
+- Inlet: SphinxDocWriter can now output the values provided in an input file
```
white238, line 42: Mention that they have been renamed

src/axom/inlet/docs/sphinx/example1_expected_documentation.rst
```diff
@@ -18,23 +19,28 @@ thermal_solver
      - Default Value
      - Range/Valid Values
      - Required
-   * - order
-     - thermal solver order
-     - 
-     - 1 to 2147483647
-     - |check|
+     - Value
```
white238, line 22: This column doesn't get output at all unless you set the option correct? Also, I find this a very confusing workflow.  The user would have to output this then to see the value in a helpful way they would have to render this w/ sphinx.

joshessman-llnl: I've removed this logic in favor of a `ConduitWriter` or something similar for dumping provided values

white238 (approved):

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

PR #449, opened 2021-02-03
Nested primitive array lookups must filter their paths

joshessman-llnl:
# Summary

- This PR is a hotfix
- It does the following:
  - Fixes a bug where paths are not filtered before calls to the `Reader` as they already are in `addPrimitive` and `addGenericContainer`

### DIFF
--- a/src/axom/inlet/examples/nested_structs.cpp
+++ b/src/axom/inlet/examples/nested_structs.cpp
@@ -34,6 +34,7 @@ struct Operator
   double x;
   double y;
   double z;
+  Vector origin;
   enum class Type
   {
     Translate,
@@ -58,6 +59,7 @@ struct Operator
     slice.addDouble("x");
     slice.addDouble("y");
     slice.addDouble("z");
+    slice.addDoubleArray("origin");
 
     // Verify that exactly one type of operator is defined
     table.registerVerifier([](const inlet::Table& table) {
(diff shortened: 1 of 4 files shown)


estebanpauli:
Changes look good. The only thing I would suggest is adding a unit test to make sure this doesn't stop working later. Of course, temporarily undo your change and make sure the test fails without your changes.

estebanpauli (approved):

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -282,6 +289,7 @@ dimensions: 2
           center: [4, 5, 6]
         - slice:
             x: 10
+            origin: [40, 50, 60]
```
kennyweiss, line 292: Side question: @estebanpauli -- how do `x` and `origin` combine to create a slicing plane?

kennyweiss (approved):
Thanks @joshessman-llnl !

white238 (approved):

src/axom/inlet/examples/nested_structs.cpp
estebanpauli: By default, if you just specify "x: 10", you get the "x=10" plane, with the origin at (10, 0, 0) and an up direction of (0, 1, 0). Setting the origin allows you to move the origin. In the example above, an error would be reported because the x component doesn't match the specified x value.

For full control, you can specify "origin" (point), "normal" (vector), and "up" (vector) to define your plane. The "x", "y", and "z" versions just serve as convenient shortcuts by giving you default values for the unspecified parts of the origin as well as the normal and up vectors.

PR #452, opened 2021-02-03
Inlet interface cleanup

joshessman-llnl:
# Summary

- This PR is a cleanup
- It does the following:
  - Restricts the public interface of `Inlet::Table`, in particular by removing duplicated functionality

Per my comments in https://github.com/LLNL/axom/pull/434#discussion_r564056514 and https://github.com/LLNL/axom/pull/434#issue-558533411, the interface has been tweaked a bit.

* The `addGenericArray`/`addGenericDictionary` aliases for `addStructArray`/`addStructDictionary` have been removed
* `addTable` is now private - `addStruct` should be used for nested data in the input file
* "Type-erased" implementations of `name()` and `sidreGroup` for `inlet::Proxy` have been implemented that forward to the underlying object. 
* The `Proxy` object (via `op[]`) is now the only method of access/retrieval (instead of with `getTable`) - this motivated the above
* Similarly, `contains` is now the only way of checking if something exists in the input file.  This eliminates ambiguity with `hasTable` et al, which don't check the input file and only indicate whether something is in the schema (only useful internally).  The type of the underlying object can still be queried from the `Proxy`.

### DIFF
--- a/src/axom/inlet/Proxy.cpp
+++ b/src/axom/inlet/Proxy.cpp
@@ -15,7 +15,7 @@ InletType Proxy::type() const
   if(m_table != nullptr)
   {
     // This is how Inlet stores array types in the datastore
-    if(m_table->hasTable(detail::CONTAINER_GROUP_NAME))
+    if(m_table->contains(detail::CONTAINER_GROUP_NAME))
     {
       return InletType::Container;
     }
@@ -47,6 +47,49 @@ Proxy Proxy::operator[](const std::string& name) const
     SLIC_ERROR("[Inlet] Cannot index a proxy that refers to a field");
   }
   return (*m_table)[name];
+}
+
+const axom::sidre::Group* Proxy::sidreGroup() const
+{
+  if(m_table != nullptr)
+  {
+    return m_table->sidreGroup();
+  }
+  else if(m_field != nullptr)
+  {
+    return m_field->sidreGroup();
+  }
+  else if(m_func != nullptr)
+  {
+    SLIC_ERROR("[Inlet] Cannot retrieve the sidre::Group for a Function");
+  }
+  else
+  {
+    SLIC_ERROR("[Inlet] Cannot retrieve the sidre::Group of an empty Proxy");
+  }
+  return nullptr;
+}
+
+std::string Proxy::name() const
+{
+  // FIXME: With C++14 we can implement a visit() method that takes a generic lambda
+  if(m_table != nullptr)
+  {
+    return m_table->name();
+  }
+  else if(m_field != nullptr)
+  {
+    return m_field->name();
+  }
+  else if(m_func != nullptr)
+  {
+    return m_func->name();
+  }
+  else
+  {
+    SLIC_ERROR("[Inlet] Cannot retrieve the name of an empty Proxy");
+  }
+  return "";
 }
 
 }  // end namespace inlet
--- a/src/axom/inlet/Proxy.hpp
+++ b/src/axom/inlet/Proxy.hpp
@@ -45,7 +45,7 @@ public:
    * \param [in] table The table to construct a proxy into
    *******************************************************************************
    */
-  Proxy(Table& table) : m_table(&table) { }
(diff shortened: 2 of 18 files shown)


src/axom/inlet/Table.cpp
```diff
@@ -721,7 +721,7 @@ Table& Table::required(bool isRequired)
   // If it's a generic container we set the individual fields as required,
```
white238, line 721: "If it's a struct container..."

src/axom/inlet/Table.hpp
```diff
@@ -939,74 +817,145 @@ class Table : public Verifiable<Table>
...
+   *****************************************************************************
+   */
+  std::string name() const;
+
+  /*!
+   *****************************************************************************
+   * \brief Add a Field to the input file schema.
    *
-   * \return Boolean value indicating whether this Table's subtree contains this Function.
+   * Adds a Field to the input file schema. It may or may not be required
```
white238, line 864: "Adds a Primitive Field"?

src/axom/inlet/Table.hpp
```diff
@@ -939,74 +817,145 @@ class Table : public Verifiable<Table>
...
-  const std::unordered_map<std::string, std::unique_ptr<Table>>& getChildTables() const;
+  bool hasField(const std::string& fieldName) const;
 
   /*!
    *****************************************************************************
-   * \return The full name of this Table.
+   * \brief Return whether a Function with the given name is present in this Table's
+   *  subtree.
+   *
+   * \return Boolean value indicating whether this Table's subtree contains this Function.
```
white238, line 955: too many this's.  possibly "this Table's subtree contains a function with the given name"

src/axom/inlet/Table.hpp
```diff
@@ -1217,6 +1165,20 @@ class Table : public Verifiable<Table>
     return map;
   }
 
+  /*!
+   *******************************************************************************
+   * \brief Adds a group containing the indices of a container to the calling 
+   * table, and a subtable for each index
```
white238, line 1171: remove comma

white238 (approved):

joshessman-llnl:
Resolves #453

src/axom/inlet/Proxy.cpp
```diff
@@ -49,5 +49,48 @@ Proxy Proxy::operator[](const std::string& name) const
...
+  {
+    return m_table->sidreGroup();
+  }
+  else if(m_field != nullptr)
+  {
+    return m_field->sidreGroup();
+  }
+  else if(m_func != nullptr)
+  {
+    SLIC_ERROR("[Inlet] Cannot retrieve the sidre::Group for a Function");
```
kennyweiss, line 64: Can this print the path for the `Proxy`?  
(This might only makes sense if `Proxy` stores a path independently of the `sidre::Group`).

EDIT: Looking that the member fields for `Proxy` -- looks like the answer is "No. That information is not available".

src/axom/inlet/Proxy.hpp
```diff
@@ -201,9 +218,9 @@ class Proxy
   }
 
 private:
-  Table* m_table = nullptr;
-  Field* m_field = nullptr;
-  Function* m_func = nullptr;
+  const Table* const m_table = nullptr;
+  const Field* const m_field = nullptr;
+  const Function* const m_func = nullptr;
```
kennyweiss, line 223: Re: printing the path -- looks like it doesn't  store any additional information.

kennyweiss (approved):
Thanks @joshessman-llnl -- the API seems to be getting more coherent and consistent!

src/axom/inlet/Proxy.cpp
joshessman-llnl: This could be added since everywhere that constructs a Proxy has path information - this should probably be optional though, since the implicit Table -> Proxy conversion appears to be useful: https://github.com/LLNL/axom/issues/453#issue-802425925

PR #467, opened 2021-02-10
Fix a bug affecting deeply nested function callbacks

joshessman-llnl:
# Summary

- This PR is a bugfix
- It does the following:
  - Fixes deeply nested function callbacks

This fix is conceptually similar to #449 - there's a lot of somewhat similar logic for iterating over nested objects and forwarding function calls to nested tables, so in the near future I'd like to look at refactoring that and keeping the iteration logic all in one place.


white238 (approved):

white238:
LGTM

src/axom/inlet/tests/inlet_function.cpp
```diff
@@ -442,6 +442,48 @@ TEST(inlet_function, dimension_dependent_result)
...
+  foo_schema.addFunction("bar",
+                         FunctionTag::Double,
+                         {FunctionTag::Double},
+                         "bar's description");
+
+  auto foos = inlet["quux"].get<std::vector<FooWithScalarFunc>>();
+  EXPECT_EQ(foos.size(), 2);
+
+  auto& first_func = foos[0].bar;
+  EXPECT_TRUE(static_cast<bool>(first_func));
```
kennyweiss, line 479: Is this a test that it's not a `nullptr`?  If so, that might more clearly show the intent.

kennyweiss (approved):

src/axom/inlet/tests/inlet_function.cpp
joshessman-llnl: `std::function` doesn't make it easy to get the actual stored function pointer when it's wrapping a lambda, but I can add a comment clarifying what the check is doing.

joshessman-llnl:
LGTM

PR #469, opened 2021-02-11
Rename container -> collection to prepare for renaming Table -> Container

joshessman-llnl:
# Summary

- This PR is a renaming
- It does the following:
  - Renames "container" to "collection"

Inlet uses "container" in its internal implementation to refer to the union of arrays and dictionaries.  Because `Table` is going to be renamed to `Container`, it needs to be renamed.  The term "container" is not part of the external interface so it will not affect user  code.

Other options (instead of collection) could be `data structure` (too vague, in my opinion) or `associative array` (as even integer-keyed arrays are associative).


white238 (approved):
Thanks, @joshessman-llnl ! This took a bit to wrap my brain around this but it is more clear than container in this case.. since we are referring to a collection of things not necessarily the container that holds them.  At least that's how I think of it.

white238:
LGTM

kennyweiss (approved):
Thanks @joshessman-llnl  -- I think this is moving in the right direction.

Should this be rebased on the docs PR, so you can update the glossary (and other docs)?

joshessman-llnl:
LGTM

PR #477, opened 2021-02-17
Consistent behavior for unexpected types in primitive arrays

### LINKED ISSUE #476: Crash on bad array input
Given the test below, I expect the `inlet.verify()` call to fail. Instead, it succeeds, allowing `auto values = inlet["expect_doubles"].get<std::vector<double>>();` to run, resulting in a crash.

```cpp
TEST(inlet_Inlet, badDoubles)
{
    reader->parseString(R"(
    expect_doubles: [a, b, c]
    )");
    axom::sidre::DataStore ds;

    inlet.addDoubleArray("expect_doubles");
    if (inlet.verify()) {
        auto values = inlet["expect_doubles"].get<std::vector<double>>();
        FAIL() << "Unexpectedly read " << values.size() << " values";
    }
}
```

Output:
```
[ERROR in line 639 of file /usr/WS1/pauli2/axom/src/axom/inlet/Table.hpp]
MESSAGE=[Inlet] Field with specified pathdoes not exist: 0
** StackTrace of 20 frames **
Frame 2: _ZNK4axom5inlet5Table3getIdEENSt9enable_ifIXsr6detail18is_inlet_primitiveIT_EE5valueES4_E4typeERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
Frame 3: std::unordered_map<int, double, std::hash<int>, std::equal_to<int>, std::allocator<std::pair<int const, double> > > axom::inlet::Table::getContainer<int, double>() const
Frame 4: _ZNK4axom5inlet5Table3getISt13unordered_mapIidSt4hashIiESt8equal_toIiESaISt4pairIKidEEEEENSt9enable_ifIXoosr6detail14is_inlet_arrayIT_EE5valuesr6detail13is_inlet_dictISE_EE5valueESE_E4typeEv
Frame 5: _ZNK4axom5inlet5Table3getISt6vectorIdSaIdEEEENSt9enable_ifIXsr6detail13is_std_vectorIT_EE5valueES7_E4typeEv
Frame 6: _ZNK4axom5inlet5Proxy3getISt6vectorIdSaIdEEEENSt9enable_ifIXaantsr6detail18is_inlet_primitiveIT_EE5valuentsr6detail15is_std_functionIS7_EE5valueES7_E4typeEv
Frame 7: inlet_Inlet_badDoubles_Test::TestBody()
Frame 8: void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*)
Frame 9: void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*)
Frame 10: testing::Test::Run()
Frame 11: testing::TestInfo::Run()
Frame 12: testing::TestSuite::Run()
Frame 13: testing::internal::UnitTestImpl::RunAllTests()
Frame 14: bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*)
Frame 15: bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*)
Frame 16: testing::UnitTest::Run()
Frame 17: RUN_ALL_TESTS()
Frame 18: main
Frame 19: __libc_start_main
Frame 20: tests/inlet_Inlet_test() [0x668fd4]
=====


Aborted (core dumped)
```

Interestingly, if I make the field required, the test passes, with the following output:
```
MESSAGE=[Inlet] Required Table not specified: expect_doubles/_inlet_container
```

The field is there, but is the wrong type, so the error is misleading.

joshessman-llnl:
# Summary

- This PR is a bugfix
- It does the following:
  - Partially addresses #476 by ignoring indices of elements that are not of the requested type

# Design review

Currently, having elements of type other than the one requested in an array or dictionary is not an error - these elements are just ignored.  This PR ensures that these other elements are always ignored, but this means that an array that contains the wrong types is indistinguishable from an array that does not exist.  Both cases will cause verification to fail.  

The logic to ignore elements of unexpected type is implemented at the `Reader` level, so one option would be to throw errors there if unexpected elements are found.  This would clash with the current convention of avoiding throwing errors from the `Reader`s (and instead letting Inlet handle things at verification time).  Deferring to verification time would require introducing and propagating a means of distinguishing an empty array due to it not being provided vs not having any elements of the requested type.

### DIFF
--- a/src/axom/inlet/Table.cpp
+++ b/src/axom/inlet/Table.cpp
@@ -437,12 +437,16 @@ namespace detail
   *****************************************************************************
   */
 template <typename T>
-void registerCollection(Table& table, const std::unordered_map<int, T>& collection)
+std::vector<VariantKey> registerCollection(Table& table,
+                                           const std::unordered_map<int, T>& collection)
 {
+  std::vector<VariantKey> result;
   for(const auto& entry : collection)
   {
+    result.push_back(entry.first);
     table.addPrimitive(std::to_string(entry.first), "", true, entry.second);
   }
+  return result;
 }
 
 /*!
@@ -451,11 +455,14 @@ void registerCollection(Table& table, const std::unordered_map<int, T>& collecti
   *****************************************************************************
   */
 template <typename T>
-void registerCollection(Table& table,
-                        const std::unordered_map<VariantKey, T>& collection)
+std::vector<VariantKey> registerCollection(
+  Table& table,
+  const std::unordered_map<VariantKey, T>& collection)
 {
+  std::vector<VariantKey> result;
   for(const auto& entry : collection)
   {
+    result.push_back(entry.first);
     auto string_key = indexToString(entry.first);
     SLIC_ERROR_IF(
       string_key.find('/') != std::string::npos,
@@ -465,6 +472,7 @@ void registerCollection(Table& table,
                   "[Inlet] Dictionary key cannot be the empty string");
     table.addPrimitive(string_key, "", true, entry.second);
   }
+  return result;
 }
 
 /*!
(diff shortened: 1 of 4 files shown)


joshessman-llnl:
After this PR is merged a follow-up PR will completely address #476 by storing additional reader error metadata in Sidre for use during verification.

src/axom/inlet/Table.cpp
```diff
@@ -487,47 +499,57 @@ struct PrimitiveArrayHelper<Key, bool>
    * \param [inout] table The table to add the collection to
    * \param [in] reader The Reader object to read the collection from
    * \param [in] lookupPath The path within the input file to the collection
+   * 
+   * \return The keys from the collection that was added
    *****************************************************************************
    */
-  PrimitiveArrayHelper(Table& table, Reader& reader, const std::string& lookupPath)
+  static std::vector<VariantKey> add(Table& table,
```
estebanpauli, line 506: This is probably out of scope for this PR, but does this need to be a class? This seems like a strange design. Can't this just be accomplished with function templates? This applies to to all the `PrimitiveArrayHelper` templates.

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -802,6 +802,58 @@ TYPED_TEST(inlet_object, primitive_arrays_as_std_vector)
...
+{
+  std::string testString = " arr = { [0] = 'a', [1] = 'b', [2] = 'c'}";
+  DataStore ds;
+  Inlet inlet = createBasicInlet<TypeParam>(&ds, testString);
+
+  // Define schema
+  inlet.addIntArray("arr");
+
+  // The array was empty (same as if it didn't exist), but *not* marked as required
+  EXPECT_TRUE(inlet.verify());
```
estebanpauli, line 815: The plan is to correct this to a `EXPECT_FALSE` in a future MR, correct? For now, this test will just verify the code won't crash.

estebanpauli (commented):
Changes look reasonable. However, I'm not sure that going from a crash to getting an empty array is an improvement, even if it is a temporary change. I think I would rather have a catastrophic failure (core dump) than have a field be silently ignored because the data was invalid.

src/axom/inlet/Table.cpp
joshessman-llnl: I agree that this looks a little suspect, but this ended up being necessary as I needed partial template specializations to minimize duplication, which C++ doesn't allow for function templates (only class templates).

src/axom/inlet/tests/inlet_object.cpp
joshessman-llnl: Yes - in the follow-up, the tests will be as follows:

https://github.com/LLNL/axom/blob/7435bba50ea921d145aa48a02c4bfd8013030f8e/src/axom/inlet/tests/inlet_object.cpp#L805-L844

white238 (approved):

estebanpauli (approved):

joshessman-llnl:
LGTM

PR #478, opened 2021-02-22
Enhanced error detection/reporting through Inlet's Reader interface

### LINKED ISSUE #476: Crash on bad array input

joshessman-llnl:
# Summary

- This PR is a refactoring
- It does the following:
  - Refactors Inlet's `Reader` interface to return a `ReaderResult` enumeration instead of a success boolean for higher-resolution error reporting during the verification stage
  - Fixes #476 by causing verification to fail when collections are not homogeneous, even if the collection is not required
 
The higher-resolution error messages also mean that providing an entry of the wrong type to an optional part of the schema also will now result in verification failure, instead of being ignored.  The increased type-checking capability during the verification stage allows Inlet users to more reliably detect incorrect input files, especially for optional fields.

I've also added a new page to the Sphinx docs on Inlet's verification process, though I'm not sure that the current structure of that page is the clearest way to present the information.

### DIFF
--- a/src/axom/inlet/ConduitReader.cpp
+++ b/src/axom/inlet/ConduitReader.cpp
@@ -159,39 +159,42 @@ void arrayToMap(const conduit::DataArray<ConduitType>& array,
 
 }  // namespace detail
 
-bool ConduitReader::getValue(const conduit::Node& node, int& value)
+ReaderResult ConduitReader::getValue(const conduit::Node& node, int& value)
 {
   // Match LuaReader functionality - narrow from floating-point but exclude bool
   if(node.dtype().is_number() && !node.dtype().is_uint8())
   {
     value = node.to_int();
-    return true;
+    return ReaderResult::Success;
   }
-  return false;
+  return node.dtype().is_empty() ? ReaderResult::NotFound
+                                 : ReaderResult::WrongType;
 }
 
-bool ConduitReader::getValue(const conduit::Node& node, std::string& value)
+ReaderResult ConduitReader::getValue(const conduit::Node& node, std::string& value)
 {
   if(node.dtype().is_string())
   {
     value = node.as_string();
-    return true;
+    return ReaderResult::Success;
   }
-  return false;
+  return node.dtype().is_empty() ? ReaderResult::NotFound
+                                 : ReaderResult::WrongType;
 }
 
-bool ConduitReader::getValue(const conduit::Node& node, double& value)
+ReaderResult ConduitReader::getValue(const conduit::Node& node, double& value)
 {
   // Match LuaReader functionality - promote from integer but not bool
   if(node.dtype().is_number() && !node.dtype().is_uint8())
   {
     value = node.to_double();
-    return true;
+    return ReaderResult::Success;
   }
-  return false;
+  return node.dtype().is_empty() ? ReaderResult::NotFound
+                                 : ReaderResult::WrongType;
 }
 
-bool ConduitReader::getValue(const conduit::Node& node, bool& value)
+ReaderResult ConduitReader::getValue(const conduit::Node& node, bool& value)
 {
   // Boolean literals don't appear to be parsed as such - they are strings
   if((m_protocol == "yaml") && node.dtype().is_string())
@@ -204,97 +207,100 @@ bool ConduitReader::getValue(const conduit::Node& node, bool& value)
     if(as_str == "true")
     {
       value = true;
-      return true;
+      return ReaderResult::Success;
     }
     else if(as_str == "false")
     {
       value = false;
-      return true;
+      return ReaderResult::Success;
     }
   }
   else if((m_protocol == "json") && node.dtype().is_uint8())
   {
     value = node.as_uint8();
-    return true;
+    return ReaderResult::Success;
   }
-  return false;
+  return node.dtype().is_empty() ? ReaderResult::NotFound
+                                 : ReaderResult::WrongType;
 }
 
-bool ConduitReader::getBool(const std::string& id, bool& value)
+ReaderResult ConduitReader::getBool(const std::string& id, bool& value)
 {
   return getValue(detail::traverseNode(m_root, id), value);
 }
 
-bool ConduitReader::getDouble(const std::string& id, double& value)
+ReaderResult ConduitReader::getDouble(const std::string& id, double& value)
 {
   return getValue(detail::traverseNode(m_root, id), value);
 }
 
-bool ConduitReader::getInt(const std::string& id, int& value)
+ReaderResult ConduitReader::getInt(const std::string& id, int& value)
 {
   return getValue(detail::traverseNode(m_root, id), value);
 }
 
-bool ConduitReader::getString(const std::string& id, std::string& value)
+ReaderResult ConduitReader::getString(const std::string& id, std::string& value)
 {
   return getValue(detail::traverseNode(m_root, id), value);
 }
 
-bool ConduitReader::getIntMap(const std::string& id,
-                              std::unordered_map<int, int>& values)
+ReaderResult ConduitReader::getIntMap(const std::string& id,
+                                      std::unordered_map<int, int>& values)
 {
   return getArray(id, values);
 }
 
-bool ConduitReader::getDoubleMap(const std::string& id,
-                                 std::unordered_map<int, double>& values)
+ReaderResult ConduitReader::getDoubleMap(const std::string& id,
+                                         std::unordered_map<int, double>& values)
 {
   return getArray(id, values);
 }
 
-bool ConduitReader::getBoolMap(const std::string& id,
-                               std::unordered_map<int, bool>& values)
+ReaderResult ConduitReader::getBoolMap(const std::string& id,
+                                       std::unordered_map<int, bool>& values)
 {
   return getArray(id, values);
 }
 
-bool ConduitReader::getStringMap(const std::string& id,
-                                 std::unordered_map<int, std::string>& values)
+ReaderResult ConduitReader::getStringMap(const std::string& id,
+                                         std::unordered_map<int, std::string>& values)
 {
   return getArray(id, values);
 }
 
-bool ConduitReader::getIntMap(const std::string& id,
-                              std::unordered_map<VariantKey, int>& values)
+ReaderResult ConduitReader::getIntMap(const std::string& id,
+                                      std::unordered_map<VariantKey, int>& values)
 {
   return getDictionary(id, values);
 }
 
-bool ConduitReader::getDoubleMap(const std::string& id,
-                                 std::unordered_map<VariantKey, double>& values)
+ReaderResult ConduitReader::getDoubleMap(const std::string& id,
+                                         std::unordered_map<VariantKey, double>& values)
 {
   return getDictionary(id, values);
 }
 
-bool ConduitReader::getBoolMap(const std::string& id,
-                               std::unordered_map<VariantKey, bool>& values)
+ReaderResult ConduitReader::getBoolMap(const std::string& id,
+                                       std::unordered_map<VariantKey, bool>& values)
 {
   return getDictionary(id, values);
 }
 
-bool ConduitReader::getStringMap(const std::string& id,
-                                 std::unordered_map<VariantKey, std::string>& values)
+ReaderResult ConduitReader::getStringMap(
+  const std::string& id,
+  std::unordered_map<VariantKey, std::string>& values)
 {
   return getDictionary(id, values);
 }
 
-bool ConduitReader::getIndices(const std::string& id, std::vector<int>& indices)
+ReaderResult ConduitReader::getIndices(const std::string& id,
+                                       std::vector<int>& indices)
 {
   indices.clear();
   const auto node = detail::traverseNode(m_root, id);
   if(node.dtype().is_empty())
   {
-    return false;
+    return ReaderResult::NotFound;
   }
   int num_elements = node.number_of_children();
   // Primitive arrays do not count as lists
@@ -306,37 +312,37 @@ bool ConduitReader::getIndices(const std::string& id, std::vector<int>& indices)
   indices.resize(num_elements);
   // Arrays in YAML/JSON are contiguous so we don't need to query the input file
   std::iota(indices.begin(), indices.end(), 0);
-  return true;
+  return ReaderResult::Success;
 }
 
-bool ConduitReader::getIndices(const std::string& id,
-                               std::vector<VariantKey>& indices)
+ReaderResult ConduitReader::getIndices(const std::string& id,
+                                       std::vector<VariantKey>& indices)
 {
   indices.clear();
   const auto node = detail::traverseNode(m_root, id);
   if(node.dtype().is_empty())
   {
-    return false;
+    return ReaderResult::NotFound;
   }
   if(!node.dtype().is_object())
   {
     // If it's not an object, try integer indexing
     std::vector<int> int_indices;
-    if(getIndices(id, int_indices))
+    const auto result = getIndices(id, int_indices);
+    if(result == ReaderResult::Success)
     {
       for(const int idx : int_indices)
       {
         indices.emplace_back(idx);
       }
-      return true;
     }
-    return false;
+    return result;
   }
   for(const auto& child : node.children())
   {
     indices.push_back(child.name());
   }
-  return true;
+  return ReaderResult::Success;
 }
 
 FunctionVariant ConduitReader::getFunction(const std::string&,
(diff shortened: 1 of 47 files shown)


white238:
https://axom.readthedocs.io/en/feature-essman-reader_result/axom/inlet/docs/sphinx/verification.html

src/axom/inlet/docs/sphinx/verification.rst
```diff
@@ -0,0 +1,55 @@
+.. _inlet_verification_page_label:
+
+############
+Verification
+############
+
+Before input file data can be accessed, it must first be verified by calling the ``verify()``
+method of the top-level ``Inlet`` object. This will return a ``bool`` indicating whether the
+provided input conformed to the schema, and specific violations of the schema are logged via
```
white238, line 9: This comma is not needed.

src/axom/inlet/docs/sphinx/verification.rst
```diff
@@ -0,0 +1,55 @@
...
+SLIC.
+
+This section describes the verification rules that apply to each possible element of the Inlet
+hierarchy, namely, ``Container``, ``Field``, and ``Function``.
+
+Container Verification
+----------------------
+
+If a ``Container`` is marked as required (via the ``required()``) method, then the ``Container`` must be nontrivial,
```
white238, line 18: not sure "nontrivial" brings anything to this sentence. Also the comma in "input, or contains a" is unneeded

src/axom/inlet/docs/sphinx/verification.rst
```diff
@@ -0,0 +1,55 @@
...
+
+.. note::
+  Since a ``Container`` represents internal nodes in the Inlet hierarchy, its verification status is
+  dependent on that of its child objects.  A ``Container`` is only considered valid if all of its child
+  ``Container``, ``Field``, and ``Function`` objects are also valid.
+
+Field Verification
+------------------
+
+If a ``Field`` is marked as required (via the ``required()``) method, then a value must be provided in the input.
```
white238, line 36: remove comma

white238: though i guess word disagrees with me.

src/axom/inlet/inlet_utils.cpp
```diff
@@ -165,5 +189,33 @@ void markAsStructCollection(axom::sidre::Group& target)
...
+  if(result != ReaderResult::Success)
+  {
+    target.createViewScalar("retrieval_status", static_cast<int>(result));
+  }
+}
+
+ReaderResult collectionRetrievalResult(const bool homogeneous, const bool empty)
+{
+  // First check if the collection was entirely the wrong type
+  if(!homogeneous && empty)
```
white238, line 203: !empty?

joshessman-llnl: The "empty" refers to the C++ container used to store the requested elements, which will be empty if none were present.  The name is a little confusing since it could also refer to the structure in the input file, so I've renamed it accordingly.

src/axom/inlet/inlet_utils.cpp
```diff
@@ -165,5 +189,34 @@ void markAsStructCollection(axom::sidre::Group& target)
...
+  {
+    target.createViewScalar("retrieval_status", static_cast<int>(result));
+  }
+}
+
+ReaderResult collectionRetrievalResult(const bool homogeneous,
+                                       const bool none_of_requested_type)
+{
+  // First check if the collection was entirely the wrong type
+  if(!homogeneous && none_of_requested_type)
```
white238, line 204: Reading this comment... should this then be "homogeneous && none_of_requested_type" or is the comment wrong?

joshessman-llnl: homogeneous is more like "none_of_other_type", I'll rename it

white238 (approved):

src/axom/inlet/ConduitReader.cpp
```diff
@@ -306,37 +312,37 @@ bool ConduitReader::getIndices(const std::string& id, std::vector<int>& indices)
   indices.resize(num_elements);
   // Arrays in YAML/JSON are contiguous so we don't need to query the input file
   std::iota(indices.begin(), indices.end(), 0);
-  return true;
+  return (num_elements == 0) ? ReaderResult::NotFound : ReaderResult::Success;
```
estebanpauli, line 315: I'm not sure if this is related to this code or not, but what happens with `foo: []`? Is it `NotFound`, or `Success`? In my mind, it should be `Success`. The field is there, but there just happens to be an empty list. Depending on the situation, the semantics might be the same, but you can't tell from this level.

src/axom/inlet/ConduitReader.cpp
```diff
@@ -306,37 +312,37 @@ bool ConduitReader::getIndices(const std::string& id, std::vector<int>& indices)
...
     }
-    return false;
+    return result;
   }
   for(const auto& child : node.children())
   {
     indices.push_back(child.name());
   }
-  return true;
+  return (indices.size() == 0) ? ReaderResult::NotFound : ReaderResult::Success;
```
estebanpauli, line 345: See previous comment about `Success` vs `NotFound` for empty lists.

src/axom/inlet/docs/sphinx/verification.rst
```diff
@@ -0,0 +1,54 @@
...
+SLIC.
+
+This section describes the verification rules that apply to each possible element of the Inlet
+hierarchy, namely, ``Container``, ``Field``, and ``Function``.
+
+Container Verification
+----------------------
+
+If a ``Container`` is marked as required (via the ``required()``) method, then the ``Container`` must
+have a ``Field`` or ``Function`` that was present in the input or contain a sub-``Container`` that does.
```
estebanpauli, line 19: Can a container not be empty? In yaml, if you tell me `foo` is required, both `foo: {}` and `foo: []` should be viewed as valid.

src/axom/inlet/docs/sphinx/verification.rst
```diff
@@ -0,0 +1,54 @@
...
+
+If a verification lambda was provided via the ``registerVerifier()`` method, this lambda must 
+return ``true`` when passed the corresponding ``Field`` object.
+
+Function Verification
+---------------------
+
+If a ``Function`` is marked as required (via the ``required()``) method, then a function must be provided in the input.
+
+If a verification lambda was provided via the ``registerVerifier()`` method, this lambda must
```
estebanpauli, line 53: Minor thing: this isn't necessarily a lambda. It's just something that is callable.

src/axom/inlet/docs/sphinx/verification.rst
```diff
@@ -0,0 +1,54 @@
...
+Container Verification
+----------------------
+
+If a ``Container`` is marked as required (via the ``required()``) method, then the ``Container`` must
+have a ``Field`` or ``Function`` that was present in the input or contain a sub-``Container`` that does.
+
+If a ``Container`` corresponds to an array or dictionary, the elements of the array must all be of the requested
+type.  This restriction applies even if the array/dictionary was not marked as ``required``.
+
+If a verification lambda was provided via the ``registerVerifier()`` method, this lambda must
```
estebanpauli, line 24: Not necessarily a lambda. Could be any callable.

src/axom/inlet/inlet_utils.cpp
```diff
@@ -165,5 +189,34 @@ void markAsStructCollection(axom::sidre::Group& target)
...
+  }
+  // Then check if some values were of the correct type, but others weren't
+  else if(!none_of_other_type)
+  {
+    return ReaderResult::NotHomogeneous;
+  }
+  // Finally check if nothing existed at all
+  else if(none_of_requested_type)
+  {
+    return ReaderResult::NotFound;
```
estebanpauli, line 216: How does this interact with empty collections?

src/axom/inlet/inlet_utils.hpp
```diff
@@ -174,6 +182,32 @@ inline bool isCollectionGroup(const std::string& name)
...
+* on whether the collection is empty and whether the corresponding collection
+* in the input file was homogeneous
+*
+* \param [in] none_of_other_type Whether any collection elements were of type
+* other than the requested type
+* \param [in] none_of_requested_type Whether the collection of requested type was empty,
+* i.e., if no elements of the requested type were present
+*****************************************************************************
+*/
+ReaderResult collectionRetrievalResult(const bool none_of_other_type,
```
estebanpauli, line 208: Booleans with negative names can be harder to work with. How about `all_of_other_type` and `all_of_requested_type`? Of course, you'd have to change the logic and call sites. Not sure it's worth it.

src/axom/inlet/ConduitReader.cpp
joshessman-llnl: I think this change should be possible, though Conduit's YAML parser appears to be behaving in a way that I wouldn't expect:  in the snippet you've provided, `foo` ends up being a completely empty `conduit::Node`, which is not the case with the equivalent JSON and Conduit's JSON parser.

white238: After speaking to @kennyweiss , I think we can add a couple verification functions to solve this and make it clear for everyone.

1) An empty, declared array is there so it should not fail verify()
2) Add two new functions called minLength() and maxLength()

@estebanpauli does this work for you?

src/axom/inlet/inlet_utils.hpp
white238: This is partial my fault... i was confused by the previous names.  They made sense at the call sites but not at the function level... at least to me.

src/axom/inlet/ConduitReader.cpp
estebanpauli: Are `minLength()` and `maxLength()` functions I call on the table to set the expected range of the array? If so, that works for me. If not, I'll need clarification.

white238: Yea the idea was you would do something like this at the defining part:

```
Foo = {}
```

```
inlet.addIntArray("Foo").required().minLength(1);
inlet.verify(); // fails due to lengh 0
```

joshessman-llnl: This logic has been updated to allow for empty collections.  The following tests demonstrate the new behavior:

https://github.com/LLNL/axom/blob/79490bbe954354a222d0cadc1979df16d32ef801/src/axom/inlet/tests/inlet_object.cpp#L165-L215

estebanpauli (approved):

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

white238 (approved):

joshessman-llnl:
LGTM

src/axom/inlet/inlet_utils.cpp
```diff
@@ -165,5 +192,28 @@ void markAsStructCollection(axom::sidre::Group& target)
   }
 }
 
+void markRetrievalStatus(axom::sidre::Group& target, const ReaderResult result)
+{
+  target.createViewScalar("retrieval_status", static_cast<int>(result));
```
kennyweiss, line 197: Not sure if this is the right place to do this, but would it make sense to add an entry for "USING_DEFAULT" when the user didn't supply an entry, so inlet uses the default value.

(this might be a separate annotation though, and/or might already be supported in another way)

src/axom/inlet/tests/inlet_Reader.cpp
```diff
@@ -132,36 +151,54 @@ TYPED_TEST(inlet_Reader, getMap)
...
+  ReaderResult retValue = reader.getIntMap("luaArray", ints);
+  EXPECT_EQ(retValue, ReaderResult::NotHomogeneous);
   std::unordered_map<int, int> expectedInts {{0, 4}, {1, 5}, {2, 6}, {5, 2}};
   EXPECT_EQ(expectedInts, ints);
 
   std::unordered_map<int, double> doubles;
-  found = reader.getDoubleMap("luaArray", doubles);
-  EXPECT_TRUE(found);
+  retValue = reader.getDoubleMap("luaArray", doubles);
+  EXPECT_EQ(retValue, ReaderResult::NotHomogeneous);
```
kennyweiss, line 161: How does casting work?  And are there tests for `ReaderResult::NotHomogenous` specifically related to casting?

E.g. if the `testString` above only contained `int` and `double` types, and you called
* `reader.getDoubleMap("luaArray", doubles);`
* `reader.getIntMap("luaArray", ints);
I'd imagine the former would cast the ints to doubles and return Success, but the latter would return `NotHomogenous`

(apologies if this is already there, it can be a bit difficult to see it via the github interface)

src/axom/inlet/docs/sphinx/verification.rst
```diff
@@ -0,0 +1,55 @@
...
+
+This section describes the verification rules that apply to each possible element of the Inlet
+hierarchy, namely, ``Container``, ``Field``, and ``Function``.
+
+Container Verification
+----------------------
+
+If a ``Container`` is marked as required (via the ``required()``) method, then the ``Container`` must
+have a ``Field`` or ``Function`` that was present in the input or contain a sub-``Container`` that does.
+This does not apply to a ``Container`` that corresponds to an array or dictionary.
```
kennyweiss, line 20: Re: `verify()` w/ `required()` -- empty vs. non-empty and containers? 
Perhaps an example help clarify this.

kennyweiss (approved):
Thanks @joshessman-llnl !

src/axom/inlet/tests/inlet_Reader.cpp
joshessman-llnl: This is a good question - to provide a little background on integral <-> floating-pt promotion/demotion in inlet, most of the implementation is restricted by Lua's decision to use the same type for integral and floating-point values.  To maintain consistency, I implemented special promotion/demotion logic in the YAML/JSON readers so they'd behave like the Lua reader for consistency.

As such, Inlet doesn't provide a way to see if an integral <-> floating-pt conversion occurred and therefore casting doesn't (can't?) affect the homogeneity of an array.

src/axom/inlet/inlet_utils.cpp
joshessman-llnl: I hadn't considered this until you mentioned it, but I think the information we're adding to the datastore here is sufficient to determine if the default value was used (which was impossible in the general case before this PR).  

I think if we wanted to later report this information (for documentation purposes, maybe), we could just check if the retrieval status was `Success` - if not, the default would have been used.

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

Issue #471, opened 2021-02-12
Unexpected fields in Inlet


estebanpauli:
There is currently no way to find unexpected fields in Inlet. For example, if a user inputs
```yaml
foo: 123
bar: 456
```
but the schema only defines the `foo` field, there is no way to see if there are extra, unexpected fields like `bar`. This can be a useful feature when validating input. In something like yaml, if a field is not defined by the schema, it can be an indication that a user miss-typed an optional field, or tried to add a field that maybe works for some objects but not others. There is currently no way to warn the user about these sorts of situations. It would be nice if there were.

cyrush:
Yes --- Typos can also be an issue when there are optional params. 

(Like if you can optionally accept `field/z`, but some one puts in `field/zz`   -- `field/zz`  isn't wrong, but its also likely not what was intended )

In cases where this helps in ascent, we provide a surprise check, here is an example:

https://github.com/Alpine-DAV/ascent/blob/88057dc29ef1cc70b32f9dd7efa07872e4378507/src/ascent/runtimes/flow_filters/ascent_runtime_vtkh_filters.cpp#L1233

joshessman-llnl:
After some discussion with @white238, we determined that it would be useful to support two options:

- A `strict` attribute for `Table`s that causes verification failure when unexpected fields are present
- Retrieving the global (across the Inlet hierarchy) list of unexpected fields for direct processing by the user

PR #479, opened 2021-02-22
Detect unexpected entries in input files

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Refactors the top-level `Inlet` object to store a set of unexpected fields that get updated every time an "access" is made through the `Reader`
  - Fixes (partially) #471
  - Adds a method to the top-level `Inlet` object to retrieve the set of unexpected names

A follow-up PR will add an option to mark a `Container` as "strict", which will fail verification if the set of unexpected names for that container is non-empty.

An implementation note:  the set of unexpected names is "global" to the Inlet hierarchy due to the ambiguity in assigning "ownership" of paths.  Without knowing the structure of the Inlet tree *a priori*, it would be fairly complex to assign paths to the correct (deepest) `Container` in the hierarchy.  The chosen implementation also makes it easy to retrieve the full set of unexpected items in the input file.


white238 (approved):

white238:
LGTM

joshessman-llnl:
LGTM

src/axom/inlet/ConduitReader.cpp
```diff
@@ -157,6 +157,31 @@ void arrayToMap(const conduit::DataArray<ConduitType>& array,
   }
 }
 
+/*!
+ *******************************************************************************
+ * \brief Recursive name retrieval function
```
estebanpauli, line 162: What names are retrieved? It's not clear from this description.

src/axom/inlet/ConduitReader.cpp
```diff
@@ -157,6 +157,31 @@ void arrayToMap(const conduit::DataArray<ConduitType>& array,
...
+ * 
+ * \param [in] node The Conduit node to "visit"
+ * \param [out] names The set of paths to add to
+ *******************************************************************************
+ */
+void nameRetrievalHelper(const conduit::Node& node,
+                         std::unordered_set<std::string>& names)
+{
+  // Conduit paths use [0] for array indices, Inlet does not, so they need
+  // to be removed - e.g., foo/[0]/bar vs foo/0/bar
```
estebanpauli, line 172: Is this the only time "[" and "]" can appear in the names? What if I make my json key "foo[bar]"?

src/axom/inlet/Container.cpp
```diff
@@ -45,6 +45,7 @@ Container& Container::addContainer(const std::string& name,
                                              "",
                                              m_reader,
                                              m_sidreRootGroup,
+                                             m_unexpected_names,
```
estebanpauli, line 48: The naming convention doesn't match the other fields here. Should be `m_unexpectedNames`.

src/axom/inlet/ConduitReader.cpp
```diff
@@ -157,6 +157,31 @@ void arrayToMap(const conduit::DataArray<ConduitType>& array,
...
+  auto filter_name = [](std::string name) {
```
estebanpauli, line 173: Should this be `filterName`? I don't remember all the axom naming conventions.

src/axom/inlet/Container.cpp
```diff
@@ -578,6 +582,24 @@ void addIndexViewToGroup(sidre::Group& group, const VariantKey& index)
   }
 }
 
+void updateUnexpectedNames(const std::string& accessed_name,
```
estebanpauli, line 585: Naming conventions: `accessed_name` -> `accessedName` and `unexpected_names` -> `unexpectedNames`.

src/axom/inlet/Container.hpp
```diff
@@ -269,6 +269,21 @@ bool matchesKeyType(const VariantKey& key)
...
+/*!
+ *******************************************************************************
+ * \brief Updates the set of unexpected names to reflect an user-requested access
+ * 
+ * \param [in] accessed_name The path within the input file that will be accessed
+ * \param [inout] unexpected_names The set of input file paths that have not yet
+ * been requested by the user
+ * 
+ * \note To maintain consistency, this function should always be followed by an
+ * access to a Reader
```
estebanpauli, line 281: "To maintain consistency, this function should always be followed by an access to a Reader". This seems easy to mess up. Is there a pattern you can use to avoid this? For example, what about an Reader which wraps another reader and keeps track of the unexpected names? With that pattern, `m_unexpected_names` would become a field of the wrapper proxy.

src/axom/inlet/LuaReader.cpp
```diff
@@ -26,6 +26,77 @@ namespace axom
...
+ * the input file
+ * \param [out] names The set of paths to add to
+ *******************************************************************************
+ */
+void nameRetrievalHelper(const std::unordered_set<std::string>& ignores,
+                         const sol::table& table,
+                         const std::string& prefix,
+                         std::unordered_set<std::string>& names)
+{
+  auto to_string = [](const VariantKey& key) {
```
estebanpauli, line 77: Local variables use snake case. Should they use camel case?

src/axom/inlet/Reader.hpp
```diff
@@ -244,6 +245,17 @@ class Reader
...
+  /*!
+   *****************************************************************************
+   * \brief Retrieves all paths present in the input file
+   *
+   * \return The set of all paths/full names in the input file - this represents
+   * a full traversal of the input file "tree" and includes paths to array/dictionary
+   * entries
+   *****************************************************************************
+   */
+  virtual std::unordered_set<std::string> getAllNames() = 0;
```
estebanpauli, line 258: This seems like an odd approach. I would have expected, for example, each Table to have a list of all the names defined at that level.

white238: Should we support both?  Having the Container (new name for Table) hold a list of all names which would allow power users to traverse the tree if they like.  But also have helper functions at a global level?

white238:
After meeting with @joshessman-llnl and @estebanpauli , we will have a follow up PR that adds a `strict()` function that will make a container required to not have unneeded names.  At the same time we will add helper functions at the Container level to return all unneeded names at that level and below.

white238:
Can you add some documentation for this under the verification page?

src/axom/inlet/Container.hpp
joshessman-llnl: I had considered something like this but was concerned about the additional complexity that it introduces.  I agree that it's possibly error-prone but on the other hand it's an internal method that currently only gets used in four places (in a type-independent context).  Any thoughts @white238?

src/axom/inlet/ConduitReader.cpp
joshessman-llnl: I've adjusted the container registration logic to check for these characters

src/axom/inlet/Container.hpp
```diff
@@ -269,6 +269,21 @@ bool matchesKeyType(const VariantKey& key)
   return false;
 }
 
+/*!
+ *******************************************************************************
+ * \brief Updates the set of unexpected names to reflect an user-requested access
+ * 
+ * \param [in] accessed_name The path within the input file that will be accessed
```
white238, line 276: casing

src/axom/inlet/Container.hpp
white238: I agree this is a little confusing/easy to mess up but it is only in a few places and those can be handled by adding unit tests specifically for this.  I'm not sure the extra complexity is worth it at this point since it is not user-facing and I can't think of anything else that would go into this class. If we do we should add this language-agnostic Reader wrapper (i hope it is named `ReaderKeeper`).

kennyweiss (approved):
This is a nice addition @joshessman-llnl 

I think it would be helpful if you could add this functionality to one of the inlet examples, which might give a better idea of how this might be used in practice.

E.g. add it to a `try`/`catch` and print an error/exception on bad input decks to the example.

joshessman-llnl:
> 
> 
> This is a nice addition @joshessman-llnl
> 
> I think it would be helpful if you could add this functionality to one of the inlet examples, which might give a better idea of how this might be used in practice.
> 
> E.g. add it to a `try`/`catch` and print an error/exception on bad input decks to the example.

This PR is intended to be an intermediate step before the addition of a `strict()` option (used in the same way as `required()`) that causes verification to fail.  I'll add a snippet to the `nested_structs` example that roughly shows how the `strict` option works with a TODO comment to remove/replace that snippet in the follow-up PR.

PR #480, opened 2021-02-22
Refactor iteration over nested collection elements

joshessman-llnl:
# Summary

- This PR is a refactoring
- It does the following:
  - Moves the duplicated iteration logic in the `inlet::Container` implementation to a single function that accepts a lambda

This should prevent bugs like #467, where changes to iteration logic are not propagated to all the places that need it.  Instead, only the unique logic is present at each call site.


white238 (approved):

src/axom/inlet/Container.hpp
```diff
@@ -1231,6 +1231,34 @@ class Container : public Verifiable<Container>
   template <typename Func>
   void forEachCollectionElement(Func&& func) const;
 
+  /*!
+   *****************************************************************************
+   * \brief Applies a provided function to nested elements of the calling table
+   * and stores the result in a range
+   * 
+   * \pre The function must accept two arguments of type Table& and
```
kennyweiss, line 1239:
```suggestion
   * \pre The function \a func must accept two arguments of type Table& and
```

src/axom/inlet/Container.hpp
```diff
@@ -1231,6 +1231,34 @@ class Container : public Verifiable<Container>
   template <typename Func>
   void forEachCollectionElement(Func&& func) const;
 
+  /*!
+   *****************************************************************************
+   * \brief Applies a provided function to nested elements of the calling table
+   * and stores the result in a range
```
kennyweiss, line 1237: On my first pas through the docs, I am not sure what "...in a range" means at the end

src/axom/inlet/Container.hpp
```diff
@@ -1231,6 +1231,34 @@ class Container : public Verifiable<Container>
...
+   * 
+   * \pre The function must accept two arguments of type Table& and
+   * const std::string&, respectively.  
+   * 
+   * This function will pass to the provided function the nested table
+   * as the first argument and the path of the nested element in the input file
+   * as the second argument, when applicable.
+   * 
+   * \param [out] output An iterator to the beginning of the output range
```
kennyweiss, line 1246: Minor / convention -- is it weird for the first parameter to be an output parameter, while the others are input parameters? Would users expect the output parameters to be at the end?

src/axom/inlet/Container.hpp
kennyweiss: Update after seeing the rest of the docs -- the results are put into an output iterator ``\a output``

src/axom/inlet/Container.hpp
kennyweiss: Also, can you please check that doxygen doesn't choke on the ampersand in `Table&`

kennyweiss (approved):
Thanks @joshessman-llnl 

This cleans up the code a bit, but also makes its internals more complex to use.

Thinking a bit about future maintainers, I think it would make sense to add a test for this function to help people see (in isolation) how it is meant to be used.

src/axom/inlet/Container.hpp
joshessman-llnl: Good question - STL algorithms like [std::transform](https://en.cppreference.com/w/cpp/algorithm/transform) typically always keep the callable at the end for readability when passing a lambda as an argument, and that's what I've opted to go with there.  I don't have a preference regarding the order of the first two, though, so I can flip those if you'd prefer.

kennyweiss: That makes sense.  I'm ok with it as is.

src/axom/inlet/Container.hpp
joshessman-llnl: I think Doxygen doesn't render private methods so this would only be for developer use

joshessman-llnl:
> 
> Thinking a bit about future maintainers, I think it would make sense to add a test for this function to help people see (in isolation) how it is meant to be used.

This might be a little difficult to implement given that it's a private method - there are of course workarounds but I'm not sure what the tradeoff is.

white238:
> > Thinking a bit about future maintainers, I think it would make sense to add a test for this function to help people see (in isolation) how it is meant to be used.
> 
> This might be a little difficult to implement given that it's a private method - there are of course workarounds but I'm not sure what the tradeoff is.

This is being tested at a high level but would be hard to test this specifically at a unit test level.  I think this is fine to merge.

PR #481, opened 2021-02-22
Required sub-objects of empty objects should not impede verification

### LINKED ISSUE #442: Have required sub-objects in inlet with an optional parent object
I've started to look into this, and recalled a relevant comment from a recent inlet PR:
https://github.com/LLNL/axom/pull/427#discussion_r556825102

I think the idea of only checking sub-objects when the parent object is defined makes sense, but it might be a little tricky to determine when the parent table is defined.  Currently this is based on the existence of at least one expected subobject, so `elements = { q = 1 }` would not count as "defined", but `elements = { x = 1 }` would.  

I suppose we could add logic to just check if the `elements` table is a valid "path" in the input file and go from there.

_Originally posted by @joshessman-llnl in https://github.com/LLNL/serac/pull/334#discussion_r562138672_

### LINKED ISSUE #474: Warning on optional fields
When running the test below, I get unexpected output to the log and an unexpected failure.

```cpp
TEST(inlet_Inlet, requiredSubItems)
{
    reader->parseString(R"(
    my_items: []
    )");
    axom::sidre::DataStore ds;

    auto &item = inlet.addStructArray("my_items");
    item.addString("some_field").required();
    item.addStruct("required_sub").required();
    auto &sub_item = item.addStruct("optional_sub");
    sub_item.addString("a_distant_field").required();

    EXPECT_TRUE(inlet.verify());
}
```

Log output:
```
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from inlet_Inlet
[ RUN      ] inlet_Inlet.requiredSubItems

MESSAGE=[Inlet] Required Field not specified: my_items/_inlet_container/optional_sub/a_distant_field

MESSAGE=[Inlet] Required Table not specified: my_items/_inlet_container/required_sub
/usr/WS1/pauli2/axom/src/axom/inlet/tests/inlet_Inlet.cpp:1650: Failure
Value of: inlet.verify()
  Actual: false
Expected: true
[  FAILED  ] inlet_Inlet.requiredSubItems (8 ms)
[----------] 1 test from inlet_Inlet (8 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (8 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] inlet_Inlet.requiredSubItems

 1 FAILED TEST
```

Since `"my_items"` is optional, I expect no output and the test to succeed. Even if `"my_items"` were required, I would still expect no output and for the test to succeed since the field is present (it is just an empty list).

Note from the output that no error is emited from `"some_field"` not being present, but it is for `"required_sub"` and `"optional_sub/a_distant_field"`.

joshessman-llnl:
# Summary

- This PR is a bugfix
- It does the following:
  - Modifies the verification logic in `inlet::Container` to not attempt to verify child objects when the current object is empty
  - Fixes #442, fixes #474 (duplicate)

As of this PR, a container is considered empty when it (or any of its sub-containers) do not have any requested fields or functions provided in the input file.  This means that a container containing (only) incorrectly named fields will count as empty:

```lua
foo = { baz = false } -- misspelled 'bar' as 'baz', for example
```

```cpp
auto& foo_table = inlet.addStruct("foo");
foo_table.addBool("bar").required();
// Verification will succeed here because "foo" is considered empty, and consequently
// its fields are not verified
assert(inlet.verify());
```

Adding a `strict` annotation in the follow-up to #479 may allow the internal definition of "empty" to be relaxed a bit such that the case above results in verification failure.


jamiebramwell (approved):
This will be very useful! Thanks!

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -77,6 +77,38 @@ TYPED_TEST(inlet_object, simple_struct_by_value)
...
+
+  auto& foo_table = inlet.addStruct("foo");
+  foo_table.addBool("bar", "bar's description").required(true);
+  foo_table.addBool("baz", "baz's description").required(true);
+
+  // It should fail because a) the struct exists, and
+  // b) one of the required fields is not present
+  // Note that the only reason the struct is marked as present
+  // is because at least one of its expected fields is present, i.e.,
+  // having a foo = { quux = true } will pass verification
```
estebanpauli, line 108: This sounds wrong. Why would that pass verification?

jamiebramwell:
I just hit another related issue: custom verifiers in optional objects.

RELEASE-NOTES.md
```diff
@@ -77,6 +77,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Inlet: Fixed a bug relating to nested table lookups of primitive arrays and functions
 - Fixed a bug relating to deeply nested callback functions in Inlet
 - Inlet: Always ignore primitive array elements that do not match the requested type
+- Inlet: Empty structs/collections of structs with required sub-elements no longer fail
```
white238, line 80: "Optional empty structs/collections of structs with required sub-elements no longer fail"?

joshessman-llnl: `verification` is on the next line - I think it renders properly: https://github.com/LLNL/axom/blob/2bbab0e52f1d7964bc70e6be13af094aca3219e1/RELEASE-NOTES.md

src/axom/inlet/tests/inlet_object.cpp
white238: It fails because while the "foo" structure is optional, the fields "bar" and "baz" are marked required.  So the logic is if "foo" is present, it MUST have both "bar" and "baz". 

If there was no "foo" at all, it would pass verification.

Did i explain that better, @estebanpauli ?

If this is not already an example in the documentation it should be added.

RELEASE-NOTES.md
white238: Sorry i was refering to adding "Optional" explicitly.

white238 (approved):

src/axom/inlet/tests/inlet_object.cpp
white238: Ah you were referring to the comment.

white238: Can this be fixed in a follow-up PR?

```
foo = {}
```

Should also fail...

joshessman-llnl: Yes, fixing this requires #479 so it will have to be in a follow-up

src/axom/inlet/Container.cpp
```diff
@@ -816,41 +816,48 @@ Container& Container::registerVerifier(std::function<bool(const Container&)> lam
...
+
+  // Checking the child objects is not needed if the container is empty
+  if(this_container_defined)
   {
-    if(!field.second->verify())
+    // Verify the child Fields of this Container
+    for(const auto& field : m_fieldChildren)
     {
-      verified = false;
+      verified &= field.second->verify();
```
kennyweiss, line 841: Technically, `&=` does not have the intended semantics of `&&=`.

It looks like it is probably ok in this case, but these should probably be converted to
```
       verified = verified && field.second->verify(); 
```

kennyweiss (approved):
Thanks @joshessman-llnl !

src/axom/inlet/Container.cpp
joshessman-llnl: I think in this context the difference is short-circuiting doesn't apply? I agree that it makes sense to change it since it could be confusing to use bitwise operators for logical operations.

estebanpauli: FWIW, I think `&=` has the intended semantics here. You're not trying to short-circuit. Each verification step can log more diagnostics, so you want to run them regardless of earlier errors.

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -780,6 +812,29 @@ TYPED_TEST(inlet_object, nested_array_of_nested_structs)
...
+TYPED_TEST(inlet_object, nested_array_of_struct_verify_pass)
+{
+  std::string testString = "quux = { }";
+  DataStore ds;
+  Inlet inlet = createBasicInlet<TypeParam>(&ds, testString);
+
+  auto& quux_schema = inlet.addStructArray("quux");
+
+  // Simple required field
+  quux_schema.addString("corge").required();
```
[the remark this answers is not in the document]
estebanpauli: `"quux"` is an array of structs. If it had a any elements, each would require a `"corge"` member. Since the array is empty, it's fine.

I think this is the kind of confusion that stems from `addStructArray()` returning the definition to the items of the array, rather than the array itself. There's no way to distinguish between the two. This is outside the scope of this PR, though.

estebanpauli (approved):

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

PR #488, opened 2021-03-03
Move Uberenv to submodule

white238:
* Moved uberenv to a submodule
* Updates to .uberenv_config.json
* Moved devtools spack configs into a new directory "scripts/spack/devtools_config"
* Moved all spack configs and packages into new directory "scripts/spack"
* Fixes various stuff that was broken in devtools
* Turns off docs on lassen since it's causing problems in CI


white238:
LGTM

src/axom/slic/interface/slic.cpp
```diff
@@ -16,7 +16,7 @@ namespace slic
 // Initialize static variables for controlling runtime behavior of asserts and
 // error macros.
 //------------------------------------------------------------------------------
-AXOM_API bool debug::checksAreErrors = false;
+bool debug::checksAreErrors = false;
```
agcapps, line 19: Why is this removed?

src/axom/slic/interface/slic.hpp
```diff
@@ -25,7 +25,7 @@ namespace slic
 {
 struct debug
 {
-  AXOM_API static bool checksAreErrors;
+  static bool checksAreErrors;
```
agcapps, line 28: Likewise, why is `AXOM_API` removed?  Are all symbols exported from Axom by default?

agcapps (commented):
Thank you, Chris.  Just two questions about AXOM_API being removed.

src/axom/slic/interface/slic.cpp
kennyweiss: I did that.  

msvc 2019 had about 100 link warnings on this line 
(LNK4286 -- https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-warning-lnk4286?view=msvc-160)

I think it is related to `dllimport`/`dllexport` for a function that is being used in the same object library.
We should confirm that this will not cause an issue with any of our Windows users, and if so, maybe add a `#pragma` to silence the warning.

scripts/vcpkg_ports/conduit/CONTROL
```diff
@@ -1,5 +1,5 @@
 Source: conduit
-Version: v0.5.1
+Version: v0.6.0
```
kennyweiss, line 2: I had to upgrade to `conduit@0.6.0` on Windows due to `node::to_string()`
I think this matches our spack version.

kennyweiss (approved):
This PR also fixes `inlet` on Windows -- it previously had compiler errors.

There are a few failing inlet tests on Windows, due to an apparent conflict between `long` vs. `int32` vs `int64`.  I'll create an issue for that. This won't impact the current CI on Windows since these `inlet`, and hence these tests, depend on `sidre`.

src/axom/slic/interface/slic.cpp
agcapps: all right

agcapps (approved):
We can resolve issues with exporting symbols later.

Thanks for the work, Chris and Kenny.

PR #492, opened 2021-03-08
fmt on xlc+nvcc

white238:
* Removes unneeded files from newer fmt
* Pulls a constexpr out of the double ternary operator to simplify logic to the point where xlc+nvcc is happy

Thanks @davidbeckingsale and @joshessman-llnl


white238:
After I run this through our battery of tests I'm gonna submit an issue or PR to fmt.. but this should also get merged into conduit's version @cyrush .

src/thirdparty/fmt/core.h
```diff
@@ -1571,12 +1571,10 @@ class format_arg_store
 
   friend class basic_format_args<Context>;
 
+  static constexpr unsigned long long type_encoding = detail::encode_types<Context, Args...>();
```
kennyweiss, line 1574: Should we put this in a patch file and/or add comments/tags in the file to make it easier to see this patch?

kennyweiss (approved):
Thanks for tracking this down @white238  (and others)!

Is there a simple reproducer we could provide to the vendors for the XL/nvcc compiler bug?

src/thirdparty/fmt/core.h
white238: Yea I'll push that in a bit.. was just so excited to finally track this down.

white238:
> Thanks for tracking this down @white238 (and others)!
> 
> Is there a simple reproducer we could provide to the vendors for the XL/nvcc compiler bug?

Josh was gonna attempt to narrow this down to a "simple" reproducer tomorrow.

cyrush:
Nice catch!

white238:
LGTM

white238:
This has passed all specs.

joshessman-llnl:
> 
> 
> > Thanks for tracking this down @white238 (and others)!
> > Is there a simple reproducer we could provide to the vendors for the XL/nvcc compiler bug?
> 
> Josh was gonna attempt to narrow this down to a "simple" reproducer tomorrow.

So I was initially just looking to how minimally the behavior we were seeing with `fmt` could be reproduced, and was able to get to this:
```cpp
constexpr int get_bar()
{
  return 0;
}

struct Foo
{
  // static constexpr int bar = 0; // Works with both foo.bar and Foo::bar
  static constexpr int bar = get_bar(); // Fails with foo.bar
};

int main()
{
  constexpr int y = Foo::bar; // Works just fine with get_bar()
  Foo foo;
  constexpr int x = foo.bar; // Fails with get_bar() - undefined reference to Foo::bar
  return 0;
}
```

The static member access through the `operator.` is a little unconventional but it's how `fmt` uses the `desc` variable that was modified in this PR.

I was still curious as to why it only failed with NVCC + XLC with `-x cu` and not with just XLC or NVCC + XLC not compiled as CUDA, so I started digging into the intermediate files.  It looks like the CUDA C++ frontend (`cudafec++`) parenthesizes the `foo.bar`:

```cpp
constexpr int x = (foo.bar);
```

and this is what triggers the XLC bug.  After some further investigation it looks like XLC (run standalone, not through NVCC) only fails to link the `cudafe++`-parenthesized version when debug symbols are enabled (`-g`) and with zero optimizations enabled (`-O0`)

In summary I think the XLC bug requires the following;
- Initializing of a `constexpr static` member with the result a `constexpr` function instead of a literal
- Access to that `constexpr static` member via an instance of the class instead of `X::s`
- Access via class instance in a parenthesized expression
- Debug symbols and no optimizations

agcapps (approved):
Heroism in the face of epic obscurity, distilled down to a few lines.

PR #495, opened 2021-03-11
Adjust Field::operator bool to ignore default values

joshessman-llnl:
# Summary

- This PR is a feature/bugfix
- It does the following:
  - Modifies `Field::operator bool` to ignore default values

This is sort of an extension of the discussion here: https://github.com/LLNL/axom/pull/478#discussion_r583117072
and a follow-up to #481.  With this latest set of changes, a `Container` will now only be verified if something was specified in the input file (the desired behavior) and *not* when defaults were provided in `Field`s within that `Container` (the current behavior).

### DIFF
--- a/src/axom/inlet/Field.hpp
+++ b/src/axom/inlet/Field.hpp
@@ -314,10 +314,18 @@ public:
 
   /*!
    *****************************************************************************
+   * \brief Returns whether a value for the Field exists, i.e., if a value 
+   * was provided in the input file or if a default was provided
+   *****************************************************************************
+   */
+  bool exists() const;
+
+  /*!
+   *****************************************************************************
    * \brief Returns whether a value was provided in the input file
    *****************************************************************************
    */
-  explicit operator bool() const;
+  bool isUserProvided() const;
 
 private:
   /*!
--- a/src/axom/inlet/Field.hpp
+++ b/src/axom/inlet/Field.hpp
@@ -314,10 +314,10 @@ public:
 
   /*!
    *****************************************************************************
-   * \brief Returns whether an actual value is stored
+   * \brief Returns whether a value was provided in the input file
    *****************************************************************************
    */
-  explicit operator bool() const { return m_sidreGroup->hasView("value"); }
+  explicit operator bool() const;
 
 private:
   /*!
--- a/src/axom/inlet/tests/inlet_object.cpp
+++ b/src/axom/inlet/tests/inlet_object.cpp
@@ -946,9 +946,10 @@ TYPED_TEST(inlet_object, default_scalar_marked_false)
   Inlet inlet = createBasicInlet<TypeParam>(&ds, testString);
 
   auto& scalar = inlet.addInt("foo").defaultValue(2);
-  // The field itself should evaluate to false because nothing was provided
+  // The field itself exists but was not provided by the user
   auto& field = static_cast<axom::inlet::Field&>(scalar);
-  EXPECT_FALSE(static_cast<bool>(field));
+  EXPECT_TRUE(field.exists());
+  EXPECT_FALSE(field.isUserProvided());
 
   // ...but it should still be possible to retrieve the default
   const int foo = inlet["foo"];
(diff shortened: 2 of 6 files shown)


src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -939,6 +939,69 @@ TYPED_TEST(inlet_object, struct_arrays_as_std_vector)
...
+TYPED_TEST(inlet_object, default_scalar_marked_false)
+{
+  std::string testString = " ";
+  DataStore ds;
+  Inlet inlet = createBasicInlet<TypeParam>(&ds, testString);
+
+  auto& scalar = inlet.addInt("foo").defaultValue(2);
+  // The field itself should evaluate to false because nothing was provided
+  auto& field = static_cast<axom::inlet::Field&>(scalar);
+  EXPECT_FALSE(static_cast<bool>(field));
```
kennyweiss, line 951: The semantics of casting `bool()` to `false` when using a default value seems odd to me, 
e.g. as opposed to a function like `Field::isUserProvided()`  (but obviously a better name than that).

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -939,6 +939,69 @@ TYPED_TEST(inlet_object, struct_arrays_as_std_vector)
   EXPECT_EQ(foos, expected_foos);
 }
 
+TYPED_TEST(inlet_object, default_scalar_marked_false)
+{
+  std::string testString = " ";
+  DataStore ds;
+  Inlet inlet = createBasicInlet<TypeParam>(&ds, testString);
+
+  auto& scalar = inlet.addInt("foo").defaultValue(2);
```
kennyweiss, line 948: It might also be good to test how this new feature interoperates with `.required()`

src/axom/inlet/Field.hpp
```diff
@@ -314,10 +314,10 @@ class Field : public VerifiableScalar
 
   /*!
    *****************************************************************************
-   * \brief Returns whether an actual value is stored
+   * \brief Returns whether a value was provided in the input file
```
kennyweiss, line 317: I like this functionality, but I am not sure that it is intuitive for it to be the implementation for `operator bool()`

There is likely a global view of how this all works that I'm missing though...

kennyweiss (approved):
Thanks @joshessman-llnl

src/axom/inlet/Field.hpp
joshessman-llnl: The motivation for this specific change is consistency with the logic added in #481 regarding when a `Container` is verified - I think it would be fine to move this to an explicitly named method (`inInputFile`?), but that raises the question of what `operator bool` should do, or if these objects even have an intuitive definition of `operator bool`.

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -939,6 +939,72 @@ TYPED_TEST(inlet_object, struct_arrays_as_std_vector)
...
+  // The field itself exists but was not provided by the user
+  auto& field = static_cast<axom::inlet::Field&>(scalar);
+  EXPECT_TRUE(field.exists());
+  EXPECT_FALSE(field.isUserProvided());
```
white238, line 952: Thanks @joshessman-llnl  i think this is much more clear.

white238 (approved):

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -968,8 +969,9 @@ TYPED_TEST(inlet_object, default_struct_field_marked_false)
   foo_container.addBool("bar", "bar's description").defaultValue(true);
   foo_container.addBool("baz", "baz's description").defaultValue(false);
 
-  // The struct itself should evaluate to false because nothing was provided
-  EXPECT_FALSE(static_cast<bool>(foo_container));
+  // The container itself exists but was not provided by the user
+  EXPECT_TRUE(foo_container.exists());
```
kennyweiss, line 973: 👍

kennyweiss (approved):
Thanks @joshessman-llnl! 

I agree that both the semantics and readability are improved with `.exists()` and `.userProvided()` instead of the implicit cast to `bool`

PR #498, opened 2021-03-21
Augments sidre_native_layout tests with zero-sized arrays

kennyweiss:
# Summary

- Attempting to track down a bug with zero-sized arrays when opening blueprint meshes in VisIt

### DIFF
--- a/src/axom/sidre/core/Buffer.cpp
+++ b/src/axom/sidre/core/Buffer.cpp
@@ -194,8 +194,11 @@ Buffer* Buffer::copyBytesIntoBuffer(void* src, IndexType nbytes)
 {
   if(src == nullptr || nbytes < 0 || nbytes > getTotalBytes())
   {
-    SLIC_CHECK_MSG(src != nullptr,
-                   "Cannot copy data into Buffer from null pointer.");
+    if(src == nullptr)
+    {
+      SLIC_CHECK_MSG(nbytes == 0,
+                     "Cannot copy data into Buffer from null pointer.");
+    }
     SLIC_CHECK_MSG(nbytes >= 0, "Cannot copy < 0 bytes of data into Buffer.");
     SLIC_CHECK_MSG(nbytes <= getTotalBytes(),
                    "Unable to copy " << nbytes << " bytes of data into Buffer with "
@@ -299,8 +302,14 @@ void Buffer::importFrom(conduit::Node& buffer_holder)
   {
     allocate();
     conduit::Node& buffer_data_holder = buffer_holder["data"];
-    copyBytesIntoBuffer(buffer_data_holder.element_ptr(0),
-                        buffer_data_holder.total_strided_bytes());
+
+    const auto num_bytes = buffer_data_holder.total_strided_bytes();
+    if(num_bytes > 0)
+    {
+      copyBytesIntoBuffer(buffer_data_holder.element_ptr(0), num_bytes);
+    }
+
+    this->print();
   }
 }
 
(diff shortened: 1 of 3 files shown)


kennyweiss:
I was trying to track down an issue that appears to be related to zero-sized sidre/conduit arrays.

I tested this both here and in conduit (https://github.com/LLNL/conduit/pull/732) and was not able to reproduce the problem (i.e. both codes were fine reading in and validating these files), so the problem appears to be elsewhere.

src/axom/sidre/tests/spio/spio_parallel.hpp
```diff
@@ -92,6 +95,10 @@ TEST(spio_parallel, parallel_writeread)
     i1_vals[i] = (i + 10) * (404 - my_rank - i);
   }
 
+  // Add an array view of size 0 under group "c"
+  Group* gc = flds2->createGroup("c");
+  gc->createView("i2")->allocate(DataType::c_int(0));
```
kennyweiss, line 100: I found something!  
Adding a zero-sized array to the `spio_parallel` test generates the following warning:

```
188: [ RUN      ] spio_parallel.parallel_writeread
188: 
188: ***********************************
188: [WARNING in line 198 of file <axom>/src/axom/sidre/core/Buffer.cpp]
188: MESSAGE=Failed Check: src != nullptr
188: Cannot copy data into Buffer from null pointer.
188: 
188: ***********************************
188: [       OK ] spio_parallel.parallel_writeread (125 ms)
```

So, the issue might be related to how we write/read these arrays to `hdf5` through `spio`.

kennyweiss: This was not the problem!

I found and fixed an overeager warning when importing zero-sized arrays from `hdf5` to `sidre` via `conduit`.
But, the code was already doing the right thing, it was just adding an extra warning.

src/axom/sidre/core/Buffer.cpp
```diff
@@ -299,8 +302,14 @@ void Buffer::importFrom(conduit::Node& buffer_holder)
   {
     allocate();
     conduit::Node& buffer_data_holder = buffer_holder["data"];
-    copyBytesIntoBuffer(buffer_data_holder.element_ptr(0),
-                        buffer_data_holder.total_strided_bytes());
+
+    const auto num_bytes = buffer_data_holder.total_strided_bytes();
+    if(num_bytes > 0)
```
kennyweiss, line 307: This checks that we're only attempting to copy bytes into a buffer when there are actually bytes to copy.
(Note: `copyBytesIntoBuffer` was already doing the right thing in this case, so this will not change behavior. The code was previously issuing a warning in this case though).

src/axom/sidre/core/Buffer.cpp
```diff
@@ -299,8 +302,14 @@ void Buffer::importFrom(conduit::Node& buffer_holder)
...
+    {
+      copyBytesIntoBuffer(buffer_data_holder.element_ptr(0), num_bytes);
+    }
+
+    this->print();
```
white238, line 312: Left in debug statement?

white238 (approved):

src/axom/sidre/core/Buffer.cpp
kennyweiss: Thanks!  Removed.

rhornung67 (approved):
Thanks @kennyweiss

cyrush (approved):

nselliott (approved):

PR #503, opened 2021-03-24
Don't rewrite generated headers if they haven't changed

joshessman-llnl:
# Summary

- This PR is a build system refactoring
- It does the following:
  - Avoids rewriting headers that would not otherwise change when CMake is reconfigured

This PR is primarily intended to address the unnecessary full rebuilds that take place every time CMake is reconfigured.
Specifically, any modification to a CMakeLists.txt anywhere within Axom will cause CMake to reconfigure (which is reasonable).
The issue is that this will "touch" the config header and per-component unified headers, which most Axom components are dependent on - meaning that everything will rebuild.

These changes improve developer efficiency by only "touch"ing the generated headers if doing so would change their contents.

### DIFF
--- a/src/cmake/AxomConfig.cmake
+++ b/src/cmake/AxomConfig.cmake
@@ -76,7 +76,7 @@ if(ENABLE_FORTRAN)
 
 endif(ENABLE_FORTRAN)
 
-configure_file(
+axom_configure_file(
     ${PROJECT_SOURCE_DIR}/axom/config.hpp.in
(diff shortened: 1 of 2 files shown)


white238 (approved):

agcapps (approved):
OK, thanks Chris

agcapps:
I mean thanks @joshessman-llnl !!  : )

kennyweiss (approved):
Cool! Thanks @joshessman-llnl !

PR #505, opened 2021-03-27
Update fmt after getting a better fix from author

white238:
https://github.com/fmtlib/fmt/pull/2190

Updates fmt to newest master branch.  This includes the alternate fix for the compiler bug we ran into with nvcc+xlc from the author.


kennyweiss (approved):
Thanks @white238 

Please also add a note to RELEASE-NOTES indicating which commit we're now using.
And remove your patch (if applicable).

white238:
The patch is already gone.

white238:
It was just merged, this is good to go

Issue #450, opened 2021-02-03
Document function callbacks in SphinxWriter


joshessman-llnl:
Are the callback functions added to the tables? (Perhaps I missed it).
Ideally, they would list their signature in addition to other metadata.

_Originally posted by @kennyweiss in https://github.com/LLNL/axom/pull/446#pullrequestreview-582832307_

PR #482, opened 2021-02-22
Documentation generation for function callbacks

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Adds logic to `Container` to record function signature information to the datastore
  - Fixes #450, fixes #436

To save space and maintain generality function arguments/return types are stored as integers (corresponding to the `FunctionTag` enum) in the datastore.

To produce legible documentation for the MFEM coefficient and nested struct examples this PR also modifies the `SphinxWriter` to only display documentation for the first element of an array/dictionary instead of the same schema for each element.  This logic is currently part of the sphinx-specific generation logic but could be simplified considerably if moved to the `Writer`-independent traversal logic.

PR #508, opened 2021-03-29
WIP: Add a new style of generated documentation

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Adds a new style of Sphinx documentation generation

Depends on #450 and should not be merged until that PR is merged.

See https://axom.readthedocs.io/en/feature-essman-nested_docs/axom/inlet/docs/sphinx/nested_structs_nested_documentation.html for an example, compared to the current style: https://axom.readthedocs.io/en/feature-essman-nested_docs/axom/inlet/docs/sphinx/mfem_coefficient_table_documentation.html

- [ ] Implement hashing or something similar to avoid collisions with hyperlinks


src/axom/inlet/SphinxWriter.cpp
```diff
@@ -114,6 +124,16 @@ void SphinxWriter::documentContainer(const Container& container)
...
   }
 
+  // Bail out if this is a collection of primitives -
+  // it doesn't make sense to document each individual primitive
+  // FIXME: Implement something analogous to the logic for struct collections that displays the
+  // "schema" for the primitive elements of the collection
+  if(isCollectionGroup(container.name()) &&
+     !sidreGroup->hasView(detail::STRUCT_COLLECTION_FLAG))
+  {
+    return;
```
white238, line 134: Does the user end up knowing this is a collection of primitives from the outputted documentation?

src/axom/inlet/SphinxWriter.cpp
```diff
@@ -213,6 +269,130 @@ void SphinxWriter::writeAllTables()
...
+        }
+        m_oss << "\n\n";
+      });
+
+    std::for_each(
+      functions.begin() + 1,
+      functions.end(),
+      [this, containerCount](const std::vector<std::string>& functionData) {
+        // Insert a hyperlink target for the name
+        m_oss << fmt::format(".. _{0}{1}:\n\n", functionData[0], containerCount);
```
white238, line 373: 👍

src/axom/inlet/SphinxWriter.hpp
```diff
@@ -39,14 +39,30 @@ namespace inlet
 class SphinxWriter : public Writer
 {
 public:
+  /*!
+  *******************************************************************************
+  * \brief Supported phinx style options
```
white238, line 44: "phinx" -> "sphinx"

src/axom/inlet/SphinxWriter.hpp
```diff
@@ -39,14 +39,30 @@ namespace inlet
...
 public:
+  /*!
+  *******************************************************************************
+  * \brief Supported phinx style options
+  *******************************************************************************
+  */
+  enum class Style
+  {
+    // FIXME: These names are not clear at all
+    Table,  // Tables for the child Fields/Functions of each Container
```
white238, line 50: I think nested is clear (at least how I see it) maybe focus on figuring out a better name for the overloaded "Table".  Singular or Individuals? or possibly just non-nested

src/axom/inlet/SphinxWriter.hpp
```diff
@@ -143,6 +161,17 @@ class SphinxWriter : public Writer
...
     std::vector<std::vector<std::string>> functionTable;
+    /**
+     * \brief Stores a minimal set of data about child containers
+     * Used only for the table of contents in nested "mode"
+     */
+    struct ChildContainerData
+    {
+      std::string name;
+      std::string description;
+      bool isTrivial;
```
white238, line 172: i know trivial is a semi-used word but I dont like it. this could be my problem though and thats okay.  Can you add a comment here?

src/axom/inlet/SphinxWriter.cpp
joshessman-llnl: It's not extremely clear (see https://axom.readthedocs.io/en/feature-essman-nested_docs/axom/inlet/docs/sphinx/mfem_coefficient_nested_documentation.html#attrs) but I can clean this up in a follow-up once we decide what that should look like (see also https://github.com/LLNL/axom/pull/482#pullrequestreview-598879079)

src/axom/inlet/SphinxWriter.hpp
joshessman-llnl: I think it's within the realm of possibility that we'd want a third style of Sphinx documentation - the other consideration is readability in the constructor call:
```cpp
SphinxWriter("docs.rst", "My Docs", true);
// vs...
SphinxWriter("docs.rst", "My Docs", SphinxWriter::Style::Nested);
```

white238 (dismissed):

src/axom/inlet/docs/sphinx/mfem_coefficient_nested_documentation.rst
```diff
@@ -0,0 +1,83 @@
...
+
+--------------------
+Collection contents:
+--------------------
+
+List of boundary conditions
+
+
+
+List of boundary conditions
```
white238, line 34: This is doubled for some reason.

joshessman-llnl: Good catch, ended up just being a copy-paste error!

src/axom/inlet/SphinxWriter.hpp
```diff
@@ -39,14 +39,30 @@ namespace inlet
...
+  /*!
+  *******************************************************************************
+  * \brief Supported Sphinx style options
+  *******************************************************************************
+  */
+  enum class Style
+  {
+    // FIXME: These names are not clear at all
+    Singular,  // Tables for the child Fields/Functions of each Container
+    Nested  // Table of contents for each Container, then sections for each child
```
agcapps, line 51: Perhaps `FlatTOC` and `NestedTOC`

agcapps (approved):
Looks good, from what I can see.  Consider renaming the Style names, but that's only a style recommendation.

src/axom/inlet/SphinxWriter.hpp
white238: `flat` is better than my previous suggestion.  After speaking with @joshessman-llnl , we are gonna move forward with the exciting answer of `Flat` and `Nested`

kennyweiss:
Quick question:  Where do the indices come from in the "Table Mode" ?
E.g. "Table 2 Fields" and "Table 3 Fields" in the following:
image

Externally, it seems like an implementation detail that the user wouldn't understand or be able to reference in their input decks.
Or am I missing something?

joshessman-llnl:
> 
> 
> Quick question: Where do the indices come from in the "Table Mode" ?
> E.g. "Table 2 Fields" and "Table 3 Fields" in the following:
> 
> Externally, it seems like an implementation detail that the user wouldn't understand or be able to reference in their input decks.
> Or am I missing something?
> 
> EDIT: Or is this a sphinx-generated index, unrelated to `inlet` ? Since they're not referenced, perhaps our sphinx should disable the indices?

Looking at the generated RST output (https://github.com/LLNL/axom/blame/4eb53eea9c7ccf85190ea8287009ac2dc5c8d164/src/axom/inlet/docs/sphinx/example1_table_documentation.rst#L34-L50), it looks like Sphinx is adding these in.  I can look into disabling them (or if there exists an option for that).

src/axom/inlet/docs/sphinx/writers.rst
```diff
@@ -0,0 +1,66 @@
...
+
+Inlet also provides a utility for generating a `JSON schema <https://json-schema.org/>`_ from your input file schema.
+This allows for integration with text editors like Visual Studio Code, which allows you to associate a JSON schema
+with an input file and subsequently provides autocompletion, linting, tooltips, and more.  VSCode and other editors
+currently support verification of JSON and YAML input files with JSON schemas.
+
+Using the same  ``documentation_generation.cpp`` example, the automatically generated schema can be used to assist
+with input file writing:
+
+.. image:: json_schema_example.gif
```
kennyweiss, line 66: This is awesome!

src/axom/inlet/docs/sphinx/nested_structs_table_documentation.rst
```diff
@@ -1,9 +1,9 @@
 .. |uncheck|    unicode:: U+2610 .. UNCHECKED BOX
 .. |check|      unicode:: U+2611 .. CHECKED BOX
 
-=========================================
-Nested Structs Output: Input file Options
-=========================================
+==================================
```
kennyweiss, line 4: Should these `rst` files have a comment that they were auto-generated by inlet?

src/axom/inlet/docs/sphinx/nested_structs_nested_documentation.rst
```diff
@@ -0,0 +1,235 @@
...
+
+------
+shapes
+------
+
+
+
+
+--------------------
+Collection contents:
```
white238, line 28: I'm not sure this is helpful from a user stand-point.

src/axom/inlet/docs/sphinx/mfem_coefficient_nested_documentation.rst
white238: Duplication is still here.

white238:
It's hard to tell the structure from the nested version.  We should talk about ways to improve this.

src/axom/inlet/docs/sphinx/nested_structs_nested_documentation.rst
```diff
@@ -0,0 +1,235 @@
...
+     - y-axis point to slice on
+   * - `x`_
+     - x-axis point to slice on
+
+
+.. _z:
+
+**z**
+
+z-axis point to slice on
```
white238, line 203: This addition doesn't provide value over what is in the table above.

src/axom/inlet/docs/sphinx/nested_structs_nested_documentation.rst
```diff
@@ -0,0 +1,235 @@
...
+x-axis point to slice on
+
+
+
+
+---------
+translate
+---------
+
+
```
white238, line 228: Is there something missing here?

src/axom/inlet/docs/sphinx/nested_structs_nested_documentation.rst
```diff
@@ -0,0 +1,235 @@
...
+--------------------
+Collection contents:
+--------------------
+
+List of shape operations to apply
+
+
+
+List of shape operations to apply
```
kennyweiss, line 144: >List of shape operations to apply
>
> List of shape operations to apply

Duplicated.

src/axom/inlet/docs/sphinx/nested_structs_nested_documentation.rst
```diff
@@ -0,0 +1,235 @@
...
+---------
+
+
+
+
+--------------------
+Collection contents:
+--------------------
+
+Translation vector
```
white238, line 235: ... and here?

src/axom/inlet/SphinxWriter.cpp
```diff
@@ -213,6 +269,130 @@ void SphinxWriter::writeAllTables()
   }
 }
 
+void SphinxWriter::writeNestedTables()
+{
+  // Used to avoid duplicate hyperlinks
+  int containerCount = 0;
```
kennyweiss, line 275: (This references a comment I made on the Conversation page)
It looks like the table indices are `inlet`-generated.

My concern is that these indices are meaningful to the system (e.g. to generate unique links), but not to the user.
Is there a way to use these numbers but not display them?

kennyweiss (changes requested):
Thanks @joshessman-llnl .

I think there are some nice ideas here, but perhaps the layout of the nested and flat tables should be improved a bit before we provide these to users?

Some specifics:
* I'm finding it difficult to understand the nested structure of the tables in both the "nested" and "flat" views. 
* Maybe adding the subtable path as part of a table's header would help?
E.g.   "`operators`:  subtable of ``geometry.shapes``"
* The "Flat" table layout seems a bit too spread out relative to the information.
* Would it be possible to have a table for each container, but for the nested containers to have links to their subtable in the RST file?
* Similarly, it would be nice to have a back reference to the parent table.

src/axom/inlet/SphinxWriter.cpp
joshessman-llnl: It looks like Axom is enabling this explicitly:

https://github.com/LLNL/axom/blob/858899f671ec6d596f48e2102a634e32215d6e4a/src/conf.py#L91-L93

See https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-numfig for a description of the option - should we just turn this off then (the default)?

joshessman-llnl:
> * Maybe adding the subtable path as part of a table's header would help?
>       E.g.   "`operators`:  subtable of `geometry.shapes`"


> 
> * The "Flat" table layout seems a bit too spread out relative to the information.

The "flat" layout is just the current style generated by the `SphinxWriter` class - after some discussion with the Serac team a while back it was determined that it might still be useful, hence why I opted to leave it as an option.  Of course, that doesn't mean it can't be improved, either as part of this PR or a separate one.

> 
> * Would it be possible to have a table for each container, but for the nested containers to have links to their subtable in the RST file?

If I understand correctly, this is what the "nested" style option does.

> 
> * Similarly, it would be nice to have a back reference to the parent table.

I think it would definitely be possible to add the parent path/link to the parent, were you thinking that this would be useful for both the new "nested" style and the existing "flat" style?

src/axom/inlet/docs/sphinx/nested_structs_nested_documentation.rst
joshessman-llnl: Because Inlet doesn't have a separate phase for defining the schema and reading from the input file, collections (arrays/dicts) are implemented by adding schema components (scalars, etc) to each element.  The docs generation logic just uses the schema from the first element, so unfortunately the documentation output will depend on what the input file looks like.  In this case, the first element of this collection didn't have a "translate" member, so the information (type, etc) on this member will be incomplete.

src/axom/inlet/docs/sphinx/nested_structs_nested_documentation.rst
joshessman-llnl: Due to Inlet's current architecture (see https://github.com/LLNL/axom/pull/508#discussion_r625747285), there isn't any additional information available for this element.  One solution would be to add logic to avoid creating a separate subsection when only the description is available - does that sound reasonable?

src/axom/inlet/docs/sphinx/nested_structs_nested_documentation.rst
joshessman-llnl: The idea is to express that the following section contains the schema for individual elements of the collection.  Would it be more helpful to make this more verbose (e.g., "Each element of the collection can contain the following:") or to remove it entirely?

src/axom/inlet/SphinxWriter.cpp
kennyweiss: Thanks @joshessman-llnl  -- I think so.

We might need to reword a few things in the mint docs, which, I think might be referencing these numbers.
(which seems a bit formal for sphinx docs to me).

joshessman-llnl: I've gone ahead and tried this out - see commit https://github.com/LLNL/axom/pull/508/commits/8834021be87f50e0e9e67fb6ddce2eb9fc682b1c

Should this maybe be moved into a separate PR?

PR #512, opened 2021-04-01
Add an option to mark Containers as "strict"

### LINKED ISSUE #471: Unexpected fields in Inlet

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Adds a `strict` option to `Container` (used like `required`) that fails verification when unexpected fields are present
  - Adds a method to retrieve the unexpected names within/below an individual `Container` (instead of just the set for the full Inlet hierarchy)
  - Fixes #471
  - Refactors the strictness/unexpected input file entries logic to use a contiguous `std::vector` instead of `std::unordered_set` (which ended up being overkill given the amount of random/sequential access vs search)


white238 (approved):

src/axom/inlet/inlet_utils.cpp
```diff
@@ -17,37 +17,43 @@ void setWarningFlag(axom::sidre::Group* root)
...
-  if(target.hasView("required"))
+  if(target.hasView(flag))
   {
     const std::string msg =
-      fmt::format("[Inlet] Required value has already been defined for: {0}",
+      fmt::format("[Inlet] '{0}' value has already been defined for: {1}",
+                  flag,
                   target.getName());
 
     SLIC_WARNING(msg);
```
kennyweiss, line 32: Should it still be a warnable offense if `target.hasView(flag)`  and its value is `value` ?

src/axom/inlet/inlet_utils.cpp
```diff
@@ -17,37 +17,43 @@ void setWarningFlag(axom::sidre::Group* root)
...
     setWarningFlag(&root);
   }
   else
   {
-    if(required)
+    if(value)
```
kennyweiss, line 37: Suggestion:
```cpp
int8 bval = value ? static_cast<int8>(1) : static_cast<int8>(0);  // or just:  int8 bval = value? 1: 0;
target.createViewScalar(flag, bval);
```

src/axom/inlet/inlet_utils.cpp
```diff
@@ -17,37 +17,43 @@ void setWarningFlag(axom::sidre::Group* root)
...
-  if(!target.hasView("required"))
+  if(!target.hasView(flag))
   {
     return false;
   }
-  const axom::sidre::View* valueView = target.getView("required");
+  const axom::sidre::View* valueView = target.getView(flag);
   if(valueView == nullptr)
   {
     //TODO: is this possible after it says it has the view?
```
kennyweiss, line 59: >    //TODO: is this possible after it says it has the view?

I agree -- this should not be possible

src/axom/inlet/inlet_utils.cpp
```diff
@@ -58,8 +64,9 @@ bool checkIfRequired(const axom::sidre::Group& target, axom::sidre::Group& root)
...
     const std::string msg = fmt::format(
       "[Inlet] Invalid integer value stored in "
-      " boolean value named {0}",
-      target.getName());
+      " boolean value named {0} for flag '{1}'",
+      target.getName(),
+      flag);
     SLIC_WARNING(msg);
     setWarningFlag(&root);
     return false;
```
kennyweiss, line 72: I wonder if this should warn about an incorrect value,
but then return `static_cast<bool>(inValue)` instead of `false`?

src/axom/inlet/examples/nested_structs.cpp
```diff
@@ -327,6 +327,9 @@ int main(int argc, char** argv)
   Shape::defineSchema(shapes_container);
   // _inlet_nested_struct_array_end
 
+  // Mark the entire input as "strict"
+  inlet.getGlobalContainer().strict();
```
kennyweiss, line 331: Since this as an axom example, it might be better to expose this as a command line parameter.
E.g. to only treat the data as strict when the user requests it.  (Perhaps w/ a default of `true`)

This can help users see how `strict` might be made optional.

kennyweiss (approved):
Looks good @joshessman-llnl !

PR #515, opened 2021-04-06
Reconstruct Inlet hierarchy from Sidre

### LINKED ISSUE #487: Add capability to reconstruct Inlet hierarchy from a saved DataStore
Since (almost) all information about the Inlet hierarchy is stored in Sidre, it should be possible to reconstruct the Inlet hierarchy from a datastore without needing an input file.

See https://github.com/LLNL/serac/pull/317#discussion_r582132597 for an example use case

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Fixes #487
  - Adds an option for an Inlet hierarchy to be reconstructed from a Sidre group

Note that this reconstruction does not apply to function callbacks - only to `Container`s and `Field`s


src/axom/inlet/tests/inlet_restart.cpp
```diff
@@ -0,0 +1,171 @@
...
+class inlet_restart : public ::testing::Test
+{ };
+
+TYPED_TEST_SUITE(inlet_restart, axom::inlet::detail::ReaderTypes);
+
+TYPED_TEST(inlet_restart, simple_scalars)
+{
+  std::string testString = "foo = true; bar = false";
+  DataStore ds;
+  Inlet inlet = createBasicInlet<TypeParam>(&ds, testString);
```
kennyweiss, line 52: Would it make sense to put the following lines in their own scope, e.g. to ensure that inlet's destructor doesn't do anything bad to the datastore hierarchy

```cpp
  Inlet inlet = createBasicInlet<TypeParam>(&ds, testString);
  // Check for existing fields
  inlet.addBool("foo", "foo's description");
  inlet.addBool("bar", "bar's description");
```

src/axom/inlet/tests/inlet_restart.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+  // Check for existing fields
+  inlet.addBool("foo", "foo's description");
+  inlet.addBool("bar", "bar's description");
+
+  // No input provided - the datastore should already contain all the data
+  Inlet restartInlet = createBasicInlet<TypeParam>(&ds);
+
+  // Check for existing fields
+  restartInlet.addBool("foo", "foo's description");
```
kennyweiss, line 93: Suggestion: Consider adding a comment here that this one has `addBool()` calls, while the previous test (`simple_scalars`) didn't ?

kennyweiss (approved):
Thanks @joshessman-llnl 

Should there be a test with a callback function?
How about nested structs?

joshessman-llnl:
> 
> Should there be a test with a callback function?

Callbacks aren't supported in this context, but it might be good to issue some sort of warning instead of just ignoring the function?  It might be possible to serialize functions, though:  I found this old project which appears to serialize Lua functions (https://github.com/richardhundt/lua-marshal), it might be possible to update/rewrite it (it's fairly small) and build that capability into Inlet.

white238 (approved):
Thanks @joshessman-llnl ! can you push the fix for the intel warnings into this branch then merge?

PR #517, opened 2021-04-08
refactor: pass Writer directly to write instead of registering

### LINKED ISSUE #499: Inlet: Allow more than one writer
We currently allow only one registered `Writer` class at a time.  If you want to write more than one, you have to register a new one then call `write()` again.

Implementation possibilities:

1) hold all of writers in a vector then call them one by one inside of `write()`
2) call writer by name: `write("jsonschema")`
3) just call write with the writer directly: `write(JSONSchemaWriter())` or the other way `myWriter.write(myInlet)`

joshessman-llnl:
fix: enumerate examples in docs

# Summary

- This PR is a small refactor
- It does the following:
  - Modifies `inlet::Inlet` to accept a `Writer` parameter to its `write` method instead of having to register it
  - Fixes #499

The parameter to `write` is an rvalue reference to allow the passing of "immediate"/temporary `Writer`s - non-const lvalue references can't be bound to temporaries.


white238 (approved):

agcapps (approved):
I like this much better.  Less complicated.

kennyweiss (approved):
Looks good @joshessman-llnl

PR #518, opened 2021-04-09
Fix typo

white238:
Super important.

### DIFF
--- a/.gitlab/build_lassen.yml
+++ b/.gitlab/build_lassen.yml
@@ -23,79 +23,83 @@
 
 ####
 # Template
-.pr_build_on_lassen:
+.src_build_on_lassen:
   stage: build
-  extends: [.build_blueos_3_ppc64le_ib_p9_script, .on_lassen, .pr_workflow]
+  variables:
+    ALLOC_COMMAND: "lalloc 1 -W 25 -q pdebug"
+  extends: [.src_build_script, .on_lassen, .src_workflow]
   needs: []
 
-.main_build_with_deps_on_lassen:
+.full_build_on_lassen:
   stage: build
-  extends: [.build_with_deps_blueos_3_ppc64le_ib_p9_script, .on_lassen, .main_workflow]
+  variables:
+    ALLOC_COMMAND: "lalloc 1 -W 45 -q pdebug"
+  extends: [.full_build_script, .on_lassen, .full_workflow]
   needs: []
 
 ####
 # PR Build jobs
-clang_8_0_1_nvcc_xlf (PR build_on_lassen):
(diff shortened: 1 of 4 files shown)


kennyweiss (approved):

agcapps (approved):
Yes.

rhornung67 (approved):

white238:
This PR also now includes a simplification of the gitlab ci pipelines.

.gitlab/build_lassen.yml
```diff
@@ -23,81 +23,83 @@
...
+  stage: build
+  variables:
+    ALLOC_COMMAND: "lalloc 1 -W 45 -q pdebug"
+  extends: [.full_build_script, .on_lassen, .full_workflow]
   needs: []
 
 ####
 # PR Build jobs
-clang_8_0_1_nvcc_xlf (PR build_on_lassen):
+lassen-clang_8_0_1_nvcc_xlf-src:
```
adrienbernede, line 42: If the job name starts with lassen, it will not help differentiate them from one another (we don't see much of the name in the UI).
`clang_9_0_0_upstream_xlf-src-lassen` ? The lassen stage helped with that too.

adrienbernede (commented):
This is looking good, I like the unified script. Maybe it could even be exported outside the yaml file.

.gitlab/build_lassen.yml
white238: i actually had it that way first, but at least on my screen it gets truncated pretty early and i felt that knowing which machine it came from was more important that getting the full compiler spec:

i see this:

axom-pipeline

white238: I can't confirm it but I think having machine specific stages was the cause of my intermittent failures.  It was triggering the allocate stage then waiting for the corresponding ruby build stage.

kennyweiss: This looks amazing!

Quick question -- w/ the new setup, do the lassen jobs have to wait until we get an allocation on ruby ? (even though they won't use that allocation)

adrienbernede: Does not look bad. But I prefer the stage separation. Maybe place the lassen stage as the last one...
Also, good question @kennyweiss, the `needs: ` in the lassen jobs (template) acheives this.

kennyweiss: @adrienbernede -- I agree that the stage separation makes it easier to see which jobs are which, but it made the pipeline/DAG really confusing since the lassen jobs have not dependency on the ruby jobs (or vice-versa).

Something about the old setup was also causing our jobs to fail.

PR #519, opened 2021-04-10
Bugfix for incorrect coloring in InOutOctree

### LINKED ISSUE #513: Edge case in InOutOctree generating incorrect results
A user reported warnings from the `InOutOctree` when using a specific input mesh. The mesh is roughly an axis aligned cylindrical shape.

After some investigation, the mesh appears to be valid (watertight, w/ properly oriented normals), so this appears to be a legitimate bug.

I tried to reproduce the problem using several different axis-aligned truncated cylinders (https://github.com/LLNL/axom_data/tree/feature/kweiss/truncated-cones/quest/cones), but none of the meshes I generated exhibited this problem.

kennyweiss:
# Summary

- A user reported seeing warnings from quest's `InOutOctree` about incorrect coloring for a specific mesh, which they provided 
- The mesh quality was really nice, but had a lot of axis-aligned triangles that grazed octree block faces
- After a few false starts, I was able to resolve this by inflating the block bounding boxes before testing for intersections
- As a consequence, we will likely have more "gray" blocks (i.e. blocks that intersect the mesh geometry) for this type of triangle
- See more details in #513 
 
This PR also has some code maintenance

- Migrates the axom regression test command line parsing to CLI11
- Loosens regression test that compares signed distance and point containment in cases where the point lies really close to the surface. I.e. the two can disagree about whether points on the surface are categorized as inside or outside
- Converts most of the complicated logging statements in the `InOutOctree` class from `iostreams` to `fmt`
 
Resolves #513

### DIFF
--- a/src/axom/quest/InOutOctree.hpp
+++ b/src/axom/quest/InOutOctree.hpp
@@ -1253,6 +1253,9 @@ protected:
   IndexRegistry m_indexRegistry;
 
   double m_vertexWeldThresholdSquared;
+
+  /// Bounding box scaling factor for dealing with grazing triangles
+  double m_boundingBoxScaleFactor {1.005};
 };
 
 template <int DIM>
@@ -1563,6 +1566,9 @@ void InOutOctree<DIM>::insertMeshTriangles()
         {
           childBlk[j] = blk.child(j);
           childBB[j] = this->blockBoundingBox(childBlk[j]);
+
+          // expand bounding box slightly to deal with grazing triangles
+          childBB[j].scale(m_boundingBoxScaleFactor);
 
           const InOutBlockData& childBlockData = broodData[j];
           if(!childBlockData.hasData())
@@ -1924,9 +1930,6 @@ bool InOutOctree<DIM>::withinGrayBlock(const SpacePt& queryPt,
   /// against this triangle's normal indicates queryPt's containment.
   /// It is inside when the dot product is positive.
 
-  // Bounding box scaling factor for dealing with grazing triangles
-  const double BB_SCALE_FACTOR = 1.005;
-
   SLIC_ASSERT(leafData.color() == InOutBlockData::Gray);
   SLIC_ASSERT(leafData.hasData());
 
@@ -1959,7 +1962,7 @@ bool InOutOctree<DIM>::withinGrayBlock(const SpacePt& queryPt,
       // produces an empty polygon.  To resolve this, clip against a
       // slightly expanded bounding box
       GeometricBoundingBox expandedBB = blockBB;
-      expandedBB.scale(BB_SCALE_FACTOR);
+      expandedBB.scale(m_boundingBoxScaleFactor);
 
       poly = primal::clip(tri, expandedBB);
 
@@ -2933,6 +2936,8 @@ public:
       "--Checking that internal blocks have no data, and that leaves satisfy "
       "all PM conditions");
 
+    const double bb_scale_factor = m_octree.m_boundingBoxScaleFactor;
+
     for(int lev = 0; lev < m_octree.m_levels.size(); ++lev)
     {
       const auto& levelLeafMap = m_octree.getOctreeLevel(lev);
@@ -2971,6 +2976,7 @@ public:
               // Check that this triangle intersects the bounding box of the
               // block
               GeometricBoundingBox blockBB = m_octree.blockBoundingBox(block);
+              blockBB.expand(bb_scale_factor);
               SLIC_ASSERT_MSG(
                 m_octree.blockIndexesElementVertex(tIdx, block) ||
                   intersect(m_octree.m_meshWrapper.trianglePositions(tIdx),
--- a/src/axom/quest/InOutOctree.hpp
+++ b/src/axom/quest/InOutOctree.hpp
@@ -949,12 +949,13 @@ public:
    *
    * \param [in] bb The spatial extent covered by the octree
    * \note We slightly scale the bounding box so all mesh elements are
-   * guaranteed to be enclosed by the octree.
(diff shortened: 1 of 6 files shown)


src/axom/quest/InOutOctree.hpp
```diff
@@ -1540,6 +1572,9 @@ void InOutOctree<DIM>::insertMeshTriangles()
           childBlk[j] = blk.child(j);
           childBB[j] = this->blockBoundingBox(childBlk[j]);
 
+          // expand bounding box slightly to deal with grazing triangles
+          childBB[j].scale(m_boundingBoxScaleFactor);
```
kennyweiss, line 1576: Note: This line is the bugfix. It inflates octree block bounding boxes by a small factor (e.g. 1.000123) before testing for intersections with the mesh triangles. This prevents these triangles from "slipping through the cracks" :)

kennyweiss:
Note: This PR depends on a PR in axom_data: https://github.com/LLNL/axom_data/pull/4
I will update the submodule here after that one is merged.

src/axom/quest/InOutOctree.hpp
```diff
@@ -1608,10 +1647,11 @@ void InOutOctree<DIM>::insertMeshTriangles()
               QUEST_OCTREE_DEBUG_LOG_IF(
                 DEBUG_BLOCK_1 == childBlk[j] || DEBUG_BLOCK_2 == childBlk[j],
                 //&& tIdx == DEBUG_TRI_IDX
-                "Added triangle " << tIdx << " @ " << spaceTri << " with verts "
-                                  << m_meshWrapper.triangleVertexIndices(tIdx)
-                                  << "\n\tinto block " << childBlk[j]
-                                  << "\n\twith data " << *(childDataPtr[j]));
+                fmt::format("Added triangle {} @ {} ", tIdx, spaceTri)
+                  << " with verts " << m_meshWrapper.triangleVertexIndices(tIdx)
```
white238, line 1651: Is there a reason you nested these fmt calls?

kennyweiss: Thanks for noticing it.

`fmt` didn't know how to format this. It is a slam array with an ostream operator. I think I can get this working by adding `fmt/ostream.h`.

I'll look into it.

src/axom/quest/tests/quest_regression.cpp
```diff
@@ -866,15 +787,24 @@ int main(int argc, char** argv)
   MPI_Init(&argc, &argv);
 
   {
-    // Note: this code is in a different context since SimpleLogger's
-    // destructor
-    //       might have MPI calls and would otherwise be invoked after
-    // MPI_Finalize()
+    // Note: this code is in a different context since SimpleLogger's destructor
```
white238, line 790: This is not true.

agcapps (approved):
Thank you for working on this, @kennyweiss .  Is the fix (inflate the octree blocks slightly) general enough?

white238 (approved):

kennyweiss:
@agcapps wrote: 
> Is the fix (inflate the octree blocks slightly) general enough?

It appears to be!
* I tested this against all of our quest regression tests by configuring with `AXOM_QUEST_ENABLE_EXTRA_REGRESSION_TESTS` (see below for details)
* I was even able to roll back my initial bugfix (see https://github.com/LLNL/axom/pull/519/commits/204297299a9500a571dc4d24fca41a034a7dbb26) which had a more expensive runtime for queries within "gray" octree blocks (i.e. those intersecting the mesh geometry).
* I'm happy about this rollback since it required additional runtime costs, but it might be necessary to bring this back in the future if we observe any problems
* I was also able to reduce the inflation factor from `1.005` to `1.000123`. The former was problematic on this mesh, presumably since it kept the bounding boxes aligned with the octree bounding boxes. The somewhat random new value is unlikely to trigger that code path.  I didn't expose this variable since users should not care about the value in general, but I did centralize it.

#### Details about AXOM_QUEST_ENABLE_EXTRA_REGRESSION_TESTS

Quest has a regression test that compares the InOut containment results and the SignedDistance results, and also cross compares the results. In general, we only test a subset of the quest regression tests (since these rarely change and the tests are expensive for Debug builts). 

We have a configuration flag -- `AXOM_QUEST_ENABLE_EXTRA_REGRESSION_TESTS` -- to enable all the tests
https://github.com/LLNL/axom/blob/8c4eb9b8c550baffaa150a05ee864948b28810b2/src/axom/quest/tests/CMakeLists.txt#L129-L141

src/axom/quest/tests/quest_regression.cpp
kennyweiss: Thanks @white238 -- I removed this.

src/axom/quest/InOutOctree.hpp
kennyweiss: Adding `fmt/ostream.h` worked -- and this looks a lot better now.  Thanks!

src/axom/quest/InOutOctree.hpp
```diff
@@ -19,6 +19,9 @@
 #include "axom/mint.hpp"
 #include "axom/spin.hpp"
 
+#include "fmt/fmt.hpp"
+#include "fmt/ostream.h"
```
white238, line 23: Is this not included in our fmt.hpp header?

kennyweiss: It is -- that's really confusing...  I wonder what the error that I saw was all about.

Checking if I can remove it...

kennyweiss: Thanks @white238 
Weird -- looks like it is not necessary. I removed it.

I wonder what the original error message was... ?   ¯\\\_(ツ)_/¯

PR #527, opened 2021-04-15
Fix full build rule

white238:
Tested this time in a test bed on gitlab.

The full build workflow now passes and will trigger every night against develop.


kennyweiss (approved):

agcapps (approved):
Thanks, @white238 .

white238:
I was also able to turn on the full builds for both intel specs.  What is in the compiler specs for gitlab is accurate for what can be turned on/off easily.

PR #532, opened 2021-05-04
SCR updates

nselliott:
# Summary

- This PR is a feature addition
- It does the following:
  - Supersedes #251 
  - Brings in all of the new SCR features from #251 and implements latest review suggestions
  - Adds necessary specs to add SCR as a TPL and includes SCR in host-config for one configuration


nselliott:
I created a new PR because things got out of sync between the #251 branch and the branch I was using to get SCR into the TPL builds, so I found it easier to merge them all into one branch.  This does pull in all code changes from #251.

scripts/spack/packages/scr/package.py
```diff
@@ -0,0 +1,163 @@
...
+    depends_on('er@master', when="@develop")
+    depends_on('rankstr@master', when="@develop")
+    depends_on('shuffile@master', when="@develop")
+
+    # SCR legacy is anything 2.x.x or earlier
+    # SCR components is anything 3.x.x or later
+    depends_on('er', when="@3:")
+    depends_on('kvtree', when="@3:")
+    depends_on('rankstr', when="@3:")
+    depends_on('filo', when="@3")
```
kennyweiss, line 49: Possible typo?
```suggestion
    depends_on('filo', when="@3:")
```

scripts/spack/packages/axom/package.py
```diff
@@ -69,6 +69,8 @@ class Axom(CachedCMakePackage, CudaPackage):
     variant("hdf5",     default=True, description="Build with hdf5")
     variant("lua",      default=True, description="Build with Lua")
     variant("scr",      default=False, description="Build with SCR")
+    variant("kvtree",   default=False, description="Build with kvtree")
+    variant("dtcmp",    default=False, description="Build with dtcmp")
     variant("umpire",   default=True, description="Build with umpire")
```
kennyweiss, line 74: Do we want the `kvtree` `dtcmp` variants?
It seems that we will always need these two when `+scr`  (unless I'm missing something)

kennyweiss: I think this will  also need the following update below:

```py
depends_on("scr", when="+scr")
depends_on("kvtree", when="+scr")
depends_on("ctcmp", when="+scr")
```

Perhaps they can all be combined into one line?

Tag: @white238

kennyweiss: I'm assuming the axom dependency is needed so we can add their paths to our host-config.

It would be even better if `scr`'s cmake export handled this for us, so axom's build system does not need to be aware of `kvtree` or `dtcmp`.
(I don't know enough details about these, but it didn't look like axom actually needed these -- only scr).

kennyweiss (approved):
Thanks @nselliott !

I had some minor questions about the build system changes. I think @white238 can help answer them.

scripts/spack/packages/axom/package.py
white238: I would remove the kvtree and dtcmp variants since we would never want a customer to turn them off.  The variant is only if we have scr support, which then makes us rely on scr, kvtree, and dtcmp

nselliott: If I don't have the variants, then the the ${KVTREE_DIR} and ${DTCMP_DIR} variables that are needed in FindSCR.cmake are not created and those libraries can't be found for the axom build.  It doesn't seem that SCR creates the cmake export files, at least not in a way that I could figure out.  It took me a while to figure out something that works here.

agcapps (approved):
Thank you, @nselliott , especially for the example code.

scripts/spack/packages/axom/package.py
white238: That's because you need to remove `scr`, `kvtree`, `dtcmp` from line 306 and give them its own section, like so:


```
       # Handle all of scr's deps
        if '+scr' in spec:
            for dep in ('scr', 'kvtree', 'dtcmp',):
                dep_dir = get_spec_path(spec, dep, path_replacements)
                entries.append(cmake_cache_path('%s_DIR' % dep.upper(),
                                                dep_dir))
       else:
           entries.append('# scr not build\n')
```

nselliott: OK, I made this change.  I should be able to merge if the CI passes.

nselliott:
gitlab CI keeps failing because it says the TPL installation can't be found.  What am I missing?  The file permissions look correct to me.

PR #540, opened 2021-05-06
Add work around for non-existant INTERFACE_INCLUDE_DIRECTORIES from imported targets

white238:
This is currently blocking a user.  They are testing it to make sure the hard CMake error goes away.


src/cmake/thirdparty/SetupAxomThirdParty.cmake
```diff
@@ -222,3 +222,37 @@ foreach(dep ${TPL_DEPS})
...
+        get_target_property(_dirs ${_target} INTERFACE_INCLUDE_DIRECTORIES)
+        set(_existing_dirs)
+        foreach(_dir ${_dirs})
+            if (EXISTS "${_dir}")
+                list(APPEND _existing_dirs "${_dir}")
+            endif()
+        endforeach()
+        if (_existing_dirs)
+            set_target_properties(${_target} PROPERTIES
+                                  INTERFACE_INCLUDE_DIRECTORIES "${_existing_dirs}" )
```
kennyweiss, line 255: Does this overwrite the existing value of `INTERFACE_INCLUDE_DIRECTORIES`?

kennyweiss (approved):
Thanks @white238

src/cmake/thirdparty/SetupAxomThirdParty.cmake
kennyweiss: (My assumption is that it does)

white238: yes.. it removes all non-existing directories and overwites the previous value

white238:
closed in preference to #550

Issue #546, opened 2021-05-08
Add github action for manual Windows TPL builds


kennyweiss:
Our Uberenv script is set up to build a library set of Axom on Windows, but our current (azure) CI tests Axom without its TPLs.

As a first effort towards automated testing of Axom with its TPLs, I am developing a github actions workflow for this.
I will do this in stages:

- [ ] Develop an initial workflow that is manually triggered and runs uberenv to build axom's TPLs
- [ ] Use the associated host-config to build and test axom's in a `debug` configuration
- [ ] Use the associated host-config to build and test axom's in a `release` configuration
- [ ] Extend this workflow to have several jobs that test two `vcpkg` triplets: `x64-windows` and `x86-windows`
- [ ] See if the generated vcpkg and host-config can be cached and if this can be incorporated into a standalone CI build of Axom, ideally for PR builds.

PR #549, opened 2021-05-08
Adds manually triggered github actions workflow for testing Windows TPLs

kennyweiss:
# Summary

- This PR adds a manually triggered github actions workflow to test our Windows TPLs
- The workflow builds our TPLs via `uberenv`/`vcpkg` using the `x64-windows` and `x86-windows` triplets
   and tests Axom against these for Axom's `Debug` and `Release` builds
- As a future extension, it might be nice if we could cache the TPLs. Then we could test these against PRs into develop. 
- See issue #546 for more details
- I'm planning to squash merge this to reduce due to all the trial and error commits

Successful run of this workflow after merging #502:  
* https://github.com/LLNL/axom/actions/runs/824679693

### DIFF
--- a/.github/workflows/test_windows_tpls.yml
+++ b/.github/workflows/test_windows_tpls.yml
@@ -36,4 +36,4 @@ jobs:
       run: dir
     - name: Run uberenv (w/ default settings) 
       run: |
-        python3 scripts/uberenv/uberenv.py
+        python3 .\scripts\uberenv\uberenv.py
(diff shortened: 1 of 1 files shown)


white238 (approved):

kennyweiss:
Note: Github actions work different from how I expected. 
* Each "job" runs in a new image, so you cannot easily share data between jobs. You can "cache" data in a job and then load it in a new separate job, so that could be one possible extension.
* I ended up creating a "matrix" of four jobs -- for combinations of {"Debug", "Release"} and {"x64", "x86"}. Each of these jobs clones axom, generates a TPL set and then builds and tests axom against this.

adrienbernede (approved):

PR #551, opened 2021-05-10
Prefer to use MFEM's CMake config when available

joshessman-llnl:
# Summary

- This PR is a build system change
- It does the following:
  - Prefers to use MFEM's CMake package when it's available

MFEM's CMake build system generates both Make and CMake packages.  Currently Axom only uses the CMake package for Windows builds.  This PR changes the find logic so that Unix CMake builds of MFEM will use the CMake config instead of the Make config.  The logic for Make-built CMake is unchanged.

A similar approach has been adopted + tested in Serac: https://github.com/LLNL/serac/pull/304


white238 (approved):

kennyweiss (approved):
Thanks @joshessman-llnl !

PR #553, opened 2021-05-11
Disables openmp on Intel host-configs

### LINKED ISSUE #544: Disable openmp in toss3 intel host-configs
Our intel builds take a really long time (e.g. around 20 minutes, compared to at most a few minutes in other configurations).
It looks like a large part of this is due to icc trying to optimize openmp calls.
I'm not aware of any users that depend on OpenMP 2.0 in our intel builds.

I'd like to disable openmp in these configurations.

kennyweiss:
# Summary

- This PR disables openmp within the Intel host-configs on LLNL's toss3 platform
- Discussed during team meeting on 5/10/2021
- Intel's openmp optimizations were taking too long and then giving up after 20+ minutes
- I built the `intel@19` host-config manually and it compiled fairly quickly
- Resolves #544

### DIFF
--- a/host-configs/rzgenie-toss_3_x86_64_ib-clang@10.0.0.cmake
+++ b/host-configs/rzgenie-toss_3_x86_64_ib-clang@10.0.0.cmake
@@ -11,11 +11,11 @@
 #------------------------------------------------------------------------------
 if(DEFINED ENV{SPACK_CC})
 
-  set(CMAKE_C_COMPILER "/usr/WS1/axom/libs/toss_3_x86_64_ib/2021_05_05_17_27_26/spack/lib/spack/env/clang/clang" CACHE PATH "")
+  set(CMAKE_C_COMPILER "/usr/WS1/axom/libs/toss_3_x86_64_ib/2021_05_10_21_44_53/spack/lib/spack/env/clang/clang" CACHE PATH "")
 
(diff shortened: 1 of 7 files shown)


host-configs/rzgenie-toss_3_x86_64_ib-clang@10.0.0.cmake
```diff
@@ -69,7 +69,7 @@ set(ENABLE_GTEST_DEATH_TESTS ON CACHE BOOL "")
 
 # Root directory for generated TPLs
 
-set(TPL_ROOT "/usr/WS1/axom/libs/toss_3_x86_64_ib/2021_05_05_17_27_26/clang-10.0.0" CACHE PATH "")
+set(TPL_ROOT "/usr/WS1/axom/libs/toss_3_x86_64_ib/2021_05_10_21_44_53/clang-10.0.0" CACHE PATH "")
```
kennyweiss, line 72: @white238 -- Do you think it would make sense to have `TPL_ROOT` be one directory higher and then use it for the spack compilers as well?

That would reduce the diffs when we generate a new host-config.

rhornung67 (approved):

kennyweiss:
Update: 
* `intel@19` compiles significantly faster than before
* unfortunately, `intel@18` still takes a while to compile

host-configs/rzgenie-toss_3_x86_64_ib-clang@10.0.0.cmake
white238: The compiler lines are output by the spack build package class.  It would be kind of a hassle and very us specific to do that.

PR #554, opened 2021-05-12
Support legacy hdf5 cmake build system

white238:
# Summary

- This PR is a bugfix
  - Adds support for legacy (?) hdf5 cmake build system which creates an `hdf5` target not `hdf5::hdf5-static` etc


kennyweiss (approved):
Thanks @white238 -- offhand, this looks good.

Testing it on Windows here: https://github.com/LLNL/axom/actions/runs/837189985
(but I think the previous condition should catch this case).

agcapps (approved):
Thank you for the work on this, @white238 .

PR #567, opened 2021-05-24
Reuse FESpace/QSpaces with same basis in MFEMSidreDataCollection

### LINKED ISSUE #533: Avoid duplicating MFEM FESpaces/QSpaces with the same basis name
Potential future optimization: Should `m_quadspaces` be a map from string name to `QuadratureSpace`? 
(This would need to be unique for each mesh, but `MFEMSidreDataCollection` is only over a single mesh)

_Originally posted by @kennyweiss in https://github.com/LLNL/axom/pull/433#discussion_r625478060_

joshessman-llnl:
# Summary

- This PR is a refactoring
  - Modifies `MFEMSidreDataCollection` to reuse FESpace/QSpace objects with the same basis
  - Fixes #533


kennyweiss (approved):
Thanks @joshessman-llnl !

My only minor concern is about whether we'll need to distinguish between `FESpace`s associated with the `mesh` and those associated with the `boundary_mesh`. Probably not an actual concern since I don't think we currently have a `RegisterBoundaryField`

rcarson3 (approved):
Nice work @joshessman-llnl ! This should definitely help reduce memory usage when several quadfuncs and gridfuncs share the same qspace and fespace and are saved off and then restored :)

joshessman-llnl:
> 
> My only minor concern is about whether we'll need to distinguish between `FESpace`s associated with the `mesh` and those associated with the `boundary_mesh`. Probably not an actual concern since I don't think we currently have a `RegisterBoundaryField`

Right now we hardcode the association:

https://github.com/LLNL/axom/blob/3fd0dcef99ed50fc4aa8f6531bd1d4f0d906d240/src/axom/sidre/core/MFEMSidreDataCollection.cpp#L1229-L1232

Do you anticipate needing the ability to register boundary fields? [`mfem::DataCollection`](https://mfem.github.io/doxygen/html/classmfem_1_1DataCollection.html) doesn't expose this interface either but I can create Axom/MFEM issues for this if that functionality would be useful.

kennyweiss:
> Do you anticipate needing the ability to register boundary fields? 

We might need to in the future, but let's hold off on thinking about it until there's an actual need.

white238 (approved):

rhornung67 (approved):

PR #570, opened 2021-05-25
Attempt to fix slic linker warning in mint test when building shared

### LINKED ISSUE #548: Fix linker warning on windows related to import/export of dll symbols
PR #502 revamped our dll import/export process on Windows.

I thought I fixed all the linker warnings, but it appears that we still have some of the form:
> LINK : warning LNK4217: symbol '?checksAreErrors@debug@slic@axom@@2_NA (public: static bool axom::slic::debug::checksAreErrors)' defined in 'slic.obj' is imported by 'mint_mesh_curvilinear_mesh.obj' in function '"public: class axom::sidre::View * __cdecl axom::sidre::View::setString(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &)" (?setString@View@sidre@axom@@QEAAPEAV123@AEBV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@Z)' [D:\a\axom\axom\build-axom\axom\mint\tests\mint_mesh_curvilinear_mesh_test.vcxproj]

I think the problem is related to our usage of object libraries: We want to export some symbols (e.g. `slic::debug::checksAreErrors` to other libraries, but it is also used by other components within the Axom library (which should not view these as external).

One possible resolution could be to add flags to the axom libraries (but not to axom's users) to disable this warning.

kennyweiss:
# Summary

- Attempt to resolve #548 by linking against all of axom, not just mint.

### DIFF
--- a/src/axom/mint/tests/CMakeLists.txt
+++ b/src/axom/mint/tests/CMakeLists.txt
@@ -40,7 +40,7 @@ set( mint_tests
      mint_util_write_vtk.cpp
    )
 
-set( mint_test_dependencies mint )
+set( mint_test_dependencies axom )
 
 blt_list_append( TO mint_test_dependencies ELEMENTS cuda IF ${ENABLE_CUDA} )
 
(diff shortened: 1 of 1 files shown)


kennyweiss:
Sadly, this did not resolve the issue.

PR #575, opened 2021-06-01
Add an option to "record" Inlet verification messages

### LINKED ISSUE #354: Inlet: Investigate using Conduit Blueprint verify functionality for reporting
Cyrus pointed out we can utilize existing verify reporting functionality inside of Conduit Blueprint instead of or on top of the current list of warnings Inlet produces.

### LINKED ISSUE #472: Better error reporting needed in Inlet
Currently, when a file does not conform to an Inlet schema, SLIC is used to report errors to the log. This is less than ideal from a couple of perspectives.

The first problem is that SLIC is better suited to report problems in the application. For example, it prints out information about the source file and line number where the message originated. This is of no use for an end-user whose files are being read by Inlet. Having the file and line number of where the problem is in their input file would be much more useful. Another way in which SLIC is ill-suited for this task is that the output from Inlet will be intermixed with other application-level logging. For parsing input files, it would be better to allow the host application to report errors to the console or another file that the user can easily check that does not contain logging messages from other libraries or the application as a whole.

The second way in which SLIC makes it hard to report errors is that a common way to use Inlet is perform two phases of validation. In the first phase, basic rules such as syntax, the presence of required fields, and statically-defined valid ranges of values are checked. Inlet does this on its own or with custom validators. In the second phase, this data is converted to structs, which are then validated for proper semantics. During this second phase, it is often convenient to stop processing user input as soon as an error is detected by throwing an exception. This greatly simplifies that code, since function can be written to either return valid values or throw exceptions rather than having to return success/failure statuses. However, throwing exceptions means that the host application must now check two separate things for errors in user input: the status of the first validation via `Inlet::verify`, and whether exceptions were thrown.

There are a few possible solutions:
1. Allow Inlet to be configured to throw exceptions rather than using SLIC to report problems. Something like `inlet::ValidationError` would be quite helpful. This would solve the problem of intermixing library- or application-level SLIC messages with messages regarding bad input files. Developers leveraging Inlet could throw this same exception when doing semantic verification to solve the problem of having to detect errors in different ways.
2. Allow `Inlet` objects to be handed a `SLIC` logger which would be used for all error reporting. The application could configure this to send errors to its desired location to keep this output from getting intermixed with application logs. It could also be used to simplify the multi-phase validation by adding a method to verify that no message have been output. This method would throw an `inlet::ValidationError`, allowing for a more unified way of reporting errors.
3. Remove SLIC as a dependency to Inlet. Instead, always through an `inlet::ValidationError`. This would simplify the implementation, but would allow only a single error to be reported. Custom validators would now throw this rather than return a boolean indicating the validity of a Table.

I'm sure there are other possibilities as well, or enhancements that could be made to the above suggestions. These are just a few possibilities.

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Modifies the schema verification logic to optionally write messages to a `std::vector` that can be processed by the end user instead of logging with SLIC
  - Fixes #472 
  - Fixes #354 (obsolete as slightly different approach was chosen)


src/axom/inlet/Container.cpp
```diff
@@ -1070,14 +1070,22 @@ bool Container::verify() const
...
-    verifyRequired(*m_sidreGroup, this_container_defined, "Container");
+    verifyRequired(*m_sidreGroup, this_container_defined, "Container", errors);
 
   // Verify this Container if a lambda was configured
   if(this_container_defined && m_verifier && !m_verifier(*this))
   {
     verified = false;
-    SLIC_WARNING(
-      fmt::format("[Inlet] Container failed verification: {0}", m_name));
+    const std::string msg =
```
kennyweiss, line 1079: Suggestion: Should this new functionality be wrapped in a macro  `INLET_WARNING(error, msg)`  which logs to the vector if `error` is not null and to slic otherwise?

This would potentially clean up the code and allow for easier changes later?

src/axom/inlet/Field.hpp
```diff
@@ -79,210 +79,39 @@ class Field : public VerifiableScalar
    */
   const axom::sidre::Group* sidreGroup() const { return m_sidreGroup; };
 
-  /*!
```
kennyweiss, line 82: Are you removing the doxygen comments b/c they are duplicates of what is already in `VerifiableScalar` ?

src/axom/inlet/Function.cpp
```diff
@@ -40,17 +40,27 @@ Function& Function::registerVerifier(std::function<bool(const Function&)> lambda
   return *this;
 }
 
-bool Function::verify() const
+bool Function::verify(std::vector<VerificationError>* errors) const
```
kennyweiss, line 43: Would it make sense to use an inlet type alias for `std::vector<VerificationError>` ?

src/axom/inlet/inlet_utils.cpp
```diff
@@ -103,7 +104,14 @@ bool verifyRequired(const axom::sidre::Group& target,
         "specified: {1}",
         type,
         target.getPathName());
-      SLIC_WARNING(msg);
+      if(errors)
+      {
+        errors->push_back({Path {target.getPathName()}, msg});
```
kennyweiss, line 109: Is `Path` the same as `axom::Path`?
Suggestion: It might be clearer to readers of the code to include the `axom` namespace.

src/axom/inlet/tests/inlet_errors.cpp
```diff
@@ -0,0 +1,350 @@
...
+  Inlet inlet = createBasicInlet<TypeParam>(testString);
+
+  inlet.addInt("foo", "foo's description").required();
+
+  std::vector<VerificationError> errors;
+  EXPECT_FALSE(inlet.verify(&errors));
+  // Need something about "foo" and "Required"
+  EXPECT_TRUE(
+    std::any_of(errors.begin(), errors.end(), [](const VerificationError& err) {
+      return (static_cast<std::string>(err.path) == "foo") &&
```
kennyweiss, line 42: Suggestion: Some of the common checks might be good candidates for functions in the `VerificationError` struct.

kennyweiss, line 42: Also, does `axom::Path` not have a way to explicitly get the path as a string?

kennyweiss: E.g. https://github.com/LLNL/axom/blob/7f935b2b0a331cced6f3b77683d83d9ca49a53f2/src/axom/core/Path.hpp#L54-L59

RELEASE-NOTES.md
```diff
@@ -22,6 +22,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 
 ### Added
 - Added new CMake option to allow users to turn off Axom created tools: `AXOM_ENABLE_TOOLS`
+- Inlet can now log verification errors to a user-processable list instead of using SLIC
```
kennyweiss, line 25: Thanks for updating the RELEASE-NOTES!

kennyweiss (approved):
Thanks @joshessman-llnl 

Can you please update the docs with an example that demonstrates how this works?

src/axom/inlet/Field.hpp
joshessman-llnl: Yes - it doesn't affect the generated documentation (probably would have been good to mention this in the PR description)

src/axom/inlet/inlet_utils.cpp
joshessman-llnl: Yep, this is just the Path class

src/axom/inlet/tests/inlet_errors.cpp
```diff
@@ -0,0 +1,350 @@
...
+{
+  std::string testString = "foo = 'hello'";
+  Inlet inlet = createBasicInlet<TypeParam>(testString);
+
+  inlet.addInt("foo", "foo's description");
+
+  std::vector<VerificationError> errors;
+  EXPECT_FALSE(inlet.verify(&errors));
+  // Need something about "foo" and "wrong type"
+  EXPECT_TRUE(
```
estebanpauli, line 57: It would be nice to provide something like the below to simplify verifying expected errors:
```cpp
ASSERT_CONTAINS_ERROR(errors, [](const VerificationError &err){...});
// with gmock
EXEPCT_THAT(error, ContainsErrorMatching([](const VerificationError &err));
```

I don't know where would be a good place to put it though, such that both other Axom libraries and non-Axom users of Inlet could access it. I don't know if you could to do a good chunk of it in Inlet itself and leave the other libraries/users to define a small bit of gmock glue code themselves.

estebanpauli (changes requested):
I agree with what Kenny suggested regarding the simplified method to either log or save the error. Other than that, looks good.

src/axom/inlet/tests/inlet_errors.cpp
joshessman-llnl: The cast here is calling that `operator std::string`, would it also be useful to have an alias method so the user could also do `path.string()`?

estebanpauli: I have an orthogonal suggestion to what operations `Path` should support: I think from a testing perspective, it would be better to do`return err.path == Path{"foo"} && ...`. That removes the cast syntax, and makes it clear that you are verifying that the `path` property is equal to a specific `Path` object rather than just verifying that the string representation matches a particular string.

kennyweiss: I like @estebanpauli's suggestion in terms of readability/comprehensibility for the tests

src/axom/inlet/tests/inlet_errors.cpp
kennyweiss: Does axom build with gmock enabled?
I don't recall us using it elsewhere in the code (and I personally have not used it).

@estebanpauli -- are you using it w/ klee?

kennyweiss: The syntax looks really clean

src/axom/inlet/Function.cpp
joshessman-llnl: I'd have to think about it but I'm not sure that an alias over a `std::vector` makes sense here; my concern is that a user would have to look it up (in the Doxygen or elsewhere) to see that it's just an STL type so they know how to iterate over it versus just seeing that it's a `std::vector'`.  Any thoughts @white238 ?

estebanpauli (approved):

src/axom/inlet/Function.cpp
white238: I am with @joshessman-llnl on this one. It's a simple STL type and user facing, I would like it clear as possible to them.

white238 (approved):

PR #578, opened 2021-06-02
Bump TPLs to conduit 0.7.2

nselliott:
# Summary

- This PR is a TPL update
  - Updates the TPLs to use conduit 0.7.2


nselliott:
Not ready to merge due to docker image builds not working.

kennyweiss (approved):
Thanks @nselliott !

white238:
Problem: github actions runners are limited to 14G of file system. This poses a problem when:

Axom build directory size:

    static: 5.7G
    shared (pre-fix): 5.1G
    shared: 1.9G

Docker image size:

    Pre-fix: 4G
    Now: 2.75G

Things done in this branch to mitigate this:

    Turns on shared libraries for axom only on docker builds and images
    Use axom target instead of component object libraries to fully utilize shared library file system improvements

Even though we delete the spack build and axom build directories they are saved in a docker image layer. Many hoops have to be jumped through to stop this.

white238:
I have reviewed the changes prior to mine and they were good.  but someone other than me should review this.

white238:
Lassen builds are failing. I'm out of willpower today. I'll keep plugging along later.

host-configs/ruby-toss_3_x86_64_ib-intel@18.0.2.cmake
```diff
@@ -53,7 +53,7 @@ set(ENABLE_MPI ON CACHE BOOL "")
 # Hardware Specifics
 #------------------------------------------------
 
-set(ENABLE_OPENMP ON CACHE BOOL "")
+set(ENABLE_OPENMP OFF CACHE BOOL "")
```
agcapps, line 56: For my information, why the change here?  Was this part of the disk space problem fix, or was OpenMP compilation taking too long, or what?

agcapps: And why is OpenMP turned off in the Intel host-configs on ruby, but on rzgenie?

src/axom/quest/tests/CMakeLists.txt
```diff
@@ -136,8 +136,6 @@ if (ENABLE_MPI AND AXOM_ENABLE_SIDRE)
 
     set(quest_regression_depends
         ${quest_tests_depends}
-        conduit::conduit
-        sidre
```
agcapps, line 140: Were these libraries simply not necessary?

agcapps:
>     * Use axom target instead of component object libraries to fully utilize shared library file system improvements
>

Why must you use the `axom` target instead of component targets to fully utilize the shared library?  Are the component targets not DLLs?  Not being DLLs, do they always get statically linked in?

host-configs/ruby-toss_3_x86_64_ib-intel@18.0.2.cmake
kennyweiss: @agcapps -- I disabled OpenMP in our intel builds due to excessive build times, and OpenMP coverage in other host-configs.  We also discussed this during a team meeting on 5/10/21.
Looks like I only got the RZ host-configs in that PR, and this picks up the CZ ones.

Disabling OpenMP helped significantly for the intel@19 configuration, which went down from around 30 minutes to around 5 minutes.  Unfortunately, the intel@18 configs still take around 30+ minutes.

See:
* https://github.com/LLNL/axom/issues/544
* https://github.com/LLNL/axom/pull/553

src/axom/quest/tests/CMakeLists.txt
kennyweiss: Yes, dependencies are transitive, and CMake handles this for us internally now that we're using the `blt_import_target` macro for our TPLs.

Our policy is to only explicitly depend on a target when we're directly using it. 
E.g. if we had a quest test that had explicit conduit calls, then that test should depend on the `conduit` target.

host-configs/ruby-toss_3_x86_64_ib-intel@18.0.2.cmake
agcapps: Thanks for the clarification, @kennyweiss .

src/axom/quest/tests/CMakeLists.txt
agcapps: OK, that makes sense.

white238:
> 
> 
> > ```
> > * Use axom target instead of component object libraries to fully utilize shared library file system improvements
> > ```
> 
> Why must you use the `axom` target instead of component targets to fully utilize the shared library? Are the component targets not DLLs? Not being DLLs, do they always get statically linked in?

the component targets are "object libraries" and are just a list of object files.  So everytime we were linking in an object file to a new executable or library we were bringing them along everytime.

agcapps (approved):
Thanks for this, @white238 .

PR #580, opened 2021-06-03
Reduce file system usage on docker containers

white238:
Problem: github actions runners are limited to 14G of file system.  This poses a problem when:

Axom build directory size:

- static: 5.7G
- shared (pre-fix): 5.1G
- shared: 1.9G

Docker image size:

- Pre-fix: 4G
- Now: 2.75G

Things done in this branch to mitigate this:

- Turns on shared libraries for axom only on docker builds and images
- Use `axom` target instead of component object libraries to fully utilize shared library file system improvements

Even though we delete the spack build and axom build directories they are saved in a docker image layer.  Many hoops have to be jumped through to stop this.


white238:
oh man.. i think i win the label war!

white238:
well this passed when it created the docker image but now fails with this:

```
/usr/bin/ld: /home/axom/axom_tpls/gcc-8.1.0/mfem-4.2.0/lib/libmfem.a(mem_manager.o): relocation R_X86_64_32 against `_ZNSt9bad_allocD1Ev' can not be used when making a shared object; recompile with -fPIC
```

fix incoming =/

kennyweiss (approved):
Thanks @white238 !

white238:
This is ready to go.  This is merging into @nselliott 's branch not develop.

PR #584, opened 2021-06-07
Support aggregating errors in custom verifiers.

estebanpauli:
Custom verifiers in Inlet can now receive a vector of
VerificationError objects and log their errors to that
rather than directly through Slic.

# Summary

- This PR is a feature
  - Add a capability for custom verifiers to get the `std::vector<VerificationError>` and report errors by appending to this rather than by calling slic.


estebanpauli:
Looks like I had lots of failures because I added gmock and many of the platforms are not configured with this. I'm not familiar with how this is all configured and managed. Could someone please help with this? I could changes these few tests to use gtest instead of gmock, but I have used gmock more heavily in Klee. When we go to integrate that eventually, having gmock will be much more important.

src/CMakeLists.txt
```diff
@@ -18,6 +18,10 @@ if (ENABLE_FORTRAN)
    enable_language(Fortran)
 endif()
 
+if (AXOM_ENABLE_TESTS)
+    set(ENABLE_GMOCK ON CACHE STRING "")
+endif()
```
kennyweiss, line 23: I don't think `AXOM_ENABLE_TESTS` exists yet here.
It is created when loading in the AxomOptions below.

If you enable tests with `ENABLE_TESTS`, then the following *should* work:
```suggestion
if (ENABLE_TESTS)
    set(ENABLE_GMOCK ON CACHE STRING "")
endif()
```

kennyweiss: @white238 -- is this where the logic for enabling GMOCK should go?

Do we want to optionally enable GMOCK and guard against it?

kennyweiss: Here's where `AXOM_ENABLE_TESTS` is created: 
* AxomOption: https://github.com/LLNL/axom/blob/858240e755466245b27d102c61bfcda480e26214/src/cmake/AxomOptions.cmake#L23

Which is called from:
* CMakeBasics: https://github.com/LLNL/axom/blob/858240e755466245b27d102c61bfcda480e26214/src/cmake/CMakeBasics.cmake#L10
* CMakeLists: https://github.com/LLNL/axom/blob/858240e755466245b27d102c61bfcda480e26214/src/CMakeLists.txt#L102

white238: yes `ENABLE_GMOCK` has to be enabled before BLT is loaded.

estebanpauli: I had tested with `ENABLE_GMOCK` already set in my cache, which caused things to work locally, but not when I pushed.

It looks like the best options might be to just set `ENABLE_GMOCK` unconditionally (where I currently have it, or further down before BLT). Is that right? I don't see a better option, and it shouldn't hurt anything.

white238: You should not be enabling gmock unconditionally. It should be guarded by ENABLE_TESTS because its only used there.

src/axom/inlet/VerifiableScalar.cpp
```diff
@@ -0,0 +1,21 @@
...
+VerifiableScalar& VerifiableScalar::registerVerifier(
+  std::function<bool(const Field&)> verifier)
+{
+  return registerVerifier(
+    [&verifier](const Field& field, std::vector<VerificationError>*) {
+      return verifier(field);
+    });
+}
+}  // namespace inlet
+}  // namespace axom
```
white238, line 21: newline please

src/axom/inlet/VerifiableScalar.hpp
```diff
@@ -78,11 +89,21 @@ class VerifiableScalar
...
+   *****************************************************************************
+   * \brief Registers the function object that will verify this object's contents
+   * during the verification stage.
+   *
+   * \param [in] verifier The function object.
    *****************************************************************************
   */
-  virtual VerifiableScalar& registerVerifier(
-    std::function<bool(const axom::inlet::Field&)> lambda) = 0;
+  virtual VerifiableScalar& registerVerifier(Verifier verifier) = 0;
```
white238, line 106: I may be in the minority, but using a type alias for a user-facing function makes it hard to know what the verifiers function signature should be. @kennyweiss / @joshessman-llnl  thoughts?

white238 (approved):
Thanks @estebanpauli !

src/axom/inlet/VerifiableScalar.hpp
kennyweiss: I enabled the docs here: https://axom.readthedocs.io/en/feature-pauli2-custom-verifier-log-to-vector/

The type alias way seems reasonable to me from the doxygen, e.g.:
* Field: https://axom.readthedocs.io/en/feature-pauli2-custom-verifier-log-to-vector/doxygen/html/classaxom_1_1inlet_1_1Field.html
* Function: https://axom.readthedocs.io/en/feature-pauli2-custom-verifier-log-to-vector/doxygen/html/classaxom_1_1inlet_1_1Function.html

kennyweiss: For clarify, we might want to group the `registerVerifier` functions together in the docs, 
e.g. see `inlet::Field` (https://axom.readthedocs.io/en/feature-pauli2-custom-verifier-log-to-vector/doxygen/html/classaxom_1_1inlet_1_1Field.html)

I think something like following should help:
```cpp
//@{
//!  @name Functions for registering and invoking verifiers

<all code related to `registerVerifier` here>
<all code related to `verify` here>

//@}

**Review by kennyweiss — APPROVED** [review]
Thanks @estebanpauli  -- this is really nice

File: src/axom/inlet/Function.cpp [hunk 4]
```diff
@@ -47,7 +47,7 @@ bool Function::verify(std::vector<VerificationError>* errors) const
   bool verified =
     verifyRequired(*m_sidreGroup, this_function_exists, "Function", errors);
   // Verify this Function if a lambda was configured
-  if(this_function_exists && m_verifier && !m_verifier(*this))
+  if(this_function_exists && m_verifier && !m_verifier(*this, nullptr))
```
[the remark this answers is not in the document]
estebanpauli: It should be. That was the whole point :). I'll fix this and make sure to add a test.

src/CMakeLists.txt
```diff
@@ -40,6 +40,8 @@ else()
     endif()
 endif()
 
+set(ENABLE_GMOCK ON CACHE STRING "")
```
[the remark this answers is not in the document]
estebanpauli: I don't think users should be able to selectively disable gmock. They should be able to disable testing in general, but why allow them to disable gmock? I think this would just complicate things without adding anything of value. If a user tries to disable it while enabling testing, I think build errors are a reasonable outcome.

src/axom/inlet/Function.cpp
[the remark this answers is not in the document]
estebanpauli: Now that I am trying to test this, I remember why I had stopped and then forgotten to go back -- the test file `inlet_function.cpp` is only included if SOL is found, which is not the default on the systems I tested.

src/axom/inlet/VerifiableScalar.hpp
```diff
@@ -45,8 +45,19 @@ class Field;
 class VerifiableScalar
 {
 public:
+  /**
+   * A function which can verify the contents of the item being verifier.
+   * It should report any errors via INLET_VERIFICATION_WARNING, passing
+   * in the given array of errors.
+   */
+  using Verifier = std::function<bool(const axom::inlet::Field&,
```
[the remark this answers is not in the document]
estebanpauli: I'd be happy to rename this, but would like to propose different names. `ListVerifier` makes it sound like it is verifying lists, when really the list refers to the fact that the verifier can add to the list of reasons why an input is bad. What do people think of the suggestions below? The first name is for the function that just takes a field, the second is for the one that takes a list.

1. Verifier, LoggingVerifier
2. Verifier, ReportingVerifier
3. SimpleVerifier, Verifier
4. ?
5. A combination of above options?

src/axom/inlet/VerifiableScalar.hpp
estebanpauli: `verify()` doesn't just call the custom verifiers -- it also validates the calls to other functions like `required()`, `range()`, `validValues()`. `registerVerifier()` is just an extra flavor of this. Given that, though, there are some like `defaultValue()` which are a little different.

An additional complications is that functions from the parent class (e.g. a few `get()` instances and a `registerVerifier()` for Field) get generated at the bottom of the class instead.

Give the above, I'd like to suggest that this is outside the scope of this PR. If a bigger rearrangement of this documentation is desired, it should be handled separately as it would touch a lot of lines.

src/axom/inlet/VerifiableScalar.hpp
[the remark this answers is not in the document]
estebanpauli: Should I make any more changes regarding this, or should I go ahead and merge?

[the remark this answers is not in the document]
kennyweiss: I'm not sure if any of the above suggestion are better than what's already in the code. The current PR seems ready to me, but @joshessman-llnl and @white238 should weigh in.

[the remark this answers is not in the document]
white238: merge away

PR #586, opened 2021-06-09
Use fmtlib to produce mint VTK output

kanye-quest:
# Summary

- Use {fmt} instead of `std::ofstream <<` to print formatted mesh data in `mint::write_vtk()`.
  A combination of faster float-to-string conversions in {fmt} and avoiding unnecessary stream flushes from calls to `std::endl` make this approach considerably faster than using ofstream's formatters.
  - Note that {fmt}'s default floating-point formatting may format some "small enough" numbers in fixed-point instead of scientific notation. I'm not sure if this would be a problem for some VTK readers.

Results measured with `time ./bin/mesh_tester -m bvh -p raja_cuda --infile ../car.stl --skipweld`, averaged over 4 runs:

| `develop` | this branch | factor improvement |
| ---------- | ----------- | -------------------- |
|  33.84s      | 3.97 s        | 8.51x


kennyweiss (approved):
Great improvement @kanye-quest !

white238 (approved):

agcapps (approved):
Thank you, @kanye-quest .  I see you avoided using the ternary operator to select between 2D and 3D, in favor of an `if` statement : ) (new lines 310 through 347).

bmhan12 (approved):

rhornung67 (approved):
Nice change @kanye-quest 👍

PR #588, opened 2021-06-10
Marks several static members with AXOM_EXPORT after changes in #578

kennyweiss:
# Summary

- Fixes msvc build after changes to dependencies in #578

### DIFF
--- a/src/axom/sidre/core/Group.hpp
+++ b/src/axom/sidre/core/Group.hpp
@@ -1588,7 +1588,7 @@ private:
   bool m_is_list;
 
   /// Character used to denote a path string passed to get/create calls.
-  static const char s_path_delimiter;
+  AXOM_EXPORT static const char s_path_delimiter;
 
   ///////////////////////////////////////////////////////////////////
   //
--- a/src/axom/slam/policies/SubsettingPolicies.hpp
+++ b/src/axom/slam/policies/SubsettingPolicies.hpp
@@ -14,8 +14,7 @@
  *   * isSubset(): bool -- returns whether the set is a subset of another set
  *   * parentSet() : ParentSetType -- returns a pointer to the parent set.
  *                                     nullptr when isSubset() is false
- *   * isValid() : bool -- indicates whether the Subsetting policy of the set
- *     is valid
+ *   * isValid() : bool -- indicates whether the Subsetting policy of the set is valid
  *   * [optional]
  *   * operator(): IntType -- alternate accessor for indirection
  */
@@ -23,6 +22,7 @@
 #ifndef SLAM_POLICIES_SUBSET_H_
 #define SLAM_POLICIES_SUBSET_H_
 
+#include "axom/config.hpp"
 #include "axom/core/Macros.hpp"
 
 #include "axom/slam/NullSet.hpp"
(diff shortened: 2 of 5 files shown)


rhornung67 (approved):
Is this change to support dynamic libraries?

src/axom/slam/policies/SubsettingPolicies.hpp
```diff
@@ -44,7 +44,7 @@ namespace policies
 
 struct NoSubset
```
cyrush, line 45: for these cases, should you export the struct that conatins these memberrs?

src/axom/sidre/core/Group.hpp
```diff
@@ -1588,7 +1588,7 @@ class Group
   bool m_is_list;
 
   /// Character used to denote a path string passed to get/create calls.
-  static const char s_path_delimiter;
```
cyrush, line 1591: same q: do we need to export group? That should take care of related static vars, maybe I am missing something.

kennyweiss:
@rhornung67 -- yes, this PR fixes our uberenv/vcpkg build on Windows, which builds the code in shared/dynamic mode.
We have a flag to export (nearly) all symbols:
https://github.com/LLNL/axom/blob/b6ec477ab6617cf18270328c588eb76f2f092291/scripts/vcpkg_ports/axom/portfile.cmake#L100
but for some reason, it doesn't export static class members, so we have to mark some of those with `AXOM_EXPORT`.

The current code is now passing our Windows-TPL tests: https://github.com/LLNL/axom/actions/runs/924399092

src/axom/slam/policies/SubsettingPolicies.hpp
kennyweiss: I'll try to add `AXOM_EXPORT` here instead of the static member and report back.

(Note that the current solution builds and passes all tests in our uberenv/vcpkg builds)

kennyweiss: @cyrush -- your suggestion also removes the link error, but introduces lots of link warnings of the form:
```
LINK : warning LNK4286: symbol '??0NoSubset@policies@slam@axom@@QEAA@XZ (public: __cdecl axom::slam::policies::NoSubset::NoSubset(void))' defined in 'OrderedSet.obj' is imported by 'State.obj' [<buid-axom>\axom\slam\examples\tinyHydro\slam_tinyHydro_sedov_ex.vcxproj]
```
This is the same thing we saw in #548.

On the other hand, marking the static members (as in the code that is currently checked in), does not appear to have these linker warnings (see the build logs for https://github.com/LLNL/axom/actions/runs/924399092)

I think we should keep the current solution (tagging the static members rather than the class/struct).

cyrush: Ok, that's a bit strange.

I suspect it has to do with the fact that these are header only and static members are going to be tricky b/c there isn't explicit storage associated with one compilation unit.  This may be a signal that future mysteries await us. 

Regardless, I agree with keeping what you have now.

white238 (approved):

agcapps (approved):

PR #589, opened 2021-06-10
Add hashmap to Axom Core

ksuarez1423:
# Summary

- This PR is a feature
- It does the following:
  - Adds a hashmap data structure to Core, at the request of Kenny Weiss.
  - Adds supporting tests for said data structure. 
  - This will be followed by PRs adding OpenMP support and GPU accelerator support. 

Initial development step for #574


src/axom/core/Map.hpp
```diff
@@ -0,0 +1,487 @@
...
+  
+  }; 
+
+  /// @}
+}
+
+/*!
+ * \class Map
+ *
+ * \brief Provides a hashmap implementation, relying upon chaining.
```
agcapps, line 266: "Relying upon chaining" to do what?  As a computer scientist, I think you mean to deal with hash collisions; please state so.

src/axom/core/tests/core_map.hpp
```diff
@@ -0,0 +1,109 @@
...
+  EXPECT_EQ(-2, ret.next);
+  test.insert(13, 900);
+  ret = test.find(13);
+  EXPECT_EQ(900, ret.value);
+}
+
+template <typename Key, typename T, int N>
+void test_rehash(Map <Key, T> &test)
+{
+  test.rehash();
```
kennyweiss, line 55: Please also test `rehash` with explicit values, rather than just the default values

kennyweiss (commented):
Looks like a great start @ksuarez1423 !

I added some initial comments.
Essentially:
* Please make the API for axom::Map match that of `std::unordered_map` to the extent possible
* Please increase the test coverage

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,487 @@
...
+#include "axom/core/Types.hpp"             // for axom types
+
+
+// C/C++ includes
+#include <functional> //for hashing until a custom method is defined
+#include <iostream>
+
+namespace axom
+{
+
```
kennyweiss, line 21: I wonder if this should be in an "experimental" namespace  until we've settled on an API that we want to expose to users.  E.g. `axom::experimental` ? (or something better)

Thoughts?

ksuarez1423: I personally think this is a great idea, but I'm uncertain if that's my decision to make -- what's the best way to have this conversation?

src/axom/core/tests/core_map.hpp
```diff
@@ -0,0 +1,109 @@
...
+  
+  for(int i = 0; i < N; i++)
+    {
+      test.insert(i, i*27-N/2);
+    }
+  
+  for(int i = 0; i < N; i++){
+    EXPECT_EQ(i*27-N/2, test.find(i).value);
+  }
+  auto ret = test.insert(N, 900);
```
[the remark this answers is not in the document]
ksuarez1423: My reading is that a phantom insertion doesn't exist by default for unordered_map, but wouldn't be too terrible to implement -- simply a matter of skipping the step of actual assignment. 

Since it requires a name, perhaps, well, phantom_insert? test_insert?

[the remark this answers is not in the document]
kennyweiss: I'm not a fan of `phantom_insert` (personally).
Some other possibilities: 
* `can_insert` 
* `check_insert`
* <not having this function>

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,487 @@
...
+  /*!
+   * \brief Returns the number of buckets in the Map.
+   * \return bucket_count 
+   */  
+  int get_bucket_count(){ return m_bucket_count; }
+  /*!
+   * \brief Returns the amount of items in the Map instance.
+   * \return size the amount of items in the Map instance.
+   */   
+  int get_size() { return m_size; }
```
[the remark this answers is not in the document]
ksuarez1423: Empty() is definitely something I need to add, thank you for catching that. Renaming to match the API is also immediately becoming a goal. For the full list of what we want to support from unordered_map, however, I think we need a longer conversation. I'll work on getting my thoughts together for that, and follow up with you.

src/axom/core/tests/core_map.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+namespace internal
+{
+
+template <typename Key, typename T, int N>
+Map <Key, T>  init()
+{
+  Map<Key, T> test(N/2, 2);
+  return test;
+
```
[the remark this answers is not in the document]
ksuarez1423: I'll make sure to, thank you!

src/axom/core/tests/core_map.hpp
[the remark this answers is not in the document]
ksuarez1423: I think 'can' if it returns a boolean, 'check' if it returns something more complex?

src/axom/core/Map.hpp
kennyweiss: Me have just had it :)

Any objections: @rhornung67 (and/or others) to placing `axom::Map` in `axom::experimental::Map` as a way of getting the code in early, without worrying about a finalized API at this stage?

src/axom/core/Map.hpp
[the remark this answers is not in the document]
kennyweiss: I agree that we do not want to replicate the entire API, and we should probably use different function names, where appropriate, when the behavior is significantly different

src/axom/core/Map.hpp
rhornung67: Sounds good.

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,487 @@
...
+        m_list[i].next = i+1;
+      }
+      m_list[len-1].next = -1;
+      m_free = 0;
+      m_head = -1;
+      m_size = 0;
+      m_capacity = len;
+      m_end.key = Key{0};
+      m_end.value = T{0};
+      m_end.next = -2;
```
kennyweiss, line 104: Should this just call `init` ?

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,487 @@
...
+     *
+     * \param [in] key the key of the key-value pair to be inserted.
+     * \param [in] value the value of the key-valure pair to be inserted.
+     *
+     * \return Returns a Pair object, with the first element being a pointer to the inserted node
+     *  and the second element being boolean value true if successful. If insertion fails,
+     *  the bool is set to False, and the node pointed to is the end node if the list was full,
+     *  or the item occupying the queried slot if an item with the given key already existed.
+     */
+    axom_map::Pair<Key, T> insert_no_update(Key key, T value){
```
kennyweiss, line 145: Efficiency/API issue: We will likely need to account for `T` types that we'd want to move rather than copy

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,487 @@
...
+    /// @}
+    /// \name Bucket Methods
+    /// @{
+    /*!
+     * \brief Inserts a given key-value pair into the linked list if an item with 
+     *  given key does not already exist. Fails if list is full or item with given key
+     *  already exists.
+     *
+     * \param [in] key the key of the key-value pair to be inserted.
+     * \param [in] value the value of the key-valure pair to be inserted.
```
kennyweiss, line 138: Typo:
```suggestion
     * \param [in] value the value of the key-value pair to be inserted.
```

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,487 @@
...
+   * \brief Removes key-value pair with given key from the Map instance.
+   *
+   * \param [in] key the key of the item to be removed from the Map instance.
+   *
+   * \note Deviation from STL unordered_map in returning a boolean instead of an iterator to next item in map.
+   *  Mostly useful on sparing logic when this moves to paralellism, since the bucket boundaries would become inconvenient.
+   *
+   * \return A bool value of true if the item was successfully removed, false otherwise. 
+   */  
+  bool remove(Key key){
```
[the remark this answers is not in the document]
ksuarez1423: API doesn't return the value, nor iterator to the value, but an iterator to the next member, when it returns an iterator, which is unfortunate at bucket boundaries.

However, now that I look, if I return a size_t, then this actually matches the behavior for erase(Keytype), because it only returns an iterator when you call erase with an iterator. Could this just be switched from a bool to a size_t (or IndexType), to match the 3rd version at that link?

src/axom/core/Map.hpp
```diff
@@ -383,7 +383,28 @@ class Map
...
+   *
+   * \param [in] key the key used to index into the Map to find/store item
+   *
+   * \note Major deviation from STL unordered_map in that this function is unsafe, due to inability to internally
+   *  call rehash function. User must be careful to ensure integrity of Map with repeated use of this function.
+   *
+   * \return A reference to the value of key-value pair in the Map instance that is associated with supplied Key.
+   *
+   */  
+  T& operator[](Key key)
```
kennyweiss, line 399: Suggestion: Maybe we should only support the `const` version of `operator[]`, at least initially

ksuarez1423: So read-only, thus not performing the auto-insertion step?

kennyweiss: That's what I was thinking (only a suggestion though).  

Perhaps we can add the non-const version down the line if/when we think of a clever way to do it.

ksuarez1423: While holding to the API of returning a reference, and avoiding exceptions, I am unaware of a more clever way to do this, yes. 

Though, if we wanted to break API and make this return a pointer, nullptr is always available.

kennyweiss: I think that breaking the API in than manner (i.e. returning a pointer instead of a reference) would be really confusing

ksuarez1423: Definitely would be. Just a thought of how to avoid this without throwing exceptions, which was indicated to me to be a bad thing for when this hits GPU.

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,739 @@
...
+  ~Map() { deallocate(); }
+
+  /// \name Map Methods
+  ///@{
+  /*!
+   * \brief Resizes Map by creating new Map instance and inserting all values
+   * from original Map.
+   *
+   * \param [in] buckets user-specified number of buckets in new Map. -1 to fall back to multiplicative factor.
+   * \param [in] factor user-specified multiplicative factor to determine size of new Map based upon existing.
```
nselliott, line 387: I think a word is left out here.  "new Map based upon existing size."?

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,739 @@
...
+      }
+    }
+    return check_rehash();
+  }
+  /*!
+   * \brief Deallocates memory resources for this Map instance, resets state as if constructed with 0 elements.
+   *  Also used in destructor.
+   *
+   */
+  void deallocate()
```
nselliott, line 446: STL containers use `clear()` for this operation.  Would it be better to follow that convention?

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,739 @@
...
+    m_bucket_len = 0;
+    m_size = 0;
+    m_load_factor = 0;
+  }
+
+  /*!
+   * \brief Initializes this Map instance, mostly in case deallocate was used and there's a desire to continue use.
+   *
+   */
+  void init(int num_buckets, int bucket_len = 10)
```
nselliott, line 463: Is this required before the user adds any new data into a Map that has been deallocated?  What happens if they deallocate, do not call this, and then try an insert?

nselliott:
Would it be a good idea to give Map a default number of buckets, or is that something we intend that the user should always be hands-on in specifying?

src/axom/core/Map.hpp
ksuarez1423: Definitely, I'll get that in.

src/axom/core/Map.hpp
ksuarez1423: Sure, that sounds like a good idea.

src/axom/core/Map.hpp
ksuarez1423: I don't think I've put any amount of protection on for that. I could put a conditional somewhere, perhaps, or better document to not clear then insert?

ksuarez1423:
@nselliott Given that there's no real way to determine what's best for applications in general, and that the empty Map constructor doesn't have a default number in STL, I think that the user should be hands-on in determining size.

kennyweiss (approved):
Thanks @ksuarez1423 -- this is a great start!

* Since I know the code/API will drift as we add more options, e.g. RAJA, host/device, I'm trying to focus my comments on higher level questions/thoughts.
* Thanks for making the API as similar as possible to `std`
* Thanks for improving the tests!
* I didn't look too closely at the internals of `Bucket`

src/axom/core/Map.hpp
nselliott: Definitely make it clear in the documentation that it is required.  I think if someone tries to insert with zero buckets it could be an ugly segfault, so it may be good to have some error checking there as well.

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,739 @@
...
+  /*!
+   *\brief Constructs a Map instance with the given number of buckets.
+   *
+   * \param [in] num_buckets the number of buckets used in the Map.
+   * \param [in] bucket_len the maximum number of items that can co-exist in a bucket.
+   *
+   * \pre num_buckets > 0
+   * \pre bucket_len > 0
+   */
+  Map(int num_buckets, int bucket_len = 10)
```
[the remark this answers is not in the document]
ksuarez1423: Current approach is a modification of STL -- in STL, if you specify a number, it's the minimum number of buckets you want, and length doesn't matter, because they use vectors. We don't have vectors, so bucket_len is there as a predefined number. 

10 is just a number. I'm open to other defaults, and don't have any strong emotions about the meaning of numbers -- 10 just seemed good enough, if there's not a ton of collisions.

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,739 @@
...
+      newlen = buckets;
+    }
+    else if(factor != -1 && factor > 1)
+    {
+      newlen = factor * m_bucket_count;
+    }
+    else
+    {
+      newlen = 2 * m_bucket_count;
+    }
```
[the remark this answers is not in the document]
ksuarez1423: The current check guarantees that it'll have more buckets than before, which strictly means it has enough space to continue from the previous state. If you had a smaller bucket count before, then you must have enough after rehash. 

I'm going to follow up with you on Teams about the asserts.

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,739 @@
...
+   *  Manual call of rehash() required.
+   *
+   * \note Can set "full bucket" flag for Map, which is only unset by rehashing.
+   *
+   * \return A pair whose first element is a pointer to the inserted item and bool set to true
+   *  if successful. Otherwise, the second element is set to false, and the first to one of two values: a sentinel node 
+   *  if the map is overfilled, and the actual node with the given key if an item with given key already exists.
+   *
+   */
+  axom_map::Pair<Key, T> insert(Key key, T val)
```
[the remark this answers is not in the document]
ksuarez1423: If we keep copying only, then I definitely see the benefits of passing a const reference, much like the C++ standard insertion this is matching. I'll need a minute to sit down to understand it, but if we want a move insertion at this stage, I can implement that too -- regardless, I'll swap this to const ref.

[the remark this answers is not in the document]
ksuarez1423: Should I go ahead and echo this change throughout find and erase as well, along with supporting functions in the Bucket?

src/axom/core/Map.hpp
ksuarez1423: I got that in, though I'm not necessarily pleased with the outcome -- may just revert to documentation, and throw an assert at it for debug builds.

agcapps (approved):
Thank you, @ksuarez1423 .  This is a solid foundation for Axom's map. I have my reservations about two design choices:
- requiring the user to specify number of buckets on Map construction
- requiring the call to init() after clear()
As @kennyweiss wrote, the current approach sounds reasonable.  However, if a better way becomes apparent later on, we should revise.

Also, please add the following tests (or point them out with comments, in case I blindly missed them):
- remove a nonexistent key
- several times insert_or_assign to the same key
- bucket overflow
- bucket underflow (remove last item that happens to land in the bucket)
- Map overflow (if that's different at all from bucket overflow)
- something to test check_rehash

nselliott (approved):
Thanks, @ksuarez1423

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,755 @@
...
+
+// C/C++ includes
+#include <functional>  //for hashing until a custom method is defined
+#include <iostream>
+
+namespace axom
+{
+namespace experimental
+{
+namespace axom_map
```
white238, line 22: was this namespace discussed? `axom::experimental::axom_map` is long and repetitive

ksuarez1423: https://github.com/LLNL/axom/pull/589#discussion_r653066151

ksuarez1423: The axom_map part is also internal -- the Map itself resides at axom::experimental::map as its own class.

src/axom/core/Map.hpp
white238: After @ksuarez1423 pointed me at this: this namespace should have the redundant `axom_` removed from it... so:

`axom::experimental::axom_map` to `axom::experimental::map`

src/axom/core/Map.hpp
```diff
@@ -0,0 +1,755 @@
...
+   * \param [in] bucket_len the maximum number of items that can co-exist in a bucket.
+   *
+   * \pre num_buckets > 0
+   * \pre bucket_len > 0
+   */
+  Map(int num_buckets, int bucket_len = 10)
+  {
+    init(num_buckets, bucket_len);
+    m_end.key = Key {0};
+    m_end.value = T {0};
```
nselliott, line 369: I know this has been merged, but I just noticed a potential problem here.  If the types `Key` or `T` are `std::string`, they cannot be assigned a value of 0.  We would use `std::string` for `Key` if we use this `Map` class in sidre, for example

This implementation is fine if we have the restriction that the types must be numerical, but I'm not sure what the correct implementation is to support both numerical types and strings.

PR #591, opened 2021-06-15
Extend isUserProvided() to accept a name parameter

joshessman-llnl:
# Summary

- This PR is a feature
  - Adds a name parameter (via overload) to `Container::isUserProvided` to allow for easy querying of subobjects

Provides a better solution for the workaround implemented in https://github.com/LLNL/serac/pull/491

### DIFF
--- a/src/axom/inlet/Inlet.hpp
+++ b/src/axom/inlet/Inlet.hpp
@@ -261,6 +261,18 @@ public:
   }
 
   /*!
+   *****************************************************************************
+   * \brief Return whether a subobject with the given name was provided by the user
+   *
+   * \see Container::isUserProvided
+   *****************************************************************************
+   */
+  bool isUserProvided(const std::string& name) const
+  {
+    return m_globalContainer.isUserProvided(name);
+  }
+
+  /*!
    *******************************************************************************
    * \brief Obtains a proxy view into the datastore.
    * 
--- a/src/axom/inlet/tests/inlet_object.cpp
+++ b/src/axom/inlet/tests/inlet_object.cpp
@@ -922,6 +922,7 @@ TYPED_TEST(inlet_object, default_scalar_user_provided)
   auto& field = static_cast<axom::inlet::Field&>(scalar);
   EXPECT_TRUE(field.exists());
   EXPECT_FALSE(field.isUserProvided());
+  EXPECT_FALSE(inlet.isUserProvided("foo"));
 
   // ...but it should still be possible to retrieve the default
   const int foo = inlet["foo"];
@@ -943,6 +944,7 @@ TYPED_TEST(inlet_object, default_struct_field_user_provided)
   // The container itself exists but was not provided by the user
   EXPECT_TRUE(foo_container.exists());
   EXPECT_FALSE(foo_container.isUserProvided());
+  EXPECT_FALSE(inlet.isUserProvided("foo"));
 
   // ...but it should still be possible to retrieve the default
   const Foo expected_foo {true, false};
@@ -965,6 +967,7 @@ TYPED_TEST(inlet_object, default_struct_field_user_provided_reqd)
   // The container itself exists but was not provided by the user
   EXPECT_TRUE(foo_container.exists());
   EXPECT_FALSE(foo_container.isUserProvided());
+  EXPECT_FALSE(inlet.isUserProvided("foo"));
 
   // ...but it should still be possible to retrieve the default
   const Foo expected_foo {true, false};
@@ -990,6 +993,7 @@ TYPED_TEST(inlet_object, default_struct_field_marked_true)
   // The container itself exists and was provided by the user
   EXPECT_TRUE(foo_container.exists());
   EXPECT_TRUE(foo_container.isUserProvided());
+  EXPECT_TRUE(inlet.isUserProvided("foo"));
 
   // ...and it should still be possible to retrieve the full struct
   const Foo expected_foo {true, false};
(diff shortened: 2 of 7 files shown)


src/axom/inlet/Inlet.hpp
```diff
@@ -260,6 +260,18 @@ class Inlet
...
   }
 
+  /*!
+   *****************************************************************************
+   * \brief Return whether a subobject with the given name was provided by the user
+   *
+   * \see Container::isUserProvided
+   *****************************************************************************
+   */
+  bool isUserProvided(const std::string& name) const
```
kennyweiss, line 270: minor: Should this be `hasUserProvided(name)` instead of `isUserProvided(name)` ?

I.e.  the full call on a container is `container.isUserProvided(name)`, but it seems like you're raelly asking if the container **has** a user provided subobject named `name`

src/axom/inlet/tests/inlet_object.cpp
```diff
@@ -990,6 +993,7 @@ TYPED_TEST(inlet_object, default_struct_field_marked_true)
   // The container itself exists and was provided by the user
   EXPECT_TRUE(foo_container.exists());
   EXPECT_TRUE(foo_container.isUserProvided());
+  EXPECT_TRUE(inlet.isUserProvided("foo"));
```
kennyweiss, line 996: Consider also adding a check for the following with comments?
* `isUserProvided("bar")` 
* `isUserProvided("baz")` 
* `isUserProvided("foo/bar")` 
* `isUserProvided("foo/baz")` 
* `isUserProvided("qux")` 

Based on the examples and comments, I'm not entirely clear on which entities will return `true` for `isUserProvided`.  
I think `foo/bar` would be `true`, and the rest `false`, but I'm not entirely sure about the former.

kennyweiss (changes requested):
I think this PR would benefit from additional comments/tests that help distinguish between `isUserProvided()`  vs.  <inlet equivalent of `isInSchema()`>  (I'm not sure offhand what the latter is called)

The comments in the tests were helpful, but I think I'm still missing something.

src/axom/inlet/Inlet.hpp
joshessman-llnl: The current functionality is `hasUserProvided()` (no args) on a `Container` or `Field` which checks the calling object - I agree that semantically `hasUserProvided(string name)` called with a subobject name makes more sense, but I'm not sure if it would be confusing to have two methods with names that are so similar.

src/axom/inlet/tests/inlet_object.cpp
kennyweiss: Thanks for updating the tests @joshessman-llnl 

My only remaining question is about how this would work if the test string has a field that is not in the schema.  E.g.  the input is ` foo = {quux = true }`, but the schema is `foo.bar`.
Would  `foo.isUserContained("quux")` be true ?

kennyweiss (approved):
Thanks for updating the tests @joshessman-llnl -- it is much clearer (to me) how `.contains()` and `.isUserProvided()` works and their distinction, but see my comments for another question.

In the future (i.e. outside the scope of this PR), it might be worthwhile considering better names for these functions, since `contains` might imply that is in the user provided data, rather than in the schema.

src/axom/inlet/tests/inlet_object.cpp
joshessman-llnl: It would not - Inlet's logic only calls out to the reader when something is added to the schema, so neither `contains` nor `isUserProvided` would return true for "quux".  However, the `unexpectedNames` logic added a little while back would include "quux".

joshessman-llnl:
> 
> In the future (i.e. outside the scope of this PR), it might be worthwhile considering better names for these functions, since `contains` might imply that is in the user provided data, rather than in the schema.

I definitely agree that there's some ambiguity with these names - `isUserProvided` I think is clear because it returns true if something was explicitly provided/specified by the user in the input file, but `contains` will also return true if a default was provided.  As of right now there's no way of checking to see if something was added to the schema but not provided either via a default or the input file, though that's what `contains` used to do.  `isUserProvidedOrDefaulted` is more accurate but less succinct, so I'll have to think about it.

kennyweiss:
@joshessman-llnl wrote:
>`contains` will also return true if a default was provided.

Now I'm confused again -- it sounds like there are several categories:
* `isUserProvided()` -- in the schema and the user input provided a value
* `contains()` -- in the schema, and either the user input provided a value **or** the schema provides a default
* ???  -- in the schema; the user did not provide a value and there is no default
* ???  -- not in the schema, but the user provided it in the input.  (I don't think this has a function, but can be obtained indirectly through `unexpectedNames()` )

Is there a function to check if a   Field/Function/Container/...   is in the schema, but doesn't have a value ?

joshessman-llnl:
> 
> 
> @joshessman-llnl wrote:
> 
> > `contains` will also return true if a default was provided.
> 
> Now I'm confused again -- it sounds like there are several categories:
> 
>   * `isUserProvided()` -- in the schema and the user input provided a value
> 
>   * `contains()` -- in the schema, and either the user input provided a value **or** the schema provides a default
> 
>    * ???  -- in the schema; the user did not provide a value and there is no default
> 
>   * ???  -- not in the schema, but the user provided it in the input.  (I don't think this has a function, but can be obtained indirectly through `unexpectedNames()` )
> 
> 
> Is there a function to check if a Field/Function/Container/... is in the schema, but doesn't have a value ?

All the above is correct - there is not currently (convenient) functionality to check if something is *just* in the schema (the third bullet pt), though if a user really wanted to they could search "manually" with `Container::getChildFields`/ `Container::getChildContainers` etc.  

I'm not sure how helpful this additional functionality would be - since the Inlet user is obviously in control of what gets added to the schema, they should "know" what's in the schema later on.  An exception to this might be if things were conditionally added to the schema based on some other (external) condition, but @white238 and I had previously discussed this and decided that that's not really something we'd want to encourage - that is, we'd prefer that the schema always be the same no matter what.  I could definitely be missing something though, was there a use case you had in mind?

white238:
Looks great @joshessman-llnl !  Can you add a section to the sphinx docs about the difference between `contains` and `isUserProvided`?

white238 (approved):

PR #593, opened 2021-06-16
Remove blueos requirement on cuda logic

white238:
This was found during getting Serac to build `+cuda` on a personal laptop.  Basically it just removes the blueos checks from the generic cuda logic.

I tested this against all specs on rzansel and they passed.

### DIFF
--- a/scripts/spack/packages/axom/package.py
+++ b/scripts/spack/packages/axom/package.py
@@ -95,7 +95,7 @@ class Axom(CachedCMakePackage, CudaPackage):
     depends_on("conduit~hdf5", when="~hdf5")
 
     # HDF5 needs to be the same as Conduit's
-    depends_on("hdf5@1.8.19:1.8.999~cxx~shared~fortran", when="+hdf5")
+    depends_on("hdf5@1.8.19:1.8.999~cxx~fortran", when="+hdf5")
 
     depends_on("lua", when="+lua")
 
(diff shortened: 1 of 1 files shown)


agcapps (approved):
Thanks for lifting this restriction, @white238

scripts/spack/packages/axom/package.py
```diff
@@ -95,7 +95,7 @@ class Axom(CachedCMakePackage, CudaPackage):
     depends_on("conduit~hdf5", when="~hdf5")
 
     # HDF5 needs to be the same as Conduit's
-    depends_on("hdf5@1.8.19:1.8.999~cxx~shared~fortran", when="+hdf5")
+    depends_on("hdf5@1.8.19:1.8.999~cxx~fortran", when="+hdf5")
```
kennyweiss, line 98: Minor: should the `~cxx` and `~fortran` restrictions on `hdf5` be moved to our `packages.yaml` ?

I.e. we do not expect problems building `axom` against an hdf5@1.8 that was built w/ `+cxx` or `+fortran`, we just choose not to build these internally for `axom`, right?

(Inspired by a relatively recent conversation w/ @scheibelp )

kennyweiss (approved):
Thanks @white238

scripts/spack/packages/axom/package.py
white238: I think with the new concretizer we dont needs this restriction in axom and can just leave it up to conduit but the clingo concretizer isn't default yet.

kanye-quest (approved):

PR #594, opened 2021-06-17
Feature/kweiss/multimat_revamp 

raineyeh:
# Summary

- This PR is to fix the multimat unit tests
- It does the following:
  - Make unit test compatible with the new layout-per-field feature
  - Fix traversal to run with just 1 component (until field2D with higher number of components are added)
  - Reduce unit test to use 1 component as that's temporarily hard-coded


kennyweiss:
Thanks @raineyeh !

Looks like there's a compiler error with our gcc8 test related to an incompatible return type for `SubMap::size()``
Could you please look into it?
https://dev.azure.com/healy22/Status%20Check%20Testiing/_build/results?buildId=3393&view=logs&j=1e949c51-e916-5ae4-436c-7a23f916c6ae&t=3cb93f51-2f5f-529e-e892-962657e1e8f0

PR #595, opened 2021-06-17
Bugfix for empty string literals in XLF

kennyweiss:
# Summary

- This PR is a bugfix for Fortran tests with XLF
- It converts empty string literals (`""`) into string literals with a space (`" "`) in some comparisons a string which is expected to be empty
- These literals are implicitly trimmed by XLF for the actual comparison
- I think we've encountered this in the past, but couldn't find it
- I also ran clang-format for two files that were somehow missed and are passing in our CI


white238 (approved):

cyrush (approved):

rhornung67 (approved):

agcapps (approved):
Thanks for the fixes.

PR #596, opened 2021-06-22
Fix compiler checks that were mistakenly removed

white238:
This PR returns us to a working state after a couple changes on the Spack side:

* Restores some necessary compiler checks that were mistakenly removed from a block of code in our spack package
* Returns the `~shared` variant to the conduit `depends_on` line and adds the comment that this can be removed when the new concretizer is enabled and that it should not be pushed back to spack proper

This PR really doesn't have a change from what was working before, so I opted to not rebuild all the TPLs again. I tested it on toss3 and blueos.


kennyweiss (approved):
Thanks @white238

rhornung67 (approved):

agcapps (approved):

PR #597, opened 2021-06-23
Upgrade our vcpkg usage

kennyweiss:
# Summary

- This PR updates our commit for vcpkg to its `2021.05.12` release: https://github.com/microsoft/vcpkg/tree/2021.05.12
- It updates our local package ports to the newer "manifest format", which uses `vcpkg.json` files instead of `CONTROL` files. See https://vcpkg.readthedocs.io/en/latest/maintainers/manifest-files/
- The newer `vcpkg` uses `hdf5@1.12`. Luckily there was already a PR in conduit for this (https://github.com/LLNL/conduit/pull/727).  I converted this to a patch for vcpkg and then did some debugging.


.uberenv_config.json
```diff
@@ -8,7 +8,7 @@
 "spack_configs_path": "scripts/spack/configs",
 "spack_packages_path": "scripts/spack/packages",
 "vcpkg_url": "https://github.com/microsoft/vcpkg",
-"vcpkg_commit": "b79f7675aaa82eb6c5a96ae764fb1ce379a9d5d6",
+"vcpkg_commit": "5568f110b509a9fd90711978a7cb76bae75bb092",
```
kennyweiss, line 11: Points to this release of `vcpkg`: https://github.com/microsoft/vcpkg/releases/tag/2021.05.12

scripts/vcpkg_ports/axom/vcpkg.json
```diff
@@ -0,0 +1,21 @@
+{
```
kennyweiss, line 1: Newer `vcpkg` is using `vcpkg.json` manifest files instead of `CONTROL` files within its ports

scripts/vcpkg_ports/conduit/hdf5-1-12.patch
```diff
@@ -0,0 +1,258 @@
+diff --git a/src/libs/relay/conduit_relay_io_hdf5.cpp b/src/libs/relay/conduit_relay_io_hdf5.cpp
```
kennyweiss, line 1: Adapted from https://github.com/LLNL/conduit/pull/727

scripts/vcpkg_ports/conduit/portfile.cmake
```diff
@@ -30,10 +28,14 @@ vcpkg_configure_cmake(
...
         -DHDF5_DIR=${CURRENT_INSTALLED_DIR}
+        -DCONDUIT_INSTALL_CONFIG_DIR="share/conduit"
+        -DCONDUIT_INSTALL_CMAKE_MODULE_DIR="share"
 )
 
 vcpkg_install_cmake()
-vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake)
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake
+                          TARGET_PATH share
+                          TOOLS_PATH tools/conduit)
```
kennyweiss, line 38: @cyrush -- This part ended up being really tricky (especially since each change required a `vcpkg remove` / `vcpkg install` step (similar to debugging spack recipes). The code here works, but doesn't seem ideal/elegant.

Some notes:
* `vcpkg` wants all the cmake installation files to be in `<install_dir>/share/conduit`
* `conduit` has some internal logic to modify the output directory for these files
* Without my above modifications, the previous port was placing them in `share/conduit/conduit`
* The corresponding logic in conduit is here: https://github.com/LLNL/conduit/blob/de9d180acf0f954d50aa0bac20fb97cb5c063362/src/config/CMakeLists.txt#L37-L85

agcapps (approved):
Thanks for protecting this particular corner from bit rot.  I see no reason to delay this PR, nor any problem with it.  The change from flat-text CONTROL files to yaml was arguably for the best.  My one substantive comment was about the `description` in axom/vcpkg.json

agcapps (approved):
👍

scripts/vcpkg_ports/conduit/portfile.cmake
kennyweiss: Looks like there's something different in my environment from that of the "github actions":
* https://github.com/LLNL/axom/actions/runs/965032062

Axom's config failed there due to bad paths in conduit's config:
```
CMake Error in axom/CMakeLists.txt:
  Imported target "conduit::conduit" includes non-existent path

    "D:/a/axom/axom/uberenv_libs/vcpkg/installed/include/"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.
  * An install or uninstall procedure did not complete successfully.
  * The installation package was faulty and references files it does not provide.
```
See: https://github.com/LLNL/axom/runs/2897044456?check_suite_focus=true#step:8:526

scripts/vcpkg_ports/axom/vcpkg.json
```diff
@@ -0,0 +1,21 @@
...
+  "name": "axom",
+  "version-string": "develop",
+  "homepage": "https://github.com/llnl/axom",
+  "description": "Host-config generator for Axom TPLs",
```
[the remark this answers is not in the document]
kennyweiss: It currently only lists the dependencies/features that we want to build and generates the host-config.
It does not build axom, so that's why wrote it that way.

This is similar to the "fake" packages that we used to use in our `spack` setup.

scripts/vcpkg_ports/conduit/portfile.cmake
cyrush: this looks just like the issue @white238  and @joshessman-llnl saw when looking for conduit in the old vs new way. 

Yes, the cmake files for conduit are in `lib/cmake/conduit`, not under share. 

Conduit uses relative location of the cmake folder to help understand everything else. Moving it relative to the dir structure in a new way isn't going to work.

I'll have to think about this.

kennyweiss:
I patched conduit's setup_deps to account for the vcpkg installation and it now passes our Windows TPL tests: 
https://github.com/LLNL/axom/actions/runs/972579755

scripts/vcpkg_ports/conduit/setup_deps_vcpkg_triplet.patch
```diff
@@ -0,0 +1,16 @@
...
+--- a/src/config/conduit_setup_targets.cmake
++++ b/src/config/conduit_setup_targets.cmake
+@@ -15,6 +15,11 @@ endif()
+ # we want the import root, which is right above the "lib" prefix
+ get_filename_component(_IMPORT_ROOT "${_IMPORT_PREFIX}" PATH)
+ 
++# Patch for VCPKG -- need to account for the config type
++if(VCPKG_TARGET_TRIPLET)
++    set(_IMPORT_ROOT "${_IMPORT_ROOT}/${VCPKG_TARGET_TRIPLET}")
++endif()
```
kennyweiss, line 12: @cyrush -- here's the patch to `conduit_setup_targets.cmake` to get this working

white238 (approved):

scripts/vcpkg_ports/conduit/portfile.cmake
kennyweiss: Thanks @cyrush  -- I got it working in `vcpkg` with a very small patch (that only `vcpkg` applies) -- see my comment below.

PR #600, opened 2021-06-25
Add cartesian mesh to sidre_generateindex example

nselliott:
# Summary

- This PR is an enhancement of an example code
- It does the following:
  - Adds some code that creates a simple multi-domain cartesian blueprint mesh in sidre_generateindex example.
  - This creates some output datasets that can be used for testing support of multi-domain blueprint meshes in viz tools.


src/axom/sidre/examples/sidre_generateindex.cpp
```diff
@@ -166,7 +199,7 @@ void setup_blueprint_fields(DataStore* ds, Group* fields)
   Group* nodefield = fields->createGroup("nodefield");
   nodefield->createViewString("association", "vertex");
   nodefield->createViewString("type", "scalar");
-  nodefield->createViewString("topology", "mesh");
+  nodefield->createViewString("topology", "cartesianesh");
```
kennyweiss, line 202: Looks like this has a typo:
```suggestion
  nodefield->createViewString("topology", "cartesian");
```

Are we calling `blueprint::verify` on this? 
* If not, can we please add it.
* If so, does it pass?

src/axom/sidre/examples/sidre_generateindex.cpp
```diff
@@ -111,6 +111,29 @@ void setup_blueprint_coords(DataStore* ds, Group* coords)
                          origv->getBuffer());
 }
 
+void setup_cartesian_coords(Group* coords, int domain_id)
+{
+  // 5x5 vertices, 4x4 elements
+  int nx = 5;
+  int ny = 5;
```
kennyweiss, line 118: Can we please make these static/global variables at the top of the file? 
That would remove the implicit assumptions about mesh size (5,5 / 4,4) throughout the file.

kennyweiss (approved):
Thanks @nselliott

agcapps (approved):
Thank you, @nselliott

PR #601, opened 2021-07-02
BVH public API changes + policy-based class

kanye-quest:
# Summary

- Refactors the `spin::BVH` public-facing API, based on user suggestions
  - Constructor only handles setting up default values, while the actual initialization of the `BVH` is done in a `BVH::initialize(...)` method, which replaces `BVH::build()`. This also allows re-initializing an existing `BVH` instance with a new set of bounding boxes.
  - Use Primal objects in the public API:
    - Initialization now uses a `BoundingBox*` array instead of array of `FloatType*`
    - `void BVH::getBounds(float* min, float* max)` is now `BoundingBox BVH::getBounds()`
    - `BVH::findPoints/Rays/BoundingBoxes` uses `Point*/Ray*/BoundingBox*`, respectively, instead of arrays of raw coordinates for each dimension
  - Corrects some misspellings (`getScaleFacor() -> getScaleFactor()`)
- `BVH` now uses functionality from `internal/linear_bvh` via a defined policy class `spin::policy::LinearBVH`; this should allow for defining alternative BVH construction or traversal methods in the future
  - `LinearBVH` consolidates most of the functionality from `BVHData.hpp`, `emit_bvh.hpp`, and the traversal code from `BVH_impl.hpp`
- Remove requirement for RAJA/Umpire from CPU-only LinearBVH
- Remove `QueryAccessor.hpp` and consolidate `BVH::find*` functions by templating on the Primal query shape type


src/axom/primal/geometry/BoundingBox.hpp
```diff
@@ -551,8 +552,8 @@ void BoundingBox<T, NDIMS>::checkAndFixBounds()
 template <typename T, int NDIMS>
 void BoundingBox<T, NDIMS>::clear()
 {
-  m_min = PointType(std::numeric_limits<T>::max());
-  m_max = PointType(std::numeric_limits<T>::lowest());
+  m_min = PointType(InvalidMin);
```
steveg21, line 555: Makes sense

src/axom/spin/BVH.hpp
```diff
@@ -125,36 +140,30 @@ class BVH
...
-   * \brief Default constructor. Disabled.
+   * \brief Default constructor.
    */
-  BVH() = delete;
+  BVH() : m_AllocatorID(axom::execution_space<ExecSpace>::allocatorID()) { }
 
   /*!
-   * \brief Creates a BVH instance, of specified dimension, over a given set
+   * \brief Initializes a BVH instance, of specified dimension, over a given set
    *  of geometric entities, each represented by its corresponding axis-aligned
```
steveg21, line 159: What are the geometric entities here?  Perhaps this is a cut & paste leftover.

src/axom/spin/BVH.hpp
```diff
@@ -168,14 +177,11 @@ class BVH
    * \pre boxes != nullptr
    * \pre numItems > 0
    */
-  BVH(const FloatType* boxes,
-      IndexType numItems,
-      int allocatorID = axom::execution_space<ExecSpace>::allocatorID());
+  int initialize(const BoxType* boxes,
```
steveg21, line 180: Looks good.  I'm OK with just adding the arguments to a ctor, but the disadvantage is that the ctor can throw exceptions with that approach.

src/axom/spin/BVH.hpp
```diff
@@ -195,40 +201,40 @@ class BVH
...
-   *
-   * \note min/max point to arrays that are at least NDIMS long.
-   *
-   * \pre min != nullptr
-   * \pre max != nullptr
+   * \return box the bounding box of the constructed BVH, or an invalid
+   *  bounding box if the BVH has not been initialized.
    */
-  void getBounds(FloatType* min, FloatType* max) const;
+  BoxType getBounds() const
```
steveg21, line 226: Much more consistent with the other parts of Axom now

src/axom/spin/BVH.hpp
```diff
@@ -336,23 +314,13 @@ class BVH
...
    * \pre candidates == nullptr
-   * \pre xmin != nullptr
-   * \pre xmax != nullptr
-   * \pre ymin != nullptr
-   * \pre ymax != nullptr
-   * \pre zmin != nullptr if dimension==3
-   * \pre zmax != nullptr if dimension==3
+   * \pre boxes != nullptr
    */
   void findBoundingBoxes(IndexType* offsets,
```
steveg21, line 319: This, and the other new find() functions are much less error-prone for the caller than the original versions.

steveg21 (approved):
I've only had a chance to review the public interface, but so far it looks good.  It's much less error-prone for callers than the original.

src/axom/spin/BVH.hpp
```diff
@@ -168,14 +177,11 @@ class BVH
    * \pre boxes != nullptr
    * \pre numItems > 0
    */
-  BVH(const FloatType* boxes,
-      IndexType numItems,
-      int allocatorID = axom::execution_space<ExecSpace>::allocatorID());
+  int initialize(const BoxType* boxes,
+                 IndexType numItems,
+                 int allocatorID = axom::execution_space<ExecSpace>::allocatorID());
```
kennyweiss, line 182: Design question: Given that BVH now has a member `allocator`, should `initialize` have a default value for `allocatorID`?

A different idea might be provide two overloads:
```cpp
BVH::initialize(const BoxType*, IndexType); // use the member allocator
BVH::initialize(const BoxType*, IndexType, int allocatorID); // overload, use the provided allocator instead
```
Where the first function would call the second using the member allocator.

The distinction would be if the user modified the `axom::allocatorID`, between calling the `BVH` constructor and calling `initialize`. I'm pretty sure that `axom::execution_space<ExecSpace>::allocatorID()` is a singleton.

src/axom/spin/BVH.hpp
```diff
@@ -195,40 +201,40 @@ class BVH
...
+  FloatType getScaleFactor() const { return m_scaleFactor; };
 
   /*!
    * \brief Sets the tolerance used for querying the BVH.
    * \param [in] TOL the tolerance to use.
    *
    * \note Default tolerance set to floating_point_limits<FloatType>::epsilon()
    */
-  void setTolerance(FloatType TOL) { m_Tolernace = TOL; };
+  void setTolerance(FloatType TOL) { m_tolerance = TOL; };
```
kennyweiss, line 212: Consistency (minor): We use `EPS` for the tolarance variable in most of `primal`. It might be good to make the change here for consistency throughout `axom`

```suggestion
  void setTolerance(FloatType EPS) { m_tolerance = EPS; };
```

src/axom/spin/BVH.hpp
```diff
@@ -275,12 +275,7 @@ class BVH
    * \param [out] counts stores the number of candidates for each ray
    * \param [out] candidates array of candidate IDs for each ray
    * \param [in] numRays the total number of rays
-   * \param [in] x0 array consisting the ray source point x-coordinates.
-   * \param [in] nx array consisting the ray normal x-components.
-   * \param [in] y0 array consisting the ray source point y-coordinates
-   * \param [in] ny array consisting the ray normal y-components
-   * \param [in] z0 array consisting the ray source point z-coorindates (in 3D)
-   * \param [in] nz array consisting the ray normal z-components (in 3D)
+   * \param [in] rays array of the rays to query against the BVH
```
kennyweiss, line 278: If a user already has the ray data as separate structs of arrays (e.g. `x0*`, `y0*`, `z0*`, `nx*`, `ny*`, `nz*`), what would be the best way for them to call this function?

Would they need to allocate a new array of `primal::Ray`s and construct them in a kernel?

Should we provide both options within the `BVH` class? ... and possibly use an internal `zip` function to compose the rays on the fly?

Note: Asking b/c I believe this is how some user codes already have their data laid out, and the extra allocation would be undesirable.

src/axom/spin/BVH.hpp
```diff
@@ -364,24 +332,195 @@ class BVH
 private:
   /// \name Private Members
   /// @{
+  static constexpr FloatType DEFAULT_SCALE_FACTOR = 1.001;
```
kennyweiss, line 335: Minor: I ran into some weird cases in the InOutOctree class that needed a less "regular" default spacing.
The following seemed to work better: `1.000123`

This might not be an issue for the BVH class, but it would be worthwhile testing this against the following mesh: 
https://github.com/LLNL/axom_data/blob/main/quest/cap.stl

See also:
 * https://github.com/LLNL/axom/blob/7e2c27b2229e2778281e755fc0d43bcf93119cdd/src/axom/quest/InOutOctree.hpp#L429
* https://github.com/LLNL/axom/issues/513

kennyweiss (approved):
I really like the new API @kanye-quest !

My only main concern is about user codes that already have their data laid out using structs of arrays, and would prefer the older API.  Could we add utility functions for them to still use the older API, but perhaps `zip` together the array elements into a `primal` primitive on the fly (i.e. without allocating and copying the query data)?

Also: Do you have a sense on any impact (positive, negative or neutral) on the performance impacts of this PR?

- [ ] Please update the `RELEASE-NOTES` to indicate changes to the API.

src/axom/spin/BVH.hpp
kanye-quest: That's definitely something worth considering.

I think it would be preferable to keep the API surface small. We might be able to make the `findRays` function templated on a `typename RayIterable`, so we could accept any object that returns a `primal::Ray` when it is dereferenced or accessed with `it[idx]` (and likewise for `findPoints/findBoundingBoxes`)

We could then define something akin to a `zip_iterator` within primal for `primal::Ray/Point/BoundingBox` so that user codes could pass in their SoA data without allocating another array for the primal primitives.

src/axom/spin/BVH.hpp
kanye-quest: Good point: if `allocatorID` is a member field, it doesn't make much sense to set it to a default `allocatorID()` in the constructor, then do it again in `initialize()`.

How do you feel about removing the allocator argument from `initialize()` entirely, and making it an optional `setAllocatorID(int)` member function? All we really do with that argument is set `m_allocatorID = allocatorID`.

kennyweiss: That works for me.

Maybe add a note to the doxygen that it uses a previously set allocator in `setAllocatorID` or the default allocator?

src/axom/spin/BVH.hpp
kennyweiss: That sounds like a really nice compromise! and a generally useful addition to primal/spin

src/axom/spin/BVH.hpp
agcapps: I agree with Kenny's suggestion.

src/axom/spin/tests/CMakeLists.txt
```diff
@@ -43,15 +43,16 @@ foreach ( test ${spin_tests} )
...
+  NAME       spin_bvh_test
+  SOURCES    spin_bvh.cpp
+  OUTPUT_DIR ${TEST_OUTPUT_DIRECTORY}
+  DEPENDS_ON ${spin_depends}
+  FOLDER     axom/spin/tests
+  )
+
+if ( RAJA_FOUND AND UMPIRE_FOUND )
 
     axom_add_test(
```
agcapps, line 57: Consider calling `axom_add_test` outside the test `if ( RAJA_FOUND AND UMPIRE_FOUND )`, so that the only thing inside the test is the call to `set_property(... "CUDA_VISIBLE_DEVICES=0")`

src/axom/spin/tests/spin_bvh.cpp
```diff
@@ -1757,44 +1437,46 @@ AXOM_CUDA_TEST(spin_bvh, use_pool_allocator)
...
-  FloatType* boxes = axom::allocate<FloatType>(6, allocID);
+  BoxType* boxes = axom::allocate<BoxType>(1, allocID);
   axom::for_all<exec>(
     0,
-    6,
-    AXOM_LAMBDA(axom::IndexType idx) { boxes[idx] = (idx < 3) ? 0.0 : 1.0; });
+    1,
+    AXOM_LAMBDA(axom::IndexType idx) {
+      boxes[idx] = {PointType(0.), PointType(1.)};
+    });
```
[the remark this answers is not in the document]
agcapps: Agreed.

agcapps (approved):
Thank you, @kanye-quest .  Much cleaner than before.

src/axom/spin/internal/linear_bvh/build_radix_tree.hpp
```diff
@@ -528,6 +532,25 @@ AXOM_HOST_DEVICE static inline void sync_store(BBoxType& box,
 #endif
 }
 
+//------------------------------------------------------------------------------
```
steveg21, line 535: Code above here produces a warning when I compile since macros aren't defined: "build_radix_tree.hpp(454): warning: variable "nreads" was declared but never referenced"

steveg21 (approved):

src/axom/spin/tests/spin_bvh.cpp
```diff
@@ -227,68 +155,70 @@ void generate_aabbs3d(const mint::Mesh* mesh, FloatType*& aabbs)
...
   EXPECT_EQ(NDIMS, mesh->getDimension());
 
   // allocate output arrays
   const IndexType ncells = mesh->getNumberOfCells();
-  aabbs = axom::allocate<FloatType>(ncells * STRIDE);
+  aabbs = axom::allocate<BoxType>(ncells);
 
 #ifdef __ibmxl__
 
   /* workaround for IBM XL compiler */
```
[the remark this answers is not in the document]
kanye-quest: I didn't get any segfaults on the `centroid` array, but there were some weird segfaults that occured when accessing the `primal::Point sum` local variable. However, they seemed to go away after I replaced a multiplication with a `constexpr double` with division by a `constexpr int`. Seems like an issue with XL's handling of constexpr?

src/axom/spin/BVH.hpp
kanye-quest: Done 👍

src/axom/spin/BVH.hpp
kanye-quest: Done 👍

src/axom/spin/tests/CMakeLists.txt
kanye-quest: Done 👍

src/axom/spin/internal/linear_bvh/build_radix_tree.hpp
kanye-quest: That warning should be suppressed now.

src/axom/spin/BVH.hpp
kanye-quest: I've added the API change for accepting array-like types for the query objects in `find*()`, so they should be able to accept any object which is indexable with `operator[]`.

As for the `zip_iterator` types, I'll work on introducing that in a future PR. I've opened up an issue here: #604

PR #605, opened 2021-07-12
Adds support for loading C2C mesh files in quest

kennyweiss:
# Summary

- This PR add support to quest for reading C2C `.contour` files
- The contours are converted to NURBS and can be sampled onto a linear segment mesh using `C2CReader::getLinearMesh()`
- When the 2D segment mesh is closed, we can pass it into `quest::InOutOctree<2>` for in/out queries
- It updates the quest `containment_driver` example to support loading in 2D contours from the command line, generating an `InOutOctree<2>` over the contours and sampling the in/out field at the vertices of a regular grid
- ~~This PR also upgrades axom's built-in CLI11 library from `v1.8.0` to `v1.9.2`~~ 
  CLI11 update was moved to #607
- It updates the `inout` C API (and driver example) to support 2D queries
- It includes a bugfix for Mint's  VTK writer to support dumping fields of type `float` and `int64`. 

### Example
For this example contour file:
```
>cat test_spline.contour 
point = spline_start
piece = rz(  units=cm, spline=cubic, 
             beginTan=-90deg, endTan=-90deg, 
             rz=6.283185307179586 1.5
		5.235987755982989 0.5
		4.1887902047863905 1.5
		3.141592653589793 0.5
		2.0943951023931953 1.5
		1.0471975511965976 0.5
		0 1.5)
point = spline_end
piece = line(end=(0cm,0cm))
piece = line(end=(6.283185307179586cm,0cm))
piece = line(end=spline_start)

```

 w/ a sinusoidal cubic spline and three linear segments:
image

We get the following linear mesh (sampled using 100 segments per contour):
image

When querying at the vertices of a `256^2` grid, using the following command line args:
```
>./examples/quest_containment_driver_ex -i test_spline.contour -v        \
                                        --min -.2 -.2  --max 6.5 2.2     \
                                        -l9 --batched
```

we construct the following In/Out field:
image

### TODO

- [x] Add MPI-based C2C loader (similar to STL / PSTL reader)
- [x] Incorporate into quest::inout C-API and Fortran API
- [x] Add some examples to axom_data -- https://github.com/LLNL/axom_data/pull/5
- [x] Update data submodule
- [x] Update RELEASE-NOTES

### DIFF
--- a/src/axom/quest/CMakeLists.txt
+++ b/src/axom/quest/CMakeLists.txt
@@ -40,7 +40,6 @@ set( quest_headers
     detail/PointInCellMeshWrapper_mfem.hpp
 
     ## File readers
-    readers/C2CReader.hpp
     readers/STLReader.hpp
 
     ## quest interface
@@ -60,7 +59,6 @@ set( quest_sources
     MeshTester.cpp
 
     ## File readers
-    readers/C2CReader.cpp
     readers/STLReader.cpp
 
     ## quest interface
@@ -69,6 +67,8 @@ set( quest_sources
     interface/signed_distance.cpp
 
    )
+
+
 
 set( quest_depends_on
      core
@@ -80,17 +80,22 @@ set( quest_depends_on
      fmt
      )
 
-blt_list_append(TO quest_depends_on IF C2C_FOUND ELEMENTS c2c)
 blt_list_append(TO quest_depends_on IF ENABLE_CUDA ELEMENTS cuda)
 blt_list_append(TO quest_depends_on IF ENABLE_OPENMP ELEMENTS openmp)
 blt_list_append(TO quest_depends_on IF SPARSEHASH_FOUND ELEMENTS sparsehash)
+     
+if(C2C_FOUND)
+     list(APPEND quest_headers readers/C2CReader.hpp)
+     list(APPEND quest_sources readers/C2CReader.cpp)
+     list(APPEND quest_depends_on c2c)
+endif()
 
 if (ENABLE_MPI)
     list(APPEND quest_headers readers/PSTLReader.hpp)
-    list(APPEND quest_sources readers/PSTLReader.cpp)
+    blt_list_append(TO quest_headers IF C2C_FOUND ELEMENTS readers/PC2CReader.hpp)
 
-    list(APPEND quest_headers readers/PC2CReader.hpp)
-    list(APPEND quest_sources readers/PC2CReader.cpp)
+    list(APPEND quest_sources readers/PSTLReader.cpp)
+    blt_list_append(TO quest_sources IF C2C_FOUND ELEMENTS readers/PC2CReader.cpp)
     
     list(APPEND quest_depends_on mpi)
     blt_list_append(TO quest_depends_on IF AXOM_ENABLE_LUMBERJACK ELEMENTS lumberjack )
--- a/src/axom/quest/CMakeLists.txt
+++ b/src/axom/quest/CMakeLists.txt
@@ -71,23 +71,23 @@ set( quest_sources
 
 
 set( quest_depends_on
-     core
-     slic
-     mint
-     slam
-     spin
-     primal
-     fmt
-     )
+    core
+    slic
+    mint
+    slam
+    spin
+    primal
+    fmt
+    )
 
 blt_list_append(TO quest_depends_on IF ENABLE_CUDA ELEMENTS cuda)
 blt_list_append(TO quest_depends_on IF ENABLE_OPENMP ELEMENTS openmp)
 blt_list_append(TO quest_depends_on IF SPARSEHASH_FOUND ELEMENTS sparsehash)
      
 if(C2C_FOUND)
-     list(APPEND quest_headers readers/C2CReader.hpp)
-     list(APPEND quest_sources readers/C2CReader.cpp)
-     list(APPEND quest_depends_on c2c)
+    list(APPEND quest_headers readers/C2CReader.hpp)
+    list(APPEND quest_sources readers/C2CReader.cpp)
+    list(APPEND quest_depends_on c2c)
 endif()
 
 if (ENABLE_MPI)
@@ -98,7 +98,6 @@ if (ENABLE_MPI)
     blt_list_append(TO quest_sources IF C2C_FOUND ELEMENTS readers/PC2CReader.cpp)
     
     list(APPEND quest_depends_on mpi)
-    blt_list_append(TO quest_depends_on IF AXOM_ENABLE_LUMBERJACK ELEMENTS lumberjack )
 endif()
 
 if (SHROUD_FOUND)
@@ -145,11 +144,11 @@ axom_install_component(NAME    quest
 # Add tests and examples
 #------------------------------------------------------------------------------
 if (AXOM_ENABLE_EXAMPLES)
-  add_subdirectory(examples)
+    add_subdirectory(examples)
 endif()
 
 if (AXOM_ENABLE_TESTS)
-  add_subdirectory(tests)
+    add_subdirectory(tests)
 endif()
 
 axom_add_code_checks(PREFIX quest)
--- a/src/axom/quest/CMakeLists.txt
+++ b/src/axom/quest/CMakeLists.txt
@@ -31,21 +31,21 @@ set( quest_headers
     detail/inout/InOutOctreeStats.hpp
     detail/inout/InOutOctreeValidator.hpp
 
-    # Mesh tester headers
+    # Mesh tester
     MeshTester.hpp
 
-    # PointInCell headers
+    # PointInCell
     PointInCell.hpp
     detail/PointFinder.hpp
     detail/PointInCellMeshWrapper_mfem.hpp
 
-    ## STL library headers
+    ## File readers
+    readers/C2CReader.hpp
     readers/STLReader.hpp
 
-    ## quest interface headers
+    ## quest interface
     interface/internal/mpicomm_wrapper.hpp
     interface/internal/QuestHelpers.hpp
-
     interface/inout.hpp
     interface/signed_distance.hpp
 
@@ -56,18 +56,15 @@ set( quest_sources
     ## All-nearest-neighbors query
     AllNearestNeighbors.cpp
 
-    ## Geometry Sources
-    # <nothing yet>
-
-    ## Mesh tester sources
+    ## Mesh tester
     MeshTester.cpp
 
-    ## STL library sources
+    ## File readers
+    readers/C2CReader.cpp
     readers/STLReader.cpp
 
-    ## quest interface sources
+    ## quest interface
     interface/internal/QuestHelpers.cpp
-
     interface/inout.cpp
     interface/signed_distance.cpp
 
@@ -83,9 +80,10 @@ set( quest_depends_on
      fmt
      )
 
-blt_list_append(TO quest_depends_on IF SPARSEHASH_FOUND ELEMENTS sparsehash)
-blt_list_append(TO quest_depends_on IF ENABLE_OPENMP ELEMENTS openmp)
+blt_list_append(TO quest_depends_on IF C2C_FOUND ELEMENTS c2c)
 blt_list_append(TO quest_depends_on IF ENABLE_CUDA ELEMENTS cuda)
+blt_list_append(TO quest_depends_on IF ENABLE_OPENMP ELEMENTS openmp)
+blt_list_append(TO quest_depends_on IF SPARSEHASH_FOUND ELEMENTS sparsehash)
 
 if (ENABLE_MPI)
     list(APPEND quest_headers readers/PSTLReader.hpp)
--- a/src/axom/quest/CMakeLists.txt
+++ b/src/axom/quest/CMakeLists.txt
@@ -89,11 +89,11 @@ if (ENABLE_MPI)
     list(APPEND quest_headers readers/PSTLReader.hpp)
     list(APPEND quest_sources readers/PSTLReader.cpp)
 
-    blt_list_append(TO quest_depends_on
-                    IF AXOM_ENABLE_LUMBERJACK
-                    ELEMENTS lumberjack )
-
+    list(APPEND quest_headers readers/PC2CReader.hpp)
+    list(APPEND quest_sources readers/PC2CReader.cpp)
+    
     list(APPEND quest_depends_on mpi)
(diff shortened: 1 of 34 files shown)


src/axom/quest/examples/containment_driver.cpp
```diff
@@ -35,88 +36,230 @@ namespace slic = axom::slic;
...
-    meshBB.addPoint(pt);
+    delete m_surfaceMesh;
+    delete m_octree;
   }
 
-  SLIC_ASSERT(meshBB.isValid());
+#ifdef AXOM_USE_C2C
+  void loadContourMesh(const std::string& inputFile, int segmentsPerKnotSpan)
+  {
+    quest::C2CReader* reader = new quest::C2CReader();
```
estebanpauli, line 75: This could just be `quest::C2CReader reader;`. Then you wouldn't have to worry about deleting it later.

src/axom/quest/examples/containment_driver.cpp
```diff
@@ -35,88 +36,230 @@ namespace slic = axom::slic;
...
+    AXOM_UNUSED_VAR(segmentsPerKnotSpan);
+    SLIC_ERROR(
+      "Configuration error: Loading contour files is only supported when Axom "
+      "is configured with C2C support.");
+  }
+#endif  // AXOM_USE_C2C
+
+  void loadSTLMesh(const std::string& inputFile)
+  {
+    quest::STLReader* reader = new quest::STLReader();
```
estebanpauli, line 99: Should be `quest::STLReader reader;`.

src/axom/quest/examples/containment_driver.cpp
```diff
@@ -386,15 +508,25 @@ struct Input
 #endif
   }
 
+  bool isInput2D() const
+  {
+    using axom::utilities::string::endsWith;
+    if(endsWith(inputFile, ".contour"))
```
estebanpauli, line 514: This is just a small style thing, but I prefer

```
return predicate();
```

to

```
if (predicate()) {
    return true;
}
return false;
```

src/axom/quest/interface/inout.cpp
```diff
@@ -19,6 +19,23 @@ namespace quest
...
+struct InOutParameters
+{
+  bool m_verbose {false};
+  int m_dimension {3};
+  int m_segmentsPerKnotSpan {25};  /// Used when linearizing curves
+  double m_vertexWeldThreshold {1E-9};
+
+  void setDefault()
+  {
+    m_verbose = false;
```
estebanpauli, line 32: To avoid repeating values, you could do `*this = InOutParameters{};`

src/axom/quest/interface/inout.cpp
```diff
@@ -428,10 +531,38 @@ int inout_mesh_center_of_mass(double* coords)
...
+  return rc;
+}
+
+//------------------------------------------------------------------------------
+
+bool inout_evaluate(double x, double y)
+{
+  if(!inout_initialized())
+  {
+    SLIC_WARNING("quest inout query must be initialized "
```
estebanpauli, line 558: Should this be an error instead?

src/axom/quest/interface/inout.cpp
```diff
@@ -444,7 +575,30 @@ bool inout_evaluate(double x, double y, double z)
...
+  const int dim = inout_get_dimension();
+  return (dim == 2) ? s_inoutHelper2D.within(x, y, z)
+                    : s_inoutHelper3D.within(x, y, z);
+}
+
+int inout_evaluate(const double* x, const double* y, int npoints, int* res)
+{
+  if(!inout_initialized())
+  {
+    SLIC_WARNING("quest inout query must be initialized "
```
estebanpauli, line 587: Error instead of warning?

estebanpauli: There are multiple more calls like this. If you change one, the rest should be made consistent.

src/axom/quest/interface/internal/QuestHelpers.cpp
```diff
@@ -356,6 +351,62 @@ int read_mesh(const std::string& file, mint::Mesh*& m, MPI_Comm comm)
...
+    return READ_FAILED;
+  }
+
+  // STEP 1: allocate output mesh object
+  m = new SegmentMesh(DIMENSION, mint::SEGMENT);
+
+  // STEP 2: allocate reader
+  quest::C2CReader* reader = nullptr;
+  #ifdef AXOM_USE_MPI
+  reader = new quest::PC2CReader(comm);
```
estebanpauli, line 381: The readers can be statically allocated, avoiding the need to delete it. If you do want to allocate dynamically, I recommend a `std::unique_ptr<>`.

src/axom/quest/interface/internal/QuestHelpers.hpp
```diff
@@ -229,9 +224,44 @@ int read_mesh_shared(const std::string& file,
...
+/*!
+ * \brief Reads in the contour mesh from the specified file
+ *
+ * \param [in] file the file consisting of the surface
+ * \param [in] segmentsPerPiece number of segments to sample per contour Piece
+ * \param [in] vertexWeldThreshold threshold for welding vertices of adjacent curves
+ * \param [out] m user-supplied pointer to point to the mesh object
+ * \param [in] comm the MPI communicator, only applicable when MPI is available
+ *
+ * \note This method currently expects the contours to be given in the C2C format
```
estebanpauli, line 241: Since the function is called `read_c2c_mesh()`, I don't think this will ever change. It can probably be left out. Unless you were trying to specify that it should be a contour file rather than an assembly file.

src/axom/quest/readers/PC2CReader.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+  virtual ~PC2CReader() = default;
+
+  /*!
+   * \brief Reads in a C2C file to all ranks in the associated communicator
+   *
+   * \note Rank 0 reads in the C2C mesh file and broadcasts to the other ranks
+   * \return status set to zero on success; non-zero otherwise
+   */
+  virtual int read() final override;
```
estebanpauli, line 38: The `virtual` is redundant here, and `override` already tells the reader (and compiler) you're overriding a function. Some tools (e.g. clion) complain about this redundancy.

estebanpauli (commented):
I left a few small suggestions, but looks good overall.

src/axom/quest/examples/containment_driver.cpp
kennyweiss: Thanks -- me too.

I wasn't sure if I'd need more development there, so I left it more general at firstl.
Since I didn't actually need the flexibility, I've adopted your suggestion

src/axom/quest/interface/inout.cpp
kennyweiss: For the inout "C" interface, we opted to have everything use return codes so the end user can decide how to do error handling.

src/axom/quest/interface/inout.cpp
kennyweiss: Same as above:
> For the inout "C" interface, we opted to have everything use return codes so the end user can decide how to do error handling.

src/axom/quest/interface/internal/QuestHelpers.cpp
kennyweiss: Good call!

src/axom/quest/readers/PC2CReader.hpp
kennyweiss: Good suggestion.
I updated `PSTLReader` as well

src/axom/quest/interface/inout.cpp
kennyweiss: Great suggestion!

kennyweiss:
Thanks for the great suggestions @estebanpauli!
I've incorporated all of them other than the ones about warnings vs. errors in the `inout` C API, as those were based on an explicit (earlier) decision to allow users to handle errors vis return codes in the C API.

estebanpauli (approved):

src/axom/quest/interface/internal/QuestHelpers.cpp
```diff
@@ -349,12 +343,59 @@ int read_mesh(const std::string& file, mint::Mesh*& m, MPI_Comm comm)
...
+
+  // STEP 2: construct C2C reader
+  #ifdef AXOM_USE_MPI
+  quest::PC2CReader reader(comm);
+  #else
+  AXOM_UNUSED_VAR(comm);
+  quest::C2CReader reader;
+  #endif
+
+  // STEP 3: read the mesh from the STL file
```
kennyweiss, line 381:
```suggestion
  // STEP 3: read the mesh from the input file
```

src/axom/quest/readers/C2CReader.cpp
```diff
@@ -0,0 +1,372 @@
...
+    : m_curve(curve)
+  {
+    int p = m_curve.order - 1;
+    int knotSize = m_curve.knots.size();
+    SLIC_ASSERT(p >= 1);
+    SLIC_ASSERT(knotSize >= 2 * p);
+    computeSpanIntervals(EPS);
+  }
+
+  // Helper function to compute the start and end index of each knot span
```
kennyweiss, line 46:
```suggestion
  // Helper function to compute the start and end parametric coordinates of each knot span
```

src/axom/quest/CMakeLists.txt
```diff
@@ -83,19 +80,25 @@ set( quest_depends_on
...
      )
 
-blt_list_append(TO quest_depends_on IF SPARSEHASH_FOUND ELEMENTS sparsehash)
-blt_list_append(TO quest_depends_on IF ENABLE_OPENMP ELEMENTS openmp)
 blt_list_append(TO quest_depends_on IF ENABLE_CUDA ELEMENTS cuda)
+blt_list_append(TO quest_depends_on IF ENABLE_OPENMP ELEMENTS openmp)
+blt_list_append(TO quest_depends_on IF SPARSEHASH_FOUND ELEMENTS sparsehash)
+     
+if(C2C_FOUND)
+     list(APPEND quest_headers readers/C2CReader.hpp)
```
white238, line 88: This is critical... there are 5 spaces instead of 4!!! THE HORROR!

src/axom/quest/CMakeLists.txt
```diff
@@ -83,19 +80,25 @@ set( quest_depends_on
...
-                    IF AXOM_ENABLE_LUMBERJACK
-                    ELEMENTS lumberjack )
+    list(APPEND quest_headers readers/PSTLReader.hpp)
+    blt_list_append(TO quest_headers IF C2C_FOUND ELEMENTS readers/PC2CReader.hpp)
 
+    list(APPEND quest_sources readers/PSTLReader.cpp)
+    blt_list_append(TO quest_sources IF C2C_FOUND ELEMENTS readers/PC2CReader.cpp)
+    
     list(APPEND quest_depends_on mpi)
+    blt_list_append(TO quest_depends_on IF AXOM_ENABLE_LUMBERJACK ELEMENTS lumberjack )
```
white238, line 101: This doesn't technically need to be here since it will be inherited by SLIC. (minus something I cant think of)

src/axom/quest/interface/inout.cpp
```diff
@@ -283,103 +288,164 @@ struct InOutHelper
...
+    s_inoutHelper2D.setVerbose(s_inoutParams.m_verbose);
+    s_inoutHelper2D.setSegmentsPerKnotSpan(s_inoutParams.m_segmentsPerKnotSpan);
+    s_inoutHelper2D.setVertexWeldThreshold(s_inoutParams.m_vertexWeldThreshold);
 
-int inout_finalize() { return s_inoutHelper.finalize(); }
+    rc = s_inoutHelper2D.initialize(mesh, comm);
+    if(rc == QUEST_INOUT_FAILED)
+    {
+      s_inoutHelper2D.restoreLoggingLevel();
+    }
```
white238, line 413: Missing a break here?

agcapps (approved):
Thanks, @kennyweiss .  Looks very serviceable: useful to codes *and* eminently maintainable.

src/axom/quest/readers/C2CReader.cpp
```diff
@@ -0,0 +1,372 @@
...
+  }
+
+  int numSpans() const { return m_spanIntervals.size(); }
+
+  double startParameter(int span = -1) const
+  {
+    const bool inRange = span >= 0 && span < numSpans();
+    return inRange ? m_spanIntervals[span].first : m_curve.knots[0];
+  }
+  double endParameter(int span = -1) const
```
white238, line 109: this is missing a line, did you run this through clangformat?

src/axom/quest/readers/C2CReader.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+          // Fix end point if necessary; check against 0th vertex in mesh
+          const int endIdx = pts.size() - 1;
+          mesh->getNode(0, meshPt.data());
+          if(primal::squared_distance(pts[endIdx], meshPt) < EPS_SQ)
+          {
+            pts[endIdx] = meshPt;
+          }
+        }
+        else  // This is the first, and possibly only span, check its endpoint, fix if necessary
```
white238, line 336: 👍  to your comments!

white238 (approved):
approved pending fixing the missing break

src/axom/quest/CMakeLists.txt
kennyweiss: Thanks -- removed.

src/axom/quest/interface/inout.cpp
kennyweiss: Indeed!

Double whammy -- I had tests for this function in 3D in the `quest_inout_interface` file, but had not extended them to 2D in this PR. 

I updated the tests to 2D, verified that the missing `break` was caught by the test and then fixed this.

Thanks!

src/axom/quest/readers/C2CReader.cpp
kennyweiss: The code was run through clang-format. 
It seems that our rules don't enforce a newline between functions.

I manually added the line.

kennyweiss:
Thanks for the thorough review! 
This PR should be ready to merge once the following axom_data PR is reviewed and merged: 
https://github.com/LLNL/axom_data/pull/6

In testing the 2D inout_interface, I discovered that the unit_circle contour file had the wrong orientation 
(It was oriented clockwise, but should have been counterclockwise).

PR #612, opened 2021-07-17
Feature/kweiss/multimat_revamp

raineyeh:
This PR is to fix the multimat unit tests

It does the following:

- Make unit test compatible with the new layout-per-field feature
- Fix traversal to run with just 1 component (until field2D with higher number of components are added)
- Reduce unit test to use 1 component as that's temporarily hard-coded
- Fix compilation errors with SubMap index type clash
- Fix tinyHydro compilation error with unmatched Map constructor


kennyweiss (approved):
Thanks @raineyeh -- I'll resolve the remaining compiler error in the `multimat-revamp` branch.

PR #614, opened 2021-07-20
fix exported conduit dependency

white238:
I missed one of the two directories that need to be moved in parallel.  This affects downstream users.

This should probably be fixed in conduit ultimately because this fails in a very subtle way (`conduit` is created instead of all the `conduit::conduit` targets)

Conduit outputs two sets of cmake config files, limited set in `${CONDUIT_DIR}\lib\cmake` and the full set in `${CONDUIT_DIR}\lib\cmake\conduit`

Limited set:

```
[white238@rzgenie5:repo]$ ls -al /usr/WS1/axom/libs/toss_3_x86_64_ib/2021_06_08_23_40_14/clang-10.0.0/conduit-0.7.2/lib/cmake/
total 24
drwxrwsr-x 3 weiss27 axom 4096 Jun  9 00:12 .
drwxrwsr-x 3 weiss27 axom 4096 Jun  9 00:12 ..
drwxrwsr-x 2 weiss27 axom 4096 Jun  9 00:12 conduit
-rw-rw-r-- 1 weiss27 axom 3472 Jun  9 00:11 conduit-relwithdebinfo.cmake
-rw-rw-r-- 1 weiss27 axom 7276 Jun  9 00:11 conduit.cmake
```

Full set:

```
[white238@rzgenie5:repo]$ ls -al /usr/WS1/axom/libs/toss_3_x86_64_ib/2021_06_08_23_40_14/clang-10.0.0/conduit-0.7.2/lib/cmake/conduit
total 44
drwxrwsr-x 2 weiss27 axom 4096 Jun  9 00:12 .
drwxrwsr-x 3 weiss27 axom 4096 Jun  9 00:12 ..
-rw-rw-r-- 1 weiss27 axom 3210 Jun  9 00:11 ConduitConfig.cmake
-rw-rw-r-- 1 weiss27 axom 1386 Jun  9 00:11 ConduitConfigVersion.cmake
-rw-rw-r-- 1 weiss27 axom 3472 Jun  9 00:11 conduit-relwithdebinfo.cmake
-rw-rw-r-- 1 weiss27 axom 7340 Jun  9 00:11 conduit.cmake
-rw-rw-r-- 1 weiss27 axom 6453 May  5 11:02 conduit_setup_deps.cmake
-rw-rw-r-- 1 weiss27 axom 4876 May  5 11:02 conduit_setup_targets.cmake
```

### DIFF
--- a/src/cmake/axom-config.cmake.in
+++ b/src/cmake/axom-config.cmake.in
@@ -76,7 +76,8 @@ if(NOT AXOM_FOUND)
     if(NOT CONDUIT_DIR) 
       set(CONDUIT_DIR ${AXOM_CONDUIT_DIR}) 
     endif()
-    find_dependency(Conduit REQUIRED NO_DEFAULT_PATH PATHS "@AXOM_CONDUIT_CONFIG_DIR@")
+    find_dependency(Conduit REQUIRED NO_DEFAULT_PATH
+                    PATHS "${CONDUIT_DIR}/@AXOM_CONDUIT_CONFIG_DIR@")
   endif()
 
   # hdf5
--- a/src/cmake/axom-config.cmake.in
+++ b/src/cmake/axom-config.cmake.in
@@ -76,7 +76,7 @@ if(NOT AXOM_FOUND)
     if(NOT CONDUIT_DIR) 
       set(CONDUIT_DIR ${AXOM_CONDUIT_DIR}) 
     endif()
-    find_dependency(Conduit REQUIRED NO_DEFAULT_PATH PATHS "${CONDUIT_DIR}/lib/cmake/conduit")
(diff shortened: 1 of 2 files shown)


white238:
@cyrush I think your previous fix of only installing a conduit config in one directory might have been half correct.  There are some files left over that can leave you in a bad state.

kennyweiss (approved):
Thanks @white238 !

agcapps (approved):
Thanks @white238

cyrush:
@white238 thanks for hunting this down!

This solves the mystery that Kenny was having. 

To summarize: 

There are few files hanging around in the old place.

Axom, when building was looking in the right place, but  the imported targets was still looking in the old place.

Since the old files where hanging around, it found a partial setup. 

@kennyweiss  this fix may also solve the Threads::Threads issue. 

I'll make the fix in conduit to remove the old files.

PR #615, opened 2021-07-21
Add missing host/device annotations to Primal/Spin functions

kanye-quest:
# Summary

- Fixes an issue identified by @steveg21, where initializing `spin::BVH` fails with CUDA.

  `nvcc` somehow doesn't warn us about certain functions with missing `AXOM_HOST_DEVICE` annotations, and instead will generate a call instruction to address `0x0`. The existing tests for the BVH code weren't able to catch the issue, since the functions in question are somehow getting inlined when our tests are being built, which avoids the problematic call instruction.


kennyweiss (approved):
Thanks @kanye-quest !

white238 (approved):

kennyweiss:
One of the LC jobs got stuck waiting for an allocation. 
I reran the pipeline and it passed the second time around.

rhornung67 (approved):
Good detective work @kanye-quest !

PR #622, opened 2021-08-02
Adds a data collection utility and improves support for loading in MFEMSidreDataCollection

kennyweiss:
# Summary

- This PR is a feature, but also includes some bugfixes
- It adds a new tool for generating distributed blueprint-compliant (high order) meshes over a Cartesian domain or from an mfem mesh: `data_collection_util`. The latter is active when Axom is configured with `mfem` and `AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION`
 
- This PR also improves support for loading data collections in `MFEMSidreDataCollection`. 
- Specifically:
    - It avoids appending the `.root` suffix to an input root file when that suffix is already present
    - It wires up the high order nodal grid function during load (when available)
    - It improves error messages when attempting to load using a different number of ranks than were used to save the file

- This PR also add a new string utility function: `removeSuffix()`
- I noticed that the string utilities did not have unit tests, so I added some
- In doing so, there were a few undocumented edge-case behaviors. I'd like the reviewers to sign off on the changes I made (which I'll point out via comments)

#### TODO:

- [x] Update RELEASE-NOTES

### DIFF
--- a/src/axom/core/tests/utils_stringUtilities.hpp
+++ b/src/axom/core/tests/utils_stringUtilities.hpp
@@ -15,8 +15,6 @@
 //------------------------------------------------------------------------------
 TEST(utils_stringUtilities, startsWith)
 {
-  std::cout << "Testing startsWith() functions" << std::endl;
-
   {
     std::string testString = "foo.bar";
     std::string testPrefix = "boo";
@@ -42,8 +40,6 @@ TEST(utils_stringUtilities, startsWith)
 
 TEST(utils_stringUtilities, endsWith)
 {
-  std::cout << "Testing endsWith() functions" << std::endl;
-
   {
     std::string testString = "foo.bar";
     std::string testSuffix = ".baz";
@@ -69,8 +65,6 @@ TEST(utils_stringUtilities, endsWith)
 
 TEST(utils_stringUtilities, removeSuffix)
 {
-  std::cout << "Testing removeSuffix() function" << std::endl;
-
   // different suffix
   {
     std::string testString = "foo.bar";
@@ -106,8 +100,6 @@ TEST(utils_stringUtilities, removeSuffix)
 
 TEST(utils_stringUtilities, toLower)
 {
-  std::cout << "Testing toLower() function" << std::endl;
-
   // already lower
   {
     std::string testString = "foo.bar";
@@ -135,8 +127,6 @@ TEST(utils_stringUtilities, toLower)
 
 TEST(utils_stringUtilities, toUpper)
 {
-  std::cout << "Testing toUpper() function" << std::endl;
-
   // already upper
   {
     std::string testString = "foo.bar";
@@ -164,10 +154,9 @@ TEST(utils_stringUtilities, toUpper)
 
 TEST(utils_stringUtilities, split)
 {
-  std::cout << "Testing split() function" << std::endl;
   using StrVec = std::vector<std::string>;
 
-  // Test w/ proper delim
+  // Test w/ proper delimiter
   {
     std::string testString = "foo/bar/baz";
     StrVec exp {"foo", "bar", "baz"};
@@ -194,7 +183,7 @@ TEST(utils_stringUtilities, split)
     EXPECT_EQ(exp, results);
   }
 
-  // Test other delimeter
+  // Test other delimiter
   {
     std::string testString = "foo.bar.baz";
     StrVec exp {"foo", "bar", "baz"};
@@ -203,7 +192,7 @@ TEST(utils_stringUtilities, split)
     EXPECT_EQ(exp, results);
   }
 
-  // Test different delimeter
+  // Test different delimiter
   {
     std::string testString = "foo.bar.baz";
     StrVec exp {"foo.bar.baz"};
@@ -215,7 +204,6 @@ TEST(utils_stringUtilities, split)
 
 TEST(utils_stringUtilities, splitLastNTokens)
 {
-  std::cout << "Testing split() function" << std::endl;
   using StrVec = std::vector<std::string>;
 
   // Test w/ sufficient tokens
@@ -295,7 +283,7 @@ TEST(utils_stringUtilities, splitLastNTokens)
     EXPECT_EQ(exp, results);
   }
 
-  // Test mixed delim: /
+  // Test mixed delim: '/'
   {
     const std::size_t N = 3;
     std::string testString = "foo.bar/baz.qux";
@@ -306,7 +294,7 @@ TEST(utils_stringUtilities, splitLastNTokens)
     EXPECT_EQ(exp, results);
   }
 
-  // Test mixed delim: .
+  // Test mixed delim: '.'
   {
     const std::size_t N = 3;
     std::string testString = "foo.bar/baz.qux";
--- a/src/axom/core/tests/utils_stringUtilities.hpp
+++ b/src/axom/core/tests/utils_stringUtilities.hpp
@@ -1,0 +1,319 @@
+// Copyright (c) 2017-2021, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level LICENSE file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include "gtest/gtest.h"
+
+#include "axom/core/utilities/StringUtilities.hpp"
+
+#include <string>
+#include <vector>
+
+//------------------------------------------------------------------------------
+// UNIT TESTS
+//------------------------------------------------------------------------------
+TEST(utils_stringUtilities, startsWith)
+{
+  std::cout << "Testing startsWith() functions" << std::endl;
+
+  {
+    std::string testString = "foo.bar";
+    std::string testPrefix = "boo";
+    EXPECT_FALSE(axom::utilities::string::startsWith(testString, testPrefix));
+  }
+
+  {
+    std::string testString = "foo.bar";
+    EXPECT_FALSE(axom::utilities::string::startsWith(testString, 'b'));
+  }
+
+  {
+    std::string testString = "foo.bar";
+    std::string testPrefix = "foo";
+    EXPECT_TRUE(axom::utilities::string::startsWith(testString, testPrefix));
+  }
+
+  {
+    std::string testString = "foo.bar";
+    EXPECT_TRUE(axom::utilities::string::startsWith(testString, 'f'));
+  }
+}
+
+TEST(utils_stringUtilities, endsWith)
+{
+  std::cout << "Testing endsWith() functions" << std::endl;
+
+  {
+    std::string testString = "foo.bar";
+    std::string testSuffix = ".baz";
+    EXPECT_FALSE(axom::utilities::string::endsWith(testString, testSuffix));
+  }
+
+  {
+    std::string testString = "foo.bar";
+    EXPECT_FALSE(axom::utilities::string::endsWith(testString, 'z'));
+  }
+
+  {
+    std::string testString = "foo.bar";
+    std::string testSuffix = ".bar";
+    EXPECT_TRUE(axom::utilities::string::endsWith(testString, testSuffix));
+  }
+
+  {
+    std::string testString = "foo.bar";
+    EXPECT_TRUE(axom::utilities::string::endsWith(testString, 'r'));
+  }
+}
+
+TEST(utils_stringUtilities, removeSuffix)
+{
+  std::cout << "Testing removeSuffix() function" << std::endl;
+
+  // different suffix
+  {
+    std::string testString = "foo.bar";
+    std::string testSuffix = ".baz";
+    EXPECT_EQ("foo.bar",
+              axom::utilities::string::removeSuffix(testString, testSuffix));
+  }
+
+  // same suffix
+  {
+    std::string testString = "foo.bar";
+    std::string testSuffix = ".bar";
+    EXPECT_EQ("foo",
+              axom::utilities::string::removeSuffix(testString, testSuffix));
+  }
+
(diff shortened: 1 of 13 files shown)


src/axom/core/utilities/StringUtilities.cpp
```diff
@@ -48,15 +48,20 @@ std::vector<std::string> splitLastNTokens(const std::string& input,
   auto last_pos = std::string::npos;
   auto pos = input.find_last_of(delim, last_pos - 1);
 
-  while((pos != std::string::npos) && (result.size() < n - 1))
+  if(n > 0 && !input.empty())
```
kennyweiss, line 51: @joshessman-llnl -- I think you added this function. Can you please confirm the new edge-case conditions.

Essentially, if the input is empty or the user asks for 0 tokens, the function now returns an empty vector (instead of a vector with an empty string, and a vector with all tokens, respectively)

src/axom/core/tests/utils_utilities.hpp
```diff
@@ -91,16 +91,9 @@ TEST(utils_Utilities, random_real_with_seed)
   constexpr double a = -5.0;
   constexpr double b = 5.0;
 
-  const double expected_reals[5] = {-1.5112829544380526,
-                                    -2.3311429024686219,
-                                    -3.6335370551231403,
-                                    -4.714431326610093,
-                                    3.6893326916732878};
```
kennyweiss, line 98: Two comments here:
1. As far as I can tell, we cannot assume that different compilers will provide the same set of values for a uniform distribution when given the same seed.  See, e.g. https://stackoverflow.com/questions/28607889/how-to-generate-the-same-random-number-sequence-over-multiple-types-of-compilers
2. For some reason, the two tests in this filer were failing when I ran the `./core_serial_tests`, but not when I ran `ctest`?

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -855,13 +855,23 @@ void MFEMSidreDataCollection::Load(const std::string& path,
...
+
   #if defined(AXOM_USE_MPI) && defined(MFEM_USE_MPI)
   if(m_comm != MPI_COMM_NULL)
   {
-    IOManager reader(m_comm);
     // The conduit abstraction appears to automatically handle the ".root"
     // suffix, but the IOManager does not, so it gets added here
-    reader.read(m_bp_grp->getDataStore()->getRoot(), path + ".root");
+    using axom::utilities::string::endsWith;
+    std::string suffixedPath = endsWith(path, ".root") ? path : path + ".root";
```
kennyweiss, line 867: DC bugfix 1: Don't append `.root` if it's already there.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -855,13 +855,23 @@ void MFEMSidreDataCollection::Load(const std::string& path,
...
+
+    IOManager reader(m_comm);
+    reader.read(m_bp_grp->getDataStore()->getRoot(), suffixedPath);
+
+    // Get some data in support of error checks below
+    numInputRanks = reader.getNumGroupsFromRoot(suffixedPath);
+    MPI_Comm_size(m_comm, &numCommRanks);
```
kennyweiss, line 874: DC bugfix/enhancement 2: Improved logging (a few lines down), if loading on a different number of ranks than were used for save.
Note @nselliott: this required making a private function from `IOManager` public (`getNumGroupsFromRoot()`)

src/axom/core/utilities/StringUtilities.cpp
joshessman-llnl: I think this is reasonable - it looks like the `result.size() < n - 1` was bogus anyways if `n == 0` since `n` is of unsigned type.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -873,21 +883,36 @@ void MFEMSidreDataCollection::Load(const std::string& path,
...
+  const std::string gfPath = "topologies/mesh/grid_function";
+  if(GetBPGroup()->hasView(gfPath))
+  {
+    const std::string nodalGFName = GetBPGroup()->getView(gfPath)->getString();
+    if(this->HasField(nodalGFName))
+    {
+      this->SetMeshNodesName(nodalGFName);
+      mesh->NewNodes(*this->GetField(nodalGFName), false);
+    }
+  }
```
kennyweiss, line 915: DC bugfix 3:  We need to account for the nodal grid function when loading an `MFEMSidreDataCollection`.

src/axom/core/utilities/StringUtilities.cpp
kennyweiss: Thanks -- it was, but had the effect of making `N` irrelevant, and returned all tokens

kennyweiss: As an aside, the API/signature for this function is different from the one for `split()` above.
This one returns a vector, the other has a vector as an `out` parameter.

In the future, we might want to make these consistent.

src/axom/core/tests/utils_stringUtilities.hpp
```diff
@@ -0,0 +1,307 @@
...
+    std::string testString = "FoO.bAr";
+    std::string expString = "foo.bar";
+    axom::utilities::string::toLower(testString);
+    EXPECT_EQ(expString, testString);
+  }
+}
+
+TEST(utils_stringUtilities, toUpper)
+{
+  // already upper
```
white238, line 130: I think these comments are wrong

src/tools/CMakeLists.txt
```diff
@@ -68,6 +68,34 @@ if(AXOM_ENABLE_QUEST)
             )
 endif()
 
+#------------------------------------------------------------------------------
+# data_collection_util is a utility that generates blueprint-conforming sidre
+# output files from an input mfem mesh
+#------------------------------------------------------------------------------
+if( AXOM_ENABLE_SIDRE AND HDF5_FOUND AND MFEM_FOUND AND AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION )
```
white238, line 75: Does HDF5_FOUND need to be checked here?

src/tools/data_collection_util.cpp
```diff
@@ -0,0 +1,620 @@
...
+  void fixBoxParams()
+  {
+    const bool dimProvided = boxDim > 0;
+    const bool rangeProvided = !(boxMins.empty() && boxMaxs.empty());
+    const bool resProvided = !boxResolution.empty();
+
+    // First check: we need at least one of: dimension, mins/maxs, resolution
+    if(!(dimProvided || rangeProvided || resProvided))
+    {
+      throw "Box mesh must have at least one of: dimension, mins and maxs or "
```
white238, line 77: Is there a reason you don't use SLIC here?  If so can you add a comment?

src/tools/data_collection_util.cpp
```diff
@@ -0,0 +1,620 @@
...
+    int szMins = boxMins.size();
+    int szMaxs = boxMaxs.size();
+    int szRes = boxResolution.size();
+
+    // Error checking on provided dimension
+    if(dimProvided)
+    {
+      if(boxDim < 1 || boxDim > 3)
+      {
+        throw "Dimension cannot exceed 3";
```
white238, line 90: Error is confusing if you end up with boxDim == 0

white238:
Is there a possible test you could add for the new tool?  If not could you at least add something to make sure that it doesn't crash on a basic input?

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -873,21 +883,36 @@ void MFEMSidreDataCollection::Load(const std::string& path,
...
+                  "dataset was created using {} ranks",
+                  numCommRanks,
+                  numInputRanks));
 
     UpdateStateFromDS();
     UpdateMeshAndFieldsFromDS();
   }
+
+  // Set the mesh nodal grid function when present
+  const std::string gfPath = "topologies/mesh/grid_function";
```
joshessman-llnl, line 906: Suggestion: I haven't gone through and replaced it everywhere, but it might be good to switch out the hardcoded "mesh" here for `MFEMSidreDataCollection::s_mesh_topology_name`.

src/tools/data_collection_util.cpp
```diff
@@ -0,0 +1,620 @@
...
+      }
+    }
+  }
+
+public:
+  Input() = default;
+
+  bool isVerbose() const { return m_verboseOutput; }
+
+  void parse(int argc, char** argv, CLI::App& app)
```
joshessman-llnl, line 149: This might be a Maslow's hammer-type situation but to me this looks like a lot of command-line options; had you given any thought to using Inlet here?

joshessman-llnl (approved):
Thanks for catching/fixing these SidreDC bugs!

src/axom/core/tests/utils_stringUtilities.hpp
kennyweiss: Thanks -- copy-paste error

src/tools/CMakeLists.txt
kennyweiss: It's necessary for writing out the files. 

- [ ] Make HDF5 optional and change the protocol when HDF5's not available

src/tools/data_collection_util.cpp
kennyweiss: I'm using exceptions since the CLI parsing is in a try/catch loop

TODO:
- [ ] Add a comment about exceptions
- [ ] Confirm that this is caught properly

src/tools/data_collection_util.cpp
kennyweiss: Agreed.

TODO:
- [ ] Fix this

agcapps (approved):
Thanks @kennyweiss

src/tools/CMakeLists.txt
kennyweiss: I wasn't actually setting the protocol -- just using the default.

It looks like the MFEMSIdreDataCollection should be updated to deal with cases where HDF5s' not enabled. I'll create a ticket for that so we can handle it later.

kennyweiss: Update: https://github.com/LLNL/axom/issues/623

src/tools/data_collection_util.cpp
kennyweiss: Thanks for the comment.

Turns out I needed to throw `CLI::ValidationError` instead of just a string.
I also updated the doxy comment to indicate that the function could throw that exception.

src/axom/core/utilities/StringUtilities.hpp
```diff
@@ -63,6 +63,21 @@ inline bool startsWith(const std::string& str, const char prefix)
   return str.size() >= 1 && str[0] == prefix;
 }
 
+/*!
+ * \brief Removes \a suffix from the end of \a str, if present
+ * \param [in] str string to be searched
+ * \param [in] suffix string to check for
+ * \return     String with (one copy of) \a suffix removed from end, if it was present  
+ */
+inline std::string removeSuffix(const std::string& str, const std::string& suffix)
```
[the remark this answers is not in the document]
kennyweiss: Thanks @joshessman-llnl 

I created https://github.com/LLNL/axom/issues/624 to track this

src/tools/data_collection_util.cpp
kennyweiss: Thanks for the suggestion.
I considered it, but I think it would be overkill in this case.
There are actually two different modes, and each one only has 3-ish main options.
e.g.
```bash
> srun -n8 ./bin/data_collection_util -f file -m star-q2.mesh --scale 5
> srun -n8 ./bin/data_collection_util --min 0 0 --max 1 1 --res 64 64 -p 2
```

I split these up into cli's `option_group`s, which provides a better help/usage message:
```bash
>./bin/data_collection_util -h
Utility tool to create a blueprint compliant data store
Usage: ./bin/data_collection_util [OPTIONS]

Options:
  -h,--help                                       Print this help message and exit
  -o,--output-name TEXT                           Name of the output mesh. Defaults to box{2,3}d for box meshes, and the mfem mesh name when loading from an mfem file
  -v,--verbose                                    Enable/disable verbose output
  -l,--ref-level INT:NONNEGATIVE=0                The number of uniform refinement levels to apply to the mesh
  -f,--mesh-form ENUM:value in {box->0,file->1} OR {0,1}=0
                                                  Determines the input type -- either box or file
[Option Group: box]
  Options for setting up a box mesh
  Options:
    --min FLOAT x 2 Needs: --max                    Min bounds for box mesh (x,y[,z])
    --max FLOAT x 2 Needs: --min                    Max bounds for box mesh (x,y[,z])
    --res INT x 2                                   Resolution of the box mesh (i,j[,k])
    -d,--dimension INT:POSITIVE                     Dimension of the box mesh
    -p,--polynomial-order INT:NONNEGATIVE=2         polynomial order of the generated mesh
[Option Group: file]
  Options for loading from an mfem mesh file
  Options:
    -m,--mfem-file TEXT:FILE                        Path to a mesh file in the mfem format
    --scale FLOAT                                   Scale factors for the file mesh. Can either be one value or mesh.Dimension() values
    --translate FLOAT x 2                           Translation for the file mesh (sx,sy[,sz])
```

src/axom/core/utilities/StringUtilities.cpp
kennyweiss: See #624

white238 (approved):

PR #626, opened 2021-08-04
Adds optional MPI dependency to data_collection_util app and test

kennyweiss:
# Summary

- This PR is a bugfix
- The `data_collection_util` test was missing an optional MPI dependency
- This PR also modifies the `data_collection_util` test to use more than one rank when MPI is available


agcapps (approved):
Thanks for the fix, @kennyweiss

white238 (approved):

PR #638, opened 2021-08-12
Handle SLURM update

white238:
Props to @adrienbernede

### DIFF
--- a/.gitlab/build_ruby.yml
+++ b/.gitlab/build_ruby.yml
@@ -44,14 +44,14 @@ ruby_release:
 .src_build_on_ruby:
   stage: build
   variables:
-    ALLOC_COMMAND: "srun -p pdebug -t 30 -N 1 ${ASSIGN_ID}"
+    ALLOC_COMMAND: "srun -p pdebug -t 30 -N 1  --interactive ${ASSIGN_ID}"
   extends: [.src_build_script, .on_ruby, .src_workflow]
   needs: [ruby_allocate]
 
 .full_build_on_ruby:
   stage: build
   variables:
-    ALLOC_COMMAND: "srun -p pdebug -t 60 -N 1"
+    ALLOC_COMMAND: "srun -p pdebug -t 60 -N 1 --interactive "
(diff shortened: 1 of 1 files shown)


kennyweiss (approved):
Enthusiastic 👍 

Thanks @white238  and @adrienbernede.
(Weird -- (I thought) that was the first thing I tried).

adrienbernede:
From the logs on ruby, it seems the jobs are not reusing the pre-allocated resource:
```
[32;1m$ ${ALLOC_COMMAND} python3 scripts/llnl_scripts/build_src.py -v --host-config ${HOST_CONFIG} --extra-cmake-options -DENABLE_DOCS=OFF[0;m
srun: job 2699928 queued and waiting for resources
srun: job 2699928 has been allocated resources
```

adrienbernede (approved):

adrienbernede:
@kennyweiss There should not be any srun message, those are here because there is an allocation actually being requested.

`ALLOC_COMMAND` is defined in `variables:` and _might_ be evaluate before the scripts starts, hence before the `ASSIGN_ID` variable is assigned a value.

adrienbernede:
Looking from `variables` in the table here (https://docs.gitlab.com/ee/ci/variables/where_variables_can_be_used.html#gitlab-runner-internal-variable-expansion-mechanism), you can see how expansion is operated.

If I understand correctly, variables expansion in the `variables:` section does not support variables from scripts per:
https://docs.gitlab.com/ee/ci/variables/where_variables_can_be_used.html#gitlab-runner-internal-variable-expansion-mechanism

adrienbernede:
Currently trying to find a solution working with your modular design.

rhornung67 (approved):
Thanks @adrienbernede I hope this fixes the issue.

adrienbernede:
Per the tests, this fixes the issue. We are discussing a pre-existing issue that is causing sub-optimal use of resource.

white238:
I am going to merge this PR because it unblocks us.  Then we can do a follow up PR to fix the alloc problem.

PR #639, opened 2021-08-12
Testing/bernede1/pre alloc fix

adrienbernede:
Following on #638. Added only the last commit.


kennyweiss (approved):
Thanks for digging into this @adrienbernede!

white238 (approved):

adrienbernede:
Ah ah! Looks like it's working great indeed, and now sharing allocation on ruby!

Issue #616, opened 2021-07-23
Build axom@0.5.0 with spack@develop and xl@16.1.1.7 on lassen


adrienbernede:
I am making some progress, but I am now facing this error:

```
[ 42%] Building CXX object axom/sidre/CMakeFiles/sidre.dir/spio/interface/c_fortran/wrapIOManager.cpp.o
[...]

PR #641, opened 2021-08-18
Workaround for C++14 feature missing from gcc@4.9.3

kennyweiss:
# Summary

- Some user configurations are built with C++14, but are pointing at a gcc@4.9.3 standard library
- The latter is not fully C++14 compliant, but is the default on some platforms, and it can be difficult to propagate the right flags in some build configuration.
- This change allows the code to build in the above configuration.

Possible resolution of #616

Note: This workaround doesn't fix the underlying problem of using `C++14` but pointing to a pre-`C++14` library. It only avoids using a feature from `C++14` that is missing from `gcc@4.9.3`'s standard library implementation, and thus not triggering the error.

### DIFF
--- a/src/axom/sidre/core/MFEMSidreDataCollection.cpp
+++ b/src/axom/sidre/core/MFEMSidreDataCollection.cpp
@@ -2419,15 +2419,15 @@ void MFEMSidreDataCollection::reconstructFields()
       {
         // The vdim is being overwritten here with the value in the basis string so we
         // can correctly construct the QuadratureFunction
-        m_quadspaces.emplace(basis_name,
-                             detail::NewQuadratureSpace(basis_name, mesh, vdim));
+        m_quadspaces[basis_name] = std::unique_ptr<mfem::QuadratureSpace>(
+          detail::NewQuadratureSpace(basis_name, mesh, vdim));
         is_gridfunc = false;
       }
       // Only need to create a new FEColl if one doesn't already exist
       else if(m_fecolls.count(basis_name) == 0)
       {
-        m_fecolls.emplace(basis_name,
-                          mfem::FiniteElementCollection::New(basis_name.c_str()));
+        m_fecolls[basis_name] = std::unique_ptr<mfem::FiniteElementCollection>(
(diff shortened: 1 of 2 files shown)


src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -2419,15 +2419,15 @@ void MFEMSidreDataCollection::reconstructFields()
       {
         // The vdim is being overwritten here with the value in the basis string so we
         // can correctly construct the QuadratureFunction
-        m_quadspaces.emplace(basis_name,
-                             detail::NewQuadratureSpace(basis_name, mesh, vdim));
+        m_quadspaces[basis_name] = std::unique_ptr<mfem::QuadratureSpace>(
```
kennyweiss, line 2422: @joshessman-llnl -- this is the issue that I was seeing that prevented me from upgrading to the `MFEMSidreDataCollection`.  Once this is in, I think I'll be able to upgrade.

white238 (approved):
Thanks @kennyweiss ! I look to us both being on the same class!

joshessman-llnl (approved):
Thanks!

src/axom/sidre/core/MFEMSidreDataCollection.cpp
joshessman-llnl: Can we flip the `AXOM_ENABLE_MFEM_SIDREDATACOLL` default then (or just remove the conditional entirely)? Or should that wait on a confirmed upgrade?

agcapps (approved):
Thanks @kennyweiss

rhornung67 (approved):
Thanks @kennyweiss I can't wait until gnu 4.9.3 is a distant memory.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
kennyweiss: I'm ok with flipping the default (in a separate PR). 

I don't think we should remove this option until we have a confirmed update.

PR #645, opened 2021-08-31
Defer element orientation checks until nodal GF is set in MFEMSidreDataColl

### LINKED ISSUE #643: Restarting with a periodic mesh using MFEMSidreDataCollection
@kennyweiss , @joshessman-llnl 

I'm getting some unexpected messages regarding the orientation of the mesh when I restart a simulation that uses a periodic mesh. The messages I get are something like this:

Elements with wrong orientation: 25 / 95 (NOT FIXED)
Boundary elements with wrong orientation: 3 / 38 (fixed)

I get those messages only when I compile in Debug mode. The restarted simulation then gives a segmentation fault, but I'm not sure if this segmentation fault is related to the mesh-orientation messages or not.

I've been able to reduce the code to the smallest set of lines that would still give those orientation messages. This code looks something like this:

```
Mesh *m_mesh;

if (cycle_to_load >= 0)
{
  m_datacoll.Load(cycle_to_load);
}
else
{
  Mesh base_mesh = Mesh::MakeCartesian2D(10, 10, Element::Type::QUADRILATERAL, false, 1.0, 1.0);
  Mesh periodic_mesh = Mesh::MakePeriodic(base_mesh, vertexMap);

  m_mesh = new ParMesh(MPI_COMM_WORLD, periodic_mesh);

  m_datacoll.SetMesh(m_mesh);
  m_datacoll.SetCycle(cycle_to_load);
  m_datacoll.SetTime(time);
  m_datacoll.Save("filename", "sidre_hdf5");
}
```

Thanks!

joshessman-llnl:
# Summary

- This PR is a bugfix
- It does the following:
  - Moves the call to `Mesh::Finalize` to after the nodal grid function is set
  - Fixes #643 @alecampos2009

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -77,6 +77,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Fixed bug in `Mint`'s VTK output for fields of type `int64` and `float`
 - Improved loading of data collections in `MFEMSidreDataCollection`
 - Added workaround to `MFEMSidreDataCollection` for `C++14` standard library feature that was not available in `gcc@4.9.3`
+- Delayed finalization of reloaded meshes in `MFEMSidreDataCollection` until after the setting
+  of the nodal `GridFunction` (when applicable)
 
 
 ## [Version 0.5.0] - Release date 2021-05-14
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -77,8 +77,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Fixed bug in `Mint`'s VTK output for fields of type `int64` and `float`
 - Improved loading of data collections in `MFEMSidreDataCollection`
 - Added workaround to `MFEMSidreDataCollection` for `C++14` standard library feature that was not available in `gcc@4.9.3`
-- Delayed finalization of reloaded meshes in `MFEMSidreDataCollection` until after the setting
-  of the nodal `GridFunction` (when applicable)
+- Delayed finalizing reloaded mesh in `MFEMSidreDataCollection` until after setting
+  the nodal `GridFunction` (when applicable)
 
 
 ## [Version 0.5.0] - Release date 2021-05-14
--- a/src/axom/sidre/core/MFEMSidreDataCollection.cpp
+++ b/src/axom/sidre/core/MFEMSidreDataCollection.cpp
@@ -914,6 +914,11 @@ void MFEMSidreDataCollection::Load(const std::string& path,
       mesh->NewNodes(*this->GetField(nodalGFName), false);
     }
   }
+
+  // Now we can finalize the mesh
+  const bool fix_orientation = false;
+  const bool refine = false;
+  mesh->Finalize(refine, fix_orientation);
 }
 
 void MFEMSidreDataCollection::LoadExternalData(const std::string& path)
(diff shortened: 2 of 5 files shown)


RELEASE-NOTES.md
```diff
@@ -77,6 +77,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Fixed bug in `Mint`'s VTK output for fields of type `int64` and `float`
 - Improved loading of data collections in `MFEMSidreDataCollection`
 - Added workaround to `MFEMSidreDataCollection` for `C++14` standard library feature that was not available in `gcc@4.9.3`
+- Delayed finalization of reloaded meshes in `MFEMSidreDataCollection` until after the setting
+  of the nodal `GridFunction` (when applicable)
```
kennyweiss, line 81: Slight suggestion:
```suggestion
- Delayed finalizing reloaded mesh in `MFEMSidreDataCollection` until after setting
  the nodal `GridFunction` (when applicable)
```

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -2399,119 +2418,127 @@ void MFEMSidreDataCollection::reconstructMesh()
...
+
+void MFEMSidreDataCollection::reconstructFields()
+{
+  sidre::Group* f = m_bp_grp->getGroup("fields");
+  for(auto idx = f->getFirstValidGroupIndex(); sidre::indexIsValid(idx);
+      idx = f->getNextValidGroupIndex(idx))
+  {
+    Group* field_grp = f->getGroup(idx);
+    // The nodal grid function will already have been reconstructed
+    if(field_grp->getName() != m_meshNodesGFName)
```
kennyweiss, line 2539: I wonder if there should still be a check since one can't guarantee the order that users might call this.

src/axom/sidre/core/MFEMSidreDataCollection.hpp
```diff
@@ -602,6 +598,9 @@ class MFEMSidreDataCollection : public mfem::DataCollection
   // Used as part of Load()
   void reconstructMesh();
 
+  // Reconstructs a single field from its corresponding Sidre group
+  void reconstructField(Group* field_grp);
+
   // Reconstructs all non-mesh-related fields using the current contents
   // of the datastore, used as part of Load()
```
kennyweiss, line 605: Suggestion: Please document the expectation that the nodal grid function has already been reconstructed/registered

src/axom/sidre/tests/sidre_mfem_datacollection.cpp
```diff
@@ -698,6 +699,11 @@ static void testParallelMeshReload(mfem::Mesh& base_mesh,
   auto reader_group_data = getGroupData(*reader_pmesh);
   EXPECT_EQ(writer_group_data, reader_group_data);
   EXPECT_TRUE(sdc_reader.verifyMeshBlueprint());
+
+  // Make sure that orientation checks pass
+  // These can fail if the nodal GF is not set up correction for a periodic (and HO?) mesh
+  EXPECT_EQ(sdc_reader.GetMesh()->CheckElementOrientation(), 0);
```
kennyweiss, line 705: Can you please expand on this a bit?

I guess we're assuming that the test meshes are orientable but possibly periodic. 
It might be good to note that (?)

kennyweiss (approved):
Thanks for fixing this @joshessman-llnl !

src/axom/sidre/core/MFEMSidreDataCollection.cpp
joshessman-llnl: I may be forgetting something but I think we should be able to guarantee the order since these methods are private and only accessible through `UpdateMeshAndFieldsFromDS` which does everything in the correct order

agcapps (approved):
Thanks for your work, @joshessman-llnl .

PR #646, opened 2021-08-31
Verify component dependencies

white238:
- clarify test dependency variable
- guard against test requiring mint
- add error check for spin in quest


kennyweiss (approved):
Thanks @white238

rhornung67 (approved):
Very nice @white238 Thank you for your service!

PR #650, opened 2021-09-13
Guards `data_collection_util` against primal being off

### LINKED ISSUE #632: Axom build errors when compiling data_collection_util.cpp
I used to be able to compile axom with
    -D AXOM_ENABLE_ALL_COMPONENTS=OFF 
    -D AXOM_ENABLE_SLIC=ON 
    -D AXOM_ENABLE_SIDRE=ON 
    -D MFEM_DIR=my_mfem_directory
    -D AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION=ON 

Now, when I use those settings, Axom tries to compile data_collection_util.cpp and then gives an error stating that axom/primal.hpp can't be found. Seems like the code is trying to compile data_collection_util.cpp even if the configuration was set up so that primal would not be included.

Does this mean that if I now want to use the MFEM Sidre data collection functionality I also need to install primal? Should the logic in axom/src/tools/CMakeLists.txt be updated so that data_collection_util.cpp is compiled only if primal was compiled as well?

white238:
closes #632

### DIFF
--- a/src/tools/CMakeLists.txt
+++ b/src/tools/CMakeLists.txt
@@ -72,7 +72,8 @@ endif()
 # data_collection_util is a utility that generates blueprint-conforming sidre
 # output files from an input mfem mesh
 #------------------------------------------------------------------------------
-if( AXOM_ENABLE_SIDRE AND MFEM_FOUND AND AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION )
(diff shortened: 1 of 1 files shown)


kennyweiss (approved):

agcapps (approved):
Thanks Chris

PR #652, opened 2021-09-17
BVH optimization for SignedDistance

kanye-quest:
# Summary

This PR adds an optimization for point queries against a BVH with the device traversal API. At a given node in the BVH, when both child nodes need to be traversed, we first traverse the node with the bounding box that is closer to the query point. This improves the performance of signed-distance queries, since the current-minimum distance for a given query point converges much faster to the actual minimum distance. As a result, more of the tree can be excluded from traversal.

This is not added to the `BVH::findCandidates` methods, because in that case, we must traverse all tentatively-intersecting nodes regardless.

## Performance

Measured with `./examples/quest_signed_distance_interface_ex --batched --box_dims 200 200 200 --file ../data/quest/plane.stl`

|          | `develop` | this branch | speedup |
| ------ | ---------- | ------------ | --------- |
| rzgenie | 109.177s | 29.8469s | 3.66x |
| rzansel | 9.75109s | 3.3926s | 2.87x |

### DIFF
--- a/src/axom/spin/internal/linear_bvh/bvh_traverse.hpp
+++ b/src/axom/spin/internal/linear_bvh/bvh_traverse.hpp
@@ -36,11 +36,21 @@ inline bool leaf_node(const int32& nodeIdx) { return (nodeIdx < 0); }
  * \param [in] p the primitive in query, e.g., a point, ray, etc.
  * \param [in] B functor that defines the check for the bins
  * \param [in] A functor that defines the leaf action
+ * \param [in] Comp functor used for determining which child node to traverse
+ *  down first if both bins are to be traversed
  *
  * \note The supplied functor `B` is expected to take the following two
  *  arguments:
  *    (1) The supplied primitive, p
  *    (2) a primal::BoundingBox< FloatType, NDIMS > of the BVH bin
+ *
+ * \note The supplied functor `Comp` is expected to take the following three
+ *  arguments:
+ *    (1) The left child bounding box
+ *    (2) The right child bounding box
+ *    (3) The primitive being queried
+ *  It should return true if the right child is closer (indicating a swap is
+ *  necessary) and false if the left child is closer.
  *
  * \see BVHData for the details on the internal data layout of the BVH.
  *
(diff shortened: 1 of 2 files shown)


kennyweiss (approved):
Neat!

src/axom/spin/internal/linear_bvh/bvh_traverse.hpp
```diff
@@ -36,26 +36,42 @@ inline bool leaf_node(const int32& nodeIdx) { return (nodeIdx < 0); }
...
+          typename TraversePref>
 AXOM_HOST_DEVICE inline void bvh_traverse(
   const primal::BoundingBox<FloatType, NDIMS>* inner_nodes,
   const int32* inner_node_children,
   const int32* leaf_nodes,
   const PrimitiveType& p,
   InBinCheck&& B,
-  LeafAction&& A)
+  LeafAction&& A,
+  TraversePref&& Comp)
```
white238, line 74: We are being rather inconsistent in casing in this function.

kanye-quest: I just went along with the different casing for the lambda arguments, but if there's another way we'd prefer to name them I'm all ears.

white238 (approved):

src/axom/spin/internal/linear_bvh/bvh_traverse.hpp
```diff
@@ -36,26 +36,42 @@ inline bool leaf_node(const int32& nodeIdx) { return (nodeIdx < 0); }
...
  *    (1) The supplied primitive, p
  *    (2) a primal::BoundingBox< FloatType, NDIMS > of the BVH bin
  *
+ * \note The supplied functor `Comp` is expected to take the following three
+ *  arguments:
+ *    (1) The left child bounding box
+ *    (2) The right child bounding box
+ *    (3) The primitive being queried
+ *  It should return true if the right child is closer (indicating a swap is
+ *  necessary) and false if the left child is closer.
```
agcapps, line 53: Unless I misunderstood, please consider the following for greater clarity.
```suggestion
 *  It should return true if the primitive is closer to the right child bounding
 *  box (indicating a swap is necessary) and false if the primitive is closer to
 *  the left child bounding box.
```

agcapps (approved):
Thanks for this, @kanye-quest !

agcapps (approved):
Thanks, @kanye-quest .  Much appreciated.

PR #666, opened 2021-10-08
Add new shroud interface for DataStore

nselliott:
# Summary

- This PR is a feature
  - Adds overloaded DataStore::generateBlueprintIndex signature to sidre_shroud.yaml and modifies shroud-generated code accordingly.


nselliott:
> Thanks @nselliott
> 
> Please test this in a non-MPI config that has sidre. I don't think we have any in our current CI, so you'd have to create one by setting `ENABLE_MPI=OFF`.

There are numerous compile failures unrelated to this PR with `ENABLE_MPI=OFF`.  I will create an issue for that, and perhaps this shouldn't be merged without that solved across the library.

PR #675, opened 2021-10-12
String utility function consistency

### LINKED ISSUE #624: Consistency in string utility functions
The string utilities have several similar functions that have different APIs.
We should centralize then and make them consistent.

#### split vs. splitLastNTokens

The API for `split` uses an `out` parameter for its results, while `splitLastNTokens` returns a vector of strings.
https://github.com/LLNL/axom/blob/833bc1b1d51c633c07b0214f452728f4e2f9b24d/src/axom/core/utilities/StringUtilities.hpp#L66-L74

https://github.com/LLNL/axom/blob/833bc1b1d51c633c07b0214f452728f4e2f9b24d/src/axom/core/utilities/StringUtilities.hpp#L88-L101

#### Move inlet's string utils to core ?

There are several string utility functions in `inlet` that do not use `inlet` functionality. 
We should consider moving these to `core::StringUtilities`

#### removeSuffix() vs. removePrefix()

We should also ensure that the order of parameters in `removeSuffix()` and `removePrefix()` are consistent.

https://github.com/LLNL/axom/blob/295747951104558efffddb0412d9280abcd1235d/src/axom/core/utilities/StringUtilities.hpp#L66-L79 

https://github.com/LLNL/axom/blob/692a74f6d35d11f20b4e99b329e0d29c64af57ea/src/axom/inlet/inlet_utils.hpp#L132-L154

joshessman-llnl:
# Summary

- This PR is a refactoring
- It does the following:
  - Unifies the signatures for `split` and `splitlastNTokens` - I chose the more readable (in my opinion) return values instead of out-params - with copy elision etc the overhead should be minimal
  - Moves string functions from `inlet_utils.cpp` into `core` as they are not inlet-specific
  - Fixes #624

# Design review

I opted to leave the signatures of `removeSuffix` and `appendPrefix` as they are - they are inconsistent with each other but a user would probably expect a suffix to follow the "main" string argument and a prefix to precede it.  Happy to change this if there's interest.

I also considered renaming `splitLastNTokens` to `rsplitn` - which is what the Rust standard library calls it (https://doc.rust-lang.org/stable/std/primitive.str.html#method.rsplitn).  See also https://pandas.pydata.org/docs/reference/api/pandas.Series.str.rsplit.html which has similar functionality.  Thoughts on this?

### DIFF
--- a/src/axom/core/utilities/StringUtilities.cpp
+++ b/src/axom/core/utilities/StringUtilities.cpp
@@ -74,7 +74,7 @@ std::string appendPrefix(const std::string& prefix,
 {
   if(prefix.empty() || name.empty())
   {
-    return {};
+    return name;
   }
   else
   {
--- a/src/axom/core/utilities/StringUtilities.hpp
+++ b/src/axom/core/utilities/StringUtilities.hpp
@@ -122,7 +122,7 @@ std::vector<std::string> rsplitN(const std::string& input,
 * \param [in] name The ending string name.
 * \param [in] delim The delimiting character inserted between the prefix and the name, if both are non-empty
 *
-* \return The appended string.
+* \return The appended string.  If @a prefix is empty, then just @a name is returned.
 *****************************************************************************
 */
 std::string appendPrefix(const std::string& prefix,


kennyweiss:
@joshessman-llnl wrote:
> I also considered renaming splitLastNTokens to rsplitn - which is what the Rust standard library calls it (https://doc.rust-lang.org/stable/std/primitive.str.html#method.rsplitn). See also https://pandas.pydata.org/docs/reference/api/pandas.Series.str.rsplit.html which has similar functionality. Thoughts on this?

I like the idea of `rsplit` and/or `rsplitn` -- 
* `splitLastNTokens` is pretty verbose
* `rsplit` appears to be standard in other languages (as demonstrated by your links)
* I couldn't find an analogue in the `C++` std library or boost

PR #677, opened 2021-10-13
ImplicitGrid port to GPU

kanye-quest:
# Summary

- Ports `spin::ImplicitGrid` to optionally run on GPU using a device-copyable query object
  - A new subobject, `ImplicitGrid::QueryObject`, contains method `visitCandidates()`, which accepts a query object and a functor that is called for each candidate intersection
  - Modifies `quest::PointInCell` to use vectorized and/or GPU-accelerated `ImplicitGrid` queries (majority of runtime still in MFEM though)
  - Adds a `quest::findTriMeshIntersections()` variant for `ImplicitGrid`
- Changes to Slam:
  - Replaces `std::vector` usage in `slam::BitSet` with `axom::Array`
  - Adds storage policy classes to enable use of `axom::Array` with `slam::Map`
- Fixes initialization of `axom::Array` containing non-POD objects

## Performance

Tested with `./mesh_tester -m implicit -p [exec_policy] --skipWeld --inFile [mesh file]`

| mesh + exec policy | ImplicitGrid init | ImplicitGrid query | total time |
| ------------------- | ----------------- | ----------- | ---------- |
| porsche.stl: Sequential | 0.021 s | 2.651 s | 3.754 s |
| porsche.stl: OpenMP 9 threads | 0.024 s | 0.544 s | 1.468 s |
| porsche.stl: OpenMP 36 threads | 0.014 s | 0.146 s | 1.255 s |
| porsche.stl: CUDA | 0.025 s | 0.017 s | 0.164 s |
| car.stl: OpenMP 9 threads | 0.169 s | 82.104 s | 96.667 s |
| car.stl: OpenMP 36 threads | 0.107 s | 16.337 s | 34.203 s |
| car.stl: CUDA | 0.102 s | 1.030 s | 1.591 s |

Node-to-node, we see roughly a speedup of 16-17x going from rzgenie to rzansel.

## TODO:

- [x] Add an example which uses GPU bitset
- [x] ~~Consider adding iterator interface to bitset to avoid needing underlying array access in `QueryObject::visitCandidates()`~~ realistically impossible because of the bounding box case
- [x] Wait for #695, determine additional changes to `axom::Array` usage needed afterwards
- [x] Update documentation
- [x] Extend ImplicitGrid tests to support GPU
- [ ] Remove dependencies on unified memory
- [x] Update CHANGELOG


src/axom/core/Array.hpp
```diff
@@ -224,7 +224,9 @@ class Array : public ArrayBase<T, DIM>
    * \post size() == num_elements
    * \post getResizeRatio() == DEFAULT_RESIZE_RATIO
    */
-  template <typename... Args>
+  template <IndexType SFINAE = DIM,
+            typename std::enable_if<SFINAE != 1>::type* = nullptr,
+            typename... Args>
```
joshessman-llnl, line 229: I had a slightly different fix for this in my current WIP branch that checks if the first argument in the parameter pack `is_integral` - though it would probably be more thorough to check the whole pack and/or add this conditional.  What was the context for this change?

src/axom/core/Array.hpp
```diff
@@ -1227,6 +1229,15 @@ inline void Array<T, DIM>::updateNumElements(IndexType new_num_elements)
 {
   assert(new_num_elements >= 0);
   assert(new_num_elements <= m_capacity);
+
+  int new_elems = new_num_elements - m_num_elements;
+  for(int ielem = 0; ielem < new_elems; ielem++)
+  {
+    // TODO: what to do here when T isn't default-constructible?
+    // we should probably do what std::vector does (zero out memory?)
+    new(&m_data[ielem + m_num_elements]) T;
```
joshessman-llnl, line 1238: For what it's worth the docstring for `axom::Array` specifically calls out the non-default-construction of its data as being one of its main differences relative to `std::vector`.  That callout dates from before I started working on the class so I can't be sure of its motivation, though.

src/axom/core/Array.hpp
kanye-quest: Yeah, this was mostly to get it building on CI -- I'll defer to whatever fix is in your branch :P

Basically, the issue I was running into was that on the CI job for gcc 8.4.0, there was some ambiguity with a call to the `Array(num_elems, capacity, allocator_id)` constructor - it seems like if any of the arguments have to be implicitly-converted, it will prefer the variadic size constructor even if the Array instance is a single-dimension array.

src/axom/core/Array.hpp
kanye-quest: I believe leaving the memory uninitialized was problematic for `axom::Array`s of non-trivial type. I think in my case, it was `axom::Array<BitSet>` - you can end up with the destructor for `BitSet` getting called on junk memory, which causes a segfault.

Although I understand the desire to avoid doing unnecessary initializations - maybe we can only do placement-new for cases where `std::is_trivially_destructible<T>` is false.

src/axom/spin/ImplicitGrid.hpp
```diff
@@ -50,7 +54,7 @@ namespace spin
  * is designed for quick indexing and searching over a static (and relatively
  * small index space) in a relatively coarse grid.
  */
-template <int NDIMS, typename TheIndexType = int>
+template <int NDIMS, typename ExecSpace = axom::SEQ_EXEC, typename TheIndexType = int>
```
kennyweiss, line 57: This change might impact users. 
Please note in the `RELEASE-NOTES` that the template parameters for `ImplicitGrid` changed.

src/axom/spin/ImplicitGrid.hpp
```diff
@@ -215,33 +246,77 @@ class ImplicitGrid
...
+          for(int j = lower; j <= upper; ++j)
+          {
+            binData[idim][j].atomicSet(elemIdx);
+#ifdef AXOM_USE_RAJA
+            RAJA::atomicMin<AtomicPol>(&minBlkBins[idim][j], word);
+            RAJA::atomicMax<AtomicPol>(&maxBlkBins[idim][j], word);
+#else
+            minBlkBins[idim][j] = std::min(minBlkBins[idim][j], word);
+            maxBlkBins[idim][j] = std::max(maxBlkBins[idim][j], word);
+#endif
```
kennyweiss, line 316: Could this be handled within axom's ExecutionSpace / AtomicPolicy implementation?
It seems a bit strange that AtomicPolicy is defined in Axom
>     using AtomicPol = typename axom::execution_space<ExecSpace>::atomic_policy;
(defined a few lines up)

but when we use it we suddenly need to know if we have RAJA.

src/axom/spin/ImplicitGrid.hpp
```diff
@@ -426,11 +528,531 @@ class ImplicitGrid
   //! The data associated with each bin
   BinBitMap m_binData[NDIMS];
 
+  //! The lowest word index in each bin with at least one bit set
+  axom::Array<IndexType> m_minBlockBin[NDIMS];
+
+  //! The highest word index in each bin with at least one bit set
+  axom::Array<IndexType> m_maxBlockBin[NDIMS];
```
kennyweiss, line 535: I like this!

kennyweiss (approved):
This is really nice @kanye-quest !

One thing that's difficult to tell from just looking at the code is how the memory transfers work. Unless I'm missing something, it seems that parts of the device code are relying on unified memory and would need to be extended to handle other memory spaces.  If so, we should probably add comments about this since users might want additional flexibility.

Please update the `RELEASE-NOTES`

Please also consider using `axom::Array` / `axom::ArrayView` throughout, where possible. (Not strictly required for this PR though).

src/axom/spin/ImplicitGrid.hpp
kanye-quest: I think the atomic policies are aliased directly to the corresponding RAJA policies (and to void if axom isn't built with RAJA). We could potentially wrap the RAJA atomic operations on our side, in order to handle the axom-minus-RAJA case more cleanly.

src/axom/core/Array.hpp
joshessman-llnl: Yep - I think what I'll end up doing is adding a `bool` member that allows the user to select at construction time whether they want data to be initialized to `T{}` when it's allocated - possibly via a new set of constructors that take a "dummy"/sentinel param:

```cpp
// Technically we could add a boolean arg to the 1D constructor instead
Array<int> foo(axom::ArrayOptions::Uninitialized{}, N);
// but this would be much more annoying for the multidimensional case, and I'd prefer to keep things consistent
Array<int, 3> bar(axom::ArrayOptions::Uninitialized{}, M, N, K);
```

That being said I have no problem with having the Array changes in this PR merged as-is for convenience/expediency.

src/axom/spin/tests/spin_implicit_grid.cpp
```diff
@@ -415,6 +434,135 @@ TYPED_TEST(ImplicitGridTest, get_candidates_pt)
...
+      AXOM_LAMBDA(IndexType i) {
+        numCandidates[i] = 0;
+        queryObj.visitCandidates(queryPts[i], [&](int candidateIdx) {
+          if(numCandidates[i] == 0)
+          {
+            candidates[i] = candidateIdx;
+          }
+          numCandidates[i]++;
+        });
+      });
```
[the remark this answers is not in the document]
kanye-quest: This was roughly equivalent to the vectorized `ImplicitGrid::getCandidatesAsArray` function - I've replaced the kernel with a call to that function, which should use the correct execution space.

src/axom/spin/tests/spin_implicit_grid.cpp
```diff
@@ -415,6 +434,135 @@ TYPED_TEST(ImplicitGridTest, get_candidates_pt)
...
+        queryObj.visitCandidates(queryPts[i], [&](int candidateIdx) {
+          int startIdx = 3 * i;
+          int slotIdx = numCandidates[i];
+          numCandidates[i]++;
+          if(slotIdx < 3)
+          {
+            candidates[startIdx + slotIdx] = candidateIdx;
+          }
+        });
+      });
```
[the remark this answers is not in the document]
kanye-quest: I've replaced this kernel with a call to the vectorized `ImplicitGrid::getCandidatesAsArray()` as well.

src/tools/mesh_tester.cpp
```diff
@@ -719,6 +721,45 @@ int main(int argc, char** argv)
         break;
       }
     }  // end of if method == 'bvh'
+    else if(params.method == "implicit")
+    {
+      switch(params.policy)
+      {
+#if defined(AXOM_USE_RAJA) && defined(AXOM_USE_UMPIRE)
+      case raja_seq:
+        quest::findTriMeshIntersectionsImplicitGrid<seq_exec, double>(
```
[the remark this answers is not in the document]
kanye-quest: No, it shouldn't - I think this part of the code was leftover from the BVH logic I copied it from.

I've modified the switch-case logic so `raja_seq` only depends on RAJA, and `seq` does the same thing as the `raja_seq` case.

src/tools/mesh_tester.cpp
```diff
@@ -719,6 +721,45 @@ int main(int argc, char** argv)
...
+          surface_mesh,
+          collisions,
+          degenerate,
+          params.resolution,
+          params.intersectionThreshold);
+        break;
+  #ifdef AXOM_USE_OPENMP
+      case raja_omp:
+        quest::findTriMeshIntersectionsImplicitGrid<omp_exec, double>(
```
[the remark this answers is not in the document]
kanye-quest: The `raja_omp` case now only depends on `AXOM_USE_RAJA` and `AXOM_USE_OPENMP`.

src/axom/quest/MeshTester.hpp
```diff
@@ -313,6 +315,200 @@ void findTriMeshIntersectionsBVH(
...
+  axom::setDefaultAllocator(axom::execution_space<ExecSpace>::allocatorID());
+
+  const int ncells = surface_mesh->getNumberOfCells();
+
+  axom::Array<detail::Triangle3> tris(ncells);
+  detail::Triangle3* p_tris = tris.data();
+
+  // Marks each cell/triangle as degenerate (1) or not (0)
+  axom::Array<int> degenerate(ncells);
+  int* p_degenerate = degenerate.data();
```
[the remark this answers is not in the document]
kanye-quest: I think we'd need something like a `View<BitSet>` to get a reference-like view of the bitset data in the GPU kernel. As it stands, a capture-by-copy of a BitSet in the device kernel would just create a new copy of the BitSet that would go away with the lambda.

src/axom/spin/tests/spin_implicit_grid.cpp
```diff
@@ -214,10 +236,7 @@ TYPED_TEST(ImplicitGridTest, insert_contains)
 
   // Insert objects into implicit grid.
   // Note: We do not insert an object with id 0
```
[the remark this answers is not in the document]
kanye-quest: It looks like that's related to this:
https://github.com/LLNL/axom/blob/df1c760fc6b770b61f70d3409b0e03d1c2f840e0/src/axom/spin/tests/spin_implicit_grid.cpp#L253-L257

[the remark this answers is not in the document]
kennyweiss: Thanks -- I probably should have added a clearer comment like:
```
// Note: Object with id 0 intentionally omitted for tests below
```

The original comment (out of context) made me thing that `ImplicitGrid` does not support objects with id 0.

src/axom/quest/PointInCell.hpp
```diff
@@ -201,6 +207,32 @@ class PointInCell
     return cellIndex;
   }
 
+  void locatePoints(axom::IndexType npts,
+                    const Point2DType* pts,
```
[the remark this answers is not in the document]
kanye-quest: Done, but with `axom::ArrayView<const Point2DType>`

src/axom/quest/PointInCell.hpp
```diff
@@ -201,6 +207,32 @@ class PointInCell
...
+  {
+    SLIC_ASSERT(npts > 0);
+    SLIC_ASSERT(pts != nullptr);
+    SLIC_ASSERT(outCellIds != nullptr);
+    SLIC_ASSERT(m_pointFinder2D != nullptr);
+
+    m_pointFinder2D->locatePoints(npts, pts, outCellIds, outIsopar);
+  }
+
+  void locatePoints(axom::IndexType npts,
```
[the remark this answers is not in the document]
kanye-quest: Done, but with `axom::ArrayView<const Point3DType>`

agcapps (approved):
Max, thank you for your hard work!

src/axom/core/Array.hpp
```diff
@@ -1001,6 +1001,15 @@ inline void Array<T, DIM, SPACE>::updateNumElements(IndexType new_num_elements)
 {
   assert(new_num_elements >= 0);
   assert(new_num_elements <= m_capacity);
+
+  int new_elems = new_num_elements - m_num_elements;
+  for(int ielem = 0; ielem < new_elems; ielem++)
+  {
+    // TODO: what to do here when T isn't default-constructible?
```
rhornung67, line 1008: Do we capture these TODO items sprinkled in the code as GitHub issues?

src/axom/core/utilities/BitUtilities.hpp
```diff
@@ -3,22 +3,23 @@
...
+ * \file BitUtilities.hpp
+ *
+ * \brief Header file containing bitwise utility functions.
  *
- * \brief Some bit twiddling operations in support of slam::BitSet
  */
 
-#ifndef SLAM_BIT_TWIDDLE_H_
-#define SLAM_BIT_TWIDDLE_H_
+#ifndef AXOM_BIT_UTILITIES_HPP
```
rhornung67, line 14: 👍

src/axom/core/Array.hpp
kennyweiss: This particular TODO is the subject of https://github.com/LLNL/axom/pull/709

rhornung67 (approved):

kanye-quest:
For now, I've opted to explicitly mark parts of the point-in-cell/mesh tester code where unified memory is actually needed, by using `axom::Array<..., MemorySpace::Unified>`. These should only be internally required; for point-in-cell we don't make any assumptions about where the user-provided output arrays are located.

I'll create issues to revisit these areas in the future.

Issue #633, opened 2021-08-09
Array Updats


joshessman-llnl:
From #627

I envision making the following changes in each PR (roughly in order)

- [ ] Extending Array to multiple dimensions - #627
- [ ] Replacing MCArray with Array (now that multiple dimensions are possible
- [ ] Adding a View (ArrayView?) class that provides indexing/access to a pointer it does not own, while removing the external data feature from Array
- [ ] Adding GPU support via a template parameter for the memory space of the Array/ArrayView

joshessman-llnl:
Some caveats/gotchas/pitfalls that have come up as I've been working through the MCArray -> Array replacement:

1. MCArray functions like `size()`/`capacity()`/etc returned the *number of tuples* and not the *number of elements*.  In the n-dimensional case this would probably correspond to the number of rank n-1 subtensors.  I've chosen to instead match `numpy.ndarray` and unconditionally return the number of *elements* - see https://numpy.org/doc/stable/reference/generated/numpy.ndarray.size.html

2. `emplace` as implemented in `MCArray` doesn't really correspond to the idea of `emplace` in the sense that the STL uses it, so I've opted not to add that functionality to `Array`.  If it's needed my preference would be to pick a different name for it.  Additionally, the `MCArray::emplace` interface is a little vague in my opinion - that is, it might not necessarily be clear to someone what it does on first glance.  Do any widely used numerical container classes provide similar functionality?

3. The semantics of "raw" `insert()` methods are pretty straightforward in the 1D case but significantly less so in the multidimensional case - if someone wants to insert of a single value (or an array of values of arbitrary length) at some arbitrary position in a multidimensional container, how would that affect the extents/dimensions/strides?  My sense is we'd want to disallow it in the multidimensional case.  I have implemented an `insert(IndexType pos, Array other)` (and a shortcut `append(Array other)`) that would allow the insertion of nonscalar data:
```cpp
int[] values = {1, 2, 3, 4}; // 2x2 {{1,2},{3,4}}
Array<int, 2> arr(1, 2);
arr(0,0) = 5;
arr(0,1) = 6;
// TODO: this will be an ArrayView since we're only temporarily wrapping a pointer
arr.insert(0, Array<int, 2>(values, 2, 2));
// will print [ 1 2 3 4 5 6 ]
std::cout << arr << "\n";
```
[...]

PR #695, opened 2021-10-27
Full GPU support for axom::Array

joshessman-llnl:
# Summary

- This PR is a feature
- It does the following:
  - Extends `Array` and in particular `ArrayView` to run on the GPU
  - Adds a `MemorySpace` parameter to allow users to optionally lock down the memory space of an `Array` at compile-time

Remaining tasks (non-exhaustive):
- [x] Multidim indexing/tests
- [x] Documentation updates

New docs:
https://axom.readthedocs.io/en/feature-essman-gpu_array/axom/core/docs/sphinx/core_containers.html#using-arrays-in-gpu-code


src/axom/core/docs/sphinx/core_containers.rst
```diff
@@ -137,6 +145,69 @@ The output of this example is::
...
+
+
+The full template signature for ``Array`` (``ArrayView`` has an analogous signature) is
+``Array<typename T, int DIM = 1, MemorySpace SPACE = MemorySpace::Dynamic>``.  Of particular interest
+is the last parameter, which specifies the memory space in which the array's data are allocated.
+The default, ``Dynamic``, means that the memory space is set via an allocator ID at runtime.
+
+.. note:: Allocating ``Array``s in different memory spaces is only possible when Umpire is available.
+
+Setting the ``MemorySpace`` to an option other than ``Dynamic`` - e.g., ``MemorySpace::Device`` provides
```
agcapps, line 169:
```suggestion
Setting the ``MemorySpace`` to an option other than ``Dynamic`` (for example, ``MemorySpace::Device``) provides
```

src/axom/core/docs/sphinx/core_containers.rst
```diff
@@ -137,6 +145,69 @@ The output of this example is::
...
+
+For example, a GPU kernel can require that its argument arrays be allocated in a specific memory space.
+
+.. literalinclude:: ../../examples/core_containers.cpp
+   :start-after: _cuda_kernel_start
+   :end-before: _cuda_kernel_end
+   :language: C++
+
+To illustrate how different memory spaces can be required, the inputs happen to be in unified memory and the output in
+device memory.
```
agcapps, line 182: Suggestion: move this idea to before the code snippet, to just after line 174, and rephrase it as something like
"This following kernel requires its input arrays ``A`` and ``B`` to be in unified memory and its output array ``C`` to be in device memory."

src/axom/core/examples/core_containers.cpp
```diff
@@ -176,8 +170,103 @@ void demoArrayBasic()
...
+{
+  //This example requires Umpire
+#if defined(AXOM_USE_UMPIRE) && defined(AXOM_USE_CUDA) && defined(__CUDACC__)
+  // _cuda_array_create_start
+  constexpr int N = 10;
+  const int allocator_id = axom::getUmpireResourceAllocatorID(
+    umpire::resource::MemoryResourceType::Unified);
+
+  // By default the memory space can be specified at runtime.
+  // If this argument is not provided host memory will be allocated.
```
agcapps, line 203: This comment is a little confusing, because you're dealing both with a template parameter which specifies the memory space at compile time (with default value `Dynamic`) and also with a constructor parameter which specifies the memory allocator at runtime (with default value "the allocator for `Dynamic`", which is host memory).  Here's a stab at a rewrite, which may or may not match your intention:

```suggestion
  // The last template parameter specifies a memory space.
  // Its default value is Dynamic, which lets the user specify the
  // memory space at runtime with a memory allocator ID.  The
  // third constructor parameter specifies the allocator.
  // If this argument is not provided host memory will be allocated.
```

src/axom/core/docs/sphinx/core_containers.rst
```diff
@@ -137,6 +145,69 @@ The output of this example is::
...
+
+The following snippet illustrates how one would create and initialize the inputs/outputs to this kernel.  Note
+how a "Dynamic" ``Array`` can be converted into an ``Array`` whose memory space is locked down.
+
+.. literalinclude:: ../../examples/core_containers.cpp
+   :start-after: _cuda_array_create_start
+   :end-before: _cuda_array_create_end
+   :language: C++
+
```
agcapps, line 191: What happens if the user does something like specify the memory space with a template parameter but provide an incompatible allocator?  (A compile-time failure would be awesome!)  I can see this happening through code modification or incremental development/debugging.

agcapps (commented):
This is very good, Josh.  I had a few questions, one of which (mismatch of memory space template parameter with constructor-specified allocator) may need more documentation or testing.  I think I'm ready to approve once you've clarified those things.

agcapps:
And thank you for the link to the branch's docs on readthedocs.  That is very helpful when reviewing documentation.

src/axom/core/docs/sphinx/core_containers.rst
joshessman-llnl: In this case the incompatible allocator is ignored:
https://github.com/LLNL/axom/blob/6644fd6526506489e9f18962253875ba5af1d696/src/axom/core/Array.hpp#L647-L649

That being said, I still have a warning in this case:
https://github.com/LLNL/axom/blob/6644fd6526506489e9f18962253875ba5af1d696/src/axom/core/Array.hpp#L651-L661

src/axom/core/docs/sphinx/core_containers.rst
```diff
@@ -137,6 +145,69 @@ The output of this example is::
...
+  3. Create an ``ArrayView`` from the ``Array`` to call the function.  For non-templated kernels
+     an implicit conversion is provided.
+
+
+The full template signature for ``Array`` (``ArrayView`` has an analogous signature) is
+``Array<typename T, int DIM = 1, MemorySpace SPACE = MemorySpace::Dynamic>``.  Of particular interest
+is the last parameter, which specifies the memory space in which the array's data are allocated.
+The default, ``Dynamic``, means that the memory space is set via an allocator ID at runtime.
+
+.. note:: Allocating ``Array``s in different memory spaces is only possible when Umpire is available.
```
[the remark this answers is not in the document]
joshessman-llnl: I think in this case it's because the closing backticks are immediately followed by an `s` - I'll add a space between them that should fix that.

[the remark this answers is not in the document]
agcapps: OK, however you choose to do that is fine with me.

src/axom/core/docs/sphinx/core_containers.rst
agcapps: Thanks for clarifying.  Just as a suggestion, consider adding a sentence like

```suggestion
.. note:: Unless the Dynamic memory space is in use, the ``Array`` constructor will
          ignore an allocator ID that doesn't match its memory space, and in debug
          builds will print a warning at runtime.
```

src/axom/core/Array.hpp
```diff
@@ -137,21 +132,32 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM>>
    * \post size() == num_elements
    * \post getResizeRatio() == DEFAULT_RESIZE_RATIO
    */
-  template <typename... Args>
+  template <typename... Args,
+            typename std::enable_if<
+              detail::first_type_is_integral<Args...>::value>::type* = nullptr>
```
kanye-quest, line 137: What would it take to extend this to check all the types in the parameter pack?

src/axom/core/Array.hpp
```diff
@@ -120,10 +105,20 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM>>
...
   Array(IndexType num_elements,
         IndexType capacity = 0,
-        int allocator_id = axom::getDefaultAllocatorID());
+        int allocator_id = axom::detail::getAllocatorID<SPACE>());
+
+  /// \overload
+  template <IndexType SFINAE_DIM = DIM,
+            MemorySpace SFINAE_SPACE = SPACE,
+            typename std::enable_if<SFINAE_DIM == 1 &&
+                                    SFINAE_SPACE != MemorySpace::Dynamic>::type* = nullptr>
```
kanye-quest, line 120: I am starting to get a little worried about relying on SFINAE in `axom::Array` from a maintainability standpoint.
- Do you have a sense of the compile time impact for instantiating `axom::Array`?
- How does SFINAE compare with alternative approaches, like putting these kinds of functions in a partially-specialized base class? Seems like alot of our functions are along the lines of "enable only for DIM == 1/DIM != 1"

(These don't have to be resolved/answered for this PR -- just something to consider)

src/axom/core/Array.hpp
```diff
@@ -494,7 +500,7 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM>>
   /*!
    * \brief Exchanges the contents of this Array with the other.
    */
-  void swap(Array<T, DIM>& other);
+  void swap(Array<T, DIM, SPACE>& other);
```
kanye-quest, line 503: Can this be extended to accept `Array other` from other memory spaces as well?

src/axom/core/Array.hpp
```diff
@@ -606,32 +614,62 @@ Array<T, DIM>::Array(Args... args)
...
+  initialize(num_elements, capacity);
+}
+
+//------------------------------------------------------------------------------
+template <typename T, int DIM, MemorySpace SPACE>
+Array<T, DIM, SPACE>::Array(const Array& other, int allocator_id)
+  : ArrayBase<T, DIM, Array<T, DIM, SPACE>>(
+      static_cast<const ArrayBase<T, DIM, Array<T, DIM, SPACE>>&>(other))
+  , m_allocator_id(SPACE == MemorySpace::Dynamic
+                     ? allocator_id
```
kanye-quest, line 648: What if `allocator_id` is from the correct memory space, but a different pool?

src/axom/core/ArrayBase.hpp
```diff
@@ -10,14 +10,38 @@
...
+ */
+enum class MemorySpace
+{
+  Dynamic,
+#ifdef AXOM_USE_UMPIRE
+  Host,
+  Device,
+  Unified,
+  Pinned,
+  Constant
```
kanye-quest, line 41: Out of curiosity - should we enforce a compile-time constraint that `Array`s or `ArrayView`s are read-only on the GPU for `MemorySpace::Constant`?

src/axom/core/Array.hpp
joshessman-llnl: I was able to replace some of its uses (for conditionally enabled member functions) with `static_assert` - but I think it's pretty close to unavoidable for the three remaining uses in constructors, specifically:

- restricting the variadic ctor (`Args... args`) to only accept parameter packs of integral type - with C++20 I'd be able to use a concept, but even then I'm not sure what the difference in compile times would be 
- allowing an allocator ID only for the 1D `(int elements, int capacity)` constructor when `MemSpace == Dynamic` - this one is technically not required since I could just add a runtime warning like I do in the copy ctor (where SFINAE is not possible), but I'd prefer to be able to restrict that at compile time.

src/axom/core/Array.hpp
joshessman-llnl: I'm not an Umpire expert by any means but I believe in this case we'd have to do some introspection on the allocator to try to track down its "parent allocator"?

src/axom/core/ArrayBase.hpp
joshessman-llnl: This should be possible via a `static_assert` in the non-const `data()` accessor in `ArrayView` - I'll go ahead and add that in

kanye-quest: Does having something like:
```c++
#ifdef AXOM_DEVICE_CODE
  static_assert(SPACE == MemorySpace::Constant);
#endif
```
in the non-const overloads of `Array::data()` and `Array::operator[]` not work? I think that should allow you to write to the array from the host.

joshessman-llnl: I think that would still restrict host usage of that function if the corresponding host code was compiled with NVCC - that is, I think that macro is defined for both the host and device passes.

src/axom/core/Array.hpp
joshessman-llnl: I can add a swap function between memory spaces but I'd be curious to hear what some use cases might be for this - in the general case you'd have to copy both sets of data between memory spaces, so I think the only thing you'd really be swapping are the housekeeping/metadata variables.

src/axom/core/ArrayBase.hpp
kanye-quest: IIRC, I think `AXOM_DEVICE_CODE` is equivalent to `__CUDA_ARCH__`, which should only be defined on the device code pass.

src/axom/core/tests/core_array.hpp
```diff
@@ -733,6 +733,174 @@ void check_external_view(ArrayView<T>& v)
...
+ */
+template <typename T, int DIM, axom::MemorySpace SPACE>
+void check_device(Array<T, DIM, SPACE>& v)
+{
+  const IndexType size = v.size();
+  // Then assign to it via a raw device pointer
+  assign_raw<<<1, 1>>>(v.data(), size);
+
+  // Check the contents of the array by assigning to a Dynamic array
+  // The default Umpire allocator should be Host, so we can access it from the CPU
```
kennyweiss, line 768: >   // The default Umpire allocator should be Host, so we can access it from the CPU

Would it be worthwhile to add a check to confirm this?

src/axom/core/tests/core_array.hpp
```diff
@@ -733,6 +733,174 @@ void check_external_view(ArrayView<T>& v)
   EXPECT_EQ(data_ptr, v.data());
 }
 
+#ifdef __CUDACC__
```
kennyweiss, line 736: Does this file need to check guard/check against Umpire's availability?

src/axom/core/docs/sphinx/core_containers.rst
```diff
@@ -137,6 +145,72 @@ The output of this example is::
...
+  3. Create an ``ArrayView`` from the ``Array`` to call the function.  For non-templated kernels
+     an implicit conversion is provided.
+
+
+The full template signature for ``Array`` (``ArrayView`` has an analogous signature) is
+``Array<typename T, int DIM = 1, MemorySpace SPACE = MemorySpace::Dynamic>``.  Of particular interest
+is the last parameter, which specifies the memory space in which the array's data are allocated.
+The default, ``Dynamic``, means that the memory space is set via an allocator ID at runtime.
+
+.. note:: Allocating ``Array`` s in different memory spaces is only possible when Umpire is available.
```
kennyweiss, line 167: Suggestion: It might be a good idea to link to Umpire. E.g. https://github.com/LLNL/umpire

src/axom/core/ArrayBase.hpp
```diff
@@ -477,6 +532,74 @@ bool allNonNegative(const T (&arr)[N])
...
+template <typename... Args>
+struct all_types_are_integral
+{
+  static constexpr bool value = all_types_are_integral_impl<Args...>::value;
+};
+
+/// \brief Translates between the MemorySpace enum and Umpire allocator IDs
+template <MemorySpace SPACE>
+inline int getAllocatorID();
+
```
kennyweiss, line 557: Minor (optional): It might be useful to have a utility function of the form:  
```cpp
bool areCompatible(MemorySpace space1, MemorySpace space2);
```
And possibly also:
```cpp
#ifdef AXOM_USE_UMPIRE
bool areCompatible(MemorySpace SPACE, int allocatorID);
#endif
```

src/axom/core/examples/core_containers.cpp
```diff
@@ -176,8 +170,106 @@ void demoArrayBasic()
...
+
+  // Declare the lambda mutable so our copy of C_view (captured by value) is mutable
+  axom::for_all<axom::CUDA_EXEC<1>>(
+    0,
+    N,
+    [=] AXOM_HOST_DEVICE(axom::IndexType i) mutable {
+      C_view[i] = A_view[i] + B_view[i] + 1;
+    });
+
+  // Finally, copy things over to host mmemory so we can display the data
```
kennyweiss, line 262: typo: 
```suggestion
  // Finally, copy things over to host memory so we can display the data
```

src/axom/core/docs/sphinx/core_containers.rst
```diff
@@ -137,6 +145,72 @@ The output of this example is::
...
+########################
+
+Instead of writing kernels and device functions that operate on raw pointers, we can use ``ArrayView``
+in device code. The basic "workflow" for this process is as follows:
+
+  1. Create an ``Array`` allocated in device-accessible memory via either specifying an allocator ID
+     or using a class template parameter for the desired memory space.
+  2. Write a kernel that accepts an ``ArrayView`` parameter **by value**, not by reference or pointer.
+  3. Create an ``ArrayView`` from the ``Array`` to call the function.  For non-templated kernels
+     an implicit conversion is provided.
```
kennyweiss, line 159: >   Create an ``ArrayView`` from the ``Array`` to call the function.

Could we please spell this out a bit more explicitly in the discussion below?

One option here is to use the following `ArrayView` constructor -- 
```cpp
Array<int, 1, DEVICE> arr(n);
ArrayView(arr);
```

Another option is the implicit conversion that was mentioned.

Are there other ways?

src/axom/core/ArrayView.hpp
```diff
@@ -54,10 +55,20 @@ class ArrayView : public ArrayBase<T, DIM, ArrayView<T, DIM>>
...
 
+  /*! 
+   * \brief Constructor for transferring between memory spaces
+   * 
+   * \param [in] other The array in a different memory space to copy from
+   * 
+   * \note The parameter is non-const because \a other can be modified through the constructed View
+   */
+  template <typename OtherArrayType>
+  ArrayView(ArrayBase<T, DIM, OtherArrayType>& other);
```
kennyweiss, line 66: My understanding is that this constructor is intentionally not marked as `explicit`.

In light of #693, please add a comment on this and any other constructors for the Array classes that are intentionally implicit.

kennyweiss (approved):
Thanks @joshessman-llnl -- this is really nice and highly anticipated!

Please see my comments for some minor suggestions/corrections.

src/axom/core/Array.hpp
kennyweiss: It might be useful for memory spaces that are compatible.
E.g. a `Dynamic` space that happens to be device memory (via umpire) and a `Device` space ?

src/axom/core/StackArray.hpp
```diff
@@ -57,9 +57,57 @@ struct StackArray
...
+template <typename T, int N>
+AXOM_HOST_DEVICE bool operator==(const StackArray<T, N>& lhs,
+                                 const StackArray<T, N>& rhs)
+{
+  for(int i = 0; i < N; i++)
+  {
+    if(lhs[i] != rhs[i])
+    {
+      return false;
+    }
```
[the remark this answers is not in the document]
joshessman-llnl: I've added this to the line item that is already present for fuzzy array comparisons in #633

src/axom/core/examples/core_containers.cpp
```diff
@@ -176,8 +170,106 @@ void demoArrayBasic()
   // _iteration_end
 }
 
+#ifdef __CUDACC__
+
+// _cuda_kernel_start
+// Aliases used for convenience
+using UnifiedIntArrayView = axom::ArrayView<int, 1, axom::MemorySpace::Unified>;
```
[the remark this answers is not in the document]
joshessman-llnl: I have adjusted this to check for Umpire and also incorporated Umpire's `UMPIRE_ENABLE_UM` to make sure that Unified Memory is available.

src/axom/core/tests/core_array.hpp
joshessman-llnl: The snippets further down check for Umpire but since these functions use the full set of memory spaces I have added that to the preproc conditional.

src/axom/core/docs/sphinx/core_containers.rst
joshessman-llnl: I've added a new section to the example + docs that gets into the explicit + implicit conversions to ArrayView before diving into the `add` kernel.

src/axom/core/ArrayBase.hpp
joshessman-llnl: I can understand the utility of the second function (in particular, relating to @kanye-quest's comment about pooled allocators and also converting from Dynamic -> another space) but I'm not sure where/how the first one would be used.

src/axom/core/ArrayView.hpp
```diff
@@ -111,17 +122,59 @@ using MCArrayView = ArrayView<T, 2>;
...
+  if(SPACE == MemorySpace::Dynamic)
+  {
+    auto& rm = umpire::ResourceManager::getInstance();
+    if(rm.hasAllocator(data))
+    {
+      auto alloc = rm.getAllocator(data);
+      m_allocator_id = alloc.getId();
+    }
+  }
+#endif
```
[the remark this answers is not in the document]
joshessman-llnl: I'm probably missing something but I'm not clear on how `getAllocatorID` comes into play here - were you envisioning an overload `getAllocatorID(T* ptr)` that encapsulates the pointer introspection?

src/axom/core/ArrayBase.hpp
```diff
@@ -477,6 +510,25 @@ bool allNonNegative(const T (&arr)[N])
...
+{ };
+
+template <typename First, typename... Rest>
+struct all_types_are_integral_impl<First, Rest...>
+{
+  static constexpr bool value = std::is_integral<First>::value &&
+    all_types_are_integral_impl<Rest...>::value;
+};
+
+/// \brief Checks if the first type in a parameter pack is integral
```
white238, line 525: Comment doesn't match the struct..

white238 (approved):

PR #706, opened 2021-11-08
Improves robustness of `SignedDistance` query

### LINKED ISSUE #346: Add testing of Release Builds in our CI
Our CI (Azure & Nightly testing on LC systems) is currently testing against "Debug" builds.

 In a production setting, host-applications are typically linking against "Release" builds or "Release with Debug info", which currently our testing is not capturing.

In addition to our current test configurations:
- We should add "Release" builds in or CI and nightly testing for "common" specs (e.g., perhaps, one spec per platform)
- Also build against "Release" versions of the corresponding TPLs, e.g., RAJA, Umpire, etc.

### LINKED ISSUE #703: Regression in signed distance query
We're seeing a regression with the signed distance query on the [`plane_simp.stl` mesh](https://github.com/LLNL/axom_data/blob/main/quest/plane_simp.stl).

For the following run:
```
> ./examples/quest_signed_distance_interface_ex  \
                -f ../data/quest/plane_simp.stl  \
                --box-dims  120  120   60        \
                --box-min  -540 -140 -190        \
                --box-max   540  940  350
```
The generated signed distance field gives the following isosurface at distance 0:
image

The topological noise appears to be due to incorrect sign determination at query points.

This can also be seen in the quest regression test, which compares the signed distance to the in/out query:
```
> ./tests/quest_regression_test         \
        -m ../data/quest/plane_simp.stl \
        -r     120  120   60            \
        --min -540 -140 -190            \
        --max  540  940  350

[INFO] Running containment queries 
[INFO] Initializing InOutOctree over mesh '../data/quest/plane_simp.stl'... 
[INFO] Initialization took 36.691606953 seconds. 
[INFO] Mesh bounding box is: { min:(-419.473,-36.3145,-33.8787); max:(419.575,845.692,198.809); range:<839.047,882.006,232.687> } 
[INFO] Query bounding box is: { min:(-540,-140,-190); max:(540,940,350); range:<1080,1080,540> } 
[INFO] Querying InOutOctree on uniform grid of resolution (120,120,60) 
[INFO] Filling coordinates array took 0.103631259 seconds 
[INFO] Querying (120,120,60)^3 containment field (InOutOctree) took 1.222712744 seconds (@ 730425.8538095355 queries per second) 

[INFO] -- 

[INFO] Running distance queries 
[INFO] Initializing linear BVH over mesh '../data/quest/plane_simp.stl'... 
[INFO] Initialization took 0.251781796 seconds. 
[INFO] Mesh bounding box is: { min:(-419.473,-36.3145,-33.8787); max:(419.575,845.692,198.809); range:<839.047,882.006,232.687> } 
[INFO] Query bounding box is: { min:(-540,-140,-190); max:(540,940,350); range:<1080,1080,540> } 
[INFO] Querying BVH on uniform grid of resolution (120,120,60) 
[INFO] Filling coordinates array took 0.105148551 seconds 
[INFO] Querying (120,120,60)^3 signed distance field (BVH) took 346.761995387 seconds (@ 2575.544644110333 queries per second) 

[INFO] -- 

[INFO] Comparing results from containment and distance queries 
[INFO] ** Disagreement between SignedDistance  and InOutOctree containment queries.  
 There were 29 differences.
 Showing first 10 results:
  Disagreement on sample 288827 @ (-540,652,-19).  Signed distance: -137.72374302780582 (inside) -- InOutOctree: outside 
  Disagreement on sample 303590 @ (-531,661,-10).  Signed distance: -129.69483711472913 (inside) -- InOutOctree: outside 
  Disagreement on sample 360095 @ (540,499,26).  Signed distance: -135.76786870997012 (inside) -- InOutOctree: outside 
  Disagreement on sample 360452 @ (486,526,26).  Signed distance: -75.70138787605912 (inside) -- InOutOctree: outside 
  Disagreement on sample 435494 @ (-405,670,71).  Signed distance: -71.67610299911681 (inside) -- InOutOctree: outside 
  Disagreement on sample 450947 @ (369,724,80).  Signed distance: -127.1654044184815 (inside) -- InOutOctree: outside 
  Disagreement on sample 494420 @ (-414,697,107).  Signed distance: -115.57885572205763 (inside) -- InOutOctree: outside 
  Disagreement on sample 495352 @ (351,760,107).  Signed distance: -175.28380757357684 (inside) -- InOutOctree: outside 
  Disagreement on sample 495386 @ (-432,769,107).  Signed distance: -174.47980674036606 (inside) -- InOutOctree: outside 
  Disagreement on sample 509182 @ (-414,706,116).  Signed distance: -128.30328371674136 (inside) -- InOutOctree: outside 

[INFO] ** Methods do not agree 
```

### LINKED ISSUE #708: Add complete set of arithmetic operators for points and vectors
In primal, we have a few free-standing operator overloads for `primal::Vector`, such as adding two vectors together to yield a new vector, or the product of a scalar and a vector.

@wnissen requested that we add the complete set of valid operators.
E.g. we are missing:
* `operator+(PointType, VectorType) -> VectorType`
* `operator+(VectorType, PointType) -> VectorType`
* `operator-(PointType, PointType) -> VectorType`

But note that some combinations are invalid, e.g.
* `operator+(PointType, PointType) ->  <invalid>`

Here's what we currently have: 
https://github.com/LLNL/axom/blob/564d97741959e239541c6f235ccf23be9b5b0bae/src/axom/primal/geometry/Vector.hpp#L33-L89

kennyweiss:
# Summary

- This PR is a bugfix
- It resolves #703 
- It improves the robustness of `primal::closest_point(pt, triangle)` by adding an `EPS` tolerance parameter and using fuzzy comparisons to better determine if/when the closest point is on an edge or vertex of the triangle
- It uses this in `SignedDistance` to better track whether the closest point is on an edge or vertex.
  For these two cases, we need to track the weighted sum of all triangles incident in the edge or vertex
- It also improves how `SignedDistance` tracks which case it is in (face, edge or vertex)
- This improvement was tested against a configuration of the `plane_simp.stl` triangle mesh that was reported in #703 
   as well as a new `tetrahedron.stl` mesh that was added for this PR
- Also resolves #708
- Also resolve #346

#### TODO

- [x] Check this code against GPU configurations. Some functions will likely need to be host/device decorated
- [x] Add regression tests for the new failures (`plane_simp` and `tetrahedron`)
- [x] Ensure that all quest regression tests are passing
- [x] Update RELEASE-NOTES
- [x] Update `data` submodule after merging https://github.com/LLNL/axom_data/pull/8
- [x] Enable extra quest regression tests in a subset of CI jobs

### DIFF
--- a/.gitlab/build_ruby.yml
+++ b/.gitlab/build_ruby.yml
@@ -63,6 +63,14 @@ ruby-clang_10_0_0-src:
     HOST_CONFIG: "ruby-toss_3_x86_64_ib-${COMPILER}.cmake"
   extends: .src_build_on_ruby
 
+ruby-clang_10_0_0-release-src:
+  variables:
+    COMPILER: "clang@10.0.0"
+    HOST_CONFIG: "ruby-toss_3_x86_64_ib-${COMPILER}.cmake"
+    BUILD_TYPE: "Release"
+    EXTRA_CMAKE_OPTIONS: "-DAXOM_QUEST_ENABLE_EXTRA_REGRESSION_TESTS:BOOL=ON"
+  extends: .src_build_on_ruby
+
 ruby-clang_9_0_0-src:
   variables:
     COMPILER: "clang@9.0.0"
--- a/scripts/azure-pipelines/linux-build_and_test.sh
+++ b/scripts/azure-pipelines/linux-build_and_test.sh
@@ -18,16 +18,19 @@ function or_die () {
 }
 
 or_die cd axom
-git submodule init 
-git submodule update 
+git submodule init
+git submodule update
 
 echo HOST_CONFIG
 echo $HOST_CONFIG
 
+export BUILD_TYPE=${BUILD_TYPE:-Debug}
+
+
 if [[ "$DO_BUILD" == "yes" ]] ; then
     echo "~~~~~~ RUNNING CMAKE ~~~~~~~~"
-    or_die ./config-build.py -hc /home/axom/axom/host-configs/docker/${HOST_CONFIG}.cmake -DENABLE_GTEST_DEATH_TESTS=ON ${CMAKE_EXTRA_FLAGS}
-    or_die cd build-$HOST_CONFIG-debug
+    or_die ./config-build.py -hc /home/axom/axom/host-configs/docker/${HOST_CONFIG}.cmake -bt ${BUILD_TYPE} -DENABLE_GTEST_DEATH_TESTS=ON ${CMAKE_EXTRA_FLAGS}
+    or_die cd build-$HOST_CONFIG-${BUILD_TYPE,,}
     echo "~~~~~~ BUILDING ~~~~~~~~"
     if [[ ${CMAKE_EXTRA_FLAGS} == *COVERAGE* ]] ; then
         or_die make -j 10
--- a/src/axom/primal/operators/closest_point.hpp
+++ b/src/axom/primal/operators/closest_point.hpp
@@ -17,7 +17,7 @@
 #include "axom/primal/geometry/Point.hpp"
 #include "axom/primal/geometry/Triangle.hpp"
 #include "axom/primal/geometry/OrientedBoundingBox.hpp"
-
+#include "axom/primal/operators/detail/intersect_impl.hpp"
 namespace axom
 {
 namespace primal
@@ -60,10 +60,14 @@ namespace primal
 template <typename T, int NDIMS>
 AXOM_HOST_DEVICE inline Point<T, NDIMS> closest_point(const Point<T, NDIMS>& P,
                                                       const Triangle<T, NDIMS>& tri,
-                                                      int* loc = nullptr)
+                                                      int* loc = nullptr,
+                                                      double EPS = 1E-8)
 {
   using PointType = Point<T, NDIMS>;
   using VectorType = Vector<T, NDIMS>;
+
+  using detail::isGeq;
+  using detail::isLeq;
 
   const PointType& A = tri[0];
   const PointType& B = tri[1];
@@ -75,7 +79,7 @@ AXOM_HOST_DEVICE inline Point<T, NDIMS> closest_point(const Point<T, NDIMS>& P,
   const VectorType ap(A, P);
   const T d1 = VectorType::dot_product(ab, ap);
   const T d2 = VectorType::dot_product(ac, ap);
-  if(d1 <= 0.0f && d2 <= 0.0f)
+  if(isLeq(d1, T(0), EPS) && isLeq(d2, T(0), EPS))
   {
     // A is the closest point
     if(loc != nullptr)
@@ -91,7 +95,7 @@ AXOM_HOST_DEVICE inline Point<T, NDIMS> closest_point(const Point<T, NDIMS>& P,
   const VectorType bp(B, P);
   const T d3 = VectorType::dot_product(ab, bp);
   const T d4 = VectorType::dot_product(ac, bp);
-  if(d3 >= 0.0f && d4 <= d3)
+  if(isGeq(d3, T(0), EPS) && isLeq(d4, d3, EPS))
   {
     // B is the closest point
     if(loc != nullptr)
@@ -105,7 +109,7 @@ AXOM_HOST_DEVICE inline Point<T, NDIMS> closest_point(const Point<T, NDIMS>& P,
   //----------------------------------------------------------------------------
   // Check if P in edge region of AB
   const T vc = d1 * d4 - d3 * d2;
-  if(vc <= 0.0f && d1 >= 0.0f && d3 <= 0.0f)
+  if(isLeq(vc, T(0), EPS) && isGeq(d1, T(0), EPS) && isLeq(d3, T(0), EPS))
   {
     const T v = d1 / (d1 - d3);
     const VectorType v_ab = ab * v;
@@ -123,7 +127,7 @@ AXOM_HOST_DEVICE inline Point<T, NDIMS> closest_point(const Point<T, NDIMS>& P,
   const VectorType cp(C, P);
   const T d5 = VectorType::dot_product(ab, cp);
   const T d6 = VectorType::dot_product(ac, cp);
-  if(d6 >= 0.0f && d5 <= d6)
+  if(isGeq(d6, T(0), EPS) && isLeq(d5, d6, EPS))
   {
     // C is the closest point
     if(loc != nullptr)
@@ -137,7 +141,7 @@ AXOM_HOST_DEVICE inline Point<T, NDIMS> closest_point(const Point<T, NDIMS>& P,
   //----------------------------------------------------------------------------
   // Check if P in edge region of AC
   const T vb = d5 * d2 - d1 * d6;
-  if(vb <= 0.0f && d2 >= 0.0f && d6 <= 0.0f)
+  if(isLeq(vb, T(0), EPS) && isGeq(d2, T(0), EPS) && isLeq(d6, T(0), EPS))
   {
     const T w = d2 / (d2 - d6);
     const VectorType w_ac = ac * w;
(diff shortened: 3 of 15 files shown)


src/axom/primal/operators/closest_point.hpp
```diff
@@ -60,146 +60,131 @@ namespace primal
...
-#define A(t) t[0]
-#define B(t) t[1]
-#define C(t) t[2]
+  using PointType = Point<T, NDIMS>;
+  using VectorType = Vector<T, NDIMS>;
+
+  using detail::isGeq;
+  using detail::isLeq;
+
+  const PointType& A = tri[0];
```
rhornung67, line 72: 👍 much better!

src/axom/quest/SignedDistance.hpp
```diff
@@ -608,34 +676,22 @@ inline double SignedDistance<NDIMS, ExecSpace>::computeSign(
...
-  {
-    // CASE 2: closest point is on an edge, use sum of normals of equidistant
-    // faces
-    // TODO: Sometimes, the traversal fails to find the opposite face, so only
-    // a single face's normal is accumulated here. The proper solution would be
-    // to precompute edge pseudo-normals during construction time, but that
-    // would also require generating cell-to-face connectivity for the surface
-    // mesh.
+    break;
+  default:  // Use precomputed normal for edges and vertices
```
kanye-quest, line 688: I remember when debugging the original SignedDistance PR that the BVH traversal sometimes only accumulated one face of an edge pseudo-normal; was that the underlying issue here?

In any case, I think it might be good to have a debug check/tracker variable here for the edge and vertex pseudo-normals:
- If we keep track of the number of normals we accumulated for the current weighted sum, an edge closest point should have exactly two accumulated normals, one for each face.
- Not sure about the angle-weighted normal for vertex closest points, but I think we could check that there at least 3? (since any angle at a triangle vertex should be less than 180 degrees)

kennyweiss: > I remember when debugging the original SignedDistance PR that the BVH traversal sometimes only accumulated one face of an edge pseudo-normal; was that the underlying issue here?

There were a few problems. Most prominently:
1.  `primal::closest_point()` was using strict inequality tests, so points that were within `EPS of an edge or vertex could be marked as interior (face), and either not counted for the edge/vertex test, or would clear out the normal
2. The logic for determining whether to clear normals and/or sum normals needed some improvement.

kennyweiss: The underlying issue is that we're using purely geometric tests, so we cannot guarantee that the we're catching all (both) faces in the star of an edge or all faces in the star of a vertex.

A potentially better, but more expensive way to handle this would be to:
1. Weld the mesh to ensure that we have a mesh instead of a collection of disjoint faces (see #705)
2. Compute the pseudo-normal at each vertex of the welded mesh
3. Only keep track of the single closest point
4. Derive the sign from the normal of the mesh at this point. 

This would robustly handle the face, edge and vertex cases for a watertight surface mesh.
The other downside of this approach would be that it would involve more memory (to store the normals) and more computation time (to weld the mesh and compute the normals)

kennyweiss: We can certainly add a debug check/variable to count the number of accumulated normals.
What would you suggest we do with that information though? 
I suppose we can add an option to add scalar fields to the output vtk mesh to visualize this (closest point type and count)

.gitlab/build_ruby.yml
```diff
@@ -63,6 +63,14 @@ ruby-clang_10_0_0-src:
     HOST_CONFIG: "ruby-toss_3_x86_64_ib-${COMPILER}.cmake"
   extends: .src_build_on_ruby
 
+ruby-clang_10_0_0-release-src:
+  variables:
+    COMPILER: "clang@10.0.0"
+    HOST_CONFIG: "ruby-toss_3_x86_64_ib-${COMPILER}.cmake"
+    BUILD_TYPE: "Release"
+    EXTRA_CMAKE_OPTIONS: "-DAXOM_QUEST_ENABLE_EXTRA_REGRESSION_TESTS:BOOL=ON"
```
bmhan12, line 71: Should options like this and `AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION` be documented in our [ReadTheDocs configuration options](https://axom.readthedocs.io/en/develop/docs/sphinx/quickstart_guide/config_build.html#cmake-configuration-options)?

kennyweiss: Thanks @bmhan12. In general, yes, we should document all our user-facing config options.

In the past, we have intentionally omitted a few niche options from our documentation, including the two that you mentioned:

* `AXOM_QUEST_ENABLE_EXTRA_REGRESSION_TESTS` is a developer option. We have a few quest regression tests that are always on and this option adds additional tests. 
* `AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION` was intended as a short-lived option -- it was added because we couldn't yet enable this feature by default, and we plan to remove this option (and enable the feature by default) as soon as we can. As far as I can tell, we should be able to do this once we have a working hip configuration.

There are a few other undocumented options in the code, like `AXOM_MINT_USE_SIDRE` that we might also want to consider
https://github.com/LLNL/axom/blob/3999cf27a5437b032f9d3e23ca4476c04d139a2e/src/axom/mint/CMakeLists.txt#L18-L19

Any of these should be easy enough to add to the docs. We might want to discuss this further as a team. I'll create an issue for it.

kennyweiss:
I've resolved the issues in #703, added more quest regression tests and enabled these in some CI jobs, so this PR should be ready for review.

For completeness, here are the two examples and their isosurfaces from #703 
```
 >./examples/quest_signed_distance_interface_ex -f ../data/quest/plane_simp.stl --box-dims  120 120 60 --box-min  -540 -140 -190 --box-max   540  940  350 --batched -e1
```
image

```
./examples/quest_signed_distance_interface_ex  -f ../data/quest/tetrahedron.stl  --box-dims 100 100 100  --box-min -5 -5 -5 --box-max 5 5 5 --batched -e 1
```
image

The plane appears to be underresolved to capture all the features, but the spurious signs (topological noise) have been resolved in both cases, and the signs agree with those of the InOutOctree.

scripts/azure-pipelines/linux-build_and_test.sh
```diff
@@ -18,16 +18,19 @@ function or_die () {
...
 
+export BUILD_TYPE=${BUILD_TYPE:-Debug}
+
+
 if [[ "$DO_BUILD" == "yes" ]] ; then
     echo "~~~~~~ RUNNING CMAKE ~~~~~~~~"
-    or_die ./config-build.py -hc /home/axom/axom/host-configs/docker/${HOST_CONFIG}.cmake -DENABLE_GTEST_DEATH_TESTS=ON ${CMAKE_EXTRA_FLAGS}
-    or_die cd build-$HOST_CONFIG-debug
+    or_die ./config-build.py -hc /home/axom/axom/host-configs/docker/${HOST_CONFIG}.cmake -bt ${BUILD_TYPE} -DENABLE_GTEST_DEATH_TESTS=ON ${CMAKE_EXTRA_FLAGS}
+    or_die cd build-$HOST_CONFIG-${BUILD_TYPE,,}
```
agcapps, line 33: What do the two commas do?

agcapps (approved):
@kennyweiss , thank you for the work.

scripts/azure-pipelines/linux-build_and_test.sh
kennyweiss: It converts the variable to lowercase -- https://askubuntu.com/a/383360

This was necessary because we're not passing in the build-path (`--bp`) in our azure pipelines
and `config-build.py` lower-cases the build type (`--bt`) here:
https://github.com/LLNL/axom/blob/3999cf27a5437b032f9d3e23ca4476c04d139a2e/config-build.py#L169-L177

agcapps: Thanks.

agcapps (approved):
Kenny, thank you for your diligent work.

src/axom/quest/examples/quest_signed_distance_interface.cpp
```diff
@@ -78,7 +78,7 @@ struct Arguments
   bool use_shared {false};
   bool use_batched_query {false};
   bool ignore_signs {false};
-  quest::SignedDistExec exec_space;
+  quest::SignedDistExec exec_space {quest::SignedDistExec::CPU};
```
kanye-quest, line 81: Good catch - I was running into some issues with wonky execution space values on optimized builds as well.

kanye-quest (approved):
LGTM - great to see robustness improvements in Primal as well 👍

rhornung67 (approved):

PR #707, opened 2021-11-08
Initialize array in Polyhedron before use

### LINKED ISSUE #704: Uninitialized variable in Polyhedron class
I noticed a warning from the intel compiler about an uninitialized variable in the `Polyhedron` class that should be addressed:

https://github.com/LLNL/axom/blob/18f7408e72e2924e71edde1938648cb0a9019c10/src/axom/primal/geometry/Polyhedron.hpp#L408-L430

As far as I can tell, the `checkedEdges` array is not initialized before the first time that is accessed, so this is a legitimate concern.

bmhan12:
This PR resolves #704.

### DIFF
--- a/src/axom/primal/geometry/Polyhedron.hpp
+++ b/src/axom/primal/geometry/Polyhedron.hpp
@@ -405,7 +405,7 @@ private:
     axom::int8 checkedSize = 0;
     axom::int8 facesAdded = 0;
     // # edges * (# vertices per edge) * (# orientation per edge)
-    axom::int8 checkedEdges[MAX_VERTS * 2 * 2];
+    axom::int8 checkedEdges[MAX_VERTS * 2 * 2] = {0};
 
     // Check each vertex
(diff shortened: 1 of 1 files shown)


white238 (approved):

src/axom/primal/geometry/Polyhedron.hpp
```diff
@@ -405,7 +405,7 @@ class Polyhedron
     axom::int8 checkedSize = 0;
     axom::int8 facesAdded = 0;
     // # edges * (# vertices per edge) * (# orientation per edge)
-    axom::int8 checkedEdges[MAX_VERTS * 2 * 2];
+    axom::int8 checkedEdges[MAX_VERTS * 2 * 2] = {0};
```
kennyweiss, line 408: 👍

kennyweiss (approved):
Thanks @bmhan12 !

PR #709, opened 2021-11-09
Initialize memory in axom::Array by default, allow for creation of ArrayView<const T>

joshessman-llnl:
# Summary

- Initializes memory by default to provide expected behavior for primitive types.
- Includes an option to leave memory uninitialized via a constructor argument
- Moves `push_back` from `ArrayBase` to `Array` so it shows up in Doxygen
- Adds support for `const Array<T>` to `ArrayView<const T>` per https://github.com/LLNL/axom/issues/633#issuecomment-964546470 including changes to `operator==` for const-agnostic comparisons

See https://github.com/LLNL/axom/pull/677#discussion_r740001979


src/axom/core/ArrayBase.hpp
```diff
@@ -529,6 +482,33 @@ struct all_types_are_integral
...
+ * The final parameter is intended to be used via tag dispatch with std::is_default_constructible
+ */
+template <typename T>
+void initializeInPlace(T* data,
+                       const IndexType current_num_elements,
+                       const IndexType new_num_elements,
+                       std::true_type)
+{
+  int new_elems = new_num_elements - current_num_elements;
+  for(int ielem = 0; ielem < new_elems; ielem++)
```
kanye-quest, line 501: Should this loop be a GPU kernel if the allocator is non-unified GPU memory?

src/axom/core/Array.hpp
```diff
@@ -1004,6 +1097,14 @@ inline void Array<T, DIM, SPACE>::updateNumElements(IndexType new_num_elements)
 {
   assert(new_num_elements >= 0);
   assert(new_num_elements <= m_capacity);
+  if(m_default_construct)
+  {
+    detail::initializeInPlace(m_data,
```
kanye-quest, line 1102: It seems `std::vector<T>` instantiated with a non-default-constructible type doesn't actually support resizing or construction with initial sizes (i.e. you have to reserve capacity and/or push_back): https://godbolt.org/z/T1zYhq54h

We could probably static_assert on resizes and construction-with-size in these cases. (That can probably be a future discussion though)

kanye-quest (approved):
Thanks @joshessman-llnl!

src/axom/core/Array.hpp
```diff
@@ -426,6 +441,35 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM, SPACE>>
   template <typename... Args>
   ArrayIterator emplace(ArrayIterator pos, Args&&... args);
 
+  /*!
+   * \brief Push a value to the back of the array.
+   *
+   * \param [in] value the value to be added to the back.
+   *
+   * \note Reallocation is done if the new size will exceed the capacity.
```
kennyweiss, line 449: Please add a note about the `DIM==1` restriction to the doxygen for these three methods, 
(perhaps using `\pre` (?))

src/axom/core/Array.hpp
```diff
@@ -926,20 +1016,23 @@ template <typename T, int DIM, MemorySpace SPACE>
...
   m_num_elements = other.m_num_elements;
   m_capacity = other.m_capacity;
   m_resize_ratio = other.m_resize_ratio;
+  m_default_construct = other.m_default_construct;
 
   other.m_data = temp_data;
   other.m_num_elements = temp_num_elements;
   other.m_capacity = temp_capacity;
   other.m_resize_ratio = temp_resize_ratio;
+  other.m_default_construct = temp_default_construct;
```
kennyweiss, line 1035: Minor: Any reason not to implement these in terms of `axom::swap()` ?

It should be effectively the same, but might read a bit easier?

See: https://github.com/LLNL/axom/blob/f18c0decad8ea1d918acb50aba9a1a11bb0bcf4b/src/axom/core/utilities/Utilities.hpp#L74-L86

src/axom/core/Array.hpp
kennyweiss: Should we create an issue to track this?
(or add it to #633 ?)

src/axom/core/tests/core_array.hpp
```diff
@@ -1583,4 +1583,76 @@ TEST(core_array, checkDevice2D)
 #endif
 }
 
+struct HasDefault
+{
+  int member = 255;
+};
+
```
kennyweiss, line 1590: It might be good to add another test for the basic case, e.g. `Array<int> arr(10)` that tests that the array is default-initialized to `0`s.

kennyweiss (approved):
Thanks @joshessman-llnl

src/axom/core/Array.hpp
joshessman-llnl: This is a good point - I think I had done it like this to stay consistent but there shouldn't be any reason we can't just swap all the members with axom::swap

src/axom/core/ArrayBase.hpp
joshessman-llnl: Good catch

PR #712, opened 2021-11-10
Use a parallel mfem in Axom's spack specs

### LINKED ISSUE #680: Use parallel mfem for Axom's TPL
Axom currently uses a serial configuration of mfem. 
We should be using a parallel configuration for all/most of these.

kennyweiss:
# Summary

- This PR uses a parallel in our spack specs
- It resolves #680
- Parallel mfem depends on `hypre` and `metis`
- I disabled mfem for our `gcc@8_no_fortran` spec on `toss3` since `hypre` seemed to require a Fortran compiler
- I also disabled mfem for our `gcc@7.3.1` spec on `blueos` due to a linking problem with `blas`/`lapack` in that configuration
- This PR also enables `AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION` by default (it was previously `OFF` by default)

#### TODO
- [x] Run the `build_tpl` script and generate host-configs on the RZ platforms
- [x] Run the `build_tpl` script and generate host-configs on the CZ platforms
- [x] Remove `klee` host-configs
- [x] Update docker host-configs


scripts/spack/configs/docker/ubuntu18/packages.yaml
```diff
@@ -101,8 +101,11 @@ packages:
   hdf5:
     variants: ~shared~mpi
   # do shared mfem to allow PIC flag in mfem
+  hypre:
+    version: [2.20.0]
```
agcapps, line 105: I think the comment should stay on top of the mfem section:

```suggestion
  hypre:
    version: [2.20.0]
  # do shared mfem to allow PIC flag in mfem
```

agcapps (approved):
Looks good, except for a line rearrangement to keep a comment with its relevant section.

src/cmake/AxomOptions.cmake
```diff
@@ -10,7 +10,7 @@ option(AXOM_ENABLE_ANNOTATIONS "Enables code annotations to facilitate performan
 option(AXOM_ENABLE_SPARSEHASH "Enables Sparsehash." ON)
 option(AXOM_ENABLE_ALL_COMPONENTS "Enables all components by default" ON)
 option(AXOM_USE_64BIT_INDEXTYPE "Use 64-bit integers for axom::IndexType" OFF)
-option(AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION "Enable Axom's version of the MFEM SidreDataCollection" OFF)
+option(AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION "Enable Axom's version of the MFEM SidreDataCollection" ON)
```
kennyweiss, line 13: I'd like to explicitly call out this change -- now that we have parallel mfem, I'm enabling `AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION` by default.
If anyone is building mfem w/ `MFEM_USE_SIDRE`, they'll want/need to turn this option off.

src/axom/quest/IntersectionShaper.hpp
```diff
@@ -33,8 +33,8 @@
 
 #include "mfem.hpp"
 
-#include "fmt/fmt.hpp"
-#include "fmt/locale.h"
+#include "axom/fmt.hpp"
```
white238, line 36: =/ my bad

white238 (approved):

src/axom/quest/IntersectionShaper.hpp
kennyweiss: I think it was merged after yours and there was an invalid guard in CMake preventing these files from being included

PR #717, opened 2021-11-13
fix some spack package build options

white238:
Found these while doing other builds. They have been upstreamed to Spack but cut a lot of time off the builds and don't want to forget next time we do a TPL build.

These are not worth rebuilding TPLs now.


kennyweiss (approved):
Thanks for pushing these back to Axom, @white238 

I was looking into this earlier this week in anticipation of installing TPLs to `/collab/usr`.
We might also want to disable applications/tools/binaries for `conduit` and/or `hdf5` to further reduce build time and installation sizes.

This could definitely be done/attempted in a separate/future PR.

agcapps (approved):

Issue #724, opened 2021-11-18
Prevent `axom::Array` from being used within a kernel


kennyweiss:
The new axom::Array class is supported on the GPU via the `axom::ArrayView` wrapper class.

Since our kernel's have pass-by-value semantics, `axom::Array` should not be used directly within them, as this will incur an allocation+copy. Furthermore, if one is writing to the array, the results will not transfer back from this copy to the original array.

Since the direct `axom::Array` usage will currently compile, it is not always obvious that one is using the incorrect class.
We should explore adding a safeguard to the code to prevent users from directly using an `axom::Array` within kernels. Ideally via a `static_assert`, to catch this at compile time.

It is worth noting that this issue arises for CPU-based kernels, and not just GPU-based kernels.

PR #725, opened 2021-11-18
Adds `axom::Array::view()` method and improves testing with kernels

kennyweiss:
# Summary

- This PR improves testing of `axom::Array` and `axom::ArrayView` when called through `axom::for_all` kernels
- It adds a `view()` method to `axom::Array` to make it easier to get an `ArrayView` corresponding to an `Array`
- It highlights the need to mark lambdas as `mutable` if you want to modify the `ArrayView`'s data
- I also added a new macro -- `AXOM_TYPED_TEST` -- to allow using gtest's `TYPED_TESTS` with device code. The only difference was that it exposes `TestBody()` as a public method instead of a private method 

Thanks @joshessman-llnl  for a helpful discussion.
I noticed that there are potentially difficult-to-diagnose problems that can arise when (accidentally) using an `axom::Array` within a kernel. I attempted to resolve this, but after a few false turns created #724 to track this for a future PR.

### DIFF
--- a/src/axom/core/Array.hpp
+++ b/src/axom/core/Array.hpp
@@ -6,16 +6,17 @@
 #ifndef AXOM_ARRAY_HPP_
 #define AXOM_ARRAY_HPP_
 
-#include "axom/config.hpp"                    // for compile-time defines
-#include "axom/core/Macros.hpp"               // for axom macros
-#include "axom/core/utilities/Utilities.hpp"  // for processAbort()
-#include "axom/core/Types.hpp"                // for IndexType definition
+#include "axom/config.hpp"
+#include "axom/core/Macros.hpp"
+#include "axom/core/utilities/Utilities.hpp"
+#include "axom/core/Types.hpp"
 #include "axom/core/ArrayBase.hpp"
 #include "axom/core/ArrayIteratorBase.hpp"
+#include "axom/core/ArrayView.hpp"
 
 // C/C++ includes
-#include <algorithm>  // for std::transform
-#include <iostream>   // for std::cerr and std::ostream
+#include <algorithm>
+#include <iostream>
 
 namespace axom
 {
@@ -73,6 +74,8 @@ public:
   using value_type = T;
   static constexpr MemorySpace space = SPACE;
   using ArrayIterator = ArrayIteratorBase<Array<T, DIM, SPACE>>;
+
+  using ArrayViewType = ArrayView<T, DIM, SPACE>;
 
 public:
   /// \name Native Storage Array Constructors
@@ -515,6 +518,18 @@ public:
    * \brief Get the ID for the umpire allocator
    */
   int getAllocatorID() const { return m_allocator_id; }
+
+  /*!
+   * \brief Returns a view of the array
+   * \sa ArrayView
+   */
+  const ArrayViewType view() const { return ArrayViewType(*this); }
+
+  /*!
+   * \brief Returns a view of the array
+   * \sa ArrayView
+   */
+  ArrayViewType view() { return ArrayViewType(*this); }
 
   /// @}
 
--- a/src/axom/core/Array.hpp
+++ b/src/axom/core/Array.hpp
@@ -523,12 +523,6 @@ public:
    * \brief Returns a view of the array
    * \sa ArrayView
    */
-  const ArrayViewType view() const { return ArrayViewType(*this); }
-
-  /*!
-   * \brief Returns a view of the array
-   * \sa ArrayView
-   */
   ArrayViewType view() { return ArrayViewType(*this); }
 
   /// @}
--- a/src/axom/core/Macros.hpp
+++ b/src/axom/core/Macros.hpp
@@ -258,7 +258,7 @@
 /*!
  * \def AXOM_TYPED_TEST(CaseName, TestName)
  * \brief Minor tweak of gtest's TYPED_TEST macro to work with device code
- * \note Can be used after including `gtest/gtest.h`
+ * \note Can be used in test files after including `gtest/gtest.h`
  */
 // Specifically, we expose the `TestBody()` method as public instead of private
 #define AXOM_TYPED_TEST(CaseName, TestName)                                         \
--- a/src/axom/core/Macros.hpp
+++ b/src/axom/core/Macros.hpp
@@ -4,11 +4,9 @@
 // SPDX-License-Identifier: (BSD-3-Clause)
 
 /*!
- *
  * \file AxomMacros.hpp
  *
  * \brief Contains several useful macros for the axom project
- *
  */
 
 #ifndef AXOM_MACROS_HPP_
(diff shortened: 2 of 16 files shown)


src/axom/core/Macros.hpp
```diff
@@ -257,4 +255,38 @@
...
+ */
+// Specifically, we expose the `TestBody()` method as public instead of private
+#define AXOM_TYPED_TEST(CaseName, TestName)                                         \
+  static_assert(sizeof(GTEST_STRINGIFY_(TestName)) > 1,                             \
+                "test-name must not be empty");                                     \
+  template <typename gtest_TypeParam_>                                              \
+  class GTEST_TEST_CLASS_NAME_(CaseName, TestName)                                  \
+    : public CaseName<gtest_TypeParam_>                                             \
+  {                                                                                 \
+  public:                                                                           \
```
kennyweiss, line 271: This macro is a copy of the `TYPED_TEST` macro from gtest.
The only difference was changing this line from `private' to `public` to allow it to work with CUDA device kernels.

Without this change, I got the following error:
> The enclosing parent function ("TestBody") for an extended __host__ __device__ lambda cannot have private or protected access within its class

src/axom/core/tests/core_array_for_all.hpp
```diff
@@ -0,0 +1,156 @@
...
+    axom::synchronize<ExecSpace>();
+  }
+
+  // Check array contents on device
+  HostArray localArr = arr;
+  for(int i = 0; i < N; ++i)
+  {
+    EXPECT_EQ(localArr[i], N - i);
+  }
+}
```
kennyweiss, line 91: This test runs with all available execution spaces.
I think we'll want to modify other axom tests to use this pattern.

joshessman-llnl (approved):
Thanks!

src/axom/core/Macros.hpp
kennyweiss: Thanks @joshessman-llnl -- I thought about  both of these.
We don't currently have another place to put test-specific utility code. 
If I placed it somewhere else, we'd need to incorporate that into the build system, so I figured this was the easiest place for now.

kennyweiss: I also thought about patching this line in blt's gtest, but that would affect all blt users, which seemed like a bigger decision.

src/axom/core/Array.hpp
```diff
@@ -516,6 +519,18 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM, SPACE>>
    */
   int getAllocatorID() const { return m_allocator_id; }
 
+  /*!
+   * \brief Returns a view of the array
+   * \sa ArrayView
+   */
+  const ArrayViewType view() const { return ArrayViewType(*this); }
```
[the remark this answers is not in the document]
kennyweiss: Should I remove this function for now and let you re-add it in #709?  or leave it as is for now?

src/axom/core/examples/core_containers.cpp
```diff
@@ -219,12 +219,17 @@ void demoArrayDevice()
   // _basic_array_device_implicit_end
 
   // _basic_array_device_explicit_start
-  axom::ArrayView<int, 1, axom::MemorySpace::Device> view_of_dynamic_array(
-    device_array_dynamic);
+  using DeviceArrayView = axom::ArrayView<int, 1, axom::MemorySpace::Device>;
```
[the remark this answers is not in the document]
kennyweiss: In general, I agree that it would be better to use the direct types, but the docs seemed somewhat difficult to read due to the long lines. I opted for a compromise with a type alias inside the snippet.

For reference, here's what it currently looks like in develop (before the change):
image

src/axom/core/Array.hpp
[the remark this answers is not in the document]
joshessman-llnl: I think it would be reasonable to remove `view() const` until we can have it do "the right thing"

joshessman-llnl (approved):

src/axom/core/docs/sphinx/core_containers.rst
```diff
@@ -241,6 +241,8 @@ via a lambda:
    :end-before: _array_w_raja_end
    :language: C++
 
+.. note:: We need to mark the lambda as `mutable` if we want to modify the array (array view) data.
```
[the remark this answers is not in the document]
agcapps: Yes, much appreciated.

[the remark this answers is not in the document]
kanye-quest: I'm guessing this is because `ArrayView<T>` is copied over as const in a lambda capture? We could probably consider just having a single `T& operator[] const` function in ArrayView - const on an ArrayView<T> should logically apply to the View itself, not the underlying data.

kanye-quest (approved):
Neat!

agcapps (approved):
Thank you, Kenny.

white238 (approved):

src/axom/core/tests/CMakeLists.txt
```diff
@@ -4,19 +4,25 @@
...
-set(gtest_utils_tests
+set(core_serial_tests
     core_array.hpp
+    core_array_for_all.hpp
     core_execution_for_all.hpp
     core_execution_space.hpp
     core_memory_management.hpp
     core_Path.hpp
     core_stack_array.hpp
+    core_map.hpp
```
kennyweiss, line 25: Note: The `core_map` tests were added to the `core_serial_tests.cpp` executable, but missing from the CMake file. As such, they were being compiled, but not run as part of our unit tests.

After I added this line, it turned out that `core_map` was broken in our Windows/MSVC build. I found and fixed the following two problems:

1. MSVC was not applying the Named Return Value Optimization (nrvo) to `axom::Map`, causing a segfault when it attempted to delete memory twice. I added a move constructor and assignment operator to `Map` as well as some tests to ensure that we can successfully return a Map from a function (as well as supporting `Node`, `Pair` and `Bucket` classes).
2. MSVC was using a different function for `std::hash<int>` than our other compilers.  Our test relied on the fact that the latter were using the identity hash, which implies that if you have `N` integer slots, you can always insert all keys from `0` to `N-1` without rehashing.  This was not the case with MSVC's default hash function.  I refactored the core_map tests to use a custom map with the identify hasher.  I think we'll want to generalize these tests in the future.

src/axom/core/tests/core_array.hpp
```diff
@@ -983,14 +983,13 @@ TEST(core_array, checkResize)
...
 {
-  const char IGNORE_OUTPUT[] = ".*";
   constexpr IndexType ZERO = 0;
   IndexType size = 100;
 
   /* Resizing isn't allowed with a ratio less than 1.0. */
   Array<int> v_int(ZERO, size);
   v_int.setResizeRatio(0.99);
-  EXPECT_DEATH_IF_SUPPORTED(internal::check_resize(v_int), IGNORE_OUTPUT);
+  EXPECT_DEATH_IF_SUPPORTED(internal::check_resize(v_int), "");
```
kennyweiss, line 992: The MSVC debugger didn't the like the initial `IGNORE_OUTPUT` version of the gtest death test, but seemed to be ok with `""` as the second parameter.

PR #738, opened 2021-12-01
Consolidate bit-twiddle functions

### LINKED ISSUE #583: Centralize bit-twiddling functions
From #577: Spin has a `clz` function, which we could combine with the functions in:
https://github.com/LLNL/axom/blob/develop/src/axom/slam/internal/BitTwiddle.hpp

These could potentially be moved into core.

kanye-quest:
# Summary

- Resolves #583 
- Adds call to `__clz()` CUDA intrinsic for `axom::utilities::leadingZeros()` function

### DIFF
--- a/src/axom/core/utilities/BitUtilities.hpp
+++ b/src/axom/core/utilities/BitUtilities.hpp
@@ -120,6 +120,48 @@ AXOM_HOST_DEVICE inline int popCount(axom::uint64 word)
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


kennyweiss (approved):
Thanks @kanye-quest !

rhornung67 (approved):
Looks good. Thanks @kanye-quest

agcapps (approved):
Thanks, @kanye-quest .

PR #741, opened 2021-12-03
Use user-provided allocator ID in quest::PointInCell

### LINKED ISSUE #729: Allocator support in PointInCell
Currently, `quest::PointInCell` takes an allocator ID argument that may ostensibly be used to allocate temporary arrays. However, in practice we just allocate most of our intermediate arrays in unified memory if we're running on the GPU, since the MFEM part of the calculation has to run on the CPU.

We should replace these unified memory usages with arrays that use the allocator ID, then explicitly copy them to host-side memory if needed.

kanye-quest:
# Summary

- Resolves #729
- Replaces unified memory allocations for candidate search results with arrays allocated with the user-provided allocator ID
- Adds an overload to `axom::Array` which allows copying data from a `axom::ArrayView<const T>`
- Adds allocator ID argument to `axom::Array` copy constructors from other array spaces


agcapps (approved):
Thank you!

src/axom/quest/detail/PointFinder.hpp
```diff
@@ -208,28 +202,51 @@ class PointFinder
...
       }
+      outCellIdsPtr = outCellIdsHost;
+      outIsoparPtr = outIsoparHost;
+    }
+    else
+    {
+      ptsHostPtr = pts;
+      candidatesHostPtr = candidates;
+      offsetsHostPtr = offsets;
+      countsHostPtr = counts;
```
kennyweiss, line 249: Are these copies necessary when `!DeviceExec`?

kennyweiss (approved):
Thanks @kanye-quest!

Are there any tests that should be added/updated for this new capability?

src/axom/quest/detail/PointFinder.hpp
kanye-quest: In this case, these assignments are actually for creating host-side `ArrayView`s directly from the original `candidates/offsets/counts` arrays. In the device-side case a little further up, we do copy these arrays to an intermediate host-side array first, before initializing the views from the intermediate arrays.

kennyweiss: Thanks -- it's a bit difficult to see which ones are shallow `ArrayView` copies and which ones are deep `Array` copies. 
(I suspect it would be easier if you know to look for it :) )

kanye-quest:
I've modified the point-in-cell tests to run for CUDA and OpenMP as well. There were some more modifications I had to make to the `PointInCell` and `Array` classes for memory correctness on the device, in case you wanted to take another look @kennyweiss @agcapps

src/axom/core/Array.hpp
```diff
@@ -792,9 +799,27 @@ Array<T, DIM, SPACE>::Array(Array&& other)
...
-Array<T, DIM, SPACE>::Array(const ArrayBase<T, DIM, OtherArrayType>& other)
+Array<T, DIM, SPACE>::Array(const ArrayBase<T, DIM, OtherArrayType>& other,
+                            int allocatorId)
   : ArrayBase<T, DIM, Array<T, DIM, SPACE>>(other)
-  , m_allocator_id(axom::detail::getAllocatorID<SPACE>())
+  , m_allocator_id(allocatorId)
+{
+  initialize(static_cast<const OtherArrayType&>(other).size(),
+             static_cast<const OtherArrayType&>(other).size());
+  // axom::copy is aware of pointers registered in Umpire, so this will handle
```
kennyweiss, line 809: Minor:  This comment might be a little misleading since `axom::copy` also works in configs that don't have Umpire.

src/axom/quest/tests/quest_point_in_cell_mfem.cpp
```diff
@@ -74,10 +74,36 @@ enum MeshType
...
+template <int BLK_SZ>
+struct ExecTraits<axom::CUDA_EXEC<BLK_SZ>>
+{
+  static int getAllocatorId()
+  {
+    return axom::getUmpireResourceAllocatorID(
+      umpire::resource::MemoryResourceType::Device);
+  }
+};
+#endif
```
kennyweiss, line 101: Minor: Is this something that should be built into axom? Or something we'd expect users to have to write?

src/axom/quest/tests/quest_point_in_cell_mfem.cpp
```diff
@@ -931,12 +975,25 @@ class ExpectedValue
...
+  axom::OMP_EXEC,
+#endif
+#ifdef AXOM_USE_CUDA
+  axom::CUDA_EXEC<256>,
+#endif
+  axom::SEQ_EXEC>;
+
+TYPED_TEST_SUITE(PointInCell2DTest, ExecTypes);
+TYPED_TEST_SUITE(PointInCell3DTest, ExecTypes);
+
```
kennyweiss, line 989: This is great -- thanks for adding support for other exec spaces @kanye-quest !

kennyweiss (approved):
Thanks @kanye-quest !

src/axom/quest/tests/quest_point_in_cell_mfem.cpp
kanye-quest: Normally, we could use `axom::execution_space<ExecSpace>::allocatorID()`, but since the default for that is unified memory for `ExecSpace == CUDA_EXEC`, I have it explicitly coded for device-only memory so we can catch improper memory accesses.

My impression of the user case is that if they aren't specifying an allocator ID in `PointInCell`, they're probably fine with our default memory space.

Issue #739, opened 2021-12-02
Improve consistency in primal's `orientation` operation


kennyweiss:
There are a few inconsistencies in our 2D and 3D treatment of orientation.

* Our 2D orientation primitive treats clockwise orientation as positive, while our 3D orientation primitive treats counter-clockwise orientations as positive.
* Neither 2D or 3D expose their fuzzy tolerance parameter to the user. 2D's fuzzy tolerance is `0`, while 3D's is `1e-9`

#### Proposed resolution
I suggest that:
* both 2D and 3D should treat counter-clockwise as positive
* both should expose their fuzzy tolerance variable as an optional parameter with a default that matches other parts of `primal`, e.g. `1e-8`

#### Reference:
* 3D `orientation(Point, Triangle)`
https://github.com/LLNL/axom/blob/033414256877a1f404a6129c1a2b08ef85dff666/src/axom/primal/operators/orientation.hpp#L31-L75
* 2D `orientation(Point, Segement)`
https://github.com/LLNL/axom/blob/033414256877a1f404a6129c1a2b08ef85dff666/src/axom/primal/operators/orientation.hpp#L77-L124

PR #742, opened 2021-12-07
Adds a Delaunay triangulation capability to Quest

kennyweiss:
# Summary
- This PR is a feature
- It adds a new Delaunay triangulation capability to quest in 2D and 3D.
The triangulation is built incrementally by inserting successive points and updating to mesh to ensure it satisfies the Delaunay condition: No vertices are in the circumsphere of any triangle/tetrahedron in the mesh. For each inserted point, we remove the triangles/tetrahedra in its "Delaunay cavity" and fill this hole with a new set of triangles/tetrahedra containing the new point (the "Delaunay ball").
- The underlying triangulation is encoded using the _**I**ndexed mesh data structure with **A**djacencies (IA)_, which is implemented using `slam`'s dynamic sets, relations and maps.

- This PR contains several other optimizations and improvements throughout axom in support of this new capability, primarily in `core`, `slam` and `primal`.
    - The main new additions are `quest::Delaunay.hpp` and slam's `IAMesh` class as well as optimizations and improvements to some geometric primitives in `primal`, such as `Triangle`, `Tetrahedron` and `Sphere`.
    - There is also a driver example `delaunay_triangulation` which generates a triangle or tetrahedral mesh from a specified number of random points within a rectilinear bounding box.

**Credits:** This capability was initially implemented by @raineyeh during Summer 2017 and has been sitting in a branch.

**Update:** Here's some rough performance data on the insertion rate (in vertices per second) while refactoring and improving the code:

| Delaunay | 2D     |            |               |                  |      3D |            |               |                 
|   ---         |  ---    |   ---     |  ---         |  ---           |  ---     |  ---      |  ---        |    ---         
|               | 1,000 | 10,000 | 100,000 | 1,000,000 | 1,000 | 10,000 | 100,000 | 1,000,000
| Initial sample        (after converting driver to CLI on 12/10/2021) | 8K | 1.5K |  .15K  | -  | 0.4K | .2K |  - | -   
| _<lots of optimizations  that I didn't measure>_ |   |   |   |   |   |   | 
| Find element using   most recently introduced element (~12/24) | 160K | 94K | 43K | 11K  | 25K | 23K | 21K | 13.5K 
| Find element using   small regular grid of vertex indices (~12/26) | 130K | 158K | 152K | 115K | 29K | 29K | 27.5 K | 23.5K 
| Use StackArray in fixVertexNeighborhood() instead of std::vector (~12/27) | 150K | 177K | 168K | 126K | 30K | 32K | 30K | 25K 
| Use vector instead   of map to fix star neighborhood (~12/28) | 165K | 190K | 180K | 134K | 33K | 31K | 32K | 27.5K 
| Refactor   fixVertexNeighborhood and findCavity() to use slam relation data without extra copies | 193K | 255K | 230K | 165K |  50K | 48K | 49K | 38.5K 
| Final (for this PR) | 180K | 250K | 233K | 170K | 50K | 53K | 47K | 38.5K 

I.e. the performance improved by ~200x.  
There are still a few opportunities for optimization:
* For finding the initial triangle/tetrahedron: I'm currently using a small grid for the start vertex. Depending on the parameters, this could still involve a lot of mesh traversal depending on the ratio of vertices per grid cell. An octree on the vertices might improve performance at very little additional overhead
* For the point insertion: I'm inserting the points in random order. Spatially sorting the vertices could improve the coherence and reduce the average number of invalidated elements per vertex insertion

**Update:** The following shows the insertion of 100 random points within the unit square:
delaunay_triangulation_100pts


src/axom/core/tests/core_bit_utilities.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+    // Explicitly find first bit and check that it matches
+    int bit = 0;
+    for(; bit < BITS; ++bit)
+    {
+      if(rand_val & shifted(BITS - bit - 1)) break;
+    }
+    EXPECT_EQ(bit, axom::utilities::leadingZeros(rand_val));
+  }
+}
```
white238, line 173:
```suggestion
}

```

src/axom/core/tests/utils_utilities.hpp
```diff
@@ -128,3 +128,33 @@ TEST(utils_Utilities, minmax)
...
+    EXPECT_EQ(5.0, axom::utilities::floor(val));
+    EXPECT_EQ(5.0, axom::utilities::ceil(val));
+  }
+
+  {
+    double val = -5.0;
+    EXPECT_EQ(-5.0, axom::utilities::floor(val));
+    EXPECT_EQ(-5.0, axom::utilities::ceil(val));
+  }
+}
```
white238, line 160:
```suggestion
}

```

src/axom/core/utilities/BitUtilities.hpp
```diff
@@ -132,43 +155,35 @@ AXOM_HOST_DEVICE inline int popCount(axom::uint64 word)
...
 #endif
+  /* clang-format off */
 }
 
 }  // namespace utilities
 }  // namespace axom
 
+
+// Undefine intrinsic defines from the top of the file
+#undef _AXOM_CORE_USE_INTRINSICS_MSVC
```
white238, line 185: 👍

src/axom/primal/tests/primal_in_sphere.cpp
```diff
@@ -6,104 +6,188 @@
...
-    PointType q2 = PointType::make_point(-1, 1, 1);
+    PointType q2 {-1, 1, 1};
     EXPECT_FALSE(in_sphere(q2, p0, p1, p2, p3));
+    EXPECT_FALSE(in_sphere(q2, tet));
   }
 
   // Test some points that are outside the circumsphere
   {
-    PointType q1 = PointType::make_point(1.1, 1, 1);
+    PointType q1 {1.1, 1, 1};
```
white238, line 107: This is much nicer!

kennyweiss:
Note: This PR got pretty big as I was optimizing the Delaunay triangulation, which depended on large parts of `core`, `slam`, `primal` and `spin`.  I'll add some comments to highlight the key parts.

src/axom/core/Array.hpp
```diff
@@ -136,13 +140,32 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM, SPACE>>
...
 
+  /// \overload
+  template <IndexType SFINAE_DIM = DIM,
+            MemorySpace SFINAE_SPACE = SPACE,
+            typename std::enable_if<SFINAE_DIM == 1 &&
+                                    SFINAE_SPACE == MemorySpace::Dynamic>::type* = nullptr>
+  Array(ArrayOptions::Uninitialized,
+        IndexType num_elements,
+        IndexType capacity = 0,
+        int allocator_id = axom::detail::getAllocatorID<SPACE>());
```
kennyweiss, line 151: Note: `axom::Array` supports two features that can only be enabled through its constructors -- uninitialized data (via the `ArrayOptions::Uninitialized` tag) and a custom allocator (via the `allocatorID` argument) -- so I needed to add some extra overloads to support this.

I also improved the testing of this feature and ensured that when you copy and array with these features, the copy gets them as well.

src/axom/core/tests/core_bit_utilities.hpp
```diff
@@ -0,0 +1,173 @@
+// Copyright (c) 2017-2021, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level LICENSE file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include "gtest/gtest.h"
```
kennyweiss, line 6: We didn't have any tests for the bit utility functions: `popCount()`, `trailingZeros()` and `leadingZeros()`, so I added new tests before making changes.

src/axom/core/tests/utils_utilities.hpp
```diff
@@ -36,7 +36,7 @@ void check_random_real(int offset)
 //------------------------------------------------------------------------------
 // UNIT TESTS
 //------------------------------------------------------------------------------
-TEST(utils_Utilities, log2)
+TEST(utils_utilities, log2)
```
kennyweiss, line 39: These tests weren't being run because the name didn't match the file name and we were filtering them out.

src/axom/core/utilities/BitUtilities.hpp
```diff
@@ -14,9 +14,21 @@
 #ifndef AXOM_BIT_UTILITIES_HPP
 #define AXOM_BIT_UTILITIES_HPP
 
-#include "axom/core/Macros.hpp"  // for AXOM_DEVICE_CODE
+#include "axom/config.hpp"
+#include "axom/core/Macros.hpp"
 #include "axom/core/Types.hpp"
```
kennyweiss, line 20: I added compiler intrinsics for our CPU-based compilers, which are used extensively in `slam::BitSet`
In some runs of the Delaunay triangulation, `popCount()` was taking >50% of the runtime.

Note: the underlying optimized assembly is only enabled when the code is built in an architecture-aware manner, e.g. via the `-march-native` flag. I'm planning to add a ticket for us to enable this, e.g. via a new config option

src/axom/primal/examples/primal_introduction.cpp
```diff
@@ -71,21 +71,20 @@
...
 // _using_start
 // "using" directives to simplify code
-using namespace axom;
-using namespace primal;
+namespace primal = axom::primal;
 
 // almost all our examples are in 3D
 constexpr int in3D = 3;
 
 // primitives represented by doubles in 3D
```
kennyweiss, line 79: While I was editing the primal tests, I modernized them a bit, e.g. type aliases via `using` instead of `typedef` and using the initializer-list constructors for primal's primitives instead of `make_point`, `make_vector`, ....

src/axom/primal/geometry/Sphere.hpp
```diff
@@ -6,14 +6,16 @@
 #ifndef AXOM_PRIMAL_SPHERE_HPP_
 #define AXOM_PRIMAL_SPHERE_HPP_
```
kennyweiss, line 8: The changes to `primal::Sphere` change its API and implementation to be based on primal primitives -- e.g. it is based on `primal::Point` (which always has the right dimensions) instead of a pointer to an array.

src/axom/primal/geometry/Vector.hpp
```diff
@@ -63,6 +63,16 @@ template <typename T, int NDIMS>
 AXOM_HOST_DEVICE Point<T, NDIMS> operator+(const Vector<T, NDIMS>& V,
                                            const Point<T, NDIMS>& P);
 
+/*!
```
kennyweiss, line 66: These changes are related to https://github.com/LLNL/axom/issues/708
I think this completes the set.

src/axom/primal/operators/in_sphere.hpp
```diff
@@ -73,19 +97,39 @@ inline bool in_sphere(const Point<T, 3>& q,
                       const Point<T, 3>& p0,
                       const Point<T, 3>& p1,
                       const Point<T, 3>& p2,
-                      const Point<T, 3>& p3)
+                      const Point<T, 3>& p3,
+                      double EPS = 1e-8)
```
kennyweiss, line 101: This change significantly improves the efficiency of `in_sphere()` in 3D by converting it from a 5x5 determinant to a 4x4 determinant.

We have an explicit implementation of 4x4 determinants, but 5x5 determinants are based on a general LU decomposition.

src/axom/primal/tests/primal_in_sphere.cpp
```diff
@@ -6,104 +6,188 @@
...
+                        tet.signedVolume(),
+                        sphere.computeSignedDistance(query)));
+
+    // check that each vertex is on the sphere boundary
+    for(int i = 0; i <= DIM; ++i)
+    {
+      EXPECT_TRUE(sphere.getOrientation(tet[i]) == primal::ON_BOUNDARY);
+    }
+
+    // check that the orientation matches expectations
```
kennyweiss, line 170: These tests ensure that we get the same result for `in_sphere()` as we do when we compute the circumsphere of a triangle/tet and then call `getOrientation()`

src/axom/primal/tests/primal_intersect.cpp
```diff
@@ -5,10 +5,7 @@
```
kennyweiss, line 5: The (numerous) changes to this file are almost entirely modernization

src/axom/quest/Delaunay.hpp
```diff
@@ -0,0 +1,885 @@
+
```
kennyweiss, line 1: This is the one of the key files in this PR: It enables building up a Delaunay triangulation by incrementally inserting points in the `insertPoint()` function.

src/axom/quest/Delaunay.hpp
```diff
@@ -0,0 +1,885 @@
...
+  /// \brief Get the IA mesh data pointer
+  const IAMeshType* getMeshData() const { return &m_mesh; }
+
+  /**
+   * \brief Checks that the underlying mesh is a valid Delaunay triangulation of the point set
+   *
+   * A Delaunay triangulation is valid when none of the vertices are inside the circumspheres 
+   * of any of the elements of the mesh
+   */
+  bool isValid(bool verboseOutput = false) const
```
kennyweiss, line 256: The validity check is using an entirely different algorithm to check for validity.
* Insert points uses the mesh connectivity to find and remove invalid elements using the (implicit) in_sphere() algorithm.
* isValid() builds up a spatial index over the (explicit) circumspheres of each element and then uses `Sphere::getOrientation()`

src/axom/quest/examples/delaunay_triangulation.cpp
```diff
@@ -0,0 +1,231 @@
+// Copyright (c) 2017-2021, Lawrence Livermore National Security, LLC and
```
kennyweiss, line 1: This is the driver application for the delaunay triangulation.

src/axom/slam/mesh_struct/IA.hpp
```diff
@@ -0,0 +1,402 @@
+// Copyright (c) 2017-2021, Lawrence Livermore National Security, LLC and
```
kennyweiss, line 1: Here is the other main addition: IA mesh is the topological mesh data structure over a triangle or tetrahedral mesh that the Delaunay class is built on. It is implemented in terms of slam's dynamic sets, relations and maps.

src/axom/core/Array.hpp
rhornung67: So the default behavior is to initialize the data and a user can choose to leave it uninitialized via an arg to the ctor?

src/axom/core/Array.hpp
```diff
@@ -667,15 +691,15 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM, SPACE>>
    */
   virtual void dynamicRealloc(IndexType new_num_elements);
 
-  T* m_data = nullptr;
+  T* m_data {nullptr};
```
rhornung67, line 694: Are changes like this going to introduce confusion in coding style? What is the benefit of doing it this way vs. '= val'?

src/axom/core/StackArray.hpp
```diff
@@ -108,6 +108,33 @@ AXOM_HOST_DEVICE bool operator!=(const StackArray<T, N>& lhs,
...
+AXOM_HOST_DEVICE bool operator<(const StackArray<T, N>& lhs,
+                                const StackArray<T, N>& rhs)
+{
+  for(int i = 0; i < N; ++i)
+  {
+    if(lhs[i] < rhs[i])
+    {
+      return true;
+    }
+    else if(lhs[i] > rhs[i])
```
rhornung67, line 130: I think it would be more efficient to change this to '>='. For example, if lhs == rsh is true on entry, then the entire array has to be traversed.

src/axom/core/utilities/BitUtilities.hpp
```diff
@@ -14,9 +14,21 @@
...
+// Check for and setup defines for platform-specific intrinsics
+#if defined(_WIN64) && (_MSC_VER >= 1600)
+  #define _AXOM_CORE_USE_INTRINSICS_MSVC
+  #include <intrin.h>
+#elif defined(__x86_64__) && defined(__GNUC__)
+  #define _AXOM_CORE_USE_INTRINSICS_GCC
+  #include <x86intrin.h>
+#elif defined(__powerpc64__) && (defined(__GNUC__) || defined(__ibmxl__))
+  #define _AXOM_CORE_USE_INTRINSICS_PPC
+#endif
```
rhornung67, line 30: we will need to add a case for AMD CPUs. please make an issue for this so we don't forget.

src/axom/primal/geometry/Tetrahedron.hpp
```diff
@@ -171,18 +187,76 @@ class Tetrahedron
    * \brief Returns the signed volume of the tetrahedron
    * \sa volume()
    */
+  AXOM_HOST_DEVICE
   double signedVolume() const
   {
-    const double scale = 1. / 6.;
+    constexpr double scale = 1. / 6.;
```
rhornung67, line 193: is this called alot in device code when it is used? since division is expensive, it may be better to collect a set of numerical constants in core; etc. one_sixth = 0.16666666666666666666666666.... and use that in places like this.

src/axom/primal/tests/primal_orientation.cpp
```diff
@@ -5,83 +5,226 @@
...
+                  primal::orientation(phys, flipped_triangle));
+        EXPECT_EQ(primal::ON_POSITIVE_SIDE,
+                  primal::orientation(phys - normal, flipped_triangle));
+        EXPECT_EQ(primal::ON_NEGATIVE_SIDE,
+                  primal::orientation(phys + normal, flipped_triangle));
+      }
+
+      // check overload with explicit tolerances
+      {
+        const double TOL = 1e-2;
```
rhornung67, line 101: make constexpr?

src/axom/core/Array.hpp
agcapps: strongly agree with Rich's comment

kennyweiss: I agree about consistency and am not aware of any semantic differences.
I'll revert this.

On a related note, perhaps we should run the code through clang-tidy (with an appropriate set of agreed upon rules).

rhornung67: especially for built-in types, I think it's less jarring with the common '=' syntax that everyone uses.

src/axom/core/StackArray.hpp
kennyweiss: At one point, this was a bottleneck in the Delaunay point insertion, so I played around with several iterations of this lexicographic ordering function. 

The implementation here appeared to be a good compromise between clarity and efficiency (the compiler can optimize this pretty well since `N` is known at compile time and typically pretty small).

Would it be ok to make this a separate issue and add benchmarks to compare several variants of a comparator for `StackArray::operator<()`?

src/axom/core/Array.hpp
rhornung67: But, nice effort trying to sneak one past the goalie.... 😆 I know @agcapps would have said something if I didn't.

src/axom/core/StackArray.hpp
rhornung67: It's your call. I probably have less faith in compilers than you.

white238: i sent kenny a teams message about this and he ignored me =*( i'm sad.

src/axom/core/Array.hpp
white238: I sent kenny a teams message about this very thing and he straight up ignored me! I'm deeply offended! (I just assumed it was some weird C++ optimization thing... but still...)

src/axom/core/utilities/BitUtilities.hpp
kennyweiss: I created this ticket for AMD: https://github.com/LLNL/axom/issues/759

kanye-quest: I think clang and gcc usually have the same intrinsics of the format `__builtin_{op}`.

Annoyingly, you can't actually use `__has_builtin` to detect their existence on GCC until like GCC 10.

src/axom/core/Array.hpp
kanye-quest: There are some subtle differences, according to [this handy table](https://josuttis.com/cpp/c++initialization.pdf)

It would seem the main benefit of using braces would be to detect unintended implicit conversions, although braces do complicate things when it comes to types with initializer list constructors.

rhornung67: hey, who is the white238 person crashing our party?!?

rhornung67: @kanye-quest thanks for the pointer to the "handy table"

white238: /sad

am i not allowed in code anymore?

/goes back to build and spack land......

rhornung67: @white238 i think the issue is that you're mixing communication mediums too much, like a couple of days ago when you responded to my email via teams  😄

src/axom/core/utilities/BitUtilities.hpp
kennyweiss: Re: clang -- It's definitely not obvious from the listed conditions, but clang defines `__GNUC__`, so it is captured by the corresponding intel and ppc guards.  

I'll add a comment to the code to clarify this.

Side note:
You can run the following to see the enabled compiler defines and/or flags:
* xlc defines:   ` > xlC -dM -qshowmacros -E <file>`
* clang: 
 Flags: `> clang++   -E - -march=native -###`  
  Defines: `> clang++ -dM -E -x c /dev/null`

src/axom/core/Array.hpp
white238: its to keep you guys guessing.. where will white238 be next! (hint probably teams)

src/axom/core/Array.hpp
kanye-quest: We did run into some issues with `axom::Array`s of non-trivial type, e.g. `axom::Array<BitSet>` in ImplicitGrid, from the array data being uninitialized - e.g. assigning a new bitset like `array[i] = BitsetType{...}` would call the destructor on junk data, which led to some weird segfaults.

I think making uninitialized data opt-in also keeps `axom::Array` more drop-in compatible with `std::vector`, since vector's behavior is to default-init all elements.

rhornung67: @kanye-quest I agree 100%

src/axom/primal/geometry/Tetrahedron.hpp
rhornung67: Nevermind. I blanked on the fact you marked it constexpr. Sorry for the noise.

kennyweiss: I'm reasonably sure that the compiler will handle this for us with `constexpr`.
The following is not perfect, but should be a reasonable approximation:

https://godbolt.org/z/aoWMTKj7e

image
e.g. with `-O0`, it gets all three expressions: `constexpr double`, `const double` and `double`

rhornung67: Yeah, I get it. I was in webex when I entered my initial comment and wasn't thinking.

src/axom/core/utilities/BitUtilities.hpp
kennyweiss: It took me a while (way too long!) to find the right formulation for the intrinsics on all the platforms/compilers we care about.

Some references:
* https://clang.llvm.org/docs/LanguageExtensions.html#intrinsics-support-within-constant-expressions
* https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html
* https://stackoverflow.com/questions/11228855/header-files-for-x86-simd-intrinsics
* https://www.ibm.com/docs/en/openxl-c-and-cpp-aix/17.1.0?topic=functions-supported-gcc-non-vector-built-in
* https://docs.microsoft.com/en-us/cpp/intrinsics/intrinsics-available-on-all-architectures?view=msvc-170
* https://en.cppreference.com/w/cpp/header/bit

src/axom/core/Array.hpp
kennyweiss: @kanye-quest -- is there a test for this?

I added this because, `Delaunay::isValid()` computation is using large bitsets, e.g. w/ millions of bit ( one per tetrahedron) and in `ImplicitGrid::findCandidates()`, the cost of initializing these bits to immediately get overwritten was substantial. I saw a substantial speedup when I disabled the initialization.

RELEASE-NOTES.md
```diff
@@ -45,7 +51,14 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - When an `inlet::Field` fails a range or valid value constraint, the provided value and
   corresponding range/set of valid values are now included in the error message
 - IOManager::write now allows the calling code to pass in the full name of
-  the root file that it will produce.
+  the root file that it will produce
+- Improved consistency of orientation operations in `primal::Plane`, `primal::Sphere`, `primal::orientation()` and `primal::in_sphere()`
```
kennyweiss, line 55: This resolves #739

src/axom/core/utilities/BitUtilities.hpp
kennyweiss: I created an issue to consider renaming our bit utility functions to match the (new) standard functions

src/axom/core/utilities/BitUtilities.hpp
kennyweiss: See: #763

src/axom/core/Array.hpp
kennyweiss: I reverted the changes to `axom::Array`

Thanks for the link @kanye-quest. 
fwiw: The `Type i {val}` form of the expression has checks across the board, so it might be the best?

@white238 -- looking at our chat, I responded that the two should be equivalent! But there were other discussions, so it might have gotten lost.

src/axom/primal/tests/primal_orientation.cpp
kennyweiss: fixed!

src/axom/slam/DynamicSet.hpp
```diff
@@ -112,26 +116,143 @@ class DynamicSet : public Set<PosType, ElemType>,
...
+    DynamicSetIterator() = default;
+
+    DynamicSetIterator(PositionType pos, DynamicSetType& dSet)
+      : IterBase(pos)
+      , m_dynamicSet(&dSet)
+    { }
+
+    DynamicSetIterator(const DynamicSetIterator& it) = default;
+
+    DynamicSetIterator& operator=(const DynamicSetIterator& it)
```
kanye-quest, line 163: Should this be `default`ed as well? (Or rule-of-zeroed?)

src/axom/slam/DynamicConstantRelation.hpp
```diff
@@ -107,12 +107,12 @@ class DynamicConstantRelation : public /*Relation,*/ CardinalityPolicy
...
     , m_toSet(toSet)
+    , m_currentFromSize(fromSet == nullptr ? 0 : m_fromSet->size())
   {
-    m_relationCardinality = CardinalityPolicy::size(0);
-    m_relationsVec.resize(m_relationCardinality * fromSet->size(), INVALID_INDEX);
+    updateSizes();
   };
 
-  ~DynamicConstantRelation() {};
+  ~DynamicConstantRelation() = default;
```
kanye-quest, line 115: It may be worthwhile to delete this destructor entirely; a user-defined destructor will inhibit generation of implicit move constructors.

kanye-quest (approved):
Looks good! Just a couple of suggestions.

src/axom/slam/Utilities.hpp
```diff
@@ -47,6 +48,120 @@ struct TypeToString<double>
   static std::string to_string() { return "double"; }
 };
 
+/**
+ * \brief A simple 3D point class similar to primal's point class,
+ * with some basic Point/Vector functionalities
+ */
+template <typename DataType = double>
+struct Point3
```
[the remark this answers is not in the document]
kennyweiss: Unfortunately it can't  -- `slam` doesn't depend on `primal` and this class exists to help test some slam features using only slam and core. 

The Delaunay tests use `primal::Point` as the Point template, but for internal testing, slam::IAMesh uses the slam::Point3 class.

src/axom/quest/Delaunay.hpp
```diff
@@ -0,0 +1,885 @@
...
+            this->getElement(element_idx).circumsphere();
+          const auto& sphere = circumspheres[element_idx];
+          const auto& center = sphere.getCenter().array();
+          const auto offset = NumericArrayType(sphere.getRadius());
+
+          BoundingBox bb;
+          bb.addPoint(PointType(center - offset));
+          bb.addPoint(PointType(center + offset));
+
+          implicitGrid.insert(bb, element_idx);  // insert valid entries into grid
```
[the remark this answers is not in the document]
kennyweiss: I held back from using the vectorized/accelerated `ImplicitGrid` for this PR in an attempt to keep the PR somewhat more manageable.  I.e. I would have had to implement a non-RAJA workaround etc..

I'd like to explore the vectorized version in a follow-up PR.

src/axom/quest/Delaunay.hpp
```diff
@@ -0,0 +1,885 @@
...
+      const int kUpper = (DIM == 2) ? 0 : res;
+      const IndexType stride[3] = {1, res, (DIM == 2) ? 0 : res * res};
+      for(IndexType k = 0; k < kUpper; ++k)
+        for(IndexType j = 0; j < res; ++j)
+          for(IndexType i = 0; i < res; ++i)
+          {
+            const IndexType vals[3] = {i, j, k};
+            const GridCell cell(vals);
+            const auto idx = dot_product(cell.data(), stride, DIM);
+            grid.getBinContents(idx) = implicitGrid.getCandidatesAsArray(cell);
```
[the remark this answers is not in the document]
kennyweiss: Definitely!  I'd like to update this in a future PR.

src/axom/slam/DynamicSet.hpp
```diff
@@ -176,21 +297,38 @@ class DynamicSet : public Set<PosType, ElemType>,
...
+    const int sz = size();
+    for(int i = 0; i < sz; ++i)
     {
       if(m_data[i] == e) return i;
     }
     return INVALID_ENTRY;
   };
 
+  /**
+   * \brief Given a value, find the index of the first entry containing it
```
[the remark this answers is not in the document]
kennyweiss: Good call!

src/axom/quest/Delaunay.hpp
```diff
@@ -0,0 +1,889 @@
...
+  BoundingBox m_bounding_box;
+  bool m_has_boundary;
+  int m_num_removed_elements_since_last_compact;
+
+  ElementFinder m_element_finder;
+
+public:
+  /**
+   * \brief Default constructor
+   * \note User need to call initializeBoundary(BoundingBox) before adding points.
```
agcapps, line 77:
```suggestion
   * \note User must call initializeBoundary(BoundingBox) before adding points.
```

src/axom/slam/mesh_struct/IA.hpp
```diff
@@ -0,0 +1,402 @@
...
+ * - the boundary relation from elements to their vertices
+ * - the partial coboundary relation from vertices to one incident element, and
+ * - the adjacency relation between elements along their facets (faces of dimension TDIM-1).
+ *
+ * PointType is required to have the following interface:
+ * - .ctor(T*)          -- constructor from an array of T
+ * - operator[](index)  -- subscript operator to access the coordinates
+ * - data() -> T*       -- direct access to the data
+ * - operator==()       -- equality operator
+ */
```
agcapps, line 57: Is this a well-known data structure?  Please cite the paper (or web page, even).

src/axom/slam/mesh_struct/IA.hpp
```diff
@@ -0,0 +1,402 @@
...
+  using ModularVertexIndex =
+    slam::ModularInt<slam::policies::CompileTimeSize<IndexType, VERTS_PER_ELEM>>;
+  using ModularFacetIndex =
+    slam::ModularInt<slam::policies::CompileTimeSize<IndexType, VERTS_PER_ELEM>>;
+
+public:
+  /// \brief Default Constructor for an empty mesh
+  IAMesh();
+
+  /// \brief Construct an IA mesh with the given point coordinate and vertex indices for each elements
```
agcapps, line 115:
```suggestion
  /// \brief Construct an IA mesh with the given point coordinate and vertex indices for each element
```

src/axom/slam/Utilities.hpp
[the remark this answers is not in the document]
agcapps: @kennyweiss , please add a sentence to the struct's Doxygen explaining what you just wrote.

src/axom/slam/OrderedSet.hpp
```diff
@@ -371,7 +421,10 @@ struct OrderedSet : public Set<PosType, ElemType>,
    * An index pos is valid when \f$ 0 \le pos < size() \f$
    * \return true if the position is valid, false otherwise
    */
-  bool isValidIndex(PositionType pos) const { return pos >= 0 && pos < size(); }
+  inline bool isValidIndex(PositionType pos) const
+  {
+    return pos >= 0 && pos < size();
```
agcapps, line 426: Should the stride affect this?

src/axom/slam/tests/slam_IA.cpp
```diff
@@ -0,0 +1,796 @@
...
+  BasicTriMeshData()
+    : points(point_arr, point_arr + 24)
+    , elem(tri_arr, tri_arr + 36)
+  { }
+
+  static int numVertices() { return 8; }
+  static int numTriangles() { return 12; }
+};
+
+// clang-format off
```
agcapps, line 47: (Here and line 89) Thanks.

src/axom/slam/DynamicConstantRelation.hpp
```diff
@@ -372,22 +365,29 @@ class DynamicConstantRelation : public /*Relation,*/ CardinalityPolicy
...
     expandSizeIfNeeded(fromSetIndex + 1);
-    m_relationsVec[m_relationCardinality * fromSetIndex + offset] = toSetIndex;
+    verifyPosition(fromSetIndex);
+    m_relationsVec[relationCardinality() * fromSetIndex + offset] = toSetIndex;
   }
 
-  /**
-   * \brief Mark all values in entry \a fromSetIndex as invalid.
-   */
+  // \brief Mark all values in entry \a fromSetIndex as invalid.
```
agcapps, line 372:
```suggestion
  /// \brief Mark all values in entry \a fromSetIndex as invalid.
```

agcapps (approved):
Well, thank you @kennyweiss .  I did find a few typos, but nothing to argue too hard about.  The modernization is welcome.

src/axom/spin/ImplicitGrid.hpp
```diff
@@ -346,6 +346,36 @@ class ImplicitGrid
     return res;
   }
 
+  /*!
+   * Finds the candidate elements in the given \a gridCell of the grid
+   *
+   * \param [in] gridCell The cell of the grid
+   * \return A bitset \a bSet whose bits correspond to the elements of the IndexSet.
```
white238, line 353: Is it appropriate to annotate `bSet` here since its not an argument?

src/axom/quest/examples/delaunay_triangulation.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+  SLIC_INFO(axom::fmt::format(
+    "It took {} seconds to create a Delaunay complex with {} "
+    "points. Mesh has {} {}. Insertion rate of {:.1f} points per second.",
+    timer.elapsedTimeInSec(),
+    dt.getMeshData()->getNumberOfValidVertices(),
+    dt.getMeshData()->getNumberOfValidElements(),
+    DIM == 2 ? "triangles" : "tetrahedra",
+    numPoints / timer.elapsedTimeInSec()));
+  axom::slic::flushStreams();
```
white238, line 174: I'm reasonably certain flushing on a `SimpleLogger` doesn't do anything because it immediately writes to `std::cout`.  Did you see a behavior that needed this or was it just added because most loggers do need to be flushed?

white238 (approved):

src/axom/slam/OrderedSet.hpp
kennyweiss: No. The _indices_ in an `OrderedSet` go from `0` to `size-1`. 
The _elements_ of the set have strides offsets and indirections.

src/axom/spin/ImplicitGrid.hpp
kennyweiss: I think so (?)
I reworked the docs to remove it.

src/axom/quest/examples/delaunay_triangulation.cpp
kennyweiss: I think it's the latter, and I got in the habit of flushing before expensive operations.
I removed these.

src/axom/quest/Delaunay.hpp
```diff
@@ -74,7 +74,7 @@ class Delaunay
 public:
   /**
    * \brief Default constructor
-   * \note User need to call initializeBoundary(BoundingBox) before adding points.
+   * \note User must to call initializeBoundary(BoundingBox) before adding points.
```
dtaller, line 77: I know that I'm not on the axom team and this is a small thing, but you have a typo here "User must to call". Probably you just want to say "User must call", no 'to'.

agcapps: Thanks @dtaller !

PR #745, opened 2021-12-10
Allow the full root file name to be provided in calls to SPIO IOManager output.

### LINKED ISSUE #195: Suggestion for sidre::IOManager API
I noticed some inconsistency in the API of the `sidre::IOManager` class, that may be confusing to some users:
- `sidre::IOManager::write()` accepts a string corresponding to the base file name with no extension. 

- `sidre::IOManager::read()`, however, expects the root file, which is `<basefile>.root`

Could both `write()` and `read()` take the same string argument, e.g., either just the base filename with no extension? or the root file?

This seems more intuitive and could enhance usability aspects of this class in application codes.

nselliott:
# Summary

- This PR is a feature modification
- It does the following:
  - Modifies IOManager::write() to allow the calling code to provide the name of the root file for the SPIO output.
  - Fixes #195 
  - Implements the suggestion from the above issue to allow the calling code to provide a string in the format `"myoutput.root"` in which case that string will be used as the name of the root file, and the `.root` will be removed to serve as the name of the subdirectory holding the output data files.


src/axom/sidre/spio/IOManager.hpp
```diff
@@ -293,10 +302,24 @@ class IOManager
...
+   * output subdirectory and files containing sidre data. It will be
+   * named as the root_base string plus the suffix ".root", unless
+   * root_base already ends with ".root", in which case no additional
+   * suffix will be added.
+   *
+   * The returned string will be identical to root_base, except when
+   * root_base is provided with the suffix ".root" already there, in
+   * which case the returned string will be the root_base string with
+   * the ".root" suffix removed from its end.
+   */
```
agcapps, line 318: Thanks for the docs.

agcapps (approved):
:htumbsup:

src/axom/sidre/spio/IOManager.cpp
```diff
@@ -463,13 +463,27 @@ void IOManager::loadExternalData(sidre::Group* datagroup,
...
+  size_t base_size = file_base.size();
+  size_t suffix_size = suffix.size();
+  if(base_size > suffix_size)
+  {
+    if(file_base.compare(base_size - suffix_size, suffix_size, suffix) == 0)
+    {
+      // file_base has ".root" suffix, so remove it.
+      file_base.erase(base_size - suffix_size);
+    }
+  }
```
kennyweiss, line 485: I think you can replace this new block with
```cpp
file_base = axom::utilities::string::removeSuffix(file_base, ".root");
```

See: 
https://github.com/LLNL/axom/blob/cab5877121dbc467a716237eef1128c861a4baba/src/axom/core/utilities/StringUtilities.hpp#L66-L79

src/axom/sidre/tests/spio/spio_serial.hpp
```diff
@@ -176,3 +176,49 @@ TEST(spio_serial, write_read_write)
...
+
+//------------------------------------------------------------------------------
+TEST(spio_serial, rootfile_suffix)
+{
+  // This test verifies the usage of the string passed to writer.write() to
+  // control the names of the root file and the output files. The usual
+  // case is that the suffix ".root" is appended to the string to name the
+  // root file, and the given string is used to name the subdirectory holding
+  // the data files. The exception is if the string already has the suffix
+  // ".root", is is used unchanged to name the root file, and the suffix is
```
kennyweiss, line 188: Typo:
```suggestion
  // ".root", it is used unchanged to name the root file, and the suffix is
```

src/axom/sidre/tests/spio/spio_serial.hpp
```diff
@@ -176,3 +176,49 @@ TEST(spio_serial, write_read_write)
...
+
+  // Write output to two file locations, with and without root suffix
+  writer.write(ds->getRoot(), num_files, suffix_name, "sidre_hdf5");
+  writer.write(ds->getRoot(), num_files, nosuffix_name, "sidre_hdf5");
+
+  // Create two more DataStores to read in the both outputs
+  DataStore ds_suffix, ds_nosuffix;
+  IOManager reader(MPI_COMM_WORLD);
+  reader.read(ds_suffix.getRoot(), suffix_name);
+  reader.read(ds_nosuffix.getRoot(), nosuffix_name + ROOT_EXT);
```
kennyweiss, line 214: Clarification: It looks like `IOManager.read()` still needs the `.root` suffix. 
Is that correct?

kennyweiss (approved):
Thanks @nselliott

rhornung67 (approved):
Thanks @nselliott

src/axom/sidre/spio/IOManager.cpp
nselliott: Thanks, I wasn't aware of this utility.

src/axom/sidre/tests/spio/spio_serial.hpp
nselliott: Strictly speaking, no. `write()` always produces a file with the `.root` suffix, so in conventional usage, that filename should be provide unchanged to `read()`. But someone could rename the file to anything else and provide the new filename to `read()`, and this should still work, as long at the file's contents remain valid.

kennyweiss: Thanks -- got it -- so, `read()` does not alter the path to append/remove a suffix, but `write()` does

PR #747, opened 2021-12-14
Sub-array indexing of multi-dimensional Array and ArrayView

kanye-quest:
# Summary

This is an exploratory PR for handling multi-dimensional array sub-slicing.
- Allow for the creation of sub-arrays from multi-dimensional `ArrayBase::operator(Args... args)` when number of arguments is less than array's number of dimensions
  - Sub-array type is `ArrayView<T, N>` where `N = DIMS - IndexDims`
  - For `DIMS == IndexDims` sub-array type decays to scalar reference of full index into data array
- Add alternative indexing function `ArrayBase::operator[](StackArray<IndexType, UDim>& idxs)` as found in the mdspan paper
- Scalar `operator[](IndexType idx)` now returns a sub-array of dimensionality `DIMS - 1`; flat indexing now achieved by `ArrayBase::flatIdx(IndexType idx)`
  - This lets us do stuff like `double_vector_3d[i][j][k]` in addition to `double_vector_3d(i,j,k)`

## Open questions

- Should we have the iterator interface return over subarrays instead of over raw values as well?
- Are there any users that are using flat indexing via `operator[]`?

### DIFF
--- a/src/axom/core/ArrayBase.hpp
+++ b/src/axom/core/ArrayBase.hpp
@@ -122,14 +122,15 @@ public:
     RealConstT&,
     typename detail::ArrayTraits<ArrayType>::template Slice<SliceDim>>::type;
 
-  ArrayBase() : m_dims {} { updateStrides(); }
+  AXOM_HOST_DEVICE ArrayBase() : m_dims {} { updateStrides(); }
 
   /*!
    * \brief Parameterized constructor that sets up the default strides
    *
    * \param [in] args the parameter pack of sizes in each dimension.
    */
-  ArrayBase(const StackArray<IndexType, DIM>& args) : m_dims {args}
+  AXOM_HOST_DEVICE ArrayBase(const StackArray<IndexType, DIM>& args)
+    : m_dims {args}
   {
     updateStrides();
   }
@@ -304,7 +305,7 @@ protected:
    * In the future, this class will support different striding schemes (e.g., column-major)
    * and/or user-provided striding
    */
-  void updateStrides()
+  AXOM_HOST_DEVICE void updateStrides()
   {
     // Row-major
     m_strides[DIM - 1] = 1;
@@ -405,9 +406,9 @@ public:
    */
   using RealConstT = typename std::conditional<is_array_view, T, const T>::type;
 
-  ArrayBase(IndexType = 0) { }
+  AXOM_HOST_DEVICE ArrayBase(IndexType = 0) { }
 
-  ArrayBase(const StackArray<IndexType, 1>&) { }
+  AXOM_HOST_DEVICE ArrayBase(const StackArray<IndexType, 1>&) { }
 
   // Empy implementation because no member data
   template <typename OtherArrayType>
@@ -607,9 +608,14 @@ namespace detail
 {
 // Takes the product of a parameter pack of T
 template <typename T, int N>
-T packProduct(const T (&arr)[N])
+AXOM_HOST_DEVICE T packProduct(const T (&arr)[N])
 {
-  return std::accumulate(arr, arr + N, 1, std::multiplies<T> {});
+  T prod = 1;
+  for(int idx = 0; idx < N; idx++)
+  {
+    prod *= arr[idx];
+  }
+  return prod;
 }
 
 template <typename T, int N>
--- a/src/axom/core/ArrayBase.hpp
+++ b/src/axom/core/ArrayBase.hpp
@@ -186,6 +186,18 @@ public:
     return (*this)[indices];
   }
 
+  AXOM_HOST_DEVICE SliceType<1> operator[](const IndexType idx)
+  {
+    const StackArray<IndexType, 1> slice {idx};
+    return (*this)[slice];
+  }
+
+  AXOM_HOST_DEVICE ConstSliceType<1> operator[](const IndexType idx) const
+  {
+    const StackArray<IndexType, 1> slice {idx};
+    return (*this)[slice];
+  }
+
   template <int UDim>
   AXOM_HOST_DEVICE SliceType<UDim> operator[](const StackArray<IndexType, UDim>& idx)
   {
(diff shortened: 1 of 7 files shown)


src/axom/core/ArrayBase.hpp
```diff
@@ -475,7 +629,7 @@ bool operator==(const ArrayBase<T1, DIM, LArrayType>& lhs,
 
   for(int i = 0; i < static_cast<const LArrayType&>(lhs).size(); i++)
   {
-    if(!(lhs[i] == rhs[i]))
+    if(!(lhs.flatIdx(i) == rhs.flatIdx(i)))
     {
       return false;
     }
```
kennyweiss, line 635: Minor: Do you expect noticeably overhead from using the `flatIdx()` here instead of directly accessing the elements?

I suppose I'm wondering how much of this gets optimized/compiled out and which parts are not optimized.

**Edit:** Clearly, it's not worse than it was before this PR

src/axom/core/ArrayIteratorBase.hpp
```diff
@@ -48,8 +48,8 @@ class ArrayIteratorBase
    */
   ValueType& operator*()
   {
-    return (*m_arrayPtr)[IteratorBase<ArrayIteratorBase<ArrayType, ValueType>,
-                                      IndexType>::m_pos];
+    return m_arrayPtr->flatIdx(
+      IteratorBase<ArrayIteratorBase<ArrayType, ValueType>, IndexType>::m_pos);
```
kennyweiss, line 52: Minor/style: The code might be a bit more readable if the base type were made a type alias for the class

e.g.
```cpp
using BaseType  = IteratorBase<ArrayIteratorBase<ArrayType, ValueType>, IndexType>;
```

Then this function could be:
```cpp
    return m_arrayPtr->flatIdx(BaseType::m_pos);
``

**Review by kennyweiss — APPROVED** [review]
This is really nice @kanye-quest 

Some thoughts:

* I'd be curious to see what the performance cost might be for the different access methods,
e.g.
  * `arr.flatIdx ( /* expression with i,j,k */)` 
  * `arr[i][j][k]`
  * `arr[ StackArray<IndexType,3>(i,j,k) ])  // does this work?` 
  * `arr( i,j,k)`
  * `arr(i,j)[k]`
* ~~Can you please add a link to the PR description to the mdspan documentation that you were looking at?~~
* Please ensure that the RELEASE-NOTES indicate the change to `operator[]`. I think it's new enough that people aren't extensively using the mcarrays, but they'll want to know about the changes

File: src/axom/core/tests/core_array.hpp [hunk 3]
```diff
@@ -1421,7 +1421,7 @@ TEST(core_array, check_multidimensional)
   for(int i = 0; i < v_int_flat.size(); i++)
   {
     // For a multidim array, op[] is a "flat" index into the raw data
```
white238, line 1423:
```suggestion
    // For a multidim array, flatIdx(i) is a "flat" index into the raw data
```

src/axom/core/tests/core_array.hpp
```diff
@@ -1451,7 +1451,22 @@ TEST(core_array, check_multidimensional)
   for(int i = 0; i < v_double.size(); i++)
   {
     // For a multidim array, op[] is a "flat" index into the raw data
```
white238, line 1453:
```suggestion
    // For a multidim array, flatIdx(i) is a "flat" index into the raw data
```

src/axom/core/tests/core_array.hpp
```diff
@@ -1519,7 +1534,22 @@ TEST(core_array, check_multidimensional_view)
   for(int i = 0; i < v_double_view.size(); i++)
   {
     // For a multidim array, op[] is a "flat" index into the raw data
```
white238, line 1536:
```suggestion
    // For a multidim array, flatIdx(i) is a "flat" index into the raw data
```

src/axom/core/ArrayBase.hpp
```diff
@@ -360,6 +490,28 @@ class ArrayBase<T, 1, ArrayType>
...
+   * \brief Accessor, returns a reference to the given value.
+   * For multidimensional arrays, indexes into the (flat) raw data.
+   *
+   * \param [in] idx the position of the value to return.
+   *
+   * \note equivalent to *(array.data() + idx).
+   *
+   * \pre 0 <= idx < m_num_elements
+   */
+  AXOM_HOST_DEVICE T& flatIdx(const IndexType idx)
```
white238, line 504: I'm pretty sure we more often use the fully spelled out word `index`.  This should be `flatIndex` especially for user facing symbols.

src/axom/core/ArrayIteratorBase.hpp
agcapps: I support this suggestion, after much thought.

agcapps (approved):

rhornung67 (approved):
This looks like something good. Hopefully, we can get some helpful user feedback on it.

kanye-quest:
@kennyweiss -- re: performance:

| Indexing type | time to fill 500x500x500 array (run 10 times) |
| ------------- | ---------------| 
| flat index | 1.136 s |
| `(i, j, k)` | 2.764 s |
| `[i][j][k]` | 4.812 s |
| `(i, j)[k]` | 2.461 s |
| `StackArray<IndexType, 3>{i, j, k}` | 2.563 s |

The main outliers that appear apparent are the flat indexing and `[i][j][k]` indexing. The latter case's results aren't that surprising to me given that indexing in this manner stresses the `ArrayBase` abstraction layers considerably.

However, the fact that flat indexing is that much faster than `(i,j,k)` indexing is concerning to me; IMO those two should reduce down to the same instructions. I'll do some more probing to see what's going on.

On the `[i][j][k]` front, I think there are some potential things we could try to reduce the performance impact, such as using a currying-based approach for the `ArrayViewProxy` type.

Issue #754, opened 2021-12-23
Use axom::Path for path operations in IOManager


nselliott:
Currently there is usage of some low-level conduit utilities and some other ad-hoc string operations for the file paths and data paths. This can likely be cleaned up by using axom::Path without changing the public API.

PR #776, opened 2022-01-28
Use axom::Path in IOManager

nselliott:
# Summary

- This PR is a small code cleanup
- It does the following:
  - Modifies some of the operations on path strings in IOManager to use axom::Path rather than conduit string utilities
  - Does not change the API
  - Addresses issue #754


src/axom/sidre/spio/IOManager.cpp
```diff
@@ -793,14 +796,18 @@ std::string IOManager::getFileNameForRank(const std::string& file_pattern,
...
-  std::string slash = "/";
-  conduit::utils::rsplit_string(root_name, slash, curr, root_dir);
+  std::string delimiter = "/";
+  axom::Path root_path(root_name);
+  std::string root_dir = root_path.dirName();
+
+  if (!root_name.empty() && root_name[0] == delimiter[0])
+  {
+    root_dir = delimiter + root_dir;
+  }
```
nselliott, line 806: One limitation of axom::Path is that it does not preserve a leading '/' delimiter, which can make a meaningful difference when used as a file system path, so lines 803-806 restore this when necessary.

white238: I would argue that's a bug and not intentional.  Did you attempt to fix this in `axom::Path` and other things broke?

nselliott: I didn't touch `axom::Path`.  I could try to make a fix there before merging this PR.

nselliott: It is fixable in `axom::Path` and I got most of the way there this afternoon.  However, I think it should be a standalone change and not bloat the scope of this PR.  I'm inclined to create another issue and PR.  I could merge this one with the 4-line workaround, or let it wait for the other one.

nselliott: Issue #777 created regarding the leading delimiter in `axom::Path`.

white238: i'm okay with that

white238 (approved):

white238:
Approved pending you fix your style problem.

src/axom/sidre/spio/IOManager.cpp
```diff
@@ -793,14 +796,18 @@ std::string IOManager::getFileNameForRank(const std::string& file_pattern,
...
   //find the directory and add it to file_name
-  std::string curr;
-  std::string root_dir;
-  std::string slash = "/";
-  conduit::utils::rsplit_string(root_name, slash, curr, root_dir);
+  std::string delimiter = "/";
+  axom::Path root_path(root_name);
+  std::string root_dir = root_path.dirName();
+
+  if(!root_name.empty() && root_name[0] == delimiter[0])
```
kennyweiss, line 803: Suggestion: It might be cleaner to replace this with `axom::utilities::string::startsWith(root, delimiter[0])`

See: https://github.com/LLNL/axom/blob/7d595645e7e04e1f88c0f5b5cb265e7328eb0859/src/axom/core/utilities/StringUtilities.hpp#L43-L64

src/axom/sidre/spio/IOManager.cpp
kennyweiss: I agree that this is a bug and that it can be resolved with the current workaround for this PR with a follow-up fix in `axom::Path`, as discussed in #777.

src/axom/sidre/spio/IOManager.cpp
```diff
@@ -793,14 +796,18 @@ std::string IOManager::getFileNameForRank(const std::string& file_pattern,
...
+  {
+    root_dir = delimiter + root_dir;
+  }
 
   if(!root_dir.empty())
   {
-    file_name = root_dir + slash + file_name;
+    file_name = root_dir + delimiter + file_name;
   }
```
kennyweiss, line 811: This block could be replaced by:
```cpp
axom::utilities::string::appendPrefix(root_dir, file_name,  delimiter[0])
```
(which handles the check for empty).

See: https://github.com/LLNL/axom/blob/7d595645e7e04e1f88c0f5b5cb265e7328eb0859/src/axom/core/utilities/StringUtilities.hpp#L117-L130

kennyweiss (approved):
Thanks for working on this @nselliott 

I had a few suggestions for using more of the string utilities to clean the code up a bit more.

agcapps (approved):

rhornung67 (approved):
Thanks @nselliott

PR #780, opened 2022-02-01
Cleanup of quest::findTriMeshIntersections implementations

kanye-quest:
# Summary

- Cleans up the `quest::findTriMeshIntersections*()` functions by consolidating the common implementation between BVH and ImplicitGrid
  - New acceleration data structures can be added by specializing `quest::detail::CandidateFinder` and implementing a candidate search overload
  - As a bonus, this seems to provide a slight speedup to host-side mesh testing operations
- Various fixes for `axom::Array`
  - Fix host-side fill of uninitialized memory with non-trivial type
  - Fix behavior of non-trivial type array resizes to conform with `std::vector`
  - Allow use of iterator pairs for empty arrays


src/axom/quest/detail/MeshTester_detail.hpp
```diff
@@ -0,0 +1,447 @@
...
+  const double* y = surface_mesh->getCoordinateArray(mint::Y_COORDINATE);
+  const double* z = surface_mesh->getCoordinateArray(mint::Z_COORDINATE);
+
+  for(int n = 0; n < 3; ++n)
+  {
+    const axom::IndexType nodeIdx = triCell[n];
+    tri[n][0] = x[nodeIdx];
+    tri[n][1] = y[nodeIdx];
+    tri[n][2] = z[nodeIdx];
+  }
```
kennyweiss, line 57: Minor: This might be a bit cleaner as:
```suggestion
  for(int n = 0; n < 3; ++n)
  {
    const axom::IndexType nodeIdx = triCell[n];
    tri[n] = Point3{ x[nodeIdx], y[nodeIdx], z[nodeIdx]};
  }
```

src/axom/quest/detail/MeshTester_detail.hpp
```diff
@@ -0,0 +1,447 @@
...
+  axom::Array<int, 1, Space> m_degenerate;
+};
+
+template <typename ExecSpace, typename FloatType>
+void CandidateFinderBase<ExecSpace, FloatType>::initialize()
+{
+  const int ncells = m_surfaceMesh->getNumberOfCells();
+
+  m_tris.resize(ncells);
+  axom::ArrayView<detail::Triangle3, 1, Space> p_tris = m_tris;
```
kennyweiss, line 155: What does the `p_` prefix indicate?

kennyweiss, line 155: Would this be easier to understand if it were written as the following equivalent:
```suggestion
  auto p_tris = m_tris.view();
```

(I can see both sides of this...)

src/axom/quest/detail/MeshTester_detail.hpp
```diff
@@ -0,0 +1,447 @@
...
+   * \param [in] intersectionThreshold The tolerance threshold to use for
+   *  triangle intersection tests.
+   */
+  CandidateFinderBase(mint::UnstructuredMesh<mint::SINGLE_SHAPE>* surface_mesh,
+                      double intersectionThreshold)
+    : m_surfaceMesh(surface_mesh)
+    , m_intersectionThreshold(intersectionThreshold)
+  { }
+
+  void initialize();
```
kennyweiss, line 107: Please add a brief doxygen comment for this function

kennyweiss (approved):
Thanks @kanye-quest -- this is a really nice cleanup/consolidation!

Would it make sense to add in implementation of the naive intersection algorithm to the new MeshTester abstraction? We originally had it as a way of checking that we get the same result using our spatial index as we do by running the full `O(n^2)` "naive" algorithm.
Perhaps it is less relevant now that we have two spatial indexes (BVH and ImplicitGrid) that work in different execution spaces.
The answer could very well be "no, it's not worth the effort".

agcapps (approved):
Thank you, @kanye-quest .

rhornung67 (approved):
Looks like a nice code improvement. Thanks @kanye-quest

bmhan12 (approved):
LGTM!

src/axom/quest/detail/MeshTester_detail.hpp
kanye-quest: Those were holdover variable names from the original ImplicitGrid-based implementation; I originally used the `p_` prefix to indicate pointers to the array data. Now that we're using `ArrayView` in that place, it might more sense to change the view names to `v_` instead.

kanye-quest:
> Would it make sense to add in implementation of the naive intersection algorithm to the new MeshTester abstraction? We originally had it as a way of checking that we get the same result using our spatial index as we do by running the full O(n^2) "naive" algorithm.

That seems like a logical addition to make - the only problem I see is that the `findTriMeshIntersections()` common implementation relies on the generation of a candidate intersection list in derived `CandidateFinder`s. We probably can't fit in O(n^2) element pairs as our candidate intersections, nor would we want to, so I'd imagine this would be implemented as a `CandidateFinder` specialization which overrides `findTriMeshIntersections` entirely.

PR #782, opened 2022-02-05
Handle leading delimiter char in axom::Path

### LINKED ISSUE #777: Preserve leading delimiter in axom::Path
The current implementation of axom::Path stores a vector of strings, the components of the path, along with a single delimiter character, usually `/`.  The result of the implementation is that there is no way to construct a `Path` that has a leading delimiter.  The following two `Path` objects will be identical, and no methods of the class can retrieve the fact that `path_1` was constructed using a string with a leading `/`.

` Path path_0("foo/bar/baz");`
` Path path_1("/foo/bar/baz");`

The leading `/` is especially important when the `Path` is used to represent a file system path, so we should modify the class to preserve the leading delimiter and use it in the various methods of the class.

nselliott:
# Summary

- This PR is a bug fix
  - Modifies class `axom::Path` to preserve the a leading delimiter in the path, such as `/foo/bar`.
  - Fixes #777

Reviewers, please especially take a look at the unit tests added here, to be sure that the expected results for the tests match the desired functionality of the class. The handling of paths such as the root path `/` and those with only one component like `/foo` could be considered new functionality.


src/axom/core/tests/core_Path.hpp
```diff
@@ -116,15 +188,25 @@ TEST(core_Path, dirname_basename_multi_component_alt_delim)
...
+  axom::Path path("/foo/bar/baz");
+  EXPECT_EQ(path.dirName(), "/foo/bar");
+  EXPECT_EQ(path.baseName(), "baz");
+}
+
 TEST(core_Path, basic_equals)
 {
   axom::Path first("foo/bar/baz");
   axom::Path second("foo/bar/baz");
   axom::Path third("foo:bar:baz", ':');
```
kennyweiss, line 202: Suggestion: We should probably also test the leading delimeter with a non-default delim, 
e.g. `Path {":foo:bar:baz", ':' }`

src/axom/core/Path.cpp
```diff
@@ -45,12 +62,29 @@ Path Path::join(std::initializer_list<Path> paths, const char delim)
...
   return result;
 }
 
 Path::operator std::string() const
 {
-  return fmt::format("{0}", fmt::join(m_components, std::string(1, m_delim)));
+  std::string result;
+  if(m_leading_delim)
+  {
+    result = "/";
```
kennyweiss, line 83: Should this account for a non-default delimiter?
E.g.
```suggestion
    result = m_delim;
```

src/axom/core/Path.cpp
```diff
@@ -45,12 +62,29 @@ Path Path::join(std::initializer_list<Path> paths, const char delim)
...
+  }
+
+  result += fmt::format("{0}", fmt::join(m_components, std::string(1, m_delim)));
```
kennyweiss, line 86: Suggestion: The following one-liner could account for leading delims and joining the path
```suggestion
  return fmt::format("{0}{1}", 
                 m_leading_delim? "" : std::string(1,m_delim), 
                 fmt::join(m_components, std::string(1, m_delim)));
```

src/axom/core/Path.cpp
```diff
@@ -16,8 +16,15 @@ namespace axom
 {
 Path::Path(const std::string& path, const char delim) : m_delim(delim)
 {
+  size_t first_position = 0;  // position of first non-delimiter char in path
+  if(!path.empty() && path[0] == m_delim)
```
kennyweiss, line 20: Suggestion: Consider using `axom::utilities::string::startsWith()` for this test

https://github.com/LLNL/axom/blob/cd5866a630ada425e9f0f0de0c54c4708543f48c/src/axom/core/utilities/StringUtilities.hpp#L55-L64

kennyweiss (approved):
Thanks @nselliott !

src/axom/core/Path.cpp
nselliott: Thanks for catching that, I overlooked this one.

src/axom/core/tests/core_Path.hpp
```diff
@@ -14,6 +14,13 @@ TEST(core_Path, trivial_construct)
   EXPECT_EQ(static_cast<std::string>(path), path_str);
 }
 
+TEST(core_Path, lead_delim_construct)
+{
+  const std::string path_str = "/foo";
+  axom::Path path(path_str);
+  EXPECT_EQ(static_cast<std::string>(path), path_str);
```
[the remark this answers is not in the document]
nselliott: The cast uses an implemented `operator std::string()`, so it shouldn't break as long as the implementation continues to be tested and maintained. This was an existing part of the API that I didn't touch. If we want an API change, perhaps that should be a separate issue?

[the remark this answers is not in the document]
agcapps: Thanks for pointing out the `operator std::string()`; I had overlooked that.  I don't want to inflict scope creep on this PR or on you!  Perhaps a separate issue would be a good idea.

src/axom/core/Path.cpp
nselliott: Thanks, I have put this in.  I needed to reverse the order of the second and third operands in the ternary operator.

rhornung67 (approved):

white238 (approved):

PR #784, opened 2022-02-06
Bugfix for sign computation in `quest::SignedDistance`

### LINKED ISSUE #783: Improve robustness of SignedDistance computation (redux)
We improved the robustness of the `computeSign()` functionality of quest's SignedDistance query in #706, but found a new set of regressions in regions with sharp angles and poorly shaped triangles (it is not clear if the problem is due to very acute angles or very obtuse angles, but both are present in the regression).

I ran the signed distance query on the following simple mesh comprised of 24 triangles
signed_distance_reproducer_tri_mesh
using the following command line:
```
> ./examples/quest_signed_distance_interface_ex -f ../data/quest/fin0.stl \
                                                --box-dims 33 33 33       \
                                                --box-min -10 -20 5       \
                                                --box-max 10 0 25
```
~~(note: `fin0.stl` is not yet checked into our data repository)~~

The computed signed distance has several regions where it computes the wrong sign, e.g. it marks points as "INSIDE" when they should be marked as "OUTSIDE".  This is demonstrated in the following visualization of the zero-level isosurface of the computed distance function,  which should not have any isosurface components outside the shape:
signed_distance_reproducer_isosurface

kennyweiss:
# Summary

- This PR is a bugfix
- It resolves #783 
- The problem was related to cases where the closest point was along an edge with an acute dihedral angle and the incident triangles had (vastly) different areas. The sign is related to a pseudo-normal computation that sums the normals of the two incident triangles, and the code was not unitizing the normals before adding them. This can cause a problem when one of the triangles had a very small area, since the larger triangles was disproportionately weighting the sum.
- I created a few regression meshes in https://github.com/LLNL/axom_data/pull/9 that were failing before my fix and passing afterwards

#### TODO

- [x] Update the data submodule again after merging https://github.com/LLNL/axom_data/pull/9 into `axom_data@main`

### DIFF
--- a/src/axom/quest/SignedDistance.hpp
+++ b/src/axom/quest/SignedDistance.hpp
@@ -214,8 +214,7 @@ public:
    */
   double computeDistance(double x, double y, double z = 0.0)
   {
-    PointType pt = PointType::make_point(x, y, z);
-    return (computeDistance(pt));
+    return computeDistance(PointType {x, y, z});
   }
 
   /*!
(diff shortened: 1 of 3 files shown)


src/axom/quest/SignedDistance.hpp
```diff
@@ -643,23 +642,22 @@ inline void SignedDistance<NDIMS, ExecSpace>::checkCandidate(
...
       switch(cpt_type)
       {
       case detail::ClosestPointLocType::edge:
-        // Candidate closest point is on an edge - add the normal of a
-        // potentially-adjacent face
-        currMin.sumNormals += norm;
+        // Add the (unitized) normal of the (potentially) incident face along this edge
+        // Note: Unitizing is important in cases where the edge is shared
+        // by triangles with different areas and an acute dihedral angle
+        currMin.sumNormals += surface_elems[ei].normal().unitVector();
```
kennyweiss, line 653: The bugfix is in this line:
* We have to use the unit normal vector for edges to account for possible differences in the triangle areas
* We were already using unit normals for vertices, as can be seen a few lines down

agcapps (approved):
gotta love those corner cases

white238 (approved):

bmhan12 (approved):

PR #789, opened 2022-02-10
Fix for ImplicitGrid performance regression

kanye-quest:
# Summary

- Fixes a performance regression on `ImplicitGrid` queries on the GPU, as a result of data not being correctly allocated in device memory
  - On the `car.stl` mesh tester run, the ImplicitGrid kernels were running ~10 times slower than before
- As a bonus, removes usage of `aligned_storage` - it wasn't used correctly in `ArrayOps`, and it's getting deprecated in C++23 anyways

### DIFF
--- a/src/axom/core/ArrayBase.hpp
+++ b/src/axom/core/ArrayBase.hpp
@@ -767,7 +767,7 @@ struct ArrayOpsBase<T, true>
   {
     // Similar to fill(), except we can allocate stack memory and placement-new
     // the object with a move constructor.
-    alignas(T) unsigned char host_buf[sizeof(T)];
+    alignas(T) axom::uint8 host_buf[sizeof(T)];
     T* host_obj = ::new(&host_buf) T(std::forward<Args>(args)...);
     axom::copy(array + i, host_obj, sizeof(T));
   }
--- a/src/axom/core/ArrayBase.hpp
+++ b/src/axom/core/ArrayBase.hpp
@@ -767,7 +767,7 @@ struct ArrayOpsBase<T, true>
   {
     // Similar to fill(), except we can allocate stack memory and placement-new
     // the object with a move constructor.
-    std::aligned_storage<sizeof(T), alignof(T)> host_buf;
+    alignas(T) unsigned char host_buf[sizeof(T)];
     T* host_obj = ::new(&host_buf) T(std::forward<Args>(args)...);
     axom::copy(array + i, host_obj, sizeof(T));
   }
--- a/src/axom/spin/ImplicitGrid.hpp
+++ b/src/axom/spin/ImplicitGrid.hpp
@@ -204,8 +204,20 @@ public:
     for(int i = 0; i < NDIMS; ++i)
     {
       m_bins[i] = BinSet(m_gridRes[i]);
-      m_binData[i] =
-        BinBitMap(&m_bins[i], BitsetType(numElts, allocatorID), 1, allocatorID);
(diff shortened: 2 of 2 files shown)


src/axom/spin/ImplicitGrid.hpp
```diff
@@ -204,8 +204,20 @@ class ImplicitGrid
     for(int i = 0; i < NDIMS; ++i)
     {
       m_bins[i] = BinSet(m_gridRes[i]);
-      m_binData[i] =
-        BinBitMap(&m_bins[i], BitsetType(numElts, allocatorID), 1, allocatorID);
```
kanye-quest, line 208: The issue is in this line - when each bin in the bin map is filled with copies of `slam::BitSet`, this copy constructor is called for each `BitSet`'s underlying word array:
https://github.com/LLNL/axom/blob/5f7869f21c4a2ee0ab0756e6f241690db6bcde86/src/axom/core/Array.hpp#L182-L188
Importantly, this constructor doesn't propagate allocator IDs, so even though we're passing a `BitSet` allocated in device memory, the resulting bitset in each bin is allocated in the default "dynamic" space - host memory.

src/axom/spin/ImplicitGrid.hpp
```diff
@@ -204,8 +204,20 @@ class ImplicitGrid
...
+
+      // Workaround because axom::Array doesn't propagate allocator ID on copy
+      // construction
+      {
+        typename BinBitMap::OrderedMap bitmaps(axom::IndexType {0},
+                                               axom::IndexType {m_gridRes[i]},
+                                               allocatorID);
+        for(int ibin = 0; ibin < m_gridRes[i]; ibin++)
+        {
+          bitmaps.emplace_back(numElts, allocatorID);
```
kanye-quest, line 217: Here's the fix - we emplace our BitSets into a temporary `axom::Array`, which avoids calling the copy constructor. Then we can just move that temporary array over the original bin map array.

src/axom/core/ArrayBase.hpp
```diff
@@ -767,7 +767,7 @@ struct ArrayOpsBase<T, true>
   {
     // Similar to fill(), except we can allocate stack memory and placement-new
     // the object with a move constructor.
-    std::aligned_storage<sizeof(T), alignof(T)> host_buf;
+    alignas(T) unsigned char host_buf[sizeof(T)];
```
kennyweiss, line 770: Minor:
Would it make sense to use a fixed width type, e.g.
```suggestion
    alignas(T) axom::uint8 host_buf[sizeof(T)];
```

src/axom/spin/ImplicitGrid.hpp
kennyweiss: Thanks for the pointer @kanye-quest 

Do we have a constructor that does propagate the allocator ID? and should we be using that here?

kennyweiss (approved):
Thanks @kanye-quest !

Two questions:
1. Do you have any performance numbers for the current diffs? 10x is a big difference, but I was wondering about the walltime differences. 
2. This fix seems a bit custom, is there a way that we can encapsulate this so other `axom::Array` users don't accidentally hit the same problem.

More details on (1) -- On a slightly longer timeframe, it would be good for us to benchmark some example runs so we can track performance regressions over time. This would likely be easier once we integrate caliper (#142)

agcapps (approved):
Once again, @kanye-quest , many thanks.

rhornung67 (approved):
Looks good @kanye-quest

PR #792, opened 2022-02-13
Bugfix for gcc@8.1 config without RAJA or Umpire

kennyweiss:
# Summary

- This PR is a bugfix for user-reported compiler errors in a `gcc@8.1` configuration without RAJA or Umpire.
   - We do not currently have a CI test for this configuration and I am not sure that we need one.
- GCC was complaining about several kernels in `MeshTester` that modified lambda-captured `axom::ArrayView`s passed in by-value (using `AXOM_LAMBDA`. The fix was to change the default lambda capture to `by-reference`
   - Update: This appears to be a compiler bug in `gcc@8.1` that is fixed in `gcc@8.3`
- I added the spec for the `gcc@8.3.1` compiler to our spack configuration on toss3
   - I tested this branch (locally) against `%gcc@8.3.1~cpp14+devtools+mfem+c2c` and `%gcc@8.3.1~cpp14+devtools+mfem+c2c~raja~umpire`, but did not add host-configs for these new specs
- Also fixes some errors and warning in configs where `OpenMP` is enabled, but not RAJA or Umpire

### DIFF
--- a/src/axom/core/Macros.hpp
+++ b/src/axom/core/Macros.hpp
@@ -79,10 +79,12 @@
  */
 #ifdef AXOM_USE_CUDA
   #define AXOM_LAMBDA [=] AXOM_HOST_DEVICE
+  #define AXOM_MUTABLE_LAMBDA [=] AXOM_HOST_DEVICE
   #define AXOM_DEVICE_LAMBDA [=] AXOM_DEVICE
   #define AXOM_HOST_LAMBDA [=] AXOM_HOST
 #else
   #define AXOM_LAMBDA [=]
+  #define AXOM_MUTABLE_LAMBDA [&]
   #define AXOM_DEVICE_LAMBDA [=]
   #define AXOM_HOST_LAMBDA [=]
 #endif
--- a/src/axom/core/Macros.hpp
+++ b/src/axom/core/Macros.hpp
@@ -79,12 +79,10 @@
  */
 #ifdef AXOM_USE_CUDA
   #define AXOM_LAMBDA [=] AXOM_HOST_DEVICE
-  #define AXOM_MUTABLE_LAMBDA [=] AXOM_HOST_DEVICE
   #define AXOM_DEVICE_LAMBDA [=] AXOM_DEVICE
   #define AXOM_HOST_LAMBDA [=] AXOM_HOST
 #else
   #define AXOM_LAMBDA [=]
-  #define AXOM_MUTABLE_LAMBDA [&]
   #define AXOM_DEVICE_LAMBDA [=]
   #define AXOM_HOST_LAMBDA [=]
 #endif
--- a/src/axom/quest/detail/MeshTester_detail.hpp
+++ b/src/axom/quest/detail/MeshTester_detail.hpp
@@ -23,6 +23,15 @@
 #include "axom/spin/BVH.hpp"
 #include "axom/spin/ImplicitGrid.hpp"
 
+// HACK: Workaround for known bug in gcc@8.1 which requires
+//       some lambdas in this file to have by-reference lambda capture
+#if defined(__GNUC__) && !defined(__llvm__) && !defined(__INTEL_COMPILER) && \
+  __GNUC__ == 8 && __GNUC_MINOR__ == 1
+  #define MESH_TESTER_MUTABLE_LAMBDA [&]
+#else
+  #define MESH_TESTER_MUTABLE_LAMBDA AXOM_LAMBDA
+#endif
+
 namespace axom
 {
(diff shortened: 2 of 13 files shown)


src/axom/quest/detail/MeshTester_detail.hpp
```diff
@@ -168,9 +168,9 @@ void CandidateFinderBase<ExecSpace, FloatType>::initialize()
   // if the Triangle is degenerate.
   mint::for_all_cells<ExecSpace, mint::xargs::coords>(
     m_surfaceMesh,
-    AXOM_LAMBDA(IndexType cellIdx,
-                numerics::Matrix<double> & coords,
-                const IndexType* nodeIds) {
+    [&](IndexType cellIdx,
```
kennyweiss, line 171: This was the bugfix:
* `AXOM_LAMBDA` adds a by-value lambda capture (`[=]`)
* Since we modify data in this kernel, we need a by-reference lambda capture (`[&]`)
* I am not sure why gcc@8.1 gave an error (related to modifying const data) but not other compilers

There are two other similar changes in this file.

src/axom/quest/examples/quest_signed_distance_interface.cpp
```diff
@@ -134,10 +139,10 @@ struct Arguments
     std::string pol_info =
       "Sets execution space of the SignedDistance query.\n";
     pol_info += "Set to \'seq\' to use sequential execution policy.";
-#ifdef AXOM_USE_OPENMP
+#if defined(AXOM_USE_OPENMP) && defined(AXOM_USE_RAJA)
```
kennyweiss, line 142: Note: `OpenMP` capability of `SignedDistance` query requires `RAJA` but was missing preprocessor guards

src/axom/core/Macros.hpp
```diff
@@ -79,10 +79,12 @@
...
   #define AXOM_LAMBDA [=] AXOM_HOST_DEVICE
+  #define AXOM_MUTABLE_LAMBDA [=] AXOM_HOST_DEVICE
   #define AXOM_DEVICE_LAMBDA [=] AXOM_DEVICE
   #define AXOM_HOST_LAMBDA [=] AXOM_HOST
 #else
   #define AXOM_LAMBDA [=]
+  #define AXOM_MUTABLE_LAMBDA [&]
   #define AXOM_DEVICE_LAMBDA [=]
   #define AXOM_HOST_LAMBDA [=]
 #endif
```
kennyweiss, line 90: I added a new macro: `AXOM_MUTABLE_LAMBDA` as discussed elsewhere in this PR.

We should discuss the necessity and/or name for this new macro.

kennyweiss: Note: nvcc does not allow the following: ` [&] AXOM_HOST_DEVICE `

agcapps (approved):
Thank you for adding the guards, @kennyweiss .

rhornung67 (approved):

nselliott (approved):

src/axom/quest/detail/MeshTester_detail.hpp
kanye-quest: That's curious, what was the error we were getting? I think longer-term we should aim to fix whatever's breaking in `ArrayView::operator[]`, but this workaround is fine for now.

kennyweiss: I confirmed that the problem is a bug in `gcc@8.1` that is fixed in `gcc@8.3`.

As such, I removed the `AXOM_MUTABLE_LAMBDA` macro from `Macros.hpp` in favor of a local macro in `MeshTester_detail.hpp` that uses by-reference lambda-capture for gcc@8.1 and `AXOM_LAMBDA` otherwise.

I tested this against all our toss3 and blueos configurations as well as against `gcc@8.1` and `gcc@8.3`, with and without RAJA + Umpire.

src/axom/core/Macros.hpp
kennyweiss: **Update:** I removed the `AXOM_MUTABLE_LAMBDA` macro from this file in favor of a local solution in `MeshTester_detail.hpp`

src/axom/quest/examples/quest_signed_distance_interface.cpp
```diff
@@ -21,13 +21,15 @@ using MPI_Comm = int;
 #endif
 
 // C/C++ includes
```
white238, line 23: In case you were confused these were C/C++ files =)

white238 (approved):
Thanks for the clean up and fixes!

PR #794, opened 2022-02-22
Bugfix/yang39/allocator propagation

### LINKED ISSUE #791: Clarify propagating behavior of axom::Array allocator
# Summary

`axom::Array` propagates the underlying allocator on move/move-assignment, but not on copy construction/assignment or on swap. I think we should change this to always propagate.

## Problem

This came up while I was looking into #789. Consider these examples for `axom::Array`:

```c++
    using ArrayType = axom::Array<int>;
    ArrayType a(N, allocID);
    ArrayType b = a; // has default host allocator ID
    ArrayType c = std::move(a) // propagates allocator ID
    ArrayType d = obj.get_some_array_value(); // propagates allocator ID, either through move or elided copy
    ArrayType e = obj.get_some_array_byref(); // has default host allocator, through copy constructor
    nested_array.some_func(a); // so what happens here?
```

Note the dependence on propagation behavior on whether the copy constructor/assignment operator or the move constructor/assignment operator is called. I think having variable propagation behavior is a recipe for confusion, as it makes Array behavior harder to understand at the callsite; we should standardize on a consistent behavior for all the cases.

[This proposal](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0177r2.html) also argues that best practice is to either always propagate, or never propagate the allocator.

## Arguing for always-propagate

Here are my arguments against no-propagation:
- No-propagation behavior doesn't feel that intuitive; the allocator ID is part of the array state, why wouldn't you copy that?
- It essentially turns the "dynamic" memory space into a "dynamic, but really host by default" space. Users that want to use allocator IDs to specify device arrays have to engage in more contortions to ensure that when arrays or objects holding arrays are copied, they stay in device space.
- The main use of not propagating is in models where allocators are passed down to nested objects - in that case, not propagating keeps the allocated space consistent. We don't do that in `axom::Array` right now, nor do I see how we'd be able to implement that for arbitrary objects.

With propagating allocator IDs, moves of memory between allocator spaces is made more explicit:
```c++
axom::Array<int> src(N, fromAllocID);
axom::Array<int> dst(N, toAllocID);
// from:
dst = src; // dependent on dst's allocator ID when this is called
// to:
dst = axom::Array<int>(src, toAllocID);
```

The only problem is that this is a breaking change -- certain areas in our code/user code may already be relying on the implicit memory moves.

kanye-quest:
# Summary

- Resolves #791
- `axom::Array` now consistently propagates the allocator ID for moves, copies, and swaps
  - This effectively corresponds to an `std::vector` instantiated with an allocator that defines the allocator traits `propagate_on_container_copy_assignment/move_assignment/swap` to `std::true_type`
  - If the destination array has an explicitly-specified memory space, the allocator ID is propagated from the source array only if the source allocator ID is compatible
  - This provides a more general fix to the regression resolved in #789

### DIFF
--- a/src/axom/core/Array.hpp
+++ b/src/axom/core/Array.hpp
@@ -210,10 +210,7 @@ public:
   /// @{
 
   /*! 
-   * \brief Copy assignment operator for Array 
-   *
-   * \note The data will be allocated using the allocator ID of the
-   *  copy-assigned Array, not the argument Array.
+   * \brief Copy assignment operator for Array
    * 
    * \pre T must be TriviallyCopyable
    */
@@ -222,6 +219,7 @@ public:
     if(this != &other)
     {
       static_cast<ArrayBase<T, DIM, Array<T, DIM, SPACE>>&>(*this) = other;
+      m_allocator_id = other.m_allocator_id;
       m_resize_ratio = other.m_resize_ratio;
       m_default_construct = other.m_default_construct;
       initialize(other.size(), other.capacity());
--- a/src/axom/core/Array.hpp
+++ b/src/axom/core/Array.hpp
@@ -810,6 +810,17 @@ Array<T, DIM, SPACE>::Array(const ArrayBase<T, DIM, OtherArrayType>& other,
   : ArrayBase<T, DIM, Array<T, DIM, SPACE>>(other)
   , m_allocator_id(allocatorId)
 {
+  // If a memory space has been explicitly set for the Array object, check that
+  // the space of the user-provided allocator matches the explicit space.
+  if(SPACE != MemorySpace::Dynamic &&
+     SPACE != axom::detail::getAllocatorSpace(m_allocator_id))
+  {
+#ifdef AXOM_DEBUG
+    std::cerr << "Incorrect allocator ID was provided for an Array object with "
+                 "explicit memory space - using default for space\n";
+#endif
+    m_allocator_id = axom::detail::getAllocatorID<SPACE>();
+  }
   initialize(static_cast<const OtherArrayType&>(other).size(),
              static_cast<const OtherArrayType&>(other).size());
   // axom::copy is aware of pointers registered in Umpire, so this will handle
@@ -827,6 +838,17 @@ Array<T, DIM, SPACE>::Array(const ArrayBase<const T, DIM, OtherArrayType>& other
   : ArrayBase<T, DIM, Array<T, DIM, SPACE>>(other)
   , m_allocator_id(allocatorId)
 {
+  // If a memory space has been explicitly set for the Array object, check that
+  // the space of the user-provided allocator matches the explicit space.
+  if(SPACE != MemorySpace::Dynamic &&
+     SPACE != axom::detail::getAllocatorSpace(m_allocator_id))
+  {
+#ifdef AXOM_DEBUG
+    std::cerr << "Incorrect allocator ID was provided for an Array object with "
+                 "explicit memory space - using default for space\n";
+#endif
+    m_allocator_id = axom::detail::getAllocatorID<SPACE>();
+  }
   initialize(static_cast<const OtherArrayType&>(other).size(),
              static_cast<const OtherArrayType&>(other).size());
   // axom::copy is aware of pointers registered in Umpire, so this will handle
--- a/src/axom/core/Array.hpp
+++ b/src/axom/core/Array.hpp
@@ -196,13 +196,12 @@ public:
    * \param [in] allocator_id the ID of the allocator to use (optional)
    */
   template <typename OtherArrayType>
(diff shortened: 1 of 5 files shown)


src/axom/core/tests/core_array.hpp
```diff
@@ -781,8 +782,9 @@ void check_device(Array<T, DIM, SPACE>& v)
   assign_raw<<<1, 1>>>(v.data(), size);
 
   // Check the contents of the array by assigning to a Dynamic array
-  // The default Umpire allocator should be Host, so we can access it from the CPU
-  Array<T, 1> check_raw_array_dynamic = v;
+  // The Umpire allocator should be Host, so we can access it from the CPU
+  int host_alloc_id = axom::getDefaultAllocatorID();
+  Array<T, 1> check_raw_array_dynamic(v, host_alloc_id);
```
kanye-quest, line 787: This PR does introduce a breaking change in that assigning from a device array to a dynamic array will not automatically copy data to the host; the intended replacement is to explicitly specify the host allocator ID as a constructor argument, or assign to an array explicitly-fixed as host by template parameter (see MeshTester_detail.hpp).

src/axom/core/Array.hpp
```diff
@@ -199,24 +196,20 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM, SPACE>>
    * \param [in] allocator_id the ID of the allocator to use (optional)
```
agcapps, line 196:
```suggestion
   * \param [in] allocator_id the ID of the allocator to use (optional).
   *     Default value of -1 causes the new Array to use the allocator
   *     of \a other.
```
or words to that effect.

src/axom/core/tests/core_array.hpp
agcapps: Please make sure the Sphinx documentation is up to date, and discusses this.

agcapps (approved):
@kanye-quest , thank you for this PR.  Please make sure the docs are up to date.

src/axom/core/Array.hpp
```diff
@@ -196,27 +193,27 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM, SPACE>>
    * \brief Constructor for transferring between memory spaces
    * 
    * \param [in] other The array in a different memory space to copy from
-   * \param [in] allocator_id the ID of the allocator to use (optional)
+   * \param [in] allocator_id the ID of the allocator to use (optional).
+   *     Default value of -1 causes the new Array to use the allocator
```
white238, line 197: This comment is confusing to me.

* default to use allocator of the other array
* otherwise you specify an allocator and it has to match the other array (then why are you specifying)
* OR you specify and it uses the default allocator, which is the allocator of other or is that the dynamic allocator

I feel like I'm missing something.

kanye-quest: So the copy constructor behavior is:
- if allocator ID is explicitly-specified, use that, otherwise default to use allocator ID of other array
- as a fallback, use default allocator for the memory space parameter if the allocator ID isn't compatible with the memory space
  - i.e. `Array<T, 1, Host> host_arr(device_arr)` - since `device_arr.getAllocatorID()` isn't compatible with the Host space, we'll use the default host allocator.
  - Note that we warn if the allocator ID is explicitly specified and isn't compatible with the memory space; if the incompatibility is because of an explicitly-specified memory space, we assume the user wants to copy into that space ala the above example.

I'll try and clarify the comments/documentation some more.

src/axom/core/tests/core_array.hpp
kennyweiss: Thanks -- this seems much more intuitive and consistent to me -- i.e. specify a new allocator ID when you need to. Use the same one when unspecified.

kennyweiss (approved):
Thanks @kanye-quest  -- I think this is a really nice improvement.

src/axom/core/Array.hpp
```diff
@@ -197,27 +194,32 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM, SPACE>>
...
+   *
+   *  If the memory space of the Array has been specified as a compile-time
+   *  template argument, and the resulting allocator does not refer to the same
+   *  memory space, the Array will instead be constructed with the default
+   *  allocator for the compile-time memory space.
    */
   template <typename OtherArrayType>
-  Array(const ArrayBase<T, DIM, OtherArrayType>& other,
-        int allocator_id = axom::detail::getAllocatorID<SPACE>());
+  Array(const ArrayBase<T, DIM, OtherArrayType>& other, int allocator_id = -1);
```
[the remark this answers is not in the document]
kanye-quest: I think we could probably split this into two constructors, one with and one without an allocator ID argument, to avoid having to define special allocator ID values.

PR #797, opened 2022-03-07
UniformGrid port to GPU

kanye-quest:
# Summary

Enables the use of `spin::UniformGrid` on the GPU:
- Template arguments are added for execution space and a storage policy
- Two storage policies are provided:
  - `DynamicGridStorage` stores the uniform grid as an array of arrays; this conforms to the previous `UniformGrid` behavior, and supports operations that directly modify bins, but isn't GPU-friendly
  - `FlatGridStorage` stores the uniform grid as a single array with offsets for each bin - preferred for GPUs
- Adds a vectorized `UniformGrid` constructor and `UniformGrid::initialize` method to run grid construction with OpenMP/GPU
- Adds a `UniformGrid::QueryObject` to support query operations on a uniform grid from within a device kernel
- Adds a `UniformGrid::getCandidatesAsArray()` method for querying candidate intersections with bounding boxes

In addition:
- A `quest::findTriMeshIntersectionsUniformGrid()` variant has been added to the mesh tester utility
- `axom::ArrayView<T>::subspan()` has been added to return subspans of the underlying data


agcapps (approved):
@kanye-quest , these look like a good set of changes.  Thanks for the improvements.

rhornung67 (approved):
Looks great @kanye-quest Thanks!

RELEASE-NOTES.md
```diff
@@ -49,6 +49,20 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Adds ability to build axom using the `hip` compiler. Support for running device
   kernels with hip will be added in the future.
 - Adds new host-configs for HIP on LLNL platforms
+- Adds GPU/OpenMP support to `spin::UniformGrid`.
+  The following functions run with a user-specified execution space (specified as a template argument
+  on `UniformGrid`):
+  - `UniformGrid::initialize()`: creates/re-creates a uniform grid with an array of objects and their
+    corresponding bounding boxes
+  - `UniformGrid::getCandidatesAsArray()`: query the uniform grid with for objects that share a grid
```
kennyweiss, line 57: Minor typo:
```suggestion
  - `UniformGrid::getCandidatesAsArray()`: query the uniform grid for objects that share a grid
```

src/axom/quest/MeshTester.cpp
```diff
@@ -118,13 +113,12 @@ void findTriMeshIntersections(detail::UMesh* surface_mesh,
     size_t checkcount = binsToCheck.size();
     for(size_t curbin = 0; curbin < checkcount; ++curbin)
     {
-      std::vector<int> ntlist = ugrid.getBinContents(binsToCheck[curbin]);
-      std::vector<int>::iterator ntlit = ntlist.begin(), ntlend = ntlist.end();
-      for(; ntlit != ntlend; ++ntlit)
+      axom::ArrayView<int> ntlist = ugrid.getBinContents(binsToCheck[curbin]);
```
kennyweiss, line 116: +1 for conversion from vector to `axom::Array` and `axom::ArrayView`.

We should consider modifying the MeshTester API to remove `std::vector` in a follow-up PR.

src/axom/spin/UniformGrid.hpp
```diff
@@ -183,7 +231,6 @@ class UniformGrid
 
 private:
   /*!
-   *****************************************************************************
```
kennyweiss, line 186: 👍 (for consistent style)

src/axom/spin/UniformGrid.hpp
```diff
@@ -451,18 +702,214 @@ void UniformGrid<T, NDIMS>::insert(const BoxType& BB, const T& obj)
...
+        {
+  #ifndef __CUDA_ARCH__
+          std::sort(candidates_view.begin() + startIdx,
+                    candidates_view.begin() + startIdx + count);
+  #endif
+        }
+      });
+  }
+
+  // Step 6: Count and flag unique intersection pairs, in order to map them
```
kennyweiss, line 807: Long-shot: Does RAJA have a direct way to sort and unique-ify an array?

kennyweiss (approved):
This is really nice @kanye-quest 

I'd love to see performance numbers and/or comparisons to the other spatial indexes if/when you have them.

src/axom/spin/UniformGrid.hpp
kanye-quest: I don't think there is; in our use case it's complicated by the fact that we only want to unique-ify within individual result slices; not sure if we can really express that generically in RAJA.

PR #799, opened 2022-03-11
Ensure slic is initialized with a minimal config.

agcapps:
# Summary

- This PR is a new feature.
- It does the following:
  - Modifies SLIC static functions to ensure SLIC is initialized.
  - Fixes issue #280 

# Design

If SLIC methods were called without slic::initialize() being called, log messages would be lost.  This PR reflects the end state of the discussion on the issue page.  It modifies SLIC to ensure at least a minimal initialization is in place so developers don't waste time wondering where the log messages went, or worse, not realizing their log messages are being lost.


rhornung67 (approved):
I think this looks right to me. Thanks @agcapps I'll defer to others as to whether this covers the cases users have run into.

kennyweiss (approved):
Thanks @agcapps!

A few quick questions:
* Do we need to update any of the Fortran API via shroud? 
  Perhaps not since this function is called internally.

* Are there any sidre tests that should be updated as part of this PR? 
  Note: If slic is not initialized when creating a `DataStore`, sidre will initialize slic.
  See: https://github.com/LLNL/axom/blob/3fb7909d36255a3cf19325221561f2a6501f70b8/src/axom/sidre/core/DataStore.cpp#L88-L103

agcapps:
@kennyweiss  wrote:
>     * Do we need to update any of the Fortran API via shroud?
>       Perhaps not since this function is called internally.

@ltaylor16 , what do you think?  I added a new function `slic::ensureInitialized()` called from slic's internals but visible externally.

agcapps:
@kennyweiss  wrote:
>     * Are there any sidre tests that should be updated as part of this PR?
>       Note: If slic is not initialized when creating a `DataStore`, sidre will initialize slic.
>       See: https://github.com/LLNL/axom/blob/3fb7909d36255a3cf19325221561f2a6501f70b8/src/axom/sidre/core/DataStore.cpp#L88-L103

I don't think we need to change any sidre tests.  We make sure to use slic properly (as your link shows); initializing before and finalizing after all uses.  Also all the sidre tests pass with the changes in this PR.

white238:
We discussed the impact on Fortran in the meeting today.  We decided to address this in a followup PR that also adds a Fortran specific example.

src/axom/slic/tests/slic_uninit.cpp
```diff
@@ -0,0 +1,118 @@
...
+  testInit("logWarningMessage", [fname, line]() {
+    axom::slic::logWarningMessage("a warning message", fname, line);
+  });
+}
+
+TEST(slic_uninit, streams)
+{
+  testInit("flushStreams", []() { axom::slic::flushStreams(); });
+  testInit("pushStreams", []() { axom::slic::pushStreams(); });
+}
```
white238, line 118:
```suggestion
}

```

src/axom/slic/interface/slic.cpp
```diff
@@ -24,74 +25,69 @@ void initialize() { Logger::initialize(); }
...
+    disableAbortOnWarning();
+    setLoggingMsgLevel(slic::message::Debug);
+    std::string format = std::string("<TIMESTAMP>\n") +
+      std::string("[<LEVEL>]: <MESSAGE> \n") + std::string("FILE=<FILE>\n") +
+      std::string("LINE=<LINE>\n\n");
+    addStreamToAllMsgLevels(new slic::GenericOutputStream(&std::cout, format));
+
+    logMessage(
+      message::Warning,
+      "slic::initialize() must be called before any other calls to SLIC\n."
```
white238, line 43: Is calling `slic::initialize()` enough?  What happens if they call it then dont' create a Logger?

agcapps: `slic::initialize()` calls `static Logger::initialize`, which creates and activates Logger "root."  So I think it is enough.

white238 (approved):

PR #801, opened 2022-03-22
Remove corona allocate/release from gitlab

bmhan12:
This PR removes the allocate and release stages from the corona job.

### DIFF
--- a/.gitlab/build_corona.yml
+++ b/.gitlab/build_corona.yml
@@ -17,36 +17,12 @@
     - when: on_success
 
 ####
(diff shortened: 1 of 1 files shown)


.gitlab/build_corona.yml
```diff
@@ -16,37 +16,13 @@
       when: always
```
white238, line 16: The above if needs to be removed as well

rhornung67 (approved):

kennyweiss (approved):
Thanks @bmhan12 !

PR #804, opened 2022-03-25
Device optimization for axom::Array

kanye-quest:
# Summary

Currently, `axom::Array<T>` initializes memory on the device by creating an array on the host, then copying it to the device. This is to work around limitations where `T` has a non-trivial default constructor, as that constructor would need to be device-annotated.

However, if `T` is trivially default-constructible, the host-allocate-and-copy workaround is unnecessary and may cost some performance. In this PR, we launch a kernel for default-initialization if the type `T` is trivially default-constructible.

As a bonus, we replace the use of SFINAE in `detail::ArrayOpsBase` with tag-dispatch.

## Performance

Measured on rzansel with `./bin/mesh_tester --method [accel_method] -p 3 --skipWeld -i [test_mesh]`:

| Test case | `develop` | this PR | speedup |
| ---------- | ---------- | ------- | -------- |
| car.stl - bvh | 0.358s | 0.248s | 1.45x |
| car.stl - implicit | 2.439s | 1.413s | 1.73x | 
| car.stl - uniform | 2.643s | 0.795s | 3.32x |
| porsche.stl - bvh | 0.131s | 0.128s | 1.03x |
| porsche.stl - implicit | 0.207s | 0.159s | 1.3x |
| porsche.stl - uniform | 0.277s | 0.172s | 1.61x |

### DIFF
--- a/src/axom/core/ArrayBase.hpp
+++ b/src/axom/core/ArrayBase.hpp
@@ -732,20 +732,18 @@ struct ArrayOpsBase;
 template <typename T>
 struct ArrayOpsBase<T, false>
 {
-  static constexpr bool DefaultCtor = std::is_default_constructible<T>::value;
+  using DefaultCtorTag = std::is_default_constructible<T>;
 
   /*!
-   * \brief Default-initializes the "new" segment of an array
+   * \brief Helper for default-initializing the "new" segment of an array
    *
    * \param [inout] data The data to initialize
    * \param [in] begin The beginning of the subset of \a data that should be initialized
    * \param [in] end The end index (exclusive) of the subset of \a data that is to be initialized
    * \note Specialization for when T is default-constructible.
    */
-  template <bool HasCtor = DefaultCtor>
-  static typename std::enable_if<HasCtor>::type init(T* data,
-                                                     IndexType begin,
-                                                     IndexType nelems)
+  template <typename U = void>
+  static void init_impl(T* data, IndexType begin, IndexType nelems, std::true_type)
   {
     for(IndexType i = 0; i < nelems; ++i)
     {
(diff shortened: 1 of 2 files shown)


src/axom/core/ArrayBase.hpp
```diff
@@ -728,20 +732,18 @@ struct ArrayOpsBase;
...
+  using DefaultCtorTag = std::is_default_constructible<T>;
 
   /*!
-   * \brief Default-initializes the "new" segment of an array
+   * \brief Helper for default-initializing the "new" segment of an array
    *
    * \param [inout] data The data to initialize
    * \param [in] begin The beginning of the subset of \a data that should be initialized
    * \param [in] end The end index (exclusive) of the subset of \a data that is to be initialized
    * \note Specialization for when T is default-constructible.
```
kennyweiss, line 743: Suggestion: It might be helpful to add a doxygen `tparam` for `U` and, e.g. how it is related to the note about default-constructibility (here and elsewhere)

src/axom/core/ArrayBase.hpp
```diff
@@ -728,20 +732,18 @@ struct ArrayOpsBase;
...
-  static constexpr bool DefaultCtor = std::is_default_constructible<T>::value;
+  using DefaultCtorTag = std::is_default_constructible<T>;
 
   /*!
-   * \brief Default-initializes the "new" segment of an array
+   * \brief Helper for default-initializing the "new" segment of an array
    *
    * \param [inout] data The data to initialize
    * \param [in] begin The beginning of the subset of \a data that should be initialized
    * \param [in] end The end index (exclusive) of the subset of \a data that is to be initialized
```
kennyweiss, line 742: Please update the doxygen from `end` to `nelems`   (here and elsewhere).

Note: It looks like this was out of date before this PR, but it would still be good to fix it here.

kennyweiss (approved):
Thanks @kanye-quest -- this is really nice!
Thanks for posting the improved timings as well.

Are there any tests related to this change that should be added/updated?

src/axom/core/ArrayBase.hpp
kanye-quest: Those dummy template parameters were originally intended to avoid instantiating the helpers with non-matching tags; however, apparently we get "lazy" instantiation of member functions for free on a template class.

kennyweiss: (Based on experience) it's worth testing that conjecture with msvc... it can sometimes exhibit different behavior than other compilers on this sort of thing

kanye-quest: We don't have a unit test for this, but the `nontrivial_ctor_obj` test in `core_array_for_all` should exercise the non-trivially constructible path for both host and device memory. (we also have tests for nontrivial destructor and copy constructor)

I think we could consider adding a test case for trivial type initialization (int or double), and a test case for non-trivally default-constructible type.

agcapps (approved):
Thanks for this work, @kanye-quest .  I'll appreciate the extra testing with MSVC and your Doxygenation.

rhornung67 (approved):
looks good @kanye-quest thank you.

PR #805, opened 2022-03-29
Guard compiler defines in fmt

### LINKED ISSUE #802: Axom's fmt still bleeds into applications
If I include fmt/core.h and axom/fmt/fmt.hpp (the latter through axom/quest.hpp), I run into compilation errors. Basically, whichever is included first wins, and since the namespaces used are different, the code will always fail to compile.

```
In file included from axom/include/axom/fmt.hpp:15:
axom/include/axom/fmt/chrono.h:312:23: error: no member named 'fmt' in namespace 'axom'; did you mean simply 'fmt'?
      using namespace axom::fmt::detail;
                      ^

axom/include/axom/fmt/chrono.h:20:1: note: 'fmt' declared here
FMT_BEGIN_NAMESPACE
^
fmt/include/fmt/core.h:238:15: note: expanded from macro 'FMT_BEGIN_NAMESPACE'
    namespace fmt {           \
              ^
```

I doubt this is the best solution, but the only foolproof way I can think of to fix this right now is to prefix all macros in Axom's fmt with AXOM_.

white238:
Fixes #802 

@adayton1

### DIFF
--- a/src/thirdparty/CMakeLists.txt
+++ b/src/thirdparty/CMakeLists.txt
@@ -57,6 +57,14 @@ mark_as_advanced(CLI11_FOUND)
 # Register the built-in fmt library for string formatting
 #------------------------------------------------------------------------------
 
+# NOTE: When updating fmt you must do the following changes to properly guard
+#  our fmt from other fmt's.
+# 
+# 1) Change the namespace from 'fmt' to 'axom::fmt'.  This is in fmt/core.h:203-204,
+#    as well as, search and replace `fmt::` to `axom::fmt`.
+# 2) Search and replace compiler define's prefix from `FMT_` to `AXOM_FMT_`.
+# 3) Keep fmt's headers installed under axom's include directory.
+
 # Headers that Axom adds to fmt
 set(axom_fmt_headers
     axom/fmt.hpp
--- a/src/thirdparty/CMakeLists.txt
+++ b/src/thirdparty/CMakeLists.txt
@@ -88,8 +88,8 @@ target_include_directories(fmt INTERFACE
             $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include> )
 
 # Setup some variables for fmt in Axom's config.hpp
-set(FMT_EXCEPTIONS FALSE PARENT_SCOPE)
-set(FMT_HEADER_ONLY TRUE PARENT_SCOPE)
+set(AXOM_FMT_EXCEPTIONS FALSE PARENT_SCOPE)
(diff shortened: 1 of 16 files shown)


davidbeckingsale (approved):
Looks good to me. I did not check every single instance since I'm assuming you used sed.

white238:
> Looks good to me. I did not check every single instance since I'm assuming you used sed.

I did a search and replace on every file.  I initially walked through several files to make sure I wasn't missing some case then blanket did every file.

kennyweiss (approved):
Thanks @white238 !

Since we'll need to reapply this whenever we update `fmt` should we have a script checked in to apply these (perhaps in a follow-up PR) ?

In either case, please document somewhere that this change needs to happen when we update `fmt`.  Some possibilities:
* `fmt.hpp` 
* `thirdparty/CMakeLists.txt`
* somewhere in the sphinx dev guide, e.g. https://axom.readthedocs.io/en/develop/docs/sphinx/dev_guide/updating_tpls.html#updating-tpls

white238:
> Thanks @white238 !
> 
> Since we'll need to reapply this whenever we update `fmt` should we have a script checked in to apply these (perhaps in a follow-up PR) ?
> 
> In either case, please document somewhere that this change needs to happen when we update `fmt`. Some possibilities:
> 
>     * `fmt.hpp`
> 
>     * `thirdparty/CMakeLists.txt`
> 
>     * somewhere in the sphinx dev guide, e.g. https://axom.readthedocs.io/en/develop/docs/sphinx/dev_guide/updating_tpls.html#updating-tpls

Done.

src/thirdparty/CMakeLists.txt
```diff
@@ -57,6 +57,14 @@ mark_as_advanced(CLI11_FOUND)
 # Register the built-in fmt library for string formatting
 #------------------------------------------------------------------------------
 
+# NOTE: When updating fmt you must do the following changes to properly guard
+#  our fmt from other fmt's.
+# 
+# 1) Change the namespace from 'fmt' to 'axom::fmt'.  This is in fmt/core.h:203-204,
+#    as well as, search and replace `fmt::` to `axom::fmt`.
+# 2) Search and replace compiler define's prefix from `FMT_` to `AXOM_FMT_`.
+# 3) Keep fmt's headers installed under axom's include directory.
```
kennyweiss, line 66: Thanks @white238 !

adayton1 (approved):
LGTM

rhornung67 (approved):
Thanks @white238 for working through this.

agcapps (approved):

rhornung67 (approved):

PR #807, opened 2022-04-04
Fix libyogrt dependency for SCR

nselliott:
# Summary

- This PR is a build system/TPL bug fix
  - SCR depends on the library `libyogrt` and we had incorrect references to the spack package name for that library. It is `libyogrt` and not `yogrt`.  This was causing configurations to skip including SCR.


kennyweiss (approved):
Thanks @nselliott !

rhornung67 (approved):
Thanks @nselliott

PR #808, opened 2022-04-08
Miscellaneous axom::Array improvements

kanye-quest:
# Summary

This is an assortment of changes for `axom::Array`.

Public interface changes:
- Adds an initializer list constructor and assignment operator for 1-D arrays
- Adds an uninitialized memory overload for `Array::resize()`
- The "uninitialized-ness" of an Array is no longer a property of the array instance; instead, memory is only left uninitialized at the point of a call to the constructor or `Array::resize()`.

Robustness fixes:

- `ArrayOps::fill_range` now accepts a memory space argument for the source array; this allows for the proper invocation of copy constructors on the host side for arrays of a non-trivially-copyable type.
- Calls to `axom::copy` in the Array copy constructor have been replaced with a call to `ArrayOp::fill_range`.
- The type trait `HasTrivialCopyCtor<T>` which aliases `std::is_trivially_copy_constructible<T>` has been replaced with `TriviallyCopyable<T>`, aliasing `std::is_trivially_copyable<T>`. This fixes device-side `Array::fill()` for some types.

### DIFF
--- a/src/axom/core/Array.hpp
+++ b/src/axom/core/Array.hpp
@@ -630,7 +630,23 @@ public:
    * \note Reallocation is done if the new size will exceed the capacity.
    */
   template <typename... Args>
-  void resize(Args... args);
+  void resize(Args... args)
+  {
+    static_assert(sizeof...(Args) == DIM,
+                  "Array size must match number of dimensions");
+    const StackArray<IndexType, DIM> dims {static_cast<IndexType>(args)...};
+    resize(dims, true);
+  }
+
+  /// \overload
+  template <typename... Args>
+  void resize(ArrayOptions::Uninitialized, Args... args)
+  {
+    static_assert(sizeof...(Args) == DIM,
+                  "Array size must match number of dimensions");
+    const StackArray<IndexType, DIM> dims {static_cast<IndexType>(args)...};
+    resize(dims, false);
+  }
 
   /*!
    * \brief Exchanges the contents of this Array with the other.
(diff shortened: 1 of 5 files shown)


src/axom/core/Array.hpp
```diff
@@ -649,22 +700,34 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM, SPACE>>
    * \post size() == num_elements
    * \post getResizeRatio() == DEFAULT_RESIZE_RATIO
    */
-  void initialize(IndexType num_elements, IndexType capacity);
+  void initialize(IndexType num_elements,
+                  IndexType capacity,
+                  bool default_construct = true);
```
kennyweiss, line 705: Suggestion (minor): it might be clearer if the new parameter were named `should_default_construct`

kennyweiss (approved):
Thanks @kanye-quest !

src/axom/core/ArrayBase.hpp
```diff
@@ -18,6 +18,9 @@
 // C/C++ includes
 #include <iostream>  // for std::cerr and std::ostream
 #include <numeric>   // for std::accumulate
+#if defined(__GLIBCXX__) && !defined(_GLIBCXX_USE_CXX11_ABI)
```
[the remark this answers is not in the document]
kanye-quest: We only use this fallback for compilers using GCC 4.9.3's standard library, since the C++11 type traits we needed weren't implemented until GCC 5. The TR1 fallback should be available for that compiler library version; for other libraries, we'd just expect full C++11 compliance.

src/axom/core/ArrayBase.hpp
```diff
@@ -19,6 +19,7 @@
 #include <iostream>  // for std::cerr and std::ostream
 #include <numeric>   // for std::accumulate
 #if defined(__GLIBCXX__) && !defined(_GLIBCXX_USE_CXX11_ABI)
+  // Workaround for unimplemented C++11 type traits in libstdc++ versions <5
```
kennyweiss, line 22: 👍

rhornung67 (approved):
Thanks @kanye-quest

kennyweiss:
Could you please also update the checklist on https://github.com/LLNL/axom/issues/633 ?

PR #809, opened 2022-04-11
Fortran Slic Example

### LINKED ISSUE #29: Add Fortran Slic Logging example
Add a Fortran example that illustrates how to use Slic for logging.

bmhan12:
This PR adds Fortran examples for Slic. Resolves #29

- Additional functions were added to `slic_shroud.yaml` to wrap the slic interface. Notably missing was `addStreamToMsgLevel/addStreamToAllMsgLevels` and the ability to create a stream.
- For now, limited to creating  `GenericOutputStream` stream objects in Fortran.
  - Added constructor overloads for GenericOutputStream that take in a `std::string` instead of `std::ostream`
- Added overloads for addStreamToMsgLevel/addStreamToAllMsgLevels that take in `GenericOutputStream` instead of `LogStream` to bypass derived types in Fortran.
  - Derived types is possible in shroud after talking to @ltaylor16, is looking into a usage example.
  - Does look to require updating Axom's shroud
    - Shroud PR that adds single-level inheritance (9/3/20): https://github.com/LLNL/shroud/pull/188
    - Currently Axom is on 0.12.2 (8/5/20): [compilers.yaml](https://github.com/LLNL/axom/blob/3fb7909d36255a3cf19325221561f2a6501f70b8/scripts/spack/configs/toss_3_x86_64_ib/packages.yaml#L253), [tagged shroud versions](https://github.com/LLNL/shroud/tags)
- Added a `replaceAllInstances` function to StringUtilities that finds all substrings in a given string and replaces it with another substring.
    - Used to fix escape characters in the Slic format string passed from Fortran to C++.

The `wrap*.***` and `slic_types.yaml` files are shroud-generated.


src/axom/slic/interface/slic.hpp
```diff
@@ -154,13 +155,28 @@ void setAbortFunction(AbortFunctionPtr abort_func);
  */
 void addStreamToMsgLevel(LogStream* ls, message::Level level);
 
+/*!
+ * \brief Adds the given GenericOutputStream to the the given level.
```
kennyweiss, line 159: Minor typo:
```suggestion
 * \brief Adds the given GenericOutputStream to the given level.
```

(The same typo appears a few lines up with the original function :)

kennyweiss (approved):
Thanks @bmhan12 !

src/axom/slic/examples/basic/flogger.f
```diff
@@ -0,0 +1,45 @@
...
+  implicit none
+
+  type(SlicGenericOutputStream) fileStream
+  type(SlicGenericOutputStream) consoleStream
+  integer n
+  real(C_DOUBLE) :: u
+  integer(C_INT) :: random_msg_level
+
+  ! STEP 0: Initialize logger
+  call slic_initialize()
```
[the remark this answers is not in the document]
kennyweiss: Based on a discussion with @agcapps earlier today, I suspect that the test suggested in my previous comment will always work within axom's tests (which use only `libaxom`. 

The problem might be when we have another library that depends on axom and there are multiple copies of the `s_logger` "singleton" (see #270)

src/axom/slic/interface/c_fortran/typesSLIC.h
```diff
@@ -14,15 +14,23 @@
...
+  void *addr; /* address of C++ memory */
+  int idtor;  /* index of destructor */
+};
+typedef struct s_SLIC_GenericOutputStream SLIC_GenericOutputStream;
+
 // helper capsule_data_helper
 struct s_SLIC_SHROUD_capsule_data
 {
-  void* addr; /* address of C++ memory */
+  void *addr; /* address of C++ memory */
```
ltaylor16, line 28: These formatting changes strike me as odd since clangformat should convert it to `void*` from the Shroud generated `void *`. I edited the file to use `void*` then clangformat moved it back. Maybe we're treating C and C++ files differently.

This is more of a general problem and should be fixed in another PR.

src/axom/core/utilities/StringUtilities.hpp
```diff
@@ -174,6 +174,23 @@ std::string removeBeforeDelimiter(const std::string& str, const char delim = '/'
 std::string removeAllInstances(const std::string& target,
                                const std::string& substr);
 
+/*!
+*****************************************************************************
+* \brief This function finds all instances of a substring and replaces it
+* with another string for the target string
+*
+* \param [in] target The string to operate on
```
ltaylor16, line 182: Should this argument be `inout`?

ltaylor16 (approved):
Looks good.

agcapps (approved):
Thanks much for this, @bmhan12 !

src/axom/core/utilities/StringUtilities.hpp
bmhan12: In the example from [cppreference](https://en.cppreference.com/w/cpp/string/basic_string/replace) @kennyweiss gave it is `inout`, but I have it as an `[in]` here to be consistent with the other functions in `StringUtilities`.

PR #812, opened 2022-04-26
Update BLT, Conduit, RAJA, Umpire + CUDA, HIP version update

white238:
* This requires C++14.
* Needed for new BLT HIP support.
* Currently on fork of Spack to bring in and test @davidbeckingsale 's Spack PR for updating a bunch of packages (RAJA, Umpire, etc)
* Updates to ubuntu 20 docker images due to new clingo requirement

Current known issues:

- Intel@18 with gcc@7.3 takes 2.5 hours to build just Axom (not including TPLs) and also fails the sidre_mfem_datacollection unit test.


rhornung67 (commented):
Apart from the CI test failures, this looks great 😄

white238:
> Apart from the CI test failures, this looks great 😄

We shouldn't let small details like not building stop us from progress!

rhornung67:
Godspeed @white238 !

.uberenv_config.json
```diff
@@ -3,8 +3,8 @@
 "package_version": "develop",
 "package_final_phase": "initconfig",
 "package_source_dir": "../..",
-"spack_url": "https://github.com/spack/spack",
-"spack_commit": "23ddfba16ddc376c1a8ec77ffee731117b4fabf8",
+"spack_url": "https://github.com/davidbeckingsale/spack.git",
```
kennyweiss, line 6: Note @davidbeckingsale -- Looks like we're temporarily pointing at your fork of spack.
Please don't remove (rewrite) this commit until we can update back to spack proper.

src/CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ if (“${PROJECT_SOURCE_DIR}” STREQUAL “${CMAKE_SOURCE_DIR}”)
     # Set some default BLT options before loading BLT only if not included in
     # another project
     if (NOT BLT_CXX_STD)
-        set(BLT_CXX_STD "c++11" CACHE STRING "")
+        set(BLT_CXX_STD "c++14" CACHE STRING "")
```
kennyweiss, line 47: Please update our `RELEASE-NOTES` that the default is now `c++14` and that `c++14` is required for or RAJA and Umpire TPLs.

src/cmake/thirdparty/SetupAxomThirdParty.cmake
```diff
@@ -27,6 +27,9 @@ if (UMPIRE_DIR)
         message(FATAL_ERROR "Given UMPIRE_DIR is not a directory: ${UMPIRE_DIR}")
     endif()
 
+    # white238 - Workaround to get around umpire's exported camp logic not working
+    set(camp_DIR "${CAMP_DIR}/lib/cmake/camp")
```
kennyweiss, line 31: Is there an issue for this?

Tag: @davidbeckingsale

kennyweiss (commented):
I might have jumped the gun on my review.. just noticed that it's still WIP...

white238:
> I might have jumped the gun on my review.. just noticed that it's still WIP...

Do I need to change the title to "KENNY CALM DOWN THIS IS A WIP: Update BLT, RAJA, Umpire"?

rhornung67:
@kennyweiss back away from @white238 slowly, no sudden movements.... 😄

src/cmake/thirdparty/SetupAxomThirdParty.cmake
davidbeckingsale: This is not really an issue - we modified our installed CMake configs to handle relocation, but that means we no longer hard code any camp_DIR in the umpire (or RAJA) packages. Best practice is to treat camp as a true separate TPL, which means providing a camp_DIR - you could run find_package(camp) in Axom to have more control, or you can defer the search for camp to the Umpire/RAJA packages (but it still requires a camp_DIR).

white238:
Got intel@18 to work by just turning off lua (which turns of Sol which is the real problem).  @kennyweiss says that's fine for his code since he doesn't use that functionality.

It does on the other-hand take 2.5 hours to build Axom now with intel@18. I am going to see if I can find the specific link that causes this really long time and selectively turn it off if it is not needed.

white238:
> Got intel@18 to work by just turning off lua (which turns of Sol which is the real problem). @kennyweiss says that's fine for his code since he doesn't use that functionality.
> 
> It does on the other-hand take 2.5 hours to build Axom now with intel@18. I am going to see if I can find the specific link that causes this really long time and selectively turn it off if it is not needed.

maybe spoke too soon. the `sidre_mfem_datacollection` unit tests are failing when reloading.

scripts/spack/packages/axom/package.py
```diff
@@ -222,7 +228,7 @@ def initconfig_compiler_entries(self):
                 entries.append(cmake_cache_string("BLT_EXE_LINKER_FLAGS", flags,
                                                   description))
 
-        if "+cpp14" in spec:
+        if "+cpp14" in spec and spec.satisfies("@0.6.2:"):
```
white238, line 231: In retrospect this is not entirely correct...

white238:
This PR is now ready for review.

scripts/spack/packages/umpire/package.py
```diff
@@ -183,6 +189,16 @@ def initconfig_hardware_entries(self):
                 arch_str = ",".join(archs)
                 entries.append(cmake_cache_string(
                     "HIP_HIPCC_FLAGS", '--amdgpu-target={0}'.format(arch_str)))
+            # BEGIN AXOM EDIT
+            # Fix blt_hip getting HIP_CLANG_INCLUDE_PATH-NOTFOUND bad include directory
+            if self.spec.satisfies('%clang'):
```
bmhan12, line 194: Whoops, I missed this. This should be guarded by the system type as done for raja.

scripts/spack/packages/raja/package.py
```diff
@@ -107,22 +120,32 @@ def initconfig_hardware_entries(self):
...
-                "HIP_ROOT_DIR", hip_root))
+                "HIP_ROOT_DIR", '{0}'.format(spec['hip'].prefix)))
+            # BEGIN AXOM EDIT
+            # Fix blt_hip getting HIP_CLANG_INCLUDE_PATH-NOTFOUND bad include directory
+            if self.spec.satisfies('%clang') and 'toss_4' in self._get_sys_type(spec):
+                hip_root = spec['hip'].prefix
+                rocm_root = hip_root + "/.."
+                clang_version= str(self.compiler.version)
+                hip_clang_include_path = rocm_root + "/llvm/lib/clang/" + clang_version + "/include"
+                entries.append(cmake_cache_path("HIP_CLANG_INCLUDE_PATH", hip_clang_include_path))
```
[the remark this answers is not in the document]
bmhan12: I'll add this change to raja, and umpire too.

kennyweiss (approved):
Thanks @white238 and @bmhan12 !

Were you able to resolve the `MFEMSidreDatacollection` issue?

src/thirdparty/axom/sol.hpp
```diff
@@ -11468,11 +11468,13 @@ namespace stack {
 	struct popper {
 		inline static decltype(auto) pop(lua_State* L) {
 			record tracking{};
-#ifdef __INTEL_COMPILER
-			auto&& r = get<T>(L, -lua_size<T>::value, tracking);
-#else
+// AXOM EDIT START: sol.hpp(11477): error: an rvalue reference cannot be bound to an lvalue
```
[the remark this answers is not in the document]
white238: This is to fix intel@19 with C++14. I could not get intel@18 to work at all with C++14.  If I remember correctly this toggled back and forth between 18 and 19 and since noone uses sol with intel@18, I decided to give up.  This was not the only error with intel@18.

white238:
> Thanks @white238 and @bmhan12 !
> 
> Were you able to resolve the `MFEMSidreDatacollection` issue?

I did not have time to put effort into it.  I am also not sure when it started to fail since CI was already off for intel@18/19 due to a previous decision about time limits.  I am willing to assist in tracking it down after the PR goes through.

kennyweiss:
> > Thanks @white238 and @bmhan12 !
> > Were you able to resolve the `MFEMSidreDatacollection` issue?
> 
> I did not have time to put effort into it. I am also not sure when it started to fail since CI was already off for intel@18/19 due to a previous decision about time limits. I am willing to assist in tracking it down after the PR goes through.

Thanks @white238 -- that sounds reasonable to me. Could you please create a ticket to track that?

src/thirdparty/axom/sol.hpp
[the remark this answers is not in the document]
kennyweiss: Thanks. 

It's a bit difficult to see from github, but it looks like we no longer have an axom-specific patch here for sol. If so, should we remove these comment lines and the associated lines from the patch file? (or am I missing something?)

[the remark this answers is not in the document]
white238: This is an axom specific patch, and it has the markings. There is also a patch file.

RELEASE-NOTES.md
```diff
@@ -70,6 +70,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Adds examples for using the BVH device traversal API
 
 ###  Changed
+- Axom now requires C++14 and will default to that if not specified via `BLT_CXX_STD`.
```
kanye-quest, line 73: 🎉

scripts/spack/packages/axom/package.py
```diff
@@ -252,10 +258,12 @@ def initconfig_hardware_entries(self):
                 entries.append(
                     "# cuda_arch could not be determined\n\n")
 
-            if "+cpp14" in spec:
-                cudaflags += " -std=c++14"
-            else:
-                cudaflags += " -std=c++11"
+            if spec.satisfies("^blt@:0.6.1"):
```
kanye-quest, line 261: Not too clear on Spack syntax: does this specify a "BLT at least version 0.6.1" check, or is this checking for BLT with the exact version number? If it's the latter, is there a way of specifying the former?

kanye-quest (approved):
LGTM overall, just a few thoughts

scripts/spack/packages/axom/package.py
kennyweiss: I think it means "If we're using a BLT up to and including 0.6.1, do the following"
(It is the `:` right after the the `@` that specifies the end of the range, but leaves the beginning of the range open).

kanye-quest: Ah gotcha, thanks!

scripts/spack/configs/toss_3_x86_64_ib/compilers.yaml
```diff
@@ -3,32 +3,32 @@ compilers:
...
     spec: clang@9.0.0
     target: x86_64
 - compiler:
     environment: {}
     extra_rpaths: []
     flags:
-      cflags: --gcc-toolchain=/usr/tce/packages/gcc/gcc-8.1.0
-      cxxflags: --gcc-toolchain=/usr/tce/packages/gcc/gcc-8.1.0
+      cflags: --gcc-toolchain=/usr/tce/packages/gcc/gcc-8.3.1
+      cxxflags: --gcc-toolchain=/usr/tce/packages/gcc/gcc-8.3.1
```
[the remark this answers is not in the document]
kennyweiss: Sounds like a good idea, unless there's a specific reason we're using `gcc@8.1.0` w/ `intel@19` instead of `gcc@8.3.1`.
(and if so, it would be good to document that).

Since the TPLs can take a bit to rebuild, perhaps it's something we can upgrade in the future though.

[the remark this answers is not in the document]
white238: We really should do a full evaluation of what compilers we are building. I have a feeling we have drifted completely away from what our customers are using.

I agree though I should have bumped it across the board.

scripts/spack/packages/axom/package.py
white238: The syntax mimics the python list slice syntax and is inclusive of the number given.  Which can be a bit odd and sometimes you have to create a version that doesn't exist.

PR #818, opened 2022-05-06
Adds first cut at Distributed Closest Point query with example application

kennyweiss:
# Summary

- This PR is a feature
- It adds a `DistributedClosestPoint` class to quest, which solves the following problem:
    - Given an "object" mesh composed of a collection of points distributed across MPI ranks
       and a "query" mesh composed of a collection of points, distributed across the MPI ranks
    - Find the closest point in the object mesh to each point in the query mesh. (Also find the rank containing the closest point and the index of the object on that rank).
    - This can be used, e.g. to compute the distance and/or direction to the object
- The class takes the dimension (2 or 3) and the execution space (sequential CPU, openmp CPU or CUDA) as runtime parameters
- The current approach to solving the distributed problem is for each rank to generate a spatial index over its "object mesh" and to send its "query mesh" data to every other rank to update the closest point. This will be optimized in future PRs by only sending data between ranks that could affect the closest point.

- This PR also adds an example executable that uses this class. 
   - The example uses a circle as the object mesh and the vertices of a mesh in the mesh blueprint format as the query mesh
   - The latter can be generated for a Cartesian mesh in 2D or 3D or from an mfem mesh using axom's `data_collection_util`.

Here are two example runs:
```bash
## Generate a 30x30 Cartesian mesh distributed over 3 ranks 
## and run the distributed closest point query against a circular object w/ 100 samples per rank
srun -n3 ./bin/data_collection_util -p1 --min -1.5 -1.5 --max 1.5 1.5 --res 30 30
srun -n3 ./examples/quest_distributed_distance_query_ex -m box_2d.root -n 100

## Generate a Star mesh over 36 ranks, query against a circular object
cp <path/to/mfem>/data/star.mesh star.mesh
srun -n36 ./bin/data_collection_util -p1 -f 1 -m  star.mesh -l 4 --scale 1.25
srun -n36 ./examples/quest_distributed_distance_query_ex -m star.root -n 100
```

Here is a visualization of a 12 rank run on LLNL's blueos (3 nodes with 4 GPUs each) on a 300x300 Cartesian mesh:
dcp_12rank_gpu_300x300
The left side shows two derived fields of the closest point -- distance and direction to the object.
The right side shows the ranks containing the object and query points.

Here is a similar visualization of a 36 rank run on LLNL's toss3 platform on the star mesh:
dcp_36rank_cpu_star

#### Known limitations
We will follow up with future PRs to:
* Add support for 3D queries to the application. I.e. the DistributedClosestPoint class is set up for 3D, but I haven't tested this yet
*  Optimize the MPI communications. The current code sends data from each rank to every other rank.
* Remove unnecessary memory transfers. The current code contains some extra memory copies when running on the CPU. These can be eliminated
* Improve the robustness of how the input meshes are laid out. The code currently assumes the coordinate data and computed closest_point fields have interleaved data, e.g. `xyzxyzxyz....`. It should also support coordinates in three separate stride-1 arrays, e.g. `xxx.....yyy...zzz....`

### DIFF
--- a/src/axom/quest/DistributedClosestPoint.hpp
+++ b/src/axom/quest/DistributedClosestPoint.hpp
@@ -370,7 +370,7 @@ public:
    * \brief Computes the closest point within the objects for each query point
    * in the provided particle mesh, provided in the mesh blueprint rooted at \a meshGroup
    *
-   * \param meshGroup The root of a mesh blueprint for the query points
+   * \param mesh_node The root node of a mesh blueprint for the query points
    * \param coordset The coordinate set for the query points
    *
    * Uses the \a coordset coordinate set of the provided blueprint mesh
@@ -845,13 +845,25 @@ private:
  * Each of these are distributed over the problem's mpi ranks. This class orchestrates passing
  * the query points to all ranks whose object meshes might contain a closest point.
  *
+ * \note This class assumes that the object mesh and the query mesh are decomposed over the same
+ * number of mpi ranks.
+ * \warning This class will need to support cases where some ranks have zero object points.
+ * This is not currently supported.
+ *
  * \note The class currently supports object meshes that are comprised of a collection of points.
  * In the future, we'd like to consider more general object meshes, e.g. triangle meshes.
+ *
+ * \note The class currently supports object meshes and query meshes with a single domain per MPI rank.
+ * We intend to add support for multiple computational domains on each rank.
  *
  * To use this class, first set some parameters, such as the runtime execution policy,
  * then pass in the object mesh and build a spatial index over this mesh.
  * Finally, compute the closest points in the object mesh to each point in a query mesh 
  * using the \a computeClosestPoint() function.
+ *
+ * \note The implementation currently assumes that the coordinates for the positions and vector field
+ * data are interleaved (i.e. xyzxyzxyz....). We will relax this assumption in the future to support both
+ * interleaved and strided data.
  */
 class DistributedClosestPoint
 {
--- a/src/axom/quest/DistributedClosestPoint.hpp
+++ b/src/axom/quest/DistributedClosestPoint.hpp
@@ -1,0 +1,180 @@
+// Copyright (c) 2017-2022, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level LICENSE file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#ifndef QUEST_DISTRIBUTED_CLOSEST_POINT_H_
+#define QUEST_DISTRIBUTED_CLOSEST_POINT_H_
+
+#include "axom/core.hpp"
+#include "axom/slic.hpp"
+#include "axom/slam.hpp"
+#include "axom/primal.hpp"
+#include "axom/mint.hpp"
+#include "axom/spin.hpp"
+
+#include "axom/fmt.hpp"
+
+#include <list>
+#include <vector>
+#include <set>
+#include <cstdlib>
+#include <cmath>
+
+namespace axom
+{
+namespace quest
+{
+template <int NDIMS = 2, typename ExecSpace = axom::SEQ_EXEC>
+class DistributedClosestPoint
+{
+public:
+  // TODO: generalize to 3D
+  static_assert(NDIMS == 2,
+                "DistributedClosestPoint only currently supports 2D");
+
+  static constexpr int DIM = NDIMS;
+  using PointType = primal::Point<double, DIM>;
+  using BoxType = primal::BoundingBox<double, DIM>;
+  using PointArray = axom::Array<PointType>;
+  using BoxArray = axom::Array<BoxType>;
+  using BVHTreeType = spin::BVH<DIM, ExecSpace>;
+
+private:
+  struct MinCandidate
+  {
+    /// Squared distance to query point
+    double minSqDist {numerics::floating_point_limits<double>::max()};
+    /// Index within mesh of closest element
+    int minElem;
+  };
+
+public:
+  DistributedClosestPoint(
+    int allocatorID = axom::execution_space<ExecSpace>::allocatorID())
+    : m_allocatorID(allocatorID)
+  { }
+
+public:  // Query properties
+  void setVerbosity(bool isVerbose) { m_isVerbose = isVerbose; }
+
+public:
+  /// Utility function to generate an array of 2D points
+  void generatePoints(double radius, int numPoints)
+  {
+    using axom::utilities::random_real;
+
+    // Generate in host because random_real is not yet ported to the device
+    PointArray pts;
+    pts.reserve(numPoints);
+    for(int i = 0; i < numPoints; ++i)
+    {
+      const double angleInRadians = random_real(0., 2 * M_PI);
+      const double rsinT = radius * std::sin(angleInRadians);
+      const double rcosT = radius * std::cos(angleInRadians);
+
+      pts.emplace_back(PointType {rcosT, rsinT});
+    }
+
+    if(m_isVerbose)
+    {
+      SLIC_INFO("Points on object:");
+      const auto& arr = m_points;
+      for(auto i : slam::PositionSet<int>(arr.size()))
+      {
+        const double mag = sqrt(arr[i][0] * arr[i][0] + arr[i][1] * arr[i][1]);
+        SLIC_INFO(axom::fmt::format("\t{}: {} -- {}", i, arr[i], mag));
+      }
+    }
+
+    m_points = PointArray(pts, m_allocatorID);  // copy to ExecSpace
+  }
+
+  bool generateBVHTree()
+  {
+    const int npts = m_points.size();
+    axom::Array<BoxType> boxesArray(npts, npts, m_allocatorID);
+    auto boxesView = boxesArray.view();
+    axom::for_all<ExecSpace>(
+      npts,
+      AXOM_LAMBDA(axom::IndexType i) { boxesView[i] = BoxType {m_points[i]}; });
+
+    // Build bounding volume hierarchy
+    m_bvh.setAllocatorID(m_allocatorID);
+    int result = m_bvh.initialize(boxesView, npts);
+
+    return (result == spin::BVH_BUILD_OK);
+  }
+
+  void computeClosestPoints(const PointArray& queryPts,
+                            axom::Array<axom::IndexType>& cpIndexes) const
+  {
+    SLIC_ASSERT(!queryPts.empty());
+
+    const int nPts = queryPts.size();
+
+    /// Create an ArrayView in ExecSpace that is compatible with cpIndexes
+    axom::Array<axom::IndexType> cp_idx(nPts, nPts, m_allocatorID);
+    auto query_inds = cp_idx.view();
+
+    /// Create an ArrayView in ExecSpace that is compatible with queryPts
+    PointArray execPoints(nPts, nPts, m_allocatorID);
+    execPoints = queryPts;
+    auto query_pts = execPoints.view();
+
+    // Get a device-useable iterator
+    auto it = m_bvh.getTraverser();
+
+    using axom::primal::squared_distance;
+    using int32 = axom::int32;
+
+    AXOM_PERF_MARK_SECTION(
+      "ComputeClosestPoints",
+      axom::for_all<ExecSpace>(
+        nPts,
+        AXOM_LAMBDA(int32 idx) mutable {
+          PointType qpt = query_pts[idx];
+
+          MinCandidate curr_min {};
+
+          auto searchMinDist = [&](int32 current_node, const int32* leaf_nodes) {
+            const int candidate_idx = leaf_nodes[current_node];
+            const PointType candidate_pt = m_points[candidate_idx];
+            const double sq_dist = squared_distance(qpt, candidate_pt);
+
+            if(sq_dist < curr_min.minSqDist)
+            {
+              curr_min.minSqDist = sq_dist;
+              curr_min.minElem = candidate_idx;
+            }
+          };
+
+          auto traversePredicate = [&](const PointType& p,
+                                       const BoxType& bb) -> bool {
+            return squared_distance(p, bb) <= curr_min.minSqDist;
+          };
+
+          // Traverse the tree, searching for the point with minimum distance.
+          it.traverse_tree(qpt, searchMinDist, traversePredicate);
+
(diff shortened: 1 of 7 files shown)


src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -0,0 +1,1093 @@
...
+  bool m_isVerbose {false};
+  int m_allocatorID;
+  int m_rank;
+  int m_nranks;
+
+  PointArray m_points;
+  BoxArray m_boxes;
+
+#if defined(AXOM_USE_RAJA) && defined(AXOM_USE_UMPIRE)
+  SeqBVHTree* m_bvh_seq {nullptr};
```
kanye-quest, line 847: I think `spin::BVH` should be usable without RAJA or Umpire in the sequential case.

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -0,0 +1,1093 @@
...
+
+    MPI_Comm_rank(MPI_COMM_WORLD, &m_rank);
+    MPI_Comm_size(MPI_COMM_WORLD, &m_nranks);
+  }
+
+  ~DistributedClosestPointImpl()
+  {
+#if defined(AXOM_USE_RAJA) && defined(AXOM_USE_UMPIRE)
+    delete m_bvh_seq;
+    m_bvh_seq = nullptr;
```
kanye-quest, line 286: Minor: maybe consider making these BVH instance pointers `std::unique_ptr`, and deleting the destructor?

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -0,0 +1,1093 @@
...
+ * using the \a computeClosestPoint() function.
+ */
+class DistributedClosestPoint
+{
+public:
+  using RuntimePolicy = DistributedClosestPointRuntimePolicy;
+
+  ~DistributedClosestPoint()
+  {
+    delete m_dcp_2;
```
kanye-quest, line 884: Same here wrt `unique_ptr`/rule of zero

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -0,0 +1,1093 @@
...
+ * \param [in] send_rank MPI rank to send \a send_node to
+ * \param [in] recv_rank MPI rank to receive \a recv_node from
+ * \param [in] tag tag for MPI messages
+ * \param [in] comm MPI communicator to use
+ * 
+ * Sends/receives a pair of conduit nodes (along with their schemas) from the current rank.
+ * Uses non-blocking send (isend) and blocking receives and ensures that the request
+ * objects associated with the send are finalized
+ */
+inline void send_and_recv_node(conduit::Node& send_node,
```
kanye-quest, line 211: Have you taken a look at the `communicate_using_schema` helper class in Relay for setting up isends/irecvs?

kanye-quest (approved):
LGTM overall 👍Just a few minor comments/suggestions.

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
```diff
@@ -0,0 +1,881 @@
...
+  queryTimer.stop();
+
+  // Output some timing stats
+  auto getMinMax =
+    [](double inVal, double& minVal, double& maxVal, double& sumVal) {
+      MPI_Allreduce(&inVal, &minVal, 1, MPI_DOUBLE, MPI_MIN, MPI_COMM_WORLD);
+      MPI_Allreduce(&inVal, &maxVal, 1, MPI_DOUBLE, MPI_MAX, MPI_COMM_WORLD);
+      MPI_Allreduce(&inVal, &sumVal, 1, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD);
+    };
+  {
```
agcapps, line 796: Why do you introduce a scope here?  And if a scope is desirable here, perhaps to group variables that aren't elsewhere used, why not put `getMinMax` inside it?

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
```diff
@@ -0,0 +1,881 @@
...
+  axom::ArrayView<T> getNodalVectorField(const std::string& fieldName) const
+  {
+    SLIC_ASSERT_MSG(hasPoints(),
+                    "Cannot extract a field from the BlueprintParticleMesh "
+                    "before adding points");
+
+    T* data = hasField(fieldName)
+      ? static_cast<T*>(
+          m_fieldsGroup->getView(axom::fmt::format("{}/values/x", fieldName))
+            ->getVoidPtr())
```
agcapps, line 327: I don't understand how the `x` values give a vector field.  Please document the assumptions on data storage (such as interleaved, separate, etc.)

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -0,0 +1,1093 @@
...
+   * \param [in] coords The root group of a mesh blueprint's coordinate values
+   * \note This function currently supports mesh blueprints with the "point" topology
+   */
+  void importObjectPoints(const conduit::Node& coords, int nPts)
+  {
+    // Extract pointers to the coordinate data
+    // Note: The following assumes the coordinates are contiguous with stride NDIMS
+    // TODO: Generalize to support other strides
+
+    // TODO: Add error checking for children 'x', 'y' and 'z' and striding
```
agcapps, line 314: Thank you!

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -0,0 +1,1093 @@
...
+ * the query points to all ranks whose object meshes might contain a closest point.
+ *
+ * \note The class currently supports object meshes that are comprised of a collection of points.
+ * In the future, we'd like to consider more general object meshes, e.g. triangle meshes.
+ *
+ * To use this class, first set some parameters, such as the runtime execution policy,
+ * then pass in the object mesh and build a spatial index over this mesh.
+ * Finally, compute the closest points in the object mesh to each point in a query mesh 
+ * using the \a computeClosestPoint() function.
+ */
```
agcapps, line 876: Please note mesh assumptions and preconditions, including that the object and point meshes should be decomposed over the same number of domains, and that the work is decomposed along the object mesh decomposition.  No redecomposition is done.  Thank you for the chat where we discussed this.

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -0,0 +1,1093 @@
...
+    auto dumpNode = [=](const conduit::Node& n,
+                        const std::string&& fname,
+                        const std::string& protocol = "json") {
+      conduit::relay::io::save(n, fname, protocol);
+    };
+
+    // create conduit node containing data that has to xfer between ranks
+    conduit::Node xfer_node;
+    {
+      // clang-format off
```
agcapps, line 427: thank goodness for this ability

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -0,0 +1,1093 @@
...
+
+    return false;
+  }
+
+  /**
+   * \brief Computes the closest point within the objects for each query point
+   * in the provided particle mesh, provided in the mesh blueprint rooted at \a meshGroup
+   *
+   * \param meshGroup The root of a mesh blueprint for the query points
+   * \param coordset The coordinate set for the query points
```
agcapps, line 395: for clarity, change `meshGroup` to `mesh_node` (see line 410)

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -0,0 +1,1093 @@
...
+
+    if(m_nranks > 1)
+    {
+      // arbitrary tags for sending data to other ranks and getting it back
+      const int tag_before = 1234;
+      const int tag_after = 4321;
+
+      // NOTE: uses a naive algorithm to computed distributed closest points
+      // Every rank sends it data to every other rank.
+      // TODO: Devise a more efficient algorithm to only send data to ranks with closer points
```
agcapps, line 483: 👍

agcapps (approved):
Thank you for this, Kenny.  Looks good.  I noted a few suggestions.  I also appreciated the bullet points you summarized as "Known Limitations," especially the last point regarding input format.  Please add comment(s) explaining this.

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
```diff
@@ -0,0 +1,881 @@
...
+  double circleRadius {1.0};
+  int circlePoints {100};
+  RuntimePolicy policy {RuntimePolicy::seq};
+
+private:
+  bool m_verboseOutput {false};
+
+  // clang-format off
+  const std::map<std::string, RuntimePolicy> s_validPolicies{
+    #if defined(AXOM_USE_RAJA) && defined(AXOM_USE_UMPIRE)
```
white238, line 74: Why is RAJA and Umpire guarding seq? Shouldn't that be there regardless?  What happens when all 4 of these are off? I don't see this example guarded by any of these in the CMake but I could have missed it.

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
```diff
@@ -0,0 +1,881 @@
...
+      pmesh->GetBoundingBox(mins, maxs);
+      numElements = pmesh->ReduceInt(numElements);
+      myRank = pmesh->GetMyRank();
+    }
+    else
+    {
+      mesh->GetBoundingBox(mins, maxs);
+    }
+
+    if(myRank == 0)
```
white238, line 627: todo to move to SLIC_INFO_ROOT when my pr goes in?

src/axom/quest/DistributedClosestPoint.hpp
kennyweiss: Thanks @kanye-quest -- that looks promising.

@gunney1, when you're improving the MPI part of this, please look into if we can use and/or adapt `communicate_using_schema` to schedule the async send/receives.

https://github.com/LLNL/conduit/blob/3edfd63095da3e7a94ea2ae4537abb5ae7b4b547/src/libs/relay/conduit_relay_mpi.hpp#L256-L266

src/axom/quest/DistributedClosestPoint.hpp
kennyweiss: done

src/axom/quest/DistributedClosestPoint.hpp
kennyweiss: done

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
kennyweiss: The scope was just to make the logging block a bit cleaner. 
I'll add a comment at the top to make it a bit clearer.

The `getMinMax` is used in a different scope below, so cannot be in this one.

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
kennyweiss: Thanks! I added a comment about this.

src/axom/quest/DistributedClosestPoint.hpp
kennyweiss: Thanks. I also added a note that the class will need to support the case where a rank has an empty object mesh (i.e. no object points), but that this is not currently supported/tested in this initial version.

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
```diff
@@ -0,0 +1,881 @@
...
+                        minPts,
+                        maxPts,
+                        sumPts / num_ranks));
+  }
+
+  mfem::Array<int> dofs;
+  for(auto i : IndexSet(nQueryPts))
+  {
+    const auto& cp = cpPositions[i];
+    (*distances)(i) = sqrt(squared_distance(qPts[i], cp));
```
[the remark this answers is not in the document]
kennyweiss: This doesn't override `i`, it just uses `i` as the loop variable.
I uses a `slam::PositionSet` as a personal preference since it has more concise notation.

I changed to variable from `i` to `idx` to (hopefully) make it a bit clearer.

rhornung67 (approved):
This looks really good @kennyweiss

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
```diff
@@ -0,0 +1,881 @@
...
+    directions->MakeOwner(vfec);
+    m_dc.RegisterField("direction", directions);
+  }
+
+  /// Prints some info about the mesh
+  void printMeshInfo()
+  {
+    switch(m_dc.GetMesh()->Dimension())
+    {
+    case 2:
```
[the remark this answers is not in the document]
kennyweiss: Thanks. I added a `SLIC_ERROR_IF` to ensure that `DIM` can only be 2 or 3.

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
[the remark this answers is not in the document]
white238: oh its the () operator.. got it.. carry on you don't need to change the variable name. i'm just dumb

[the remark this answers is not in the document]
kennyweiss: I agree that syntax is a bit confusing since it is a pointer to a `GridFunction`, and it uses `operator()` instead of `operator[]`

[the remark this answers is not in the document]
white238: This is what happens when i look at PRs past my bedtime.

white238 (approved):

kennyweiss:
Thanks for the great feedback everyone.

I've addressed all concerns other than the unnecessary RAJA guards on the sequential policy. 
Working on that now before merging...

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
kennyweiss: Thanks -- I fixed the guards -- this query no longer requires RAJA or Umpire for the sequential policy

src/axom/quest/DistributedClosestPoint.hpp
kennyweiss: Thanks -- I fixed the guards -- this query no longer requires RAJA or Umpire for the sequential policy

PR #822, opened 2022-05-20
Correct error in example of BVH usage.

gunney1:
# Summary

- This PR fixes a code documentation error

I think the example should not use double to represent a BoundingBox.  Can't see how we can represent a bounding box with a single number.  The fix is more wordy, but hopefully correct.


agcapps (approved):
`aabbs` gets passed as the first argument to `bvh.initialize()`.  That parameter is a template type, and BoundingBoxes get read out of `aabbs` using square brackets.  So yes, I think this change is a good one that will increase clarity.

kanye-quest (approved):
Nice catch. Thanks @gunney1!

rhornung67 (approved):

src/axom/spin/BVH.hpp
```diff
@@ -94,7 +94,7 @@ struct BVHPolicy<FloatType, NDIMS, ExecType, BVHType::LinearBVH>
  *     constexpr int DIMENSION = 3;
  *
  *     // get a list of axis-aligned bounding boxes in a flat array
- *     const double* aabbs = ...
+ *     const primal::BoundingBox<float, DIMENSION>* aabbs = ...
```
kennyweiss, line 97: Thanks @gunney1 -- this correction is certainly an improvement.

The situation is a bit more complicated since the code will work over any type that satisfies the `BoxIndexable` requirements. This includes `primal::BoundingBox` as well as `ZipIndexable<primal::BoundingBox>`.

We should probably document this better (in a future PR), e.g. here or in the BVH docs: https://axom.readthedocs.io/en/develop/axom/spin/docs/sphinx/bvh.html

kennyweiss (approved):
Thanks @gunney1

PR #565, opened 2021-05-22
Adds the multimat component to Axom

kennyweiss:
# Summary

- This PR adds a new component to Axom: Multmat
- Multimat helps manage multimaterial field data
- The vast majority of multimat was developed by @raineyeh over Summer 2018 in collaboration with @kennyweiss and @agcapps 
- I rebased the code on top of our current develop branch

###### TODO:

- [x] Merge in @raineyeh's branch -- https://github.com/LLNL/axom/pull/612
- [x] Fix compiler errors and warnings on LC host-configs
- [x] Rebase against axom@develop
- [ ] Update RELEASE-NOTES
- [x] Add basic docs

PR #823, opened 2022-05-24
[Multimat 2/n] Fix lifetime issues in Slam

kanye-quest:
# Summary

- Fixes a lifetime issue with `slam::BivariateMap`:
  - Adds functionality to `slam::Map` to store its corresponding `Set` by value if passed a reference to a set
  - Changes `slam::BivariateMap` to use `slam::Map` with mapped-from set stored by-value; this ensures that `slam::BivariateMap` is trivially-relocatable
  - Removes ability to set `BivariateSet` pointer as internally-managed for now; this could lead to a potential use-after-free

This is related to the Windows test failures on #565 -- MSVC is apparently less aggressive at eliding the copies out of return statements compared to gcc or clang. This behavior can be replicated in gcc or clang by adding the `-fno-elide-constructors` switch.

The test added in this PR demonstrates the issue as well, by creating a `BivariateMap` within a scoped-block, then copy-assigning it to a variable outside of the block and attempting to use the outer variable.


src/axom/slam/Map.hpp
```diff
@@ -89,6 +89,42 @@ class Map : public MapBase<typename S::PositionType>, public StrPol
...
+      }
+      else
+      {
+        return &m_set;
+      }
+    }
+
+    const USet* m_pSet {nullptr};
+    USet m_set;
+  };
```
kennyweiss, line 126: This is nice! 
I wonder if it is more generally useful, e.g. outside of Map's implementation.

src/axom/slam/Map.hpp
```diff
@@ -114,6 +150,26 @@ class Map : public MapBase<typename S::PositionType>, public StrPol
...
+      DataType defaultValue = DataType(),
+      SetPosition stride = StridePolicyType::DEFAULT_VALUE,
+      int allocatorID = axom::getDefaultAllocatorID())
+    : StridePolicyType(stride)
+    , m_set(theSet)
+  {
+    static_assert(std::is_same<SetType, USet>::value,
+                  "Argument set is of a more-derived type than the Map's set "
+                  "type. This may lead to object slicing. Use Map's pointer "
+                  "constructor instead to store polymorphic sets.");
```
kennyweiss, line 168: 👍

src/axom/slam/tests/slam_map_BivariateMap.cpp
```diff
@@ -414,6 +414,74 @@ TEST(slam_bivariate_map, iterate)
...
+        EXPECT_EQ(*ptr, getVal<T>(idx1, idx2, i));
+      }
+}
+
+TEST(slam_bivariate_map, testScopedMapBehavior)
+{
+  using BSet = BivariateSetType;
+  using IndPol = STLIndirection<double>;
+
+  testScopedCopyBehavior<double, BSet, IndPol, StrideOneType>(1);
```
kennyweiss, line 474: Thanks for tracking this down and adding these tests!

kennyweiss (approved):
Thanks @kanye-quest !

white238 (approved):

PR #824, opened 2022-05-25
Implement threshold to reduce computations in closest point query.

gunney1:
# Summary

- This PR is a feature
  - Adds a distance threshold to the distributed closest point query so users can avoid searching spaces known to yield unproductive results, at the request of @kennyweiss 

Added method setDistanceThreshold is to DistributedClosestPoint class.
Added command-line switch, --dist-threshold to quest_distributed_distance_query_example so user can specify the threshold when running the example.

### DIFF
--- a/src/axom/quest/DistributedClosestPoint.hpp
+++ b/src/axom/quest/DistributedClosestPoint.hpp
@@ -22,6 +22,7 @@
 #include "conduit_relay_io.hpp"
 
 #include <memory>
+#include <limits>
 #include <cstdlib>
 #include <cmath>
 
@@ -280,11 +281,22 @@ public:
   DistributedClosestPointImpl(RuntimePolicy runtimePolicy, bool isVerbose)
     : m_runtimePolicy(runtimePolicy)
     , m_isVerbose(isVerbose)
+    , m_sqDistanceThreshold(std::numeric_limits<double>::max())
   {
     setAllocatorID();
 
     MPI_Comm_rank(MPI_COMM_WORLD, &m_rank);
     MPI_Comm_size(MPI_COMM_WORLD, &m_nranks);
+  }
+
+  /**
+   * \brief Sets the threshold for the query
+   *
+   * \param [in] threshold Ignore distances greater than this value.
+   */
+  void setSquaredDistanceThreshold(double sqThreshold)
+  {
+    m_sqDistanceThreshold = sqThreshold;
   }
 
 public:
@@ -754,7 +766,9 @@ public:
 
           auto traversePredicate = [&](const PointType& p,
                                        const BoxType& bb) -> bool {
-            return squared_distance(p, bb) <= curr_min.minSqDist;
+            auto sqDist = squared_distance(p, bb);
+            return sqDist <= curr_min.minSqDist &&
+              sqDist <= m_sqDistanceThreshold;
           };
 
           // Traverse the tree, searching for the point with minimum distance.
@@ -829,6 +843,7 @@ public:
 private:
   RuntimePolicy m_runtimePolicy;
   bool m_isVerbose {false};
+  double m_sqDistanceThreshold;
   int m_allocatorID;
   int m_rank;
   int m_nranks;
@@ -932,6 +947,17 @@ public:
     m_dimension = dim;
   }
 
+  /**
+   * \brief Sets the threshold for the query
+   *
+   * \param [in] threshold Ignore distances greater than this value.
+   */
+  void setDistanceThreshold(double threshold)
+  {
+    SLIC_ERROR_IF(threshold < 0.0, "Distance threshold must be non-negative.");
+    m_sqDistanceThreshold = threshold * threshold;
+  }
+
   /// Sets the logging verbosity of the query. By default the query is not verbose
   void setVerbosity(bool isVerbose) { m_isVerbose = isVerbose; }
 
@@ -1021,9 +1047,11 @@ public:
     switch(m_dimension)
     {
     case 2:
+      m_dcp_2->setSquaredDistanceThreshold(m_sqDistanceThreshold);
       m_dcp_2->computeClosestPoints(query_node, cooordset);
       break;
     case 3:
+      m_dcp_3->setSquaredDistanceThreshold(m_sqDistanceThreshold);
       m_dcp_3->computeClosestPoints(query_node, cooordset);
       break;
     }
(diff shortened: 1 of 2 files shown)


src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -280,13 +281,24 @@ class DistributedClosestPointImpl
...
   }
 
+  /**
+   * \brief Sets the threshold for the query
+   *
+   * \param [in] threshold Ignore distances greater than this value.
+   */
+  void setSquaredDistanceThreshold(double sqThreshold)
+  {
+    m_sqDistanceThreshold = sqThreshold;
```
agcapps, line 299: Consider a check for non-negative value (such as at (new) line 957 in `setDistanceThreshold`.

agcapps (approved):
Thanks, @gunney1 .

src/axom/quest/DistributedClosestPoint.hpp
gunney1: I considered it and originally decided that the check would be redundant because we have a check in the public interface already.  Do it anyway?

agcapps: Well, `setSquaredDistanceThreshold()` is a `public` member function and does have Doxygen, so I'd recommend checking here.  Some enterprising user will be sure to find it and try calling it, doubtless with values that "for sure" won't be negative : )

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
```diff
@@ -61,6 +61,8 @@ struct Input
   int circlePoints {100};
   RuntimePolicy policy {RuntimePolicy::seq};
 
+  double distThreshold {1e100};
```
kennyweiss, line 64: Any reason not to make this default to `std::numeric_limits<double>::max()` ?

kennyweiss (commented):
Thanks @gunney1!

What does the code return for the "closest_point" when there are no object points within the threshold radius? Do we need any special treatment? Similarly, do we need special treatment for cases where there are no closest points within the threshold on the initial rank, but there are closest points on other ranks?

I'd also like to see a visualization of the results for these cases.
I'd be happy to meet to show you how to generate these.

src/axom/quest/DistributedClosestPoint.hpp
gunney1: Okay.  Will do.  Thanks.

gunney1: I pushed the check.

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
gunney1: The reason is that we'd have to square it and compare the squares -> floating point overflow.  We could set it to sqrt(numeric_limits<double>::max()) - epsilon.  But I thought that would be a weird number that may be confusing to look at in a debugger.  What do you think?

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
```diff
@@ -109,6 +111,10 @@ struct Input
       ->description("Radius for circle")
       ->capture_default_str();
 
+    app.add_option("-d,--dist-threshold", distThreshold)
+      ->description("Distance threshold to search")
+      ->capture_default_str();
```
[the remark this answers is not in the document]
gunney1: Will do.  Thanks for the pointer.  I'm using `axom::CLI::NonNegativeNumber`.  I thought that zero may be good value to set when debugging.

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
kennyweiss: It would be good to check that passing in `numeric_limits<double>::max` doesn't do anything stange.

Does the following help test against expected behavior?

```cpp
#include <iostream>
#include <limits>


int main () {

  double max_1 = std::numeric_limits<double>::max();
  double max_2 = max_1 * max_1;

  std::cout << "numeric limits max: " << max_1 << '\n';
  std::cout << "numeric_limits max squared: " << max_2 << '\n';

  for(double d : {-5., 0., 100., std::numeric_limits<double>::max(), max_1, max_2})
  {
    for(double m : {max_1, max_2})
    {
        if( d < m) { std::cout << d << " < " << m << std::endl; }
        else if( d == m) { std::cout << d << " == " << m << std::endl; }
        else if( d > m) { std::cout << d << " > " << m << std::endl; }
    }
  }

  return 0;
}
```
https://godbolt.org/z/99rzoTao7

Here is the output of some comparisons (using gcc@12.1):
```
numeric limits max: 1.79769e+308
numeric_limits max squared: inf
-5 < 1.79769e+308
-5 < inf
0 < 1.79769e+308
0 < inf
100 < 1.79769e+308
100 < inf
1.79769e+308 == 1.79769e+308
1.79769e+308 < inf
1.79769e+308 == 1.79769e+308
1.79769e+308 < inf
inf > 1.79769e+308
inf == inf
```

That might imply that if we pass in `numeric_limits<double>::max()` then our comparisons for squared distances will be against `inf`.

gunney1: Yes.  That looks like using `max()` should work.  Thanks!

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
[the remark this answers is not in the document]
kennyweiss: Thanks @gunney1 -- that's what I was hoping to see.

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
```diff
@@ -58,9 +59,12 @@ struct Input
   std::string meshFile;
 
   double circleRadius {1.0};
+  std::vector<double> circleCenter {0.0, 0.0};
```
kennyweiss, line 62: Minor: Will this need three default values for 3D? We can add it later when we add support for 3D to the example.

kennyweiss, line 62: 👍

kennyweiss (approved):
Thanks @gunney1 !

src/axom/quest/examples/quest_distributed_distance_query_example.cpp
gunney1: Yes.  Thanks for the catch.  I'll add a TODO note.

PR #825, opened 2022-05-26
Set  number of output files MFEMSidreDataCollection

dtaller:
# Summary
Allow the user to set the number of output files for Parallel IO in MFEMSidreDataCollection

- This PR is a feature
- It does the following:
  - Add setter method for number of files
  - use this user-specified number of files (if given) when writing out the files to disk. If not provided, use num_procs as before.

### DIFF
--- a/src/axom/sidre/core/MFEMSidreDataCollection.cpp
+++ b/src/axom/sidre/core/MFEMSidreDataCollection.cpp
@@ -162,8 +162,8 @@ void MFEMSidreDataCollection::SetComm(MPI_Comm comm)
 
 void MFEMSidreDataCollection::SetNumFiles(int num_files)
 {
-   SLIC_ASSERT_MSG(num_files > 0, "Output must be to at least 1 file");
-   m_num_files = num_files;
+  SLIC_ASSERT_MSG(num_files > 0, "Output must be to at least 1 file");
+  m_num_files = num_files;
 }
   #endif
 
@@ -841,10 +841,7 @@ void MFEMSidreDataCollection::SetMesh(MPI_Comm comm, Mesh* new_mesh)
 {
   // use MFEMSidreDataCollection's custom SetMesh, then set MPI info
   SetMesh(new_mesh);
-
-  m_comm = comm;
-  MPI_Comm_rank(comm, &myid);
-  MPI_Comm_size(comm, &num_procs);
+  SetComm(comm);
 }
   #endif
 
@@ -1088,9 +1085,12 @@ void MFEMSidreDataCollection::Save(const std::string& filename,
     temp_domain_grp->copyGroup(domain_grp);
 
     int num_files = num_procs;
-    if (m_num_files > 0) {
-       SLIC_ASSERT_MSG(num_files <= num_procs, "Save output must have num_files less than or equal to number of mpi ranks");
-       num_files = m_num_files;
+    if (m_num_files > 0) 
+    {
+      SLIC_ASSERT_MSG(num_files <= num_procs, 
+                      "Save output must have num_files less than or equal to "
+                      "number of mpi ranks");
+      num_files = m_num_files;
     }
     writer.write(temp_root, num_files, file_path, protocol);
 
--- a/src/axom/sidre/core/MFEMSidreDataCollection.cpp
+++ b/src/axom/sidre/core/MFEMSidreDataCollection.cpp
@@ -86,6 +86,7 @@ MFEMSidreDataCollection::MFEMSidreDataCollection(const std::string& collection_n
   , m_owns_datastore(true)
   , m_owns_mesh_data(own_mesh_data)
   , m_meshNodesGFName("mesh_nodes")
+  , m_num_files(-1)
 {
   m_datastore_ptr = new sidre::DataStore();
 
@@ -126,6 +127,7 @@ MFEMSidreDataCollection::MFEMSidreDataCollection(const std::string& collection_n
   , m_owns_datastore(false)
   , m_owns_mesh_data(own_mesh_data)
   , m_meshNodesGFName("mesh_nodes")
+  , m_num_files(-1)
   , m_datastore_ptr(nullptr)
   , m_bp_index_grp(bp_index_grp)
 {
@@ -155,6 +157,13 @@ void MFEMSidreDataCollection::SetComm(MPI_Comm comm)
   appendRankToFileName = true;
   MPI_Comm_rank(m_comm, &myid);
   MPI_Comm_size(m_comm, &num_procs);
+}
+
+
+void MFEMSidreDataCollection::SetNumFiles(int num_files)
+{
+   SLIC_ASSERT_MSG(num_files > 0, "Output must be to at least 1 file");
+   m_num_files = num_files;
 }
   #endif
 
@@ -1078,7 +1087,12 @@ void MFEMSidreDataCollection::Save(const std::string& filename,
     }
     temp_domain_grp->copyGroup(domain_grp);
 
-    writer.write(temp_root, num_procs, file_path, protocol);
+    int num_files = num_procs;
+    if (m_num_files > 0) {
+       SLIC_ASSERT_MSG(num_files <= num_procs, "Save output must have num_files less than or equal to number of mpi ranks");
+       num_files = m_num_files;
+    }
+    writer.write(temp_root, num_files, file_path, protocol);
 
     // Now that we've written the data, we can delete the temporary group
     temp_root->getParent()->destroyGroup("_sidre_tmp_save");
--- a/src/axom/sidre/core/MFEMSidreDataCollection.cpp
+++ b/src/axom/sidre/core/MFEMSidreDataCollection.cpp
@@ -1085,9 +1085,9 @@ void MFEMSidreDataCollection::Save(const std::string& filename,
     temp_domain_grp->copyGroup(domain_grp);
 
     int num_files = num_procs;
-    if (m_num_files > 0) 
+    if (m_num_files > 0)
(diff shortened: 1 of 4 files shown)


src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -832,10 +840,7 @@ void MFEMSidreDataCollection::SetMesh(MPI_Comm comm, Mesh* new_mesh)
 {
   // use MFEMSidreDataCollection's custom SetMesh, then set MPI info
   SetMesh(new_mesh);
-
-  m_comm = comm;
-  MPI_Comm_rank(comm, &myid);
-  MPI_Comm_size(comm, &num_procs);
+  SetComm(comm);
```
dtaller, line 843: Please not that this changes the behavior, since SetComm also sets  serial = false;
  appendRankToFileName = true; (in addition to setting comm and num_procs and myid as done here). I believe that my change gives us the desired behavior and makes things consistent.

src/axom/sidre/core/MFEMSidreDataCollection.hpp
```diff
@@ -226,6 +226,11 @@ class MFEMSidreDataCollection : public mfem::DataCollection
   /** If no mesh was associated with the collection, this method should be
       called before using any of the Load() methods to read parallel data. */
   void SetComm(MPI_Comm comm);
+
+  /// Set number of files for parallel writing
+  /** Allows the user to consolidate data for N ranks into M files, 0 < M <= N
```
kennyweiss, line 231: Please add that when `num_files` is less than or equal to 0, there will be one rank per file. 
And that setting `num_files` should be less than or equal to `N` (where `N` is the number of ranks)

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1078,7 +1083,15 @@ void MFEMSidreDataCollection::Save(const std::string& filename,
...
     temp_domain_grp->copyGroup(domain_grp);
 
-    writer.write(temp_root, num_procs, file_path, protocol);
+    int num_files = num_procs;
+    if(m_num_files > 0)
+    {
+      SLIC_ASSERT_MSG(num_files <= num_procs,
+                      "Save output must have num_files less than or equal to "
+                      "number of mpi ranks");
+      num_files = m_num_files;
```
kennyweiss, line 1092: Since asserts are only active in debug configurations, please set `num_files` to at most `num_procs`

We have several utility functions that might be helpful, e.g. `axom::utilities::min()` or `axom::utiliities::clampVal()`
See: https://github.com/LLNL/axom/blob/57558853d1cc9193387d332faa1e4269b6589827/src/axom/core/utilities/Utilities.hpp

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -156,6 +158,12 @@ void MFEMSidreDataCollection::SetComm(MPI_Comm comm)
   MPI_Comm_rank(m_comm, &myid);
   MPI_Comm_size(m_comm, &num_procs);
 }
+
+void MFEMSidreDataCollection::SetNumFiles(int num_files)
+{
+  SLIC_ASSERT_MSG(num_files > 0, "Output must be to at least 1 file");
```
kennyweiss, line 164: We might want to also check that `num_files` is not larger than `num_ranks`?

But I think that would only be valid after calling `SetComm`, so perhaps that's not possible here.

kennyweiss (changes requested):
Thanks @dtaller !

Please also add tests that exercise this new functionality, e.g. here: https://github.com/LLNL/axom/blob/develop/src/axom/sidre/tests/sidre_mfem_datacollection.cpp

and incorporate this into a data collection example, e.g. here: https://github.com/LLNL/axom/blob/develop/src/axom/sidre/examples/sidre_mfem_datacollection_restart.cpp

src/axom/sidre/core/MFEMSidreDataCollection.cpp
kennyweiss: How does `appendRankToFileName` interact with the new functionality that modifies the number of output files?  Please verify that this works as expected.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
dtaller: Yes, exactly, that is why I only do that check when we are right about to Save.

I will likely get to making the examples/tests and finishing things next week, thanks for your feedback and sorry for the delay!

src/axom/sidre/core/MFEMSidreDataCollection.cpp
nselliott: I'm not familiar with how MFEM uses this symbol, but if it affects the string passed as the third argument in writer.write(), it may cause some trouble.  That string should be identical across all ranks making the call (which is something we should document better in axom).  SPIO will append an integer onto the output filenames, but it will be distinct from the rank number when `num_files < num_ranks`.

nselliott:
Holding off on approval due to uncertainty about `appendRankToFIlename`, but otherwise this looks like correct usage of the SPIO feature for `num_files < num_ranks`.  I also concur with comments from @kennyweiss about making sure `num_files` does not exceed `num_ranks`.

src/axom/sidre/core/MFEMSidreDataCollection.hpp
dtaller: Added info to comment

src/axom/sidre/core/MFEMSidreDataCollection.cpp
dtaller: I just added two tests, and played with it. With the change as indicated here, I see the correct output:

rzgenie2{taller1}69: ls dc_save_single_file/
dc_save_single_file_0000000.hdf5

rzgenie2{taller1}70: ls dc_save_two_files/
dc_save_two_files_0000000.hdf5	dc_save_two_files_0000001.hdf5

If I go back to the previous code, I see the same thing. So it seems that this change is ok, but I could switch it back if you want.

dtaller:
> Thanks @dtaller !
> 
> Please also add tests that exercise this new functionality, e.g. here: https://github.com/LLNL/axom/blob/develop/src/axom/sidre/tests/sidre_mfem_datacollection.cpp
> 
> and incorporate this into a data collection example, e.g. here: https://github.com/LLNL/axom/blob/develop/src/axom/sidre/examples/sidre_mfem_datacollection_restart.cpp

Thanks @kennyweiss and @nselliott , I added tests and examples, and tried to address your other comments also. I appreciate your patience on my first axom MR! Next time I'll be sure to add tests and examples and do the style check from the get-go. Fighting with the style check will be easier next time :D .

src/axom/sidre/core/MFEMSidreDataCollection.cpp
nselliott: Yes, those are the filenames I would expect, with the integers appended by SPIO.  Looks like `appendRankToFileName` isn't affecting anything that is passed to SPIO, and maybe it isn't doing anything at all.  The numbers are going to be `0` to `num_files-1` and will only have a direct mapping to the rank numbers in the `num_files == num_ranks` case.

nselliott (approved):

src/axom/sidre/core/MFEMSidreDataCollection.hpp
```diff
@@ -530,6 +537,10 @@ class MFEMSidreDataCollection : public mfem::DataCollection
   // Default value: "mesh_nodes".
   std::string m_meshNodesGFName;
 
+  // For Parallel IO, the user can specify a number of files less tnan or equal to the
```
kennyweiss, line 540: Typo (minor):
```suggestion
  // For Parallel IO, the user can specify a number of files less than or equal to the
```

src/axom/sidre/examples/sidre_mfem_datacollection_restart.cpp
```diff
@@ -220,15 +220,28 @@ int main(int argc, char* argv[])
...
+      "simulations (default one output file per rank)")
+    ->capture_default_str();
   CLI11_PARSE(app, argc, argv);
 
+#ifdef EXAMPLE_USES_MPI
+  if(num_files > 0)
+  {
+    dc.SetNumFiles(num_files);
+  }
+#endif
```
kennyweiss, line 243: Suggestion (minor, not required, but possibly helpful):
Would it make sense to have the `num_files` option only appear when `EXAMPLE_USES_MPI` ?

Also, it might make sense to either add a note that `num_files` must be greater than 0, or to add a check that `num_files` is between 1 and `num_ranks`, e.g. using  `CLI::range()` 
(similar to https://github.com/LLNL/axom/blob/790d3254d59fd99cde3a7dfdce1a23304724aea3/src/thirdparty/tests/cli11_smoke.cpp#L49)

kennyweiss (approved):
Thanks @dtaller !

src/axom/sidre/core/MFEMSidreDataCollection.hpp
dtaller: Fixed!

src/axom/sidre/examples/sidre_mfem_datacollection_restart.cpp
dtaller: I asked mpi for the number of procs, and then set the range from 1 to num_procs.

If this is run without mpi, then num_procs will be 1, and it will only accept values in the range from 1 to 1, so the behavior is as the user expects

PR #826, opened 2022-05-27
Inlet function cleanup

kanye-quest:
# Summary

Replaces the use of `std::tuple` in `inlet::FunctionWrapper` with a type-erased block of memory to store arbitrary `std::function`s. This lets us remove a significant amount of the template metaprogramming mechanics in inlet/Function.hpp, and shaves about a second off of the cost from including `inlet.hpp`.


white238:
@samuelpmishLLNL

samuelpmishLLNL:
Sorry to ask a basic question but: what is this code (and the code it is replacing) designed to do?

white238:
I tested this against Serac and it passes all our tests.  Did a small timing comparison in Serac between current develop of axom and this branch:

```
~~Develop~~
~Try 1
582.33user
85.96system
1:39.31elapsed

~Try 2
568.07user
76.68system
1:39.77elapsed

~~This Branch~~
~Try 1
536.66user
79.44system
1:25.73elapsed

~Try 2
526.78user
71.87system
1:25.60elapsed
```

white238:
> Sorry to ask a basic question but: what is this code (and the code it is replacing) designed to do?
> 
> All improvements to compile time are greatly appreciated!

This bit of code allows users of Inlet to call Lua functions inside their input files with a limited amount of types for parameters and return types. We currently allow up to 2 parameters and the three following types: `double`, `vector`, and `string` and of course a `void` return type and no parameters as well.

The reason this is such a compile intensive operation is because of the combination of return types and parameters.

@kanye-quest  was able to find a drop in replacement for the previous solution that is much faster in regards to compile time.

white238 (approved):

kennyweiss (approved):
Thanks @kanye-quest !

@white238 -- Is there sufficient test coverage for all the required/desired cases?

white238:
> Thanks @kanye-quest !
> 
> @white238 -- Is there sufficient test coverage for all the required/desired cases?

Yes there was a pretty extensive set of tests in Axom and some extra's in Serac.

PR #827, opened 2022-05-28
Clean up fortran variant logic

white238:
Found a missed case in the `fortran` variant when building Serac.  This fixes that and just cleans up the logic.  This does not change any TPL builds so I won't be rebuilding.


kennyweiss (approved):
Thanks @white238 !

scripts/spack/packages/axom/package.py
```diff
@@ -297,9 +299,9 @@ def initconfig_hardware_entries(self):
             # These flags are already part of the wrapped compilers on TOSS4 systems
             #hip_link_flags = "-Wl,--disable-new-dtags -L{0}/lib -L{0}/../lib64 -L{0}/../lib -Wl,-rpath,{0}/lib:{0}/../lib:{0}/../lib64 -lamdhip64 -lhsakmt -lhsa-runtime64".format(hip_root)
 
-            if self.compiler.fc is not None:
+            if "+fortran":
```
ltaylor16, line 302: `if "+fortran" in spec` ?

white238: i mean if you want to be all picky and correct....

white238: Fixed! It's like it was never wrong.  Thanks @ltaylor16 !

agcapps (approved):
thank you Chris

PR #829, opened 2022-05-31
Guard Sol State and Sol header

white238:
`sol.hpp` is a very heavily templated header.  This causes large compile times of any users of Inlet. We originally exposed the Sol `state` class for a user but that never went anywhere.

This moves the `LuaReader::solState()` getter to protected in case someone does want it and documents it that they need to derive there own class and include `sol.hpp` in their own code.

**Axom Build Times**
develop: 2m08s
This branch: 1m51s

### DIFF
--- a/src/axom/inlet/LuaReader.hpp
+++ b/src/axom/inlet/LuaReader.hpp
@@ -15,16 +15,19 @@
 #define INLET_LUAMAP_HPP
 
 #include "axom/inlet/Reader.hpp"
-#include "axom/sol.hpp"
-
-extern "C" {
-#include "lua.h"
-#include "lualib.h"
-#include "lauxlib.h"
-}
+#include "axom/sol_forward.hpp"
 
 namespace axom
 {
+
+// Forward declarations to avoid having to include "sol.hpp" in everything
+// that depends on Inlet
+namespace sol
+{
+class state;
+enum class type;
+}
+
 namespace inlet
 {
 /*!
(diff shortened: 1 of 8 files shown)


src/axom/inlet/examples/lua_library.cpp
```diff
@@ -11,6 +11,16 @@
 #include "axom/inlet.hpp"
 #include "axom/slic/core/SimpleLogger.hpp"
 
+// _inlet_sol_state_start
+#include "axom/sol.hpp"
```
kennyweiss, line 15: Suggestion: Would it make sense to add a comment about this header inclusion being necessary?

kennyweiss (approved):
Thanks @white238 !

Would it make sense to add a line to the `RELEASE_NOTES` about this change?

src/axom/inlet/LuaReader.hpp
```diff
@@ -92,16 +94,23 @@ class LuaReader : public Reader
...
+   *   be sure to include `axom/sol.hpp` and derive a new class to access this
+   *   publically. See the `lua_library.cpp` example. Including `sol.hpp` can
+   *   increase compile time significantly.
    *
-   * \return Reference to the Sol Lua state
+   * \return Shared pointer to the Sol Lua state
    *****************************************************************************
    */
-  axom::sol::state& solState() { return m_lua; }
+  std::shared_ptr<axom::sol::state> solState() { return m_lua; }
```
kanye-quest, line 113: Is there any reason/use case for the Sol state lifetime to extend beyond the lifetime of `LuaReader`? If not, I'd suggest making the `m_lua` member field an `std::unique_ptr` and just returning a reference to the state as before.

kanye-quest (approved):
LGTM, just one suggestion:

src/axom/inlet/LuaReader.hpp
white238: I could see someone wanting the sol state to extend past when the LuaReader is done. The person that wanted this before was using quite a few things down the road after the input parsing was done.

agcapps (approved):
Thanks for this, @white238 .

PR #831, opened 2022-06-01
Adds scattered interpolation capability to quest

kennyweiss:
# Summary

- This PR is a feature
- It adds a scattered data interpolation capability to `quest`
- Given a point mesh with scalar fields, following the [Conduit mesh blueprint schema](https://llnl-conduit.readthedocs.io/en/latest/blueprint_mesh.html) and a collection of query points, the class enables interpolating the scalar fields at the query points 
- This PR also adds an example application to quest that either reads in a mesh blueprint file from disk, or generates a random collection of points and uses the point coordinate positions as scalar fields. 

#### Implementation notes
- The implementation builds a Delaunay triangulation over the input point set
- It then locates the simplex (triangle or tetrahedron) containing each query point and performs linear interpolation to interpolate the scalar value at its location
- If the query point is not contained in any cell of the triangulation, no interpolation is done, and the code returns a user-provided value (e.g. `NaN`) to signify this

#### Example  runs and visualizations
```
# Generate 100 random samples and query at 500 random points
> ./examples/quest_scattered_interpolation_ex -d2 -n100 -q500
```
Visualizing cell ids for the query points:
scattered_2d_n100_q500_cell_id
Points that are not in a triangle are marked with red squares.

Visualizing the interpolated component position fields: `pos_x` and `pos_y`:
scattered_2d_n100_q500_pos_x
scattered_2d_n100_q500_pos_y

We can also load the input mesh from a blueprint file on disk:
```
# Generate a blueprint point mesh with 100 vertices using `gen2D` as the output mesh prefix
> ./examples/quest_scattered_interpolation_ex -d2 -n100 -o gen2d
# Load in the blueprint mesh from the previous run, and query at 5000 random points
> ./examples/quest_scattered_interpolation_ex -d2 -i gen2d_mesh_after_generate.root -q5000
```
Here is the `pos_x` field from the previous run
scattered_2d_n100_q5000_pos_x_from_file

This also works in 3D. Here is  the `pos_z` field on a mesh with 1000 input points and 5000 query points:
```
> ./examples/quest_scattered_interpolation_ex -d3 -n1000 -q5000
```
scattered_3d_n1000_q5000_pos_z

#### Next steps
When running on a user-provided input mesh, I noticed a few issues related to duplicated and/or co-planar points. I think these can be resolved by welding vertices that are sufficiently close, by rearranging the insertion order of the points. These will be the subject of a future PR.

### DIFF
--- a/src/axom/quest/Delaunay.hpp
+++ b/src/axom/quest/Delaunay.hpp
@@ -395,9 +395,9 @@ public:
     return valid;
   }
 
-private:
   /// \brief Find the index of the element that contains the query point, or the element closest to the point.
-  IndexType findContainingElement(const PointType& query_pt)
+  IndexType findContainingElement(const PointType& query_pt,
+                                  bool warnOnInvalid = true)
   {
     if(m_mesh.isEmpty())
     {
@@ -451,15 +451,23 @@ private:
       // Logically, this should never happen.
       if(!m_mesh.isValidElement(element_i))
       {
-        SLIC_WARNING(fmt::format(
-          "Entered invalid element in "
-          "Delaunay::findContainingElement(). Underlying mesh {} valid",
-          m_mesh.isValid() ? "is" : "is not"));
+        SLIC_WARNING_IF(
+          warnOnInvalid,
+          fmt::format(
+            "Entered invalid element in "
+            "Delaunay::findContainingElement(). Underlying mesh {} valid",
+            m_mesh.isValid() ? "is" : "is not"));
         return INVALID_INDEX;
       }
     }
   }
 
+  /**
+   * \brief helper function to retrieve the barycentric coordinate of the query point in the element
+   */
+  BaryCoordType getBaryCoords(IndexType element_idx, const PointType& q_pt) const;
+
+private:
   /// \brief Predicate for when to compact internal mesh data structures after removing elements
   bool shouldCompactMesh() const
   {
@@ -484,11 +492,6 @@ private:
   void generateInitialMesh(std::vector<DataType>& points,
                            std::vector<IndexType>& elem,
                            const BoundingBox& bb);
-
-  /**
-   * \brief helper function to retrieve the barycentric coordinate of the query point in the element
-   */
-  BaryCoordType getBaryCoords(IndexType element_idx, const PointType& q_pt) const;
 
 private:
   /// Helper struct to find the first element near a point to be inserted
--- a/src/axom/quest/Delaunay.hpp
+++ b/src/axom/quest/Delaunay.hpp
@@ -401,14 +401,16 @@ public:
   {
     if(m_mesh.isEmpty())
     {
-      SLIC_ERROR(
+      SLIC_ERROR_IF(
+        warnOnInvalid,
         "Attempting to insert point into empty Delaunay triangulation."
         "Delaunay::initializeBoundary() needs to be called first");
       return INVALID_INDEX;
     }
     if(!m_bounding_box.contains(query_pt))
     {
-      SLIC_WARNING(
+      SLIC_WARNING_IF(
+        warnOnInvalid,
         "Attempting to locate element at location outside valid domain");
       return INVALID_INDEX;
     }
--- a/src/axom/quest/ScatteredInterpolation.hpp
+++ b/src/axom/quest/ScatteredInterpolation.hpp
@@ -105,6 +105,12 @@ public:
   using PointType = typename DelaunayTriangulation::PointType;
   using BoundingBoxType = typename DelaunayTriangulation::BoundingBox;
 
+  /**
+   * \brief Builds a Delaunay triangulation over the point set from \a mesh_node
+   *
+   * \param [in] mesh_node Conduit node for the input mesh
+   * \param [in] coordset The name of the coordinate set for the input mesh
+   */
   void buildTriangulation(conduit::Node& mesh_node, const std::string& coordset)
   {
     // Perform some simple error checking
@@ -126,6 +132,7 @@ public:
     {
       bb.addPoint(pt);
     }
+    // Scale the bounding box to ensure that all input points are contained
     bb.scale(1.5);
 
     m_delaunay.initializeBoundary(bb);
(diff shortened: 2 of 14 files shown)


src/axom/quest/Delaunay.hpp
```diff
@@ -395,20 +395,22 @@ class Delaunay
     return valid;
   }
 
-private:
   /// \brief Find the index of the element that contains the query point, or the element closest to the point.
```
kennyweiss, line 398: Note: The only real change to the Delaunay class to support this capability was in exposing the `findContainingElement()` and the `getBaryCoords()` functions as public instead of private.

Since we now expect some points to be outside of a valid cell, I added a second optional parameter to disable warnings in this case, but kept this warning on by default.

src/axom/quest/ScatteredInterpolation.hpp
```diff
@@ -0,0 +1,402 @@
...
+namespace quest
+{
+namespace detail
+{
+/**
+ * \brief Utility class to enable processing an array of points whose layout
+ * is either interleaved or separated strided arrays
+ */
+template <int NDIMS>
+struct InterleavedOrStridedPoints
```
kennyweiss, line 96: Note: This class is a kludge to allow processing of point arrays whose layout is either strided by coordinate, or whose coordinates are interleaved. Axom should provide a better native solution for this common situation.

I'll create an issue to track this.

src/axom/quest/examples/scattered_interpolation.cpp
```diff
@@ -0,0 +1,788 @@
...
+namespace internal
+{
+namespace blueprint
+{
+/**
+ *  \brief Simple wrapper to a blueprint particle mesh
+ *
+ *  Given a sidre Group, creates the stubs for a mesh blueptint particle mesh
+ */
+struct PointMesh
```
kennyweiss, line 41: This `PointMesh` class is very similar to the one in the distributed closest point example: https://github.com/LLNL/axom/blob/bd43ea33a14608a72519cf64bdea7143ac32ea38/src/axom/quest/examples/quest_distributed_distance_query_example.cpp#L131-L395

I'd like to combine these and move them into a `sidre/blueprint` directory that simplifies processing of mesh blueprint hierarchies.

I'll create an issue for this as well.

src/axom/quest/ScatteredInterpolation.hpp
```diff
@@ -0,0 +1,402 @@
...
+        }
+
+        out_fld[idx] = res;
+      }
+    }
+  }
+
+  /// TODO: Add a function that takes x,y,z coordinates
+  /// and returns the vertex indices and the barycentric coords
+  /// of the point w.r.t. its containing element
```
kennyweiss, line 347: @dtaller -- should I add this to the current PR, or save it for the next one?

src/axom/quest/examples/scattered_interpolation.cpp
```diff
@@ -0,0 +1,788 @@
...
+  case 2:
+    scattered_2d = std::unique_ptr<quest::ScatteredInterpolation<2>>(
+      new quest::ScatteredInterpolation<2>);
+    scattered_2d->buildTriangulation(bp_input, inputMesh.coordsName());
+    scattered_2d->locatePoints(bp_query, query_coords_name);
+
+    for(const auto& fld : inputMesh.getFieldNames())
+    {
+      scattered_2d->interpolateField(bp_query, query_coords_name, bp_input, fld, fld);
+    }
```
kennyweiss, line 763: Note: The bulk of this example relates to generating and loading mesh blueprint files.
The essential user-facing API is here:
* `buildTriangulation( input_mesh)  // as conduit node in blueprint format`
* `locatePoints( query_mesh) // as conduit node in blueprint format`
* // for each field
* `interpolateField( query_mesh, input_mesh, names of fields in query and input mesh)
  (the results are stored in the query mesh)

Based on a user-request, there will also be another function that, given a point in space, returns the vertex indices of the cell containing that point, and the barycentric coordinates of the point in its containing cell.

src/axom/quest/ScatteredInterpolation.hpp
dtaller: For me this function would be useful for initial testing on my end, but I will trust your judgement if you think this MR is already too long (you touched 14 files already I think). I could wait until the next one also if that makes sense.

gunney1:
Maybe I'm just a noob, but I didn't know the difference between `mesh_after_generate` and `mesh_after_query` was.  I thought it was the same mesh, but it looks like one has the nodes in the Delaunay mesh and the other is the query mesh.  Suggesting different names for the output files.

src/axom/quest/ScatteredInterpolation.hpp
kennyweiss: Thanks @dtaller -- I think I'll do a quick focused follow-up.

kennyweiss:
@gunney1 wrote (in part):
> .. I didn't know the difference between `mesh_after_generate` and `mesh_after_query` was. Suggesting different names for the output files.

Good call! Those were just names that made sense to me while I was developing/debugging the code. 
The former contains the generated (or read in) "input" dataset, while the latter contains the the "output" dataset.

I'll try to come up with something clearer.

gunney1 (approved):
Thanks @kennyweiss

rhornung67:
> @gunney1 wrote (in part):
> 
> > .. I didn't know the difference between `mesh_after_generate` and `mesh_after_query` was. Suggesting different names for the output files.
> 
> Good call! Those were just names that made sense to me while I was developing/debugging the code. The former contains the generated (or read in) "input" dataset, while the latter contains the the "output" dataset.
> 
> I'll try to come up with something clearer.
> 
> **Update:** I renamed the files to `*_input` and `*_interpolated`, respectively.

Would *_input_mesh and *_output_mesh be cleaner (certainly more symmetric)?

src/axom/quest/Delaunay.hpp
```diff
@@ -395,20 +395,22 @@ class Delaunay
     return valid;
   }
```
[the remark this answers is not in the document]
kennyweiss: Sure -- I can do that!

I've been using the vtk file to look at the triangulation, but the (recently renamed) `_input` mesh to look at the fields.
It should be easy enough to add the fields directly to the vtk file.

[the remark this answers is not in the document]
dtaller: Thanks!

src/axom/quest/ScatteredInterpolation.hpp
```diff
@@ -0,0 +1,400 @@
...
+   *
+   * \param [in] mesh_node The root conduit node of a valid mesh blueprint
+   * \param [in] field_name The name of the field to return; Can either be either the name of the field,
+   *  or the path to the field, e.g. fields/<field_name>/values
+   *
+   * \pre The field named \a field_name must be present in \a mesh_node or code will error out
+   * \note Only currently supports scalar fields
+   */
+  template <typename T>
+  ArrayView<T> getField(conduit::Node& mesh_node, const std::string& field_name)
```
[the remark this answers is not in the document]
kennyweiss: Good call -- I moved `getField()` and `isValidBlueprint()` out of the class into the anonymous namespace, since that was more consistent.

src/axom/quest/Delaunay.hpp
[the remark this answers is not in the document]
kennyweiss: Thanks for the suggestion -- the vtk output now includes the scalar fields*. 

I also converted the vtk output into a user-callable function (rather than automatically getting called as part of the `buildTriangulation()` step).

* I only added the scalar fields of type `float64`, since other types seemed to not be of interest. I can add others as necessary/desired.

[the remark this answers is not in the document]
dtaller: Looks good to me, thanks!

dtaller:
Thanks for pushing this through! I appreciate it!

PR #843, opened 2022-06-15
Attempted fix for CI flakiness on blueos

kanye-quest:
# Summary

On some of our open PRs, a couple of our GPU tests were failing on the blueos CI runs. While these failures were somewhat spurious, they always happened on the same set of tests.

As it turns out, some of our tests were trying to access host memory from device kernels; the address translation service on blueos systems would normally bail us out on these, but that capability seems to have gotten flakier recently. This PR tries to fix that by explicitly copying the offending memory to/from the device.

### DIFF
--- a/src/axom/quest/tests/quest_point_in_cell_mfem.cpp
+++ b/src/axom/quest/tests/quest_point_in_cell_mfem.cpp
@@ -379,6 +379,9 @@ public:
   /*! Tests PointInCell class using isoparametric points within each cell */
   void testIsoGridPointsOnMesh(const std::string& meshTypeStr)
   {
+    int devAllocID = axom::execution_space<ExecSpace>::allocatorID();
+    int hostAllocID = axom::execution_space<axom::SEQ_EXEC>::allocatorID();
+
     std::string filename =
       axom::fmt::format("quest_point_in_cell_{}_quad", meshTypeStr);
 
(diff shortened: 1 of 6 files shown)


kennyweiss (approved):
Thanks @kanye-quest !

src/axom/quest/tests/quest_point_in_cell_mfem.cpp
```diff
@@ -391,15 +394,21 @@ class PointInCellTest : public ::testing::Test
...
-    axom::Array<SpacePt> isoPts = generateIsoParTestPoints(::TEST_GRID_RES);
+    SpacePtArray isoPts = generateIsoParTestPoints(::TEST_GRID_RES);
 
     const auto SZ = isoPts.size();
     axom::Array<SpacePt> spacePts(SZ, SZ);
     axom::Array<SpacePt> foundIso(SZ, SZ);
     axom::Array<IndexType> foundIDs(SZ, SZ);
 
+    axom::Array<SpacePt> foundIsoDevice(SZ, SZ, devAllocID);
+    axom::Array<IndexType> foundIDsDevice(SZ, SZ, devAllocID);
```
white238, line 410: Why not use type alias you created above?

white238 (approved):

src/axom/quest/tests/quest_point_in_cell_mfem.cpp
kanye-quest: I originally added those type aliases to simplify switching between unified memory and host memory arrays, before realizing that unified memory didn't work :(

I'll just remove those aliases for consistency.

PR #844, opened 2022-06-17
Adds a curved polygon class to primal

kennyweiss:
# Summary

- This PR is a feature
- It adds `primal::CurvedPolygon` representing a polygon with `BezierCurves` as edges
- It also adds functions to compute the `sector_area` and `sector_centroid` of a BezierCurve. These are used to compute the area and centroid of a `CurvedPolygon`, respectively.
- This branch was cherry picked from `feature/kweiss/bezier-polygon-revamp` which uses the above functionality to intersect two curved polygons. 
    - The latter is still in development (https://github.com/LLNL/axom/pull/590) due to the need to handle tangential and corner intersections. 
    - After this branch is merged, I will rebase the latter on this.

Credits: Much of this code was developed by @davidgunderman 

#### TODO:

- [x] Update `RELEASE-NOTES`

PR #848, opened 2022-06-21
Adds integral evaluation on Bezier curves/bounded regions

jcs15c:
- This PR is a feature
- It adds the `evaluate_integrals.hpp` file to `primal/operators`, which contains methods to
  - Evaluate scalar line integrals on collections of Bezier curves. 
  - Evaluate vector field line integrals on collections of Bezier curves.
  - Evaluate 2D area integrals for regions defined by Bezier curves.
- It adds additional tests in primal for the above methods.
- It adds an optional dependency for primal on `mfem`, used to evaluate quadrature algorithms.

edit: See https://en.wikipedia.org/wiki/Line_integral#Line_integral_of_a_scalar_field and https://en.wikipedia.org/wiki/Line_integral#Line_integral_of_a_vector_field for information about the different types of line integrals.


src/axom/primal/CMakeLists.txt
```diff
@@ -65,6 +66,13 @@ set( primal_sources
      operators/detail/intersect_impl.cpp
    )
 
+blt_list_append( TO primal_headers IF MFEM_FOUND ELEMENTS
+
+    ## operators
+     operators/evaluate_integral.hpp
+     operators/detail/evaluate_integral_impl.hpp  
+    )
```
kennyweiss, line 74:
```suggestion
blt_list_append( 
   TO       primal_headers 
   ELEMENTS operators/evaluate_integral.hpp
            operators/detail/evaluate_integral_impl.hpp
   IF       MFEM_FOUND 
    )
```

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -0,0 +1,197 @@
...
+ * 
+ * Uses the de Casteljau algorithm. Functionality will eventually
+ * be supplanted by Bezier curve class method.
+ * 
+ * \param [in] c the Bezier curve
+ * \param [in] t on [0, 1] the parameter value on which to evaluate the derivative
+ * \return the tangent vector
+ */
+template <typename T>
+primal::Vector<double, 2> temp_dt(const primal::BezierCurve<double, 2>& c, T t)
```
kennyweiss, line 40: I merged #844, so please remove this function in favor of the one in the `BezierCurve` class

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -0,0 +1,197 @@
...
+  // Store/compute quadrature result
+  double full_quadrature = 0.0;
+  for(int q = 0; q < quad->GetNPoints(); q++)
+  {
+    // Get intermediate quadrature point
+    //  at which to evaluate tangent vector
+    x_q = c.evaluate(quad->IntPoint(q).x);
+    dx_q = detail::temp_dt(c, quad->IntPoint(q).x);
+
+    full_quadrature += quad->IntPoint(q).weight * integrand(x_q) * dx_q.norm();
```
kennyweiss, line 99: Minor (here and elsewhere): 
The compiler might be better able to optimize with temporaries for `x_q` and `dx_q`.
E.g.
```suggestion
  // Store/compute quadrature result
  double full_quadrature = 0.0;
  for(int q = 0; q < quad->GetNPoints(); q++)
  {
    // Get intermediate quadrature point
    //  at which to evaluate tangent vector
    auto x_q = c.evaluate(quad->IntPoint(q).x);
    auto dx_q = detail::temp_dt(c, quad->IntPoint(q).x);

    full_quadrature += quad->IntPoint(q).weight * integrand(x_q) * dx_q.norm();
```

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  return full_quadrature;
+}
+
+/*!
+ * \brief Evaluate the integral across a single component of the curved polygon.
+ *
+ * Uses a Spectral Mesh-Free Quadrature derived from Green's theorem, evaluating
+ * the area integral as a line integral of the antiderivative over the curve.
```
kennyweiss, line 146: Please clarify that this function is meant to be called for each `BezierCurve` of a closed polygon, and add a (simplified) reference to the paper

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -0,0 +1,197 @@
...
+ * \param [in] cs the array of Bezier curve objects that bound the region
+ * \param [in] integrand the lambda function representing the integrand. 
+ * Must accept a 2D point as input and return a double
+ * \param [in] The lower bound of integration for the antiderivatives
+ * \param [in] quad_Q the quadrature rule for the line integral
+ * \param [in] quad_P the quadrature rule for the antiderivative
+ * \return the value of the integral, which is mathematically meaningless.
+ */
+template <class Lambda>
+double evaluate_area_integral_component(const primal::BezierCurve<double, 2>& c,
```
kennyweiss, line 157: Aside: We might want to generalize the BezierCurve type to `CoordType`s other than `double`.  Should be fine for now though.

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -0,0 +1,176 @@
...
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/*!
+ * \file evaluate_integral.hpp
+ *
+ * \brief Consists of methods that evaluate integrals over regions defined
+ * by Bezier curves, such as 2D area integrals and scalar/vector field line integrals
+ *
+ * Line integrals are computed with simple 1D Gaussian Quadrature using rules generated
```
kennyweiss, line 12: Thinking ahead, we might not want to prescribe the type of quadrature in this high level description.

```suggestion
 * Line integrals are computed with 1D quadrature rules supplied
```

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -0,0 +1,176 @@
...
+ * by MFEM. 2D area integrals computed with "Spectral Mesh-Free Quadrature for Planar 
+ * Regions Bounded by Rational Parametric Curves" by David Gunderman et al.
+ */
+
+#ifndef PRIMAL_EVAL_INTEGRAL_HPP_
+#define PRIMAL_EVAL_INTEGRAL_HPP_
+
+// Axom includes
+#include "axom/config.hpp"  // for compile-time configuration options
```
kennyweiss, line 21: Minor suggestion:  Please remove unnecessary/overly specific comments

```suggestion
#include "axom/config.hpp"
```

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -0,0 +1,176 @@
...
+#include "axom/primal.hpp"
+
+#include "axom/primal/operators/detail/evaluate_integral_impl.hpp"
+
+// MFEM includes
+#ifdef AXOM_USE_MFEM
+  #include "mfem.hpp"
+#else
+  #error "Primal's integral evaluation functions require mfem library."
```
kennyweiss, line 30: 👍

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -0,0 +1,176 @@
...
+#endif
+
+// C++ includes
+#include <cmath>
+
+// Create type definitions the different field types (vector vs scalar)
+//   to allow for proper overloading of line integrals
+using Point2D = axom::primal::Point<double, 2>;
+using Vector2D = axom::primal::Vector<double, 2>;
```
kennyweiss, line 39: Please remove these type aliases

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -0,0 +1,176 @@
...
+
+namespace axom
+{
+namespace primal
+{
+/*!
+ * \brief Evaluate a line integral along a collection of Bezier curves
+ *
+ * The curve need not be connected, simply adding the portion along each segment.
```
kennyweiss, line 48: Please clarify this line.

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -0,0 +1,176 @@
...
+  const mfem::IntegrationRule* quad =
+    &(my_IntRules.Get(mfem::Geometry::SEGMENT, 2 * npts - 1));
+
+  double total_integral = 0.0;
+  for(int i = 0; i < cs.size(); i++)
+  {
+    // Compute the line integral along each component
+    total_integral +=
+      detail::evaluate_line_integral_component(cs[i], integrand, quad);
+  }
```
kennyweiss, line 76: Minor: This could use the range-for pattern:

```suggestion
  for(const auto& curve : cs)
  {
    // Compute the line integral along each component
    total_integral +=
      detail::evaluate_line_integral_component(curve, integrand, quad);
  }
```

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -0,0 +1,176 @@
...
+  {
+    npts_P = npts_Q;
+    quad_P = quad_Q;
+  }
+  else
+  {
+    quad_P = &(my_IntRules.Get(mfem::Geometry::SEGMENT, 2 * npts_P - 1));
+  }
+
+  // Define lower bound as lowest control node for all curves
```
kennyweiss, line 153: Clarity: Please add that we're only looking for the y-coordinate.
E.g.
```suggestion
  // Use minimum y-coordinate of curves as lower bound for integration
```

kennyweiss (changes requested):
Thanks @jcs15c -- this is great!

I added some suggestions for minor changes.
Please also add a line to the RELEASE-NOTES.

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * \param [in] c the Bezier curve object
+ * \param [in] integrand the lambda function representing the integrand. 
+ * Must accept a 2D point as input and return a double
+ * \param [in] quad the mfem integration rule containing nodes and weights
+ * \return the value of the integral
+ */
+inline double evaluate_line_integral_component(
+  const primal::BezierCurve<double, 2>& c,
+  std::function<double(Point2D)> integrand,
```
kennyweiss, line 42: I think the lambda should be passed by r-value ref (here and elsewhere):

```suggestion
  std::function<double(Point2D)>&& integrand,
```

(@kanye-quest -- could you please confirm?)

kennyweiss (approved):
Thanks @jcs15c  -- this is really nice!

Tag: @davidgunderman

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
jcs15c: I'm very unclear on the details, but passing the `std::function`'s by R-value reference is causing some issues with some to-be-implemented winding number code. Specifically, if we pass to `evaluate_line_integral` a lambda function that is the return value for some other function (see below), then `evaluate_line_integral_component` can no longer fit the argument to either overloaded definition. This doesn't happen when the `std::function`'s are passed by constant reference, or when the lambda function is not the return value for another function, so I am unsure how to keep this as general as possible.

```C++
// An example of function that returns argument to evaluate_line_integral
std::function<Vector2D(Point2D)> get_vector_field(Point2D p) 
{
  return [p](Point2D x) -> Vector2D { return Vector2D({p[1], p[0]}); };
}
```

kennyweiss: Thanks -- ok, sounds like something to think about and try to fix in the future.

src/axom/primal/CMakeLists.txt
```diff
@@ -68,6 +68,13 @@ set( primal_sources
      operators/detail/intersect_impl.cpp
    )
 
+blt_list_append( 
+   TO       primal_headers 
+   ELEMENTS operators/evaluate_integral.hpp
+            operators/detail/evaluate_integral_impl.hpp
+   IF       MFEM_FOUND 
+    )
```
kanye-quest, line 76: Nit: alignment?

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
kanye-quest: This might have something to do with this function (and the one below) accepting an `std::function` instead of a `Lambda&&` like in the other methods. The latter is a universal reference, so it can bind to both lvalues and rvalues.

For what it's worth, I think accepting `Lambda&&` might be better from an inlineability perspective -- it would be much more difficult for the compiler to inline through a type-erased `std::function`.

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -0,0 +1,246 @@
...
+  //  Use the same one for every curve in the polygon
+  static mfem::IntegrationRules my_IntRules(mfem::Quadrature1D::GaussLegendre);
+  const mfem::IntegrationRule *quad_Q, *quad_P;
+
+  if(npts_P <= 0) npts_P = npts_Q;
+
+  // Get the quadrature for the line integral.
+  //  Quadrature order is equal to 2*N - 1
+  quad_Q = &(my_IntRules.Get(mfem::Geometry::SEGMENT, 2 * npts_Q - 1));
+  quad_P = &(my_IntRules.Get(mfem::Geometry::SEGMENT, 2 * npts_P - 1));
```
kanye-quest, line 171: Minor: any reason to use pointers to the `IntegrationRule`s instead of references?

kanye-quest (approved):
LGTM, and thanks for the tests! Just a few comments to think about.

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -0,0 +1,246 @@
...
+/*!
+ * \file evaluate_integral.hpp
+ *
+ * \brief Consists of methods that evaluate integrals over regions defined
+ * by Bezier curves, such as 2D area integrals and scalar/vector field line integrals
+ *
+ * Line integrals are computed with 1D quadrature rules supplied
+ * by MFEM. 2D area integrals computed with "Spectral Mesh-Free Quadrature for Planar 
+ * Regions Bounded by Rational Parametric Curves" by David Gunderman et al.
+ */
```
agcapps, line 15: I appreciate this whole comment!  I see you wrote "with ... rules supplied by MFEM," as is accurate.  Please add the additional note, to make things extra clear:

```suggestion
 * Regions Bounded by Rational Parametric Curves" by David Gunderman et al.
 *
 * \note This requires the MFEM third-party library.
 */
```

src/axom/primal/operators/evaluate_integral.hpp
jcs15c: Only that the quest test I used for reference used them, changing it to references makes it much cleaner. Thank you!

src/axom/primal/tests/primal_integral.cpp
```diff
@@ -0,0 +1,231 @@
...
+  EXPECT_NEAR(evaluate_line_integral(connected_curve, const_integrand, npts),
+              8.28968500196,
+              abs_tol);
+  EXPECT_NEAR(evaluate_line_integral(connected_curve, poly_integrand, npts),
+              -5.97565740064,
+              abs_tol);
+  EXPECT_NEAR(evaluate_line_integral(connected_curve, transc_integrand, npts),
+              -0.574992518405,
+              abs_tol);
+}
```
agcapps, line 157: Please add a test with disconnected line segments, as this is noted as possible in the doxygen for `evaluate_line_integral`.  Perhaps displace the middle, linear segment by (0, 2) or some such.

src/axom/primal/tests/primal_integral.cpp
```diff
@@ -0,0 +1,231 @@
...
+  // This vector field is conservative, so it should evaluate to zero
+  EXPECT_NEAR(evaluate_line_integral(parabola_shape, conservative_field, npts),
+              0.0,
+              abs_tol);
+
+  // This vector field is generated by a in/out query, should return 1 (inside)
+  EXPECT_NEAR(evaluate_line_integral(parabola_shape, winding_field, npts),
+              1.0,
+              abs_tol);
+}
```
agcapps, line 220: Please add a test for disconnected segments.

agcapps (approved):
Thanks for this!

PR #855, opened 2022-06-25
Fix outdated references to "signed" distances in doxygen comments.

gunney1:
# Summary

- This PR is a bugfix
  - Fixes doxygen comments in code to compute squared distances between primal objects.

The comments referred to "signed" distances and "signed" squared distances, which don't make sense and are probably outdated.  I removed the "signed" and added "squared" where needed.


kennyweiss (approved):
Thanks @gunney1 !

agcapps (approved):
Thanks!

PR #856, opened 2022-06-27
HIP policy for additional tests and examples

bmhan12:
This PR:

-  Adds the HIP policy to more unit tests and examples that previously had only CUDA as the device policy.
    - Notably `mesh_tester` is excluded from this PR as it looks to require more significant refactoring to compile and link for hip ( Error `lld: error: local memory (249776) exceeds limit (65536)` encountered when trying to link with hip and all available spatial indexes)
`


src/axom/core/Array.hpp
```diff
@@ -1364,7 +1364,15 @@ inline void Array<T, DIM, SPACE>::dynamicRealloc(IndexType new_num_elements)
...
-  m_data = axom::reallocate<T>(m_data, new_capacity, m_allocator_id);
+  if(m_num_elements == 0)
+  {
+    axom::deallocate(m_data);
+    m_data = axom::allocate<T>(new_capacity, m_allocator_id);
+  }
+  else
+  {
+    m_data = axom::reallocate<T>(m_data, new_capacity, m_allocator_id);
+  }
```
bmhan12, line 1375: This was changed to fix a memory error in `spin_uniform_grid_test` in corona gitlab CI and occurred sporadically when I self ran it:
https://github.com/LLNL/axom/blob/cf9ae83a827b6a568f797459e45cb9f3ffd4e893/src/axom/spin/tests/spin_uniform_grid.cpp#L51-L56

The error occurs when resizing axom::Array `m_bins` from size 0 to 1000000 (?):
https://github.com/LLNL/axom/blob/cf9ae83a827b6a568f797459e45cb9f3ffd4e893/src/axom/spin/policy/UniformGridStorage.hpp#L53

Umpire throws an exception for `reallocate()`, but not a new `allocate()`.

@kanye-quest, I'm wary of adding this change to `axom::Array`. Thoughts on 
what could be the cause?

Backtrace of the error:
[spin_uniform_grid_backtrace.txt](https://github.com/LLNL/axom/files/8994739/spin_uniform_grid_backtrace.txt)

src/axom/primal/CMakeLists.txt
```diff
@@ -64,8 +64,8 @@ set( primal_headers
 
 set( primal_sources
 
-     ## operators
-     operators/detail/intersect_impl.cpp
+     ## Empty source file
+     dummy.cpp
```
bmhan12, line 68: To fix `lld: error: undefined hidden symbol` errors when linking HIP kernels to functions inside 
`intersect_impl.hpp`, I consolidated `intersect_impl.cpp` into the header file.

That left `primal` without any source files...

I tried making `primal` a header-only library as described in blt docs [here](https://llnl-blt.readthedocs.io/en/develop/api/target.html#blt-add-library), but that does not work for object libraries. So, I added an empty source file to satisfy the requirement.

@white238 I think this is not the ideal workaround. Any suggestions?

src/axom/quest/IntersectionShaper.hpp
```diff
@@ -326,6 +329,9 @@ class IntersectionShaper : public Shaper
                                                                 m_octs,
                                                                 m_octcount);
 
+    // Oddities required by hip
```
agcapps, line 332: I agree, this is an oddity.  But the comment should explain why it's needed, such as:
```suggestion
    // Oddities required by hip: get a local copy to avoid capturing `this`.
```

src/axom/quest/IntersectionShaper.hpp
```diff
@@ -499,6 +504,10 @@ class IntersectionShaper : public Shaper
     m_hexes = axom::allocate<PolyhedronType>(NE);
     m_hex_bbs = axom::allocate<BoundingBoxType>(NE);
 
+    // Oddities required by hip
```
agcapps, line 507: See my comment at (new) line 332.  Possibly use the following, if you want, and if it's accurate:
```suggestion
    // Oddities required by hip to avoid capturing `this`
```

src/axom/core/Array.hpp
agcapps: @bmhan12 , I agree that this seems strange and should be resolved before committing.

kanye-quest: I tried playing around with this a little while ago -- it seems like we'll sometimes get junk addresses with `malloc/realloc()`:

When the test passes:
```
[ RUN      ] spin_uniform_grid.indexing
[Instance 0x7fffffffbc80] Pointer address before realloc: 0x0
[Instance 0x7fffffffbc80] Pointer address after realloc: 0x660ebe0
[Instance 0x7fffffffbc80] Pointer address before realloc: 0x660ebe0
[Instance 0x7fffffffbc80] Pointer address after realloc: 0x660f4c0
``` 

When the test fails:
```
[ RUN      ] spin_uniform_grid.indexing
[Instance 0x7fffffffbc80] Pointer address before realloc: 0x0
[Instance 0x7fffffffbc80] Pointer address after realloc: 0x15554703d760
[Instance 0x7fffffffbc80] Pointer address before realloc: 0x15554703d760
--- segfaults here ---
```

FWIW, it doesn't seem like this problem appears in rznevada. On corona, our `malloc/realloc` seems to come from `mvapich2-2.3.6-aocc-4.5.2.0/lib/libmpi.so.12`, while on rznevada we're using `cray-mpich-8.1.16-rocmcc-5.1.1`. Maybe a bug in corona's MPI library?

kennyweiss: Can we switch our CI from `corona` to `tioga` as part of this PR? Or perhaps as a standalone PR that merges before this one?

We've been getting a number of system-related failures on `corona` in the past few weeks, and we're much more interested in the behavior on `tioga`.

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -735,6 +775,12 @@ class DistributedClosestPointImpl
     auto it = bvh->getTraverser();
     const int rank = m_rank;
 
+    double* sqDistThresh =
+      axom::allocate<double>(1, axom::execution_space<ExecSpace>::allocatorID());
```
kanye-quest, line 779: Could we use`axom::Array` here instead?

src/axom/primal/CMakeLists.txt
kennyweiss: @white238 would have the definitive answer here, but since he's currently working on removing object libraries from axom, the current solution seems reasonable to me.

src/axom/primal/operators/detail/intersect_impl.hpp
```diff
@@ -1209,6 +1209,490 @@ AXOM_HOST_DEVICE bool intersect_plane_seg(const Plane<T, 3>& plane,
   return false;
```
kennyweiss, line 1209: @bmhan12 -- can you please confirm that you used `git mv` to preserve the history on this file?

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -656,11 +695,12 @@ class DistributedClosestPointImpl
     const int npts = m_points.size();
     axom::Array<BoxType> boxesArray(npts, npts, m_allocatorID);
     auto boxesView = boxesArray.view();
+    auto pointsView = m_points.view();
 
     /// GOT TO HERE -- fix for templated ExecSpace!
     axom::for_all<ExecSpace>(
       npts,
-      AXOM_LAMBDA(axom::IndexType i) { boxesView[i] = BoxType {m_points[i]}; });
+      AXOM_LAMBDA(axom::IndexType i) { boxesView[i] = BoxType {pointsView[i]}; });
```
kennyweiss, line 703: Thanks for fixing this!

src/axom/core/Array.hpp
bmhan12: @kennyweiss  I think switching to tioga is a good idea, that can be done in a standalone PR. I am concerned with 
tioga/rzvernal currently being unstable, with nodes permanently going down on segfaults and requiring a restart by LC.  A fix looks to be coming today, so we'll have to see...

I will open a separate PR for tioga, once the tioga update goes through and gitlab CI comes back online to verify it works.

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -656,11 +695,12 @@ class DistributedClosestPointImpl
     const int npts = m_points.size();
     axom::Array<BoxType> boxesArray(npts, npts, m_allocatorID);
     auto boxesView = boxesArray.view();
+    auto pointsView = m_points.view();
 
     /// GOT TO HERE -- fix for templated ExecSpace!
```
kennyweiss, line 700: Minor: 
Could you please remove this line?

src/axom/core/Array.hpp
bmhan12: @kanye-quest  In the meantime, I've bypassed this error on `corona ` and reverting this axom::Array change by initializing the Uniform Grid in unified memory.

src/axom/sidre/tests/sidre_buffer.cpp
```diff
@@ -442,7 +442,8 @@ TEST_P(UmpireTest, reallocate)
 {
   constexpr int SIZE = 100;
 
-  #if defined(AXOM_USE_CUDA) && defined(UMPIRE_ENABLE_CONST)
+  #if(defined(AXOM_USE_CUDA) || defined(AXOM_USE_HIP)) && \
```
kennyweiss, line 445: Should we define an `AXOM_USE_DEVICE` which is set when either `AXOM_USE_CUDA` or `AXOM_USE_HIP`?

src/axom/spin/UniformGrid.hpp
```diff
@@ -796,7 +796,7 @@ void UniformGrid<T, NDIMS, ExecSpace, StoragePolicy>::getCandidatesAsArray(
         int count = counts_view[i];
         if(count > 0)
         {
-  #ifndef __CUDA_ARCH__
+  #ifndef AXOM_DEVICE_CODE
```
kennyweiss, line 799: Please confirm that this is not affected by #734

src/axom/sidre/tests/sidre_buffer.cpp
kennyweiss: Update: Looks like we have `AXOM_USE_GPU`.
Please use this consistently throughout.

See:
https://github.com/LLNL/axom/blob/223142718829f9590f652684a8ccb294fdfaf567/src/axom/core/Macros.hpp#L77-L84

src/axom/spin/tests/spin_implicit_grid.cpp
```diff
@@ -476,26 +481,28 @@ TYPED_TEST(ImplicitGridExecTest, get_candidates_pt_vectorized)
 
   GridT grid(bbox, &res, maxElts);
 
-  BBox objBox1(SpacePt {.15, .25, .05}, SpacePt {.45, .25, .35});
-  grid.insert(objBox1, 1);
+  BBox* objBox1 = axom::allocate<BBox>(1, kernelAllocID);
```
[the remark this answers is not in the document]
kennyweiss: I agree with both sides here.
@agcapps -- on the other hand, our implicit grid typically takes an entire array of objects, not just one, and I'm reasonably sure that we can still pass in an array (axom::ArrayView) of objects. (@bmhan12 -- please confirm).

Specifically, for a situation where we'd want to use GPUs, one wouldn't want to insert one object at a time, so this example was more about testing the code than in a realistic (performance-sensitive) use case.

src/axom/spin/tests/spin_implicit_grid.cpp
```diff
@@ -801,28 +823,28 @@ TYPED_TEST(ImplicitGridExecTest, get_candidates_box_vectorized)
 
     for(int i = 0; i < res[0]; ++i)
     {
-      grid.insert(BBox(SpacePt {0.025 + 0.1 * i, .05, .05},
-                       SpacePt {0.075 + 0.1 * i, .95, .95}),
-                  i);
+      BBoxes[i] = BBox {SpacePt {0.025 + 0.1 * i, .05, .05},
```
kennyweiss, line 826: Hmm -- this got a bit more awkward than I'd like...

Could we modify the code to add all the points to the BBox array and then call a single `insert` at the end?
(This test was checking functionality for `get_candidates` rather than `insert`, so I don't think it will change the behavior at all).

src/axom/spin/tests/spin_implicit_grid.cpp
```diff
@@ -833,32 +855,36 @@ TYPED_TEST(ImplicitGridExecTest, get_candidates_box_vectorized)
...
-                       // Box covers middle of domain
-                       BBox {SpacePt(0.25), SpacePt(0.75)},
-                       // Box is inverted -- BoundingBox constructor fixes this
-                       BBox {SpacePt(2.), SpacePt(-1)}};
+  axom::deallocate(BBoxes);
+  axom::deallocate(queryPt);
 
+  //// Run some queries
   constexpr int N_QUERIES = 8;
+  BBox* queryBoxes = axom::allocate<BBox>(N_QUERIES, kernelAllocID);
```
kennyweiss, line 870: Would it make sense to switch the tests in this file over to `axom::Array`
(perhaps this was suggested by others as well?)

kennyweiss (approved):
Thanks @bmhan12 -- this is a nice improvement and very comprehensive.

Did you confirm that all the hip examples work as expected? There are several that do not (yet) have associated tests, e.g. the distributed closest point.

Some observations:
* Our support for different execution spaces and runtime selection does not appear to be consistent across axom. We should discuss this as a team (after this PR) to see if we can agree on a preferred style.
* We should confirm that the `AXOM_DEVICE_CODE` usage isn't affected by #734 
* We should probably prefer `#if defined(AXOM_USE_GPU)` to `#if (defined(AXOM_USE_CUDA) || defined(AXOM_USE_HIP)` where appropriate
* The mint policies appear to have a lot of duplication w.r.t. cuda vs. hip. We might want to consider consolidating this if/where possible (?) to reduce the duplication.

src/axom/quest/DistributedClosestPoint.hpp
kennyweiss: That might give the wrong impression -- since it is a single scalar value for entire kernel.

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -735,6 +775,12 @@ class DistributedClosestPointImpl
     auto it = bvh->getTraverser();
     const int rank = m_rank;
 
+    double* sqDistThresh =
+      axom::allocate<double>(1, axom::execution_space<ExecSpace>::allocatorID());
+    sqDistThresh[0] = m_sqDistanceThreshold;
```
kennyweiss, line 780: ... but perhaps this one should be 

```suggestion
    *sqDistThresh = m_sqDistanceThreshold;
```

src/axom/core/Array.hpp
bmhan12: @kennyweiss  Yes, we could definitely do that.

rhornung67 (approved):
Nice work @bmhan12 !

src/axom/primal/operators/detail/intersect_impl.hpp
bmhan12: > @bmhan12 -- can you please confirm that you used `git mv` to preserve the history on this file?

No, I essentially appended the contents of `intersect_impl.cpp` to the end of `intersect_impl.hpp`. 

Is there a way to use `git mv` to preserve that migration and history of `intersect_impl.cpp`?

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -777,7 +777,8 @@ class DistributedClosestPointImpl
 
     double* sqDistThresh =
       axom::allocate<double>(1, axom::execution_space<ExecSpace>::allocatorID());
-    sqDistThresh[0] = m_sqDistanceThreshold;
+    *sqDistThresh = m_sqDistanceThreshold;
+    ;
```
kennyweiss, line 781: Minor: Extra semicolon

src/axom/spin/UniformGrid.hpp
bmhan12: > Please confirm that this is not affected by #734

Yes, this was okay. I don't have `mesh_tester` working with hip yet, but testing with CUDA on blueos, and sequential on blueos and toss 4 were fine.

src/axom/quest/DistributedClosestPoint.hpp
bmhan12: good catch!

src/axom/spin/tests/spin_implicit_grid.cpp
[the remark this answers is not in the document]
bmhan12: > @agcapps -- on the other hand, our implicit grid typically takes an entire array of objects, not just one, and I'm reasonably sure that we can still pass in an array (axom::ArrayView) of objects. (@bmhan12 -- please confirm).

Yes, you can pass in an array of objects. The unit test for Implicit Grid constructors was modified to test with different execution policies.

PR #860, opened 2022-06-30
Multimat ownership cleanup

kanye-quest:
# Summary

- Replaces raw pointers in Multimat used to hold Slam relational types with `axom::Array`
  - The future intent is to have these arrays be allocated in unified/device memory, allowing our Slam-based `MMFields` to be accessed on the device
  - This also makes the `MultiMat` class trivially-relocatable, and allows us to default move constructors and the destructor
- Avoid having to construct both cell-to-material and material-to-cell relations, as specified in a TODO comment
- Adds set accessor methods to `slam::DynamicVariableRelation` to more closely match API of `slam::StaticRelation`


src/axom/multimat/multimat.hpp
```diff
@@ -487,40 +511,123 @@ class MultiMat
...
+  RangeSetType& getMatSet() { return m_sets[1]; }
+  const RangeSetType& getMatSet() const { return m_sets[1]; }
+
+  /*!
+   * \brief Returns a reference to the corresponding array of offsets for a
+   *        static relation corresponding to a layout.
+   *
+   * \param layout The layout type of the relation (cell- or mat-dominant)
+   */
+  std::vector<SetPosType>& getRelBeginVec(DataLayout layout);
```
kennyweiss, line 534: Minor: Should the return type be a type alias (or possibly an existing type from the relation?) so it'll be easier to switch it over to `axom::Array`?

src/axom/multimat/multimat.hpp
```diff
@@ -487,40 +511,123 @@ class MultiMat
...
+   */
+  bool hasValidDynamicRelation(DataLayout layout) const;
 
 private:
   unsigned int m_ncells, m_nmats;
 
   //slam set variables
-  RangeSetType m_cellSet;
-  RangeSetType m_matSet;
+  axom::Array<RangeSetType> m_sets;
```
kennyweiss, line 608: Nice change! Since there are only two (I think), should this be an `axom::StackArray`? Or is it intentionally an `Array` for some other reason?

src/axom/multimat/multimat.cpp
```diff
@@ -649,20 +674,12 @@ void MultiMat::makeOtherRelation()
...
-  {
-    m_matCellNZSet = new RelationSetType(newRel);
-    m_matCellProdSet = new ProductSetType(&set2, &set1);
-  }
-  else
-  {
-    m_cellMatNZSet = new RelationSetType(newRel);
-    m_cellMatProdSet = new ProductSetType(&set2, &set1);
-  }
+  getRelSparseSet(layout) = RelationSetType(&newRel);
```
kennyweiss, line 681: Suggestion (minor) -- since `getRelSparseSet()` isn't just an accessor, but can also be a mutator, would it make sense to just name the function `relSparseSet()` (i.e. without the `get`)

src/axom/multimat/multimat.cpp
```diff
@@ -1147,24 +1144,25 @@ bool MultiMat::isValid(bool verboseOutput) const
...
       auto volfrac_sparsity = m_fieldSparsityLayoutVec[0];
       auto volfrac_layout = m_fieldDataLayoutVec[0];
 
       //Check Volfrac values match the relation value (if dense)
       if(volfrac_sparsity == SparsityLayout::DENSE && !m_dynamic_mode)
       {
-        const StaticVariableRelationType* relPtr =
-          (volfrac_layout == DataLayout::CELL_DOM ? m_cellMatRel : m_matCellRel);
+        const StaticVariableRelationType& relPtr =
+          m_staticRelations[(int)volfrac_layout];
```
kennyweiss, line 1156: I like this change (here and throughout)!

RELEASE-NOTES.md
```diff
@@ -94,6 +94,9 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Adds functions to compute the moments (area, centroid) of a `CurvedPolygon`
 - Adds functions to evaluate integrals over `BezierCurve` and `CurvedPolygon` objects 
 - Adds a `ArrayViewIndirection` storage policy to Slam
+- Adds set accessor methods to `slam::DynamicVariableRelation`
+- Adds a new component to Axom, `multimat`, to simplify the handing of multi-material meshes and
```
kennyweiss, line 98: 👍 !

kennyweiss (approved):
Thanks @kanye-quest -- this is a nice cleanup/simplification.

Thanks also for fleshing out the `DynamicVariableRelation` interface a bit more.

As you're making changes, could you please also run the examples w/ performance counters to ensure that we retain (or improve) the relative results between native, templated multimat and untemplated multimat as we presented in the paper -- https://zenodo.org/record/5559021 ? I think everything necessary should be checked in, but can dig deeper if necessary.

src/axom/multimat/multimat.hpp
kanye-quest: Using `axom::Array` was intentional, since alot of Slam objects are constructed with other Slam objects as pointers -- that way, we (hopefully) just have to allocate the Slam objects in unified memory, and can defer the value-based modifications to the Slam API till later. (I think we might still need/want to go through with the latter, to avoid dealing with virtual function calls on the device)

It also allows us to just `default` our move constructors for the `MultiMat` class, since we don't have a pointer-to-instance that we need to rebind.

PR #861, opened 2022-07-01
Set value number_of_domains in blueprint index to the correct value

nselliott:
# Summary

- This PR is a bug fix
- It does the following:
  - The value `number_of_domains` in the blueprint index produced by MFEMSidreDataCollection was not set correctly because it was set prior to the data collection getting access to the MPI communicator (the true number of domains is equal to the number of procs in the communicator).  This fix sets the value correctly immediately prior to writing the index to the root file.


kennyweiss (approved):
Thanks @nselliott !

dtaller (approved):
Just tried this in my code, it fixed things! Thanks!

rhornung67 (approved):
Thanks @nselliott

PR #862, opened 2022-07-05
Add tioga to Gitlab CI

bmhan12:
This PR:
- Adds tioga, removes corona from gitlab CI
  - Adds a `--test-serial` option to build_src.py to run unit tests serially (`make test -j16` is current default) 
    - This is to get around a srun/mpibind issue on tioga, where issuing multiple srun's in parallel results in `Fatal error in PMPI_Init` errors (recent discussion on Mattermost on why this is happening)

EDIT: Tioga job allowed to fail for now to safeguard against future, regular system updates.


kennyweiss (approved):
Thanks @bmhan12 !

rhornung67 (approved):
Looks good @bmhan12 Thank you.

PR #864, opened 2022-07-05
Fixes for building on Windows with Intel OneAPI.

agcapps:
# Summary

- This PR consists of two bugfixes from Al Nichols.
- It does the following:
  - Forward declares class Group within sidre/core/Buffer.hpp
  - Guards use of __builtin_ctz(n) and __builtin_ctzll(n) so against use with Clang-based Windows compilers

### DIFF
--- a/src/axom/sidre/core/Buffer.hpp
+++ b/src/axom/sidre/core/Buffer.hpp
@@ -33,6 +33,7 @@ namespace axom
 namespace sidre
 {
 class DataStore;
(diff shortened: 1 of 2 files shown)


rhornung67 (approved):
Tell Al, thank you!

kennyweiss (approved):
Thanks @agcapps and Al!

white238 (approved):

PR #866, opened 2022-07-13
Update devtools

white238:
Updates devtools on toss3 and blueos to really recent spack develop + my fixed cppcheck package which is sitting in a PR.

I spent enough time on this before I realized that we don't have toss4 devtools and that is going to have to be a Future Chris problem.

This also includes a new uberenv that has a non-backwards log.  It's amazing how long we lived with that.  Thanks @adrienbernede !


kennyweiss (approved):
Thanks @white238 !

bmhan12 (approved):
LGTM!

agcapps (approved):
Thank you for the update, @white238 .

scripts/spack/devtools_configs/blueos_3_ppc64le_ib_p9/packages.yaml
```diff
@@ -15,121 +15,121 @@
...
+  # Lock in versions of Devtools
   cmake:
-    version: [3.14.5]
-    buildable: false
     externals:
-    - spec: cmake
-      prefix: /usr/tce/packages/cmake/cmake-3.14.5
+    - spec: cmake@3.18.0
+      prefix: /usr/tce/packages/cmake/cmake-3.18.0
+    buildable: false
```
bmhan12, line 128: I'm curious: are just the devtool packages being built with cmake 3.18?

I see in the blueos host-config the header is still
`# CMake executable path: /usr/tce/packages/cmake/cmake-3.14.5/bin/cmake`

white238: Yes this is only for the devtools.  Due to not wanting to rebuild the devtools everytime (it more than doubles the builds), we have to split the configs like this.  I really wish there was a better way but I don't know of one.

PR #870, opened 2022-07-14
SLIC Loggers initialization

agcapps:
# Summary

- This PR is a bugfix to improve reliability of SLIC when compiled into multiple libraries 
- It does the following:
  - Modifies slic.hpp, slic.cpp, and Logger.cpp to store s_Logger (the active Logger) and s_loggers (the collection of Logger objects) as scoped to Logger.cpp instead of static in slic.hpp
  - Addresses issue #270


kennyweiss:
@agcapps -- did you also modify `slic.*pp`? Is so, they might be missing from this PR.

As posted, it's not clear that there will only be one copy of `s_logger` when slic is being used by multiple libraries.

src/axom/slic/core/Logger.cpp
```diff
@@ -16,8 +16,16 @@ namespace axom
...
-std::map<std::string, Logger*> Logger::s_loggers;
+static Logger* s_Logger = nullptr;
+
+using Loggermap = std::map<std::string, Logger*>;
+
+//------------------------------------------------------------------------------
+Loggermap & getLoggers()
+{
+  static Loggermap s_loggers;
+  return s_loggers;
```
agcapps, line 27: Here is the static singleton

agcapps:
> Thanks @agcapps This is better. Does it resolve all the initialization issues folks have run into?

I tested this with initialization of a statically linked C++ library; it does work as we wish.  I have not tested this with any Fortran libraries or codes.

src/axom/slic/core/Logger.cpp
```diff
@@ -16,8 +16,16 @@ namespace axom
 {
 namespace slic
 {
-Logger* Logger::s_Logger = nullptr;
-std::map<std::string, Logger*> Logger::s_loggers;
+static Logger* s_Logger = nullptr;
```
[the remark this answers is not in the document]
kennyweiss: I'm not sure that `s_Logger` is just a cache though, since it's used in `slic::isInitialized()` and `Logger::getActiveLogger()`. Please see my overall PR comment.

kennyweiss (changes requested):
Thanks @agcapps 

This seems closer, but I think part of the problem is still here (perhaps I'm missing something though).

We have `slic::isInitialized()` which checks that `Logger::getActiveLogger() != nullptr`. But the latter is defined as returning `s_Logger`, which, I think, still has the static initialization problem.

I think `s_Logger` also needs to be a singleton, or we need to modify `getActiveLogger()` and/pr `isInitialized()`.

See:
* https://github.com/LLNL/axom/blob/00729c226addccb6802f7bc3cbe9d193154f1bbf/src/axom/slic/interface/slic.cpp#L30
* https://github.com/LLNL/axom/blob/00729c226addccb6802f7bc3cbe9d193154f1bbf/src/axom/slic/core/Logger.cpp#L356

agcapps:
> Thanks @agcapps
> 
> This seems closer, but I think part of the problem is still here (perhaps I'm missing something though).
> 
> We have `slic::isInitialized()` which checks that `Logger::getActiveLogger() != nullptr`. But the latter is defined as returning `s_Logger`, which, I think, still has the static initialization problem.
> 
> I think `s_Logger` also needs to be a singleton, or we need to modify `getActiveLogger()` and/pr `isInitialized()`.

I singletonized `s_Logger`.  Please re-review.

agcapps:
@kennyweiss , I encapsulated the current logger pointer as a singleton, as requested.

src/axom/slic/core/Logger.cpp
```diff
@@ -322,9 +336,11 @@ bool Logger::createLogger(const std::string& name, char imask)
 //------------------------------------------------------------------------------
 bool Logger::activateLogger(const std::string& name)
 {
-  if(s_loggers.find(name) != s_loggers.end())
+  Loggermap& loggers = getLoggers();
+  if(loggers.find(name) != loggers.end())
   {
-    s_Logger = s_loggers[name];
+    Logger*& logger = getLogger();
+    logger = loggers[name];
```
kennyweiss, line 343: Could this be:
```suggestion
    getLogger() = loggers[name];
```

src/axom/slic/core/Logger.cpp
```diff
@@ -16,8 +16,21 @@ namespace axom
 {
 namespace slic
 {
-Logger* Logger::s_Logger = nullptr;
-std::map<std::string, Logger*> Logger::s_loggers;
+using Loggermap = std::map<std::string, Logger*>;
+
+//------------------------------------------------------------------------------
+Loggermap& getLoggers()
```
kennyweiss, line 22: Since it's not in the function name and this is a free function that's limited to the cpp file, could you please add a comment that this is a singleton (for `getLoggers` and `getLogger`)?

kennyweiss (approved):
Thanks @agcapps -- this looks right to me now

src/axom/slic/core/Logger.cpp
agcapps: done.

src/axom/slic/core/Logger.cpp
agcapps: done.

agcapps:
> Thanks @agcapps -- this looks right to me now

I'm very glad.

PR #871, opened 2022-07-15
Fix non-trivial axom::Array<T> insertion behavior

### LINKED ISSUE #869: Seg fault with insertions into nested axom::Array objects
Instances of `axom::Array` containing `std::vector` or other `axom::array` objects produce segmentation faults when additional objects are inserted. Issue is present on the current `develop` branch built on Windows machine, as well as on LC machine compiled with clang. For example:

```C++
axom::Array<double> a1({1.0, 1.1, 1.2});
axom::Array<double> a2({2.0, 2.1, 2.2});
axom::Array<double> a3({3.0, 3.1, 3.2});
axom::Array<double> a4({4.0, 4.1, 4.2});

axom::Array<axom::Array<double>> array_array({a1, a2, a3});

array_array.insert(0, a4);    // exception thrown in deallocate
array_array.insert(1, a4);    // exception thrown in deallocate
array_array.insert(3, a4);    // no exception thrown
```
Errors persist if `.insert` overload with iterator arguments are used.

Similarly, an `axom::Array` of `std::vector` objects throws a similar error in a different location:
```C++
std::vector<double> v1({1.0, 1.1, 1.2});
std::vector<double> v2({2.0, 2.1, 2.2});
std::vector<double> v3({3.0, 3.1, 3.2});
std::vector<double> v4({4.0, 4.1, 4.2});

axom::Array<std::vector<double>> array_vector({v1, v2, v3});

array_vector.insert(array_vector.begin(), v4);         // exception thrown in _Orphan_all_unlocked_v3
array_vector.insert(array_vector.begin() + 1, v4);     // exception thrown in _Orphan_all_unlocked_v3
array_vector.insert(array_vector.end(), v4);           // no exception thrown
```
with identical errors if `.insert` overload with array indices are used instead. The same functions called on a `std::vector` object containing `axom::Array` objects does not have the same issues. Looking at the call stack, it appears the issue involves the call to `std::move_backwards` by `move` in `ArrayBase.hpp`.

kanye-quest:
# Summary

Fixes a segfault caused by calling `axom::Array<T>::insert()/::erase()` for a non-trivial type.

In order to create a slot for a new element in the middle of an array, or compact an array after erasing elements in the middle, we have an `ArrayOps::move()` helper to move a block of elements within an array. The resulting call to `std::move/move_backwards` will move-assign elements in the array to an uninitialized index, containing a junk pointer-to-data. The move assignment operator then attempts to deallocate this junk pointer, which causes the segfault.

The fix is to use `std::uninitialized_copy` with move iterators to correctly handle move-construction of objects in uninitialized memory.

Resolves #869.

### DIFF
--- a/src/axom/core/tests/core_array_for_all.hpp
+++ b/src/axom/core/tests/core_array_for_all.hpp
@@ -499,6 +499,91 @@ AXOM_TYPED_TEST(core_array_for_all, dynamic_array_initializer_list)
 }
 
 //------------------------------------------------------------------------------
+AXOM_TYPED_TEST(core_array_for_all, dynamic_array_of_arrays)
+{
+  using ExecSpace = typename TestFixture::ExecSpace;
+  using DynamicArray = typename TestFixture::DynamicArray;
+  using HostArray = typename TestFixture::HostArray;
+  using ArrayOfArrays =
+    typename TestFixture::template DynamicTArray<DynamicArray>;
+  using HostArrayOfArrays =
+    typename TestFixture::template HostTArray<DynamicArray>;
+
+  int kernelAllocID = axom::execution_space<ExecSpace>::allocatorID();
+#if defined(AXOM_USE_GPU) && defined(AXOM_USE_UMPIRE)
+  if(axom::execution_space<ExecSpace>::onDevice())
+  {
+    kernelAllocID = axom::getUmpireResourceAllocatorID(
+      umpire::resource::MemoryResourceType::Device);
+  }
+#endif
+
+  constexpr axom::IndexType N = 5;
+  ArrayOfArrays arr_2d(0, N, kernelAllocID);
+
+  // Insert some arrays
+  arr_2d.push_back(DynamicArray({1, 2, 3}, kernelAllocID));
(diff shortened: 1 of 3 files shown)


src/axom/core/tests/core_array_for_all.hpp
```diff
@@ -498,6 +498,91 @@ AXOM_TYPED_TEST(core_array_for_all, dynamic_array_initializer_list)
...
+      {
+        EXPECT_EQ(subarr_host[j], i + j + 1);
+      }
+    }
+  }
+
+  // Insert an array at the beginning
+  // A move operation will be triggered to allocate a slot for the new
+  // element
+  arr_2d.insert(arr_2d.begin(), DynamicArray({0, 1, 2}));
```
kennyweiss, line 548: Could you please also add a test that inserts into a middle position?
(I'd expect this to be captured by writing to the beginning, but it'd be good to verify that).

kennyweiss (approved):
Thanks @kanye-quest !

jcs15c (approved):
Thank you for fixing this!

PR #873, opened 2022-07-20
Fix sign error in primal::Triangle::signedArea()

jcs15c:
# Summary

This PR fixes a sign error in the `signedArea()` method of `primal::Triangle` objects, and updates the relevant test in `primal_triangle` that expects the incorrect behavior. Currently, triangles whose vertices are oriented counter-clockwise incorrectly return a negative area. This is also inconsistent with the behavior described by the Doxygen header. 

This is fixed by adjusting the determinant calculated in `signedArea()` to be consistent with the ["Shoelace Theorem"](https://artofproblemsolving.com/wiki/index.php/Shoelace_Theorem) for finding the oriented area of a polygon. The relevant test in `primal_triangle` is adjusted so that the points are oriented counter-clockwise.

### DIFF
--- a/src/axom/primal/geometry/Triangle.hpp
+++ b/src/axom/primal/geometry/Triangle.hpp
@@ -138,8 +138,8 @@ public:
     const PointType& C = m_points[2];
 
     // clang-format off
-    return 0.5 * determinant(B[0]-A[0], B[1]-A[1],
-                             C[0]-A[0], C[1]-A[1]);
+    return 0.5 * determinant(B[0]-A[0], C[0]-A[0],
+                             B[1]-A[1], C[1]-A[1]);
     // clang-format on
   }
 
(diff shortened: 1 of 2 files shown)


rhornung67 (approved):
wow. how was that working?

bmhan12 (approved):
Nice catch!

kennyweiss (approved):
Thanks @jcs15c !

kennyweiss:
Thanks for fixing this @jcs15c -- looks like I introduced the bug a few months ago (https://github.com/LLNL/axom/commit/28d481c1e373071e5a07a8bb097c077a69980aef) while cleaning up `Triangle::area()`

@rhornung67 wrote:
> wow. how was that working?

The code was giving the opposite sign (and the tests were checking this). Luckily, it looks like we were only using this in the unit tests. The other usage takes the absolute value of the `signedArea()` to get the (unsigned) `area()` of a triangle.

jcs15c:
Thank you all for the reviews!

PR #878, opened 2022-07-22
Misc cleanup

### LINKED ISSUE #874: Allow users to create their own `mfem` target
Serac currently allows a "codevelop" build which allows us to include MFEM and Axom as part of the Serac build system, included via `add_subdirectory` in Serac.

If a CMake target exists by the name of `mfem` when we go to find MFEM ourselves, we should skip the find and creation of our `mfem` target.

white238:
This PR does two cleanup things:

1. Allows users to provide their own `mfem` CMake target.  This is useful for projects that co-develop multiple projects together. For example, Serac has the option to co-develop MFEM and Axom in one build. Fixes #874
2.  Removes the now unnecessary sparsehash header check. `std::tr1` was deprecated pre-c++11


rhornung67 (approved):

kennyweiss (approved):
Thanks @white238

white238:
Still working on smoothing out some CMake details...

kennyweiss (approved):

PR #881, opened 2022-07-26
Adds winding number computations for BezierCurve objects

jcs15c:
# Summary

- This PR is a feature

It implements winding number computations along arbitrary Bezier curves that can be used to perform in/out queries for CurvedPolygon objects. 

The winding number is a measure of how many times a curve loops around a given query point, and can be used for containment queries. Given a closed curve, if the winding number evaluates to a nonzero integer, the query point is on the interior. If the winding number evaluates to zero, the query point is interior. For curves that are not closed/watertight/continuous, the winding number will be a fractional value that corresponds heuristically to "closeness to the interior/exterior" and the orientation of the closest curve. For points exactly on the boundary, the winding number is mathematically ambiguous, evaluating to a value between the winding numbers on either side of the curve.

This is performed using a novel bisection algorithm that splits BezierCurve objects until the query point is outside of its convex control polygon, at which point a direct formula is used. It also adds relevant operators to `primal` to facilitate this algorithm, including

- `bool in_polygon( Point, Polygon )` Determines if a query point is contained in a polygon using winding numbers
- `bool is_convex( Polygon )` Determines if a polygon is convex

Additionally, adjustments have been made to `intersect_ray(Ray, Segment)` to account for edge cases involving parallel rays and segments. Nonfunctional changes have also been made to `primal::BezierCurve` and `primal::CurvedPolygon` to use `axom::Array` objects instead of `std::vector`.

### Examples
The main function `double primal::winding_number( Point, CurvedPolygon )` computes the winding number for a single point. To generate these examples, it was run for each point along a uniform 2D grid.
example2

example3

example1


src/axom/primal/operators/is_convex.hpp
```diff
@@ -0,0 +1,58 @@
...
+namespace axom
+{
+namespace primal
+{
+/*!
+ * \brief Determines if a polygon defined by ordered vertices is convex
+ * 
+ * \param [in] poly The polygon
+ * 
+ * Uses algorithm adapted from [Maa 99], which uses dot products to detect
```
kennyweiss, line 22: Please expand the reference to `[Ma 99]`.

I think it is
> Y. L. Ma, W.T. Hewitt. "Point inversion and projection for NURBS curve and surface: Control polygon approach"
Computer Aided Geometric Design 20(2):79-99, May 2003.

Is that right?

src/axom/primal/operators/is_convex.hpp
```diff
@@ -0,0 +1,58 @@
...
+ * 
+ * \return A boolean value indicating convexity
+ */
+template <typename T>
+bool is_convex(const Polygon<T, 2>& poly)
+{
+  int n = poly.numVertices() - 1;
+  if(n + 1 < 3) return true;  // Triangles and lines are convex
+
+  for(int i = 1; i < n; i++)
```
kennyweiss, line 35: Please add a (brief) comment about why we don't need to check the edge from vertex 0 to vertex N.

src/axom/primal/tests/primal_curved_polygon.cpp
```diff
@@ -53,8 +53,8 @@ void checkMoments(const primal::CurvedPolygon<CoordType, 2>& bPolygon,
  */
 template <typename CoordType, int DIM>
 primal::CurvedPolygon<CoordType, DIM> createPolygon(
-  const std::vector<primal::Point<CoordType, DIM>> ControlPoints,
-  const std::vector<int> orders)
+  const axom::Array<primal::Point<CoordType, DIM>> ControlPoints,
```
kennyweiss, line 56: Thanks for converting these tests from `std::vector` to `axom::Array` and detailing the missing features described in #869

src/axom/primal/tests/primal_curved_polygon.cpp
```diff
@@ -74,13 +74,14 @@ primal::CurvedPolygon<CoordType, DIM> createPolygon(
   int iter = 0;
   for(int j = 0; j < num_edges; ++j)
   {
-    std::vector<PointType> subCP;
-    subCP.assign(ControlPoints.begin() + iter,
-                 ControlPoints.begin() + iter + orders[j] + 1);
+    axom::Array<PointType> subCP(orders[j] + 1);
+    for(int i = 0; i < orders[j] + 1; i++) subCP[i] = ControlPoints[i + iter];
```
kennyweiss, line 78: Are we missing `axom::Array::assign()`? 
Would it be worthwhile adding it?

If so, please create an issue for that.

src/axom/primal/tests/primal_curved_polygon.cpp
```diff
@@ -235,18 +236,26 @@ TEST(primal_curvedpolygon, split_edge)
...
                                PointType {0.3, 2.0},
                                PointType {0.0, 1.6},
                                PointType {0.6, 1.2}};
 
-  std::vector<int> orders32 = {1, 1, 1};
+  std::cout << CP << std::endl;
+  CP.insert(0, CP[0]);
+  std::cout << CP << std::endl;
+
+  return;
```
kennyweiss, line 248: Is there a reason for the early return?
Please try to remove it and fix any problems that you encounter.

kennyweiss: ... and associated debug code/comments

src/axom/primal/tests/primal_polygon.cpp
```diff
@@ -16,6 +16,133 @@ TEST(primal_polygon, polygon_empty)
...
+  axom::Array<PointType> vertices({PointType {0, 0}, PointType {1, 1}});
+  PolygonType poly(vertices);
+
+  // Segments and Triangles are always convex
+  EXPECT_TRUE(is_convex(poly));
+
+  poly.addVertex(PointType {0, 1});
+  EXPECT_TRUE(is_convex(poly));
+
+  // Duplicate points should not affect convexity
```
kennyweiss, line 114: 👍

kennyweiss (approved):
Thanks @jcs15c -- this is a really nice feature addition that also improves other parts of the code as necessary.

Approved, pending some minor changes:
* Please update the RELEASE_NOTES file as necessary. At the very least, there should be a note about the change to intersect(ray, segment) since it changes behavior (for the better)
* Please also fix the early return in the `primal_curved_polygon` test
* Thanks for all the new unit tests! For some of these, would it make sense to add test cases/geometry that is not aligned with the coordinate axes and origin?

src/axom/primal/operators/detail/in_curved_polygon_impl.hpp
```diff
@@ -0,0 +1,159 @@
...
+{
+namespace primal
+{
+namespace detail
+{
+template <typename T>
+int sgn(T val)
+{
+  return (T(0) < val) - (val < T(0));
+}
```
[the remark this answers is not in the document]
gunney1: I agree.  Good to match wide-spread norms.

src/axom/primal/geometry/Polygon.hpp
```diff
@@ -14,6 +14,9 @@
 
 #include "axom/core/Array.hpp"
 #include "axom/primal/geometry/Point.hpp"
+#include "axom/primal/geometry/Segment.hpp"
+#include "axom/primal/operators/orientation.hpp"
+#include "axom/primal/operators/detail/intersect_ray_impl.hpp"
```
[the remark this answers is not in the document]
jcs15c: Nope, those should definitely be gone now. Thanks for pointing it out!

src/axom/primal/operators/detail/in_curved_polygon_impl.hpp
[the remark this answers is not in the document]
jcs15c: This was actually only a temporary function I added that I forgot to remove. That being said, I believe that there should be a sign function somewhere in axom, as it'd be useful to not redefine it every time it's needed. 

As for the name, I've mostly seen it in math contexts abbreviated as `sgn`, but I think `sign` is just as reasonable, especially if it's used as such elsewhere.

src/axom/primal/operators/detail/intersect_ray_impl.hpp
```diff
@@ -41,23 +42,48 @@ inline bool intersect_ray(const primal::Ray<T, 2>& R,
 {
   AXOM_STATIC_ASSERT(std::is_floating_point<T>::value);
 
-  // STEP 0: Find the parameterized direction of the segment
+  // STEP 0: Find the parameterized direction of the segment.
+  //         efine lower/upper threshold
```
gunney1, line 46: Delete this line?

src/axom/primal/tests/primal_curved_polygon.cpp
kanye-quest: We do have a method `axom::Array::set()` which sets a range of values in the `Array`. Maybe it would be worth looking into renaming that to `assign()` to correspond with the `std::vector` interface?

src/axom/primal/tests/primal_curved_polygon.cpp
jcs15c: Oops! Leftovers from when I was investigating the buggy behavior with `axom::Arrays`. All cleaned up now.

src/axom/primal/operators/in_polygon.hpp
```diff
@@ -0,0 +1,151 @@
...
+  // If the query is a vertex, return a value interpreted
+  //  as "inside" by evenodd or nonzero protocols
+  if(R == P[0]) return !strict;
+
+  int winding_num = 0;
+  for(int i = 0; i < nverts; i++)
+  {
+    int j = (i == nverts - 1) ? 0 : i + 1;
+
+    if(P[j][1] == R[1])
```
kennyweiss, line 65: Should these equality tests be converted to fuzzy equality tests?
E.g.
```diff
-    if(P[j][1] == R[1])
+    if(axom::utilities::isNearlyEqual(P[j][1],R[1],EPS))
```
Or perhaps, the equality test is related to the `strict` parameter?

(throughout this function)

src/axom/primal/operators/in_polygon.hpp
```diff
@@ -0,0 +1,151 @@
...
+ * Uses different protocols to determine containment from the winding number.
+ *   nonzero (true): If the winding number is nonzero, the point is interior.
+ *   evenodd (false): If the winding number is odd, it is interior. Exterior otherwise.
+ 
+ * \return boolean value indicating containment.
+ */
+template <typename T>
+bool in_polygon(const Point<T, 2>& query,
+                const Polygon<T, 2>& poly,
+                const bool nonzero = true,
```
kennyweiss, line 139: Suggestion: 
This parameter might be easier to decipher if it were called something like `useNonZeroRule`:
e.g.
```suggestion
                const bool useNonzeroRule = true,
```
(w/ updated documentation)

src/axom/primal/operators/in_polygon.hpp
```diff
@@ -0,0 +1,151 @@
...
+                const bool strict = false,
+                const double EPS = 1e-8)
+{
+  if(nonzero == true) return winding_number(query, poly, strict, EPS) != 0;
+  // else, use evenodd rule
+  return (winding_number(query, poly, strict, EPS) % 2) == 1;
```
kennyweiss, line 145: Suggestion (subjective): This is fine as is, but please consider this minor tweak for clarity

```suggestion
   // use either "nonzero" rule or "evenodd" rule
   return  nonzero ?  winding_number(query, poly, strict, EPS) != 0
                   :  (winding_number(query, poly, strict, EPS) % 2) == 1;
```

kennyweiss (approved):
Thanks for the updates @jcs15c -- the point-in-polygon tests looks significantly more robust!

src/axom/primal/operators/in_polygon.hpp
jcs15c: It realistically should be a fuzzy equality, yes. The initial concern was that `EPS` in this context has a different interpretation than the `EPS` used for the determinant below. For these equalities, it's the literal distance between the points, but for the determinant, it's less direct, as it isn't the distance from the point to the edge.

For now though, I think the `isNearlyEqual` works.

src/axom/primal/operators/in_polygon.hpp
jcs15c: This is much more clear, thank you for the suggestion!

src/axom/primal/operators/detail/intersect_ray_impl.hpp
jcs15c: Fixed now, along with some of the other debugging comments. Thank you!

rhornung67 (approved):

src/axom/primal/operators/in_curved_polygon.hpp
```diff
@@ -0,0 +1,114 @@
...
+inline bool in_curved_polygon(const Point<T, 2>& query,
+                              const CurvedPolygon<T, 2>& cpoly,
+                              const bool useNonzeroRule = true,
+                              const double linear_tol = 1e-8,
+                              const double edge_tol = 1e-8)
+{
+  double winding_num = winding_number(query, cpoly, linear_tol, edge_tol);
+
+  // Else, use EvenOdd rule
+  return useNonzeroRule ? std::round(winding_num)
```
kennyweiss, line 57: Minor: 
(a) should this use `lround` to convert to an integer?
(b) after converting to an integer, it might be clearer to explicitly test against zero.

E.g.
```suggestion
  return useNonzeroRule ? (std::lround(winding_num) != 0)
```

src/axom/primal/operators/in_polygon.hpp
```diff
@@ -0,0 +1,153 @@
...
+int winding_number(const Point<T, 2>& R,
+                   const Polygon<T, 2>& P,
+                   const bool strict = false,
+                   const double EPS = 1e-8)
+{
+  const int nverts = P.numVertices();
+
+  // If the query is a vertex, return a value interpreted
+  //  as "inside" by evenodd or nonzero protocols
+  if(axom::utilities::isNearlyEqual(P[0][0], R[1], EPS) &&
```
kennyweiss, line 58: Minor suggestion (feel free to ignore):

If you add the following to the top of the function, you can remove the namespaces a bit.
e.g.
```cpp
// at the top
using axom::utilities::isNearlyEqual;
using axom::numerics::determinant;
```

Then later, you can write:
```cpp
   if(isNearlyEqual(P[j][1], R[1], EPS)) {...}
   ...
   det = determinant(P[i][0] - R[0], P[j][0] - R[0],
                                P[i][1] - R[1], P[j][1] - R[1]);
```

src/axom/primal/tests/primal_curved_polygon.cpp
kennyweiss: See #885

src/axom/primal/operators/in_polygon.hpp
jcs15c: I'll keep it as is for this PR, but I'll definitely keep this in mind for the future. Thank you!

PR #887, opened 2022-08-03
More submodule work

white238:
I am getting closer but I am beginning to think it's not really possible to toggle between `add_subdirectory()` and `find_package()` in a clean way.

This at least allows me to build and test Axom/MFEM inside of Serac but not install.  Though that is not our normal usage so this seems okay for now.

### DIFF
--- a/src/cmake/thirdparty/SetupAxomThirdParty.cmake
+++ b/src/cmake/thirdparty/SetupAxomThirdParty.cmake
@@ -138,9 +138,14 @@ endif()
 # MFEM
 #------------------------------------------------------------------------------
 if (TARGET mfem)
-    # Case: included axom in project that also creates an mfem target, no need to recreate mfem
+    # Case: Axom included in project that also creates an mfem target, no need to recreate mfem
+    # Note - white238: I can't seem to get this to pass install testing due to mfem being included
+    # in multiple export sets
     message(STATUS "MFEM support is ON, using existing mfem target")
-    blt_list_append(TO TPL_DEPS ELEMENTS mfem)
+    # Add it to this export set but don't prefix it with axom::
+    install(TARGETS              mfem
+            EXPORT               axom-targets
+            DESTINATION          lib)
     set(MFEM_FOUND TRUE CACHE BOOL "" FORCE)
 elseif (MFEM_DIR)
     include(cmake/thirdparty/FindMFEM.cmake)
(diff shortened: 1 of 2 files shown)


rhornung67 (approved):

bmhan12 (approved):

src/cmake/thirdparty/SetupAxomThirdParty.cmake
```diff
@@ -138,9 +138,14 @@ endif()
 # MFEM
 #------------------------------------------------------------------------------
 if (TARGET mfem)
-    # Case: included axom in project that also creates an mfem target, no need to recreate mfem
+    # Case: Axom included in project that also creates an mfem target, no need to recreate mfem
+    # Note - white238: I can't seem to get this to pass install testing due to mfem being included
+    # in multiple export sets
     message(STATUS "MFEM support is ON, using existing mfem target")
-    blt_list_append(TO TPL_DEPS ELEMENTS mfem)
+    # Add it to this export set but don't prefix it with axom::
```
kennyweiss, line 145: I'm sure you played around with this, but would it make sense for axom to export this with the `axom::` prefix, but for the "host code" (of which axom and mfem are submodules) to export it as `mfem` ?

Specifically, my suggestion here is for axom to export the target as `axom::mfem` whether it was imported or already a target.

kennyweiss (approved):
Thanks @white238

src/cmake/thirdparty/SetupAxomThirdParty.cmake
white238: In this case, `mfem` was a real target not an imported target and CMake had checks that it did not exist in multiple exported configs.

cyrush:
For the record: `add_subdirectory()`  to fuse CMake build systems remains an painful path to maintain. 

I also think trying to magically export targets vs having the imported cmake logic just call find_package creates issues that are unique to how we are doing things vs most other cmake projects.

kanye-quest:
Spitballing here: could we use `ExternalProject_Add()` in lieu of `add_subdirectory()`? I think that'd let us independently build and install our submodule TPLs without pulling stuff into the axom cmake context. Then we could just add it like any other TPL with `find_package()`.

white238:
> Spitballing here: could we use `ExternalProject_Add()` in lieu of `add_subdirectory()`? I think that'd let us independently build and install our submodule TPLs without pulling stuff into the axom cmake context. Then we could just add it like any other TPL with `find_package()`.

`ExternalProject_Add()` is also very painful. I have tried it in the past, though its been a while, and was very unhappy with it.

This is for when Serac is heavily codeveloping Axom and (mainly) MFEM. It is not how most of the developers work or how we expect users to regularly build our code.  It isn't even recommended unless you specifically need to do so. It is for the specific developers who need to bridge the gap between the 3 projects. It also goes up one more layer to another project as well now and has been extremely helpful to speed up development.

PR #893, opened 2022-08-09
Add rational Bezier curve functionality to `BezierCurve`

jcs15c:
# Summary

- This PR is a feature
- It enhances the `BezierCurve` class to allow for rational Bezier curves, which adds a weight to each control node.
- It adds tests for rational Bezier curves, and ensures compatibility with winding number algorithms.


src/axom/primal/tests/primal_rational_bezier.cpp
```diff
@@ -0,0 +1,394 @@
...
+      EXPECT_DOUBLE_EQ(-1., c1[0][0]);
+      EXPECT_DOUBLE_EQ(-5., c1[0][1]);
+      EXPECT_DOUBLE_EQ(0.0, c1[1][0]);
+      EXPECT_DOUBLE_EQ(0.0, c1[1][1]);
+
+      EXPECT_DOUBLE_EQ(0.0, c2[0][0]);
+      EXPECT_DOUBLE_EQ(0.0, c2[0][1]);
+      EXPECT_DOUBLE_EQ(1.0, c2[1][0]);
+      EXPECT_DOUBLE_EQ(5.0, c2[1][1]);
+    }
```
kennyweiss, line 274: These lines caused some test failures on LLNL's blueos host-configs:

```
145: [INFO] Testing rational Bezier splitting for order 1 
145: ///axom/src/axom/primal/tests/primal_rational_bezier.cpp:268: Failure
145: Expected equality of these values:
145:   0.0
145:     Which is: 0
145:   c1[1][1]
145:     Which is: -1.3877787807814457e-16
145: ///axom/src/axom/primal/tests/primal_rational_bezier.cpp:271: Failure
145: Expected equality of these values:
145:   0.0
145:     Which is: 0
145:   c2[0][1]
145:     Which is: -1.3877787807814457e-16
145: ///axom/src/axom/primal/tests/primal_rational_bezier.cpp:268: Failure
145: Expected equality of these values:
145:   0.0
145:     Which is: 0
145:   c1[1][1]
145:     Which is: -1.3877787807814457e-16
145: ///axom/src/axom/primal/tests/primal_rational_bezier.cpp:271: Failure
145: Expected equality of these values:
145:   0.0
145:     Which is: 0
145:   c2[0][1]
145:     Which is: -1.3877787807814457e-16
145: [  FAILED  ] primal_rationalbezier.split_linear (0 ms)
```

Might be resolved using `EXPECT_NEAR` with an exlicit `EPS` around 1e-12 ?

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -174,11 +264,18 @@ class BezierCurve
   /// Retrieves the control point at index \a idx
   const PointType& operator[](int idx) const { return m_controlPoints[idx]; }
 
+  /// Get specific weight
+  const T& getWeight(int idx) const { return m_weights[idx]; }
+
+  /// Change specific weight
+  void setWeight(int idx, T weight) { m_weights[idx] = weight; };
```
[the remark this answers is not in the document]
kennyweiss: Perhaps, e.g. by adding an `SLIC_ASSERT(this->isRational())` and a `\pre` precondition clause to the doxygen comment.

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -190,7 +287,10 @@ class BezierCurve
   /// Returns a copy of the Bezier curve's control points
   CoordsVec getControlPoints() const { return m_controlPoints; }
 
-  /// Reverses the order of the Bezier curve's control points
+  /// Returns a copy of the Bezier curve's control points
+  axom::Array<T> getWeights() const { return m_weights; }
```
kennyweiss, line 291: Do we want a deep copy of the weights?
(Looks like you were keying off how `getControlPoints()` was implemented a few lines up)

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -199,6 +299,12 @@ class BezierCurve
     {
       axom::utilities::swap(m_controlPoints[i], m_controlPoints[ord - i]);
     }
+
+    if(m_weights[0] > 0)
```
kennyweiss, line 303: Suggestion:
```suggestion
    if(isRational())
```

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -232,26 +338,57 @@ class BezierCurve
...
+        for(int p = 1; p <= ord; ++p)
+        {
+          const int end = ord - p;
+          for(int k = 0; k <= end; ++k)
+          {
+            dCarray[k] = lerp(dCarray[k], dCarray[k + 1], t);
+            dWarray[k] = lerp(dWarray[k], dWarray[k + 1], t);
+          }
+        }
+        ptval[i] = dCarray[0] / dWarray[0];
```
kennyweiss, line 364: Do we have do guard against division by zero?

src/axom/primal/tests/primal_winding_number.cpp
```diff
@@ -116,6 +116,58 @@ TEST(primal_winding_number, simple_cases)
               abs_tol);
 }
 
+TEST(primal_winding_number, rational_cases)
```
kennyweiss, line 119: It would be cool to also add a test with several rational quadratics or cubics defining a circle and compare the winding number of those against point containment in a `primal::Circle` of the same radius and center for a variety of test points

kennyweiss (changes requested):
Thanks @jcs15c -- this is great!

Can you please also add tests for the primal operators that use Bezier curves and ensure that those work properly. E.g. `is_linear()`, `intersect()`, `evaluate_integral()`, ... `<others?>`

Please also add a documentation note and a `SLIC_ASSERT` for the ones that will not directly work with rational Beziers, e.g. `compute_moments()`.

If you'd prefer to merge this sooner and add those in a follow-up PR, that's fine, but please add an issue to track this.

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -81,6 +87,9 @@ class BezierCurve
     SLIC_ASSERT(ord >= -1);
     const int sz = utilities::max(-1, ord + 1);
     m_controlPoints.resize(sz);
+
+    m_weights.resize(1);
+    m_weights[0] = -1.0;
```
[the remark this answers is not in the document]
jcs15c: This is much more clear. Thank you!

src/axom/primal/geometry/BezierCurve.hpp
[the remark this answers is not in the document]
jcs15c: These `SLIC_ASSERT` statements don't get run on a release build, correct? Would a proper `if( !isRational() )` statement for each call just be too expensive at runtime?

src/axom/primal/geometry/BezierCurve.hpp
jcs15c: You're right, I was just following the model of `getControlPoints()`, so I'm not sure if we want a deep copy of the weights. Does this one do something different than `getControlPoints()`? I would think that both return a deep copy.

src/axom/primal/geometry/BezierCurve.hpp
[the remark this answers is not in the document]
kennyweiss: The idea is that we guard against this in debug builds but don't pay the performance hit during release builds.

I.e. calling this function on a non-rational curve is a developer error, and we assume that at runtime people know what they're doing, but we want to help them debug if there's a problem.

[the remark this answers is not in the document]
kennyweiss: Another way to think about it is that it might be worse to silently not set the weight for non-rational curves than to allow the code to segfault, since the latter might cause the code to crash in a later, unrelated place.

src/axom/primal/geometry/BezierCurve.hpp
kennyweiss: Thanks, I was questioning if either of these should be a deep copy. 
I think it's fine for now and we can change it in the future as necessary/desired.

src/axom/primal/geometry/BezierCurve.hpp
jcs15c: The recursive algorithms from [Farin 83] assume positive weights at each node. Is that too strong of an assumption for rationals generated from NURBS curves? 

In this case, even if we allow for nonnegative weights in general, the value of `dWarray[0]` can only mathematically be zero if the weight at one or both of the endpoints is zero, which shouldn't be allowed in general.

jcs15c: For the time being at least, I've just made the requirements that all weights be positive, as it makes the computation in `.split()` more straightforward in the end, but this could change in the future.

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -37,8 +37,9 @@ namespace detail
  * \param [in] quad the mfem integration rule containing nodes and weights
  * \return the value of the integral
  */
+template <typename T, int NDIMS>
 inline double evaluate_line_integral_component(
-  const primal::BezierCurve<double, 2>& c,
+  const primal::BezierCurve<T, NDIMS>& c,
   std::function<double(Point2D)> scalar_integrand,
```
kennyweiss, line 43: Should this still be a `Point2D`? Or does the type `T` and/or dimension `NDIMS` need to update as well?

jcs15c: You're right, while the line integrals should work in arbitrary dimensions, I'd forgotten that the way we define the `std::function` integrands restricts them to 2D. I still feel as if it should work for arbitrary dimensions, or at least be templated instead of `double` precision, but I'll revert these changes and make a more thorough adjustment later if it comes to it.

src/axom/primal/tests/primal_rational_bezier.cpp
```diff
@@ -0,0 +1,560 @@
...
+  CPolygon quarter_circle;
+  quarter_circle.addEdge(circle_arc);
+  quarter_circle.addEdge(leg1);
+  quarter_circle.addEdge(leg2);
+
+  for(double theta = 0.01; theta < 1.5; theta += 0.05)
+    for(int i = 1; i < 9; i++)
+    {
+      double ri = 1.0 - std::pow(10, -i);
+      double ro = 1.0 + std::pow(10, -i);
```
kennyweiss, line 417: Minor suggestion (for readability):
```suggestion
      const double offset = std::pow(10, -i);
      const double ri = 1.0 - offset;
      const double ro = 1.0 + offset;
```

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -49,6 +49,12 @@ std::ostream& operator<<(std::ostream& os, const BezierCurve<T, NDIMS>& bCurve);
  * The order of a Bezier curve with N+1 control points is N.
  * The curve is approximated by the control points,
  * parametrized from t=0 to t=1.
+ * 
+ * Contains an array of weights to represent a rational Bezier curve.
+ * Nonrational Bezier curves are identified by an empty weights array.
+ * Algorithms for Rational Bezier curves derived from
```
kennyweiss, line 55: Please add a note that the weights for rational Bezier curves must be strictly positive

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -174,11 +275,40 @@ class BezierCurve
...
+    m_weights[idx] = weight;
+  };
+
   /// Checks equality of two Bezier Curve
   friend inline bool operator==(const BezierCurve<T, NDIMS>& lhs,
                                 const BezierCurve<T, NDIMS>& rhs)
   {
-    return lhs.m_controlPoints == rhs.m_controlPoints;
+    return (lhs.m_controlPoints == rhs.m_controlPoints) &&
+      (lhs.m_weights == rhs.m_weights);
```
kennyweiss, line 311: Note -- this implementation would return false when comparing a nonrational curve and a rational curve whose weights are all 1? 

I think this is the right call, but wanted to highlight it.

src/axom/primal/geometry/BezierCurve.hpp
kennyweiss: Thanks. I agree that, assuming that weights must be strictly positive, we should be ok w.r.t. division by zero.

kennyweiss (approved):
Thanks for incorporating all the suggestions @jcs15c -- this looks great!

The new set of tests for rational Bezier curves is impressive!

src/axom/primal/operators/compute_moments.hpp
```diff
@@ -66,6 +69,9 @@ T sector_area(const primal::BezierCurve<T, 2>& curve)
 template <typename T>
```
kennyweiss, line 69: Please also add a note to the doxygen for these functions that they only work on rational Bezier curves.

kennyweiss: Actually, perhaps we should implement the `area()` and `centroid()` for rational polygons  in terms of `evaluate_integral` ?
Can you please create an issue to track this?

src/axom/primal/operators/compute_moments.hpp
```diff
@@ -103,6 +109,9 @@ T area(const primal::CurvedPolygon<T, 2>& poly, double tol = 1e-8)
   {
     for(int ed = 0; ed < ngon; ++ed)
     {
+      // Algorithm works only on nonrational Bezier curves
+      SLIC_ASSERT(!ed.isRational());
```
kennyweiss, line 113: Typo:
Looks like this should be:
```suggestion
      SLIC_ASSERT(!poly[ed].isRational());
```

src/axom/primal/operators/compute_moments.hpp
```diff
@@ -10,7 +10,7 @@
  * \file compute_moments.hpp
  *
  * \brief Consists of a set of methods to compute areas/volumes and centroids 
- * for Polygons and CurvedPolygons composed of BezierCurves
+ * for Polygons and CurvedPolygons composed of nonrational BezierCurves
```
rhornung67, line 13:
```suggestion
 * for Polygons and Curved Polygons composed of non-rational Bezier Curves
```

src/axom/primal/operators/compute_moments.hpp
```diff
@@ -28,7 +28,7 @@ namespace axom
 namespace primal
 {
 /*!
-   * \brief Calculates the sector area of a planar Bezier Curve
+   * \brief Calculates the sector area of a planar, nonrational Bezier Curve
```
rhornung67, line 31:
```suggestion
   * \brief Calculates the sector area of a planar, non-rational Bezier Curve
```

rhornung67 (commented):
One question about word choice (see comments). Otherwise, looks good.

src/axom/primal/geometry/BezierCurve.hpp
jcs15c: This is what I had in mind, yes. I interpreted it this way since it seems like, without any tolerances on the control nodes, this is more a check of if they are the same C++ object, rather than the same mathematical object.

src/axom/primal/operators/compute_moments.hpp
kennyweiss: I think @jcs15c was keying off of a paper that he used for reference when implementing some of these routines, which uses `nonrational` vs. `rational` (see below). I've also seen "polynomial Bezier curves" vs. "rational Bezier curves" when distinguishing between the two.
I think we should avoid `irrational` since that implies "cannot be expressed as a ratio" (e.g. `sqrt(2)`) rather than "is not currently expressed as a ratio".  

See: https://github.com/LLNL/axom/blob/8129e1ea791aa4de51e9fde52941ebfed53bd1d5/src/axom/primal/geometry/BezierCurve.hpp#L53-L57

jcs15c: I'm not sure why, but it seems like the convention is to use the phrase "Nonrational Bezier Curves" over "Irrational" or even "Polynomial Bezier Curves." Or at least, the both the Farin 83 reference reference I cite above and the NURBS book both prefer the phrase "nonrational."

src/axom/primal/operators/compute_moments.hpp
jcs15c: I've added this to issue #897, as I agree that it should work in the more general case. That being said, the current methods for nonrational polygons will be significantly faster than invoking quadrature, and won't require the MFEM include. I think a thorough implementation would proceed largely as before, and only compute areas/centroids of sectors using quadrature when necessary, rather than just broadly to the entire polygon.

src/axom/primal/operators/compute_moments.hpp
rhornung67: OK. It seems a bit odd to me, but if it's common usage.

rhornung67 (approved):

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -158,14 +228,45 @@ class BezierCurve
...
+  /// Make nonrational by shrinking array of weights
+  void makeNonrational() { m_weights.resize(0); }
+
+  /// Use array size as flag for rationality
+  bool isRational() const { return (m_weights.size() != 0); }
+
+  /// Check that the weights used are valid
+  bool isValid() const
+  {
+    if(isRational()) return true;
```
agcapps, line 251: It looks like this method checks that a *rational* Bezier curve is valid.  Perhaps rename the method to `isValidRational` or similar.

Further, the test at line 251 seems to be inverted. Non-rational curves would then go on to pull values out of `m_weights`.  In addition to inverting the test at line 251, consider adding a check before the loop at line 254 that `m_weights` is the same length as `m_controlPoints`.

jcs15c: Thanks for catching that, that check is definitely backwards as it is now.

src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -158,14 +228,45 @@ class BezierCurve
   /// Returns the order of the Bezier Curve
   int getOrder() const { return static_cast<int>(m_controlPoints.size()) - 1; }
 
-  /// Clears the list of control points
+  /// Make trivially rational. If already rational, do nothing
+  void makeRational()
```
agcapps, line 231: `makeRational`, `makeNonrational`, `isRational`, `getOrder`, `isValid`, `clear`, `getWeight`, and `operator[]` are part of the public interface and I don't see tests for them (perhaps I overlooked them).  Please add tests for these methods.

agcapps:
Looks very good.  I requested some tests and a rename of `isValid`.  Also, I think I caught a reversed test.  After these are addressed, I'll be happy to approve.

src/axom/primal/geometry/BezierCurve.hpp
jcs15c: Thank you for the suggestions, I've added a test that verifies `makeRational`, `makeNonrational`, `isRational`, and `getWeight`. The other methods you've mentioned are unchanged between rational and nonrational Bezier curves, and already have tests in `primal_bezier_curve.cpp`.

As for `isValidRational`, I think the best course would be to remove it from the public interface altogether, and make the method private. As of now, it is only run within `SLIC_ASSERT` statements, as I don't think users should be allowed to form invalid rational Bezier curves in the first place, as this would cause a whole slew of issues with the evaluation functions.

agcapps (approved):
Thank you.  Looks good.  I think it was a good idea to make `isValidRational` private.

PR #902, opened 2022-08-16
Adds templating of `evaluate_integral` for arbitrary dimensions

jcs15c:
# Summary

- This PR is a feature/refactoring that allows existing `evaluate_integral` methods to evaluate integrals over 3D Bezier curves, when appropriate.
- Additionally, it simplifies the user interface.
-- Overloaded function calls to `evaluate_line_integral` (which would previously take a lambda function that returns either a `double` or `Vector` object from `Point` input) are split between `evaluate_scalar_line_integral` and `evaluate_vector_line_integral`, which assume that the lambda function returns the appropriate type.
-- Overloads for `evaluate_line_integral( axom::Array<BezierCurve> )` are removed, as the functionality is identical to that of `evaluate_line_integral( CurvedPolygon )`, but with more conceptual clarity as to why the method adds together the result from each component.
- An additional test for line integrals along a 3D curve are added.


src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -40,23 +40,23 @@ namespace axom
...
- * \param [in] cs the array of Bezier curve objects
- * \param [in] integrand the lambda function representing the integrand. 
- * Must accept a 2D point as input, and return a double (scalar field) or 
- * 2D vector object (vector field).
- * \param [in] npts the number of Gaussian quadrature nodes for each component
+ * 
+ * \param [in] cpoly the CurvedPolygon object
+ * \param [in] scalar_integrand the lambda function representing the integrand. 
+ * Must accept a Point<T, NDIM> as input and return a double
+ * \param [in] npts the number of quadrature points to evaluate the line integral
```
kennyweiss, line 53: Clarification: `npts` is the number of quadrature points per `BezierCurve` in the `CurvedPolygon`, right?

src/axom/primal/tests/primal_integral.cpp
```diff
@@ -231,20 +249,64 @@ TEST(primal_integral, evaluate_line_integral_vector)
...
+    return std::sin(x[0] * x[1] * x[2]);
+  };
+
+  auto vector_field = [](Point3D x) -> Vector3D {
+    return Vector3D({4 * x[1] * x[1], 8 * x[0] * x[1], 1.0});
+  };
+
+  // Test line integral on scalar domain
+  EXPECT_NEAR(evaluate_scalar_line_integral(spatial_arc, const_integrand, npts),
+              4.09193268998,
```
kennyweiss, line 300: How were these 3D integral values obtained?

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -154,8 +180,8 @@ double evaluate_line_integral(const primal::BezierCurve<double, 2>& c,
  * \param [in] npts_P the number of quadrature points to evaluate the antiderivative
  * \return the value of the integral
  */
-template <class Lambda>
-double evaluate_area_integral(const axom::Array<primal::BezierCurve<double, 2>>& cs,
+template <class Lambda, typename T, int NDIMS>
+double evaluate_area_integral(const axom::Array<primal::BezierCurve<T, NDIMS>>& cs,
```
kennyweiss, line 184: Some quick gut checks on this change -- 
* Does `evaluate_area_integral` work on `BezierCurve`s in 3D?
* Are there restrictions? e.g. that the polygon needs to be planar?
* Should this take a `CurvedPolygon` instead of an array of `BezierCurve`s?
* Should the function below be generalized similarly?

```cpp
template <class Lambda> 
double evaluate_area_integral(const primal::CurvedPolygon<double, 2> cpoly,
                              Lambda&& integrand,
                              int npts_Q,
                              int npts_P = 0)
```

kennyweiss (approved):
Thanks @jcs15c -- I like how you were able to generalize all these integration functions from 2D to 3D with almost no changes to the code!
Thanks also for adding new tests.

src/axom/primal/tests/primal_integral.cpp
jcs15c: I computed these originally in desmos, but I did also compare them against a more robust scientific programming resource (mathematica) to verify them. I'll add a comment to clarify that.

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -66,32 +66,60 @@ double evaluate_line_integral(const axom::Array<primal::BezierCurve<double, 2>>&
...
   return total_integral;
 }
 
 /*!
- * \brief Evaluate a line integral along the boundary of a CurvedPolygon object.
+ * \brief Evaluate a line integral on a single Bezier curve on a scalar field
  *
- * See above definition for details.
+ * \param [in] c the Bezier curve object
+ * \param [in] integrand the lambda function representing the integrand.
```
bmhan12, line 86:
```suggestion
 * \param [in] scalar_integrand the lambda function representing the integrand. 
```

src/axom/primal/operators/evaluate_integral.hpp
jcs15c: Ah I see, I thought I had removed the `evaluate_area_integral( Array<BezierCurve> )` overload, but it slipped past somehow. There is another function down the line that actually does evaluate on `CurvedPolygon` objects how you describe here. 

In general, this method should only be applied to 2D curved polygons, since that's what David's paper describes. It should be able to work on 3D, planar curved polygons as well, but the formulas would have to be a little different for math reasons, namely that the ultimate set of quadrature nodes are axis aligned, and it isn't clear where that axis should be in 3D.

jcs15c: I had a comment here, but I think I deleted it somehow. Here was the gist of it:

Oh yes, this function shouldn't be there at all anymore. I thought I had removed the overload for thing function for `Array<BezierCurve>` input, but it somehow slipped past me. There's a function later on that operates directly on curved polygons that has the behavior you're describing, being that it only operates in 2 dimensions.

In general, these algorithms should only work in 2D. They would work on 3D planar curved polygons in principle, but the formulas get a little murkier. David's paper puts the quadrature nodes on the shape aligned with the axes, and for 3D planar shapes, it isn't clear where that axis would be right away. Unlike the other methods that generalized immediately, this one would require a more sophisticated change of variables to get the nodes in the right place.

bmhan12 (approved):

PR #921, opened 2022-08-30
Refactor/gunney/closest point comm loop

gunney1:
# Simplify communication loop in DistributedClosestPoint class

- This PR is a refactoring
  - Removes the "skip" messages in the communication loop.  By all-gathering the query mesh bounding boxes, we can compute the number of messages to receive, removing the need for "skip" messages.
  - Improves communication performance, because the all-gather added is likely much faster than the p2p skip messages removed.
  - Simplifies the communication logic, making it easier to maintain.

When reviewing the changes, it is probably easier to read the new `computeClosestPoints` rather than look at the diff of the changes to that method.

This PR resolves issue https://github.com/LLNL/axom/issues/908#issue-1345389433


src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -559,6 +559,11 @@ class DistributedClosestPointImpl
       PointType lower(&recvbuf[i * 2 * DIM]);
       PointType upper(&recvbuf[i * 2 * DIM + DIM]);
       all_aabbs.emplace_back(BoxType(lower, upper));
+      // Correct the auto-correction in the BoundingBox constructor.
```
gunney1, line 562: This is needed because the `BoundingBox` constructor switches the `lower` and `upper` points to make the box valid.  We are actually interested in communicating that the box is invalid--it indicates that the domain has no bounding box (domain underloading).

kennyweiss: Suggestion:

Could you please modify the `BoundingBox` constructor to take a third optional parameter which controls whether the bounds get fixed? The default should be `true` to preserve the current behavior. 

E.g. something like this:
```cpp
AXOM_HOST_DEVICE
  BoundingBox(const PointType& lowerPt, 
              const PointType& upperPt,
              bool shouldFixBounds = true)
    : m_min(lowerPt)
    , m_max(upperPt)
  {
    if(shouldFixBounds) { this->checkAndFixBounds(); }
  }
```

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -680,207 +680,197 @@ class DistributedClosestPointImpl
...
-        m_isVerbose,
-        fmt::format("=======  Starting round {}/{} =======", round, m_nranks));
-
-      if(xferNodePtr && m_nranks > 1)
+      /*
+        Send local query mesh to next rank with close-enough object
+        partition, if any.  Increase remainingRecvs, because this data
+        will come back.
+      */
+      int firstRecipForMyQuery = next_recipient(*xferNodes[m_rank]);
```
kennyweiss, line 731: Do the queries have to go in ring order? Or is this just a convenient way to order them?

I was wondering if it would make sense to sort the recipients by increasing (max) distance that way we're likely to get a pretty good approximation in the first few rounds, and might be able to skip (or at least return early) from the later rounds.

(max distance == the max possible distance between the bounding box of a query mesh and the bounding box of an object mesh)

kennyweiss (approved):
Thanks @gunney1 !

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -680,207 +680,197 @@ class DistributedClosestPointImpl
...
-                            fmt::format("round_{}_r{}_begin.json", 0, m_rank));
+        const auto& otherQueryBb = allQueryBbs[r];
+        double sqDistance =
+          axom::primal::squared_distance(otherQueryBb, myObjectBb);
+        if(sqDistance <= m_sqDistanceThreshold)
+        {
+          ++remainingRecvs;
+        }
       }
     }
```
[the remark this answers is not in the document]
gunney1: Pre-computing the next recipient is a bit inefficient.  We'd have to assess N `BoundingBox` distances and store up to N ranks indices.  We have to communicate that O(N) data as well, if we don't want to re-compute it on each rank the query `Node` lands on.  Moreover, a rank never needs to know all the other ranks a query `Node` is close to.  It just needs the next rank.  The loop in `next_recipient` `m_nranks` long, but we expect it to break very early.

src/axom/quest/DistributedClosestPoint.hpp
gunney1: The ring order is a convenient way to order them and has the advantage that the sender and receiver, being close in rank-space, are likely to be close on the physical network.  Sorting by increasing distance will scramble this order.  I don't think the early-stop will be more performant because sorting requires an O(N) or O(N logN) algorithm, which is no faster than not trying to stop early.

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -680,207 +680,197 @@ class DistributedClosestPointImpl
...
-        }
-      }  // Locally process *xferNodePtr
-
-      SLIC_INFO_IF(
-        m_isVerbose,
-        fmt::format("=======  End of round {}/{} =======", round, m_nranks));
-    }  // round loop
+      break;
+    }
+  }
```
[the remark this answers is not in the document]
gunney1: Thank you.

src/axom/quest/DistributedClosestPoint.hpp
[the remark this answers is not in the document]
kennyweiss: I was thinking that if you sort it as a preprocess, you could avoid work later on by not sending data to ranks that provably cannot have the closest point.

I suppose there is a tradeoff between how many ranks (`N`) we expect to run on  and how many ranks (on average) we expect to send data to.  For degenerate configurations, we might need to send data to all ranks (e.g. a point near the center of a sphere), but for many configs, we can probably prune most of the ranks during this preprocessing step.

In any event, this can be refactored/optimized in the future, as necessary/desired.

src/axom/quest/DistributedClosestPoint.hpp
gunney1: Done.

src/axom/quest/DistributedClosestPoint.hpp
kennyweiss: True, but I was thinking that `O(N log N)` _on rank_ (where N is hundreds or even thousands) is probably faster than an `~O(1)` algorithm that has to be run `N` times, and each of those might have additional MPI comms.

I think the difference might be that the algorithm I was suggesting would account for the max distance among the object bounding boxes (and possibly also the local approximation to the distance) and not just the distance threshold.

src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -680,207 +680,197 @@ class DistributedClosestPointImpl
...
+    for(int i = 1; i < m_nranks; ++i)
     {
-      check_send_requests(isendRequests, true);
+      int maybeNextRecip = (m_rank + i) % m_nranks;
+      if(maybeNextRecip == homeRank)
+      {
+        return maybeNextRecip;
+      }
+      double sqDistance =
+        primal::squared_distance(bb, m_objectPartitionBbs[maybeNextRecip]);
```
kennyweiss, line 864: ... as a follow up to discussions in other threads -- this check for `maybeNextRecip` only accounts for the distance between the query mesh's bounding box and the object meshes bounding box (with respect to the distance threshold).

Perhaps it can also account for the approximate max distances that we've already computed.
E.g. we might already know that the max distance from a query point on this rank to its closest point is at most `_conservative_max`. Then, if the max distance from `bb` to `m_objectPartitionBbs[maybeNextRecip]` is more than this, we can skip rank `maybeNextRecip`.

Since this is an optimization, and it can get messy with the bookkeeping related to the number of ranks to check, we might want to think about this for future PRs.

src/axom/quest/DistributedClosestPoint.hpp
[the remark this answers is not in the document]
gunney1: For the default infinite threshold, we don't gain anything by preprocessing, because there's no knowledge to be gained.  The `next_recipient` loop would return at once.
If the number allowed by the threshold << number of ranks, there's lots go gain.  However, If I read your suggestions right, it amounts to calling `next_recipient` ahead of time instead of as needed.  I think we'd end up doing the same amount of computations.
My main concern is clarity, since this refactoring is supposed to improve that.  Do you think the preprocessing would help with clarity?  I see that it only replaces calling `next_recipient` with a look-up, but maybe I'm missing something.

src/axom/quest/DistributedClosestPoint.hpp
gunney1: I get the gist of this, and I agree.

src/axom/quest/DistributedClosestPoint.hpp
kennyweiss: I think a main use-case for the distance threshold is that the end-user/application only needs to process items that are withing the given threshold and can safely ignore farther distances.

gunney1: So you're saying that with threshold and without are two distinct use cases.  If so, then sorting would buy us **a lot** in the latter case.

agcapps (approved):
Thank you, @gunney1 .  Ease of maintenance is always almost always worth the refactor.  Much appreciated.

PR #929, opened 2022-09-08
Fix Group::copyToConduitNode so that it handles lists correctly

### LINKED ISSUE #928: Group::copyToConduitNode doesn't work with lists
This line:

`      Node& g = n["groups"].fetch(group->getName());`

fails when `this` Group holds its item in the list format.  Here `group` is an unnamed child node and `getName` returns the empty string.  When using lists, we need to use `append` to create new unnamed child Nodes in a conduit list, rather than `fetch` which is used to created named child Nodes.

The same fix is needed when handling unnamed Views in a list.

nselliott:
# Summary

- This PR is a bug fix
  - Modifies `Group::copyToConduitNode` to handle Groups using the list format.j
  - Fixes #928 

This method had the assumption that all child items have a name.  This fix adds handling of unnamed items held in a list, and creates lists of unnamed child Nodes in the conduit output.

### DIFF
--- a/src/axom/sidre/core/Group.cpp
+++ b/src/axom/sidre/core/Group.cpp
@@ -1340,7 +1340,7 @@ void Group::copyToConduitNode(Node& n) const
   while(indexIsValid(vidx))
   {
     const View* view = getView(vidx);
-    if (isUsingMap())
+    if(isUsingMap())
     {
       Node& v = n["views"].fetch(view->getName());
       view->copyToConduitNode(v);
@@ -1357,7 +1357,7 @@ void Group::copyToConduitNode(Node& n) const
   while(indexIsValid(gidx))
   {
     const Group* group = getGroup(gidx);
-    if (isUsingMap())
+    if(isUsingMap())
     {
       Node& g = n["groups"].fetch(group->getName());
       group->copyToConduitNode(g);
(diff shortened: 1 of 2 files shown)


nselliott:
I need to add a suitable test before merging.

gunney1 (approved):
Thanks @nselliott

rhornung67 (approved):
Thanks for the quick fix @nselliott

kennyweiss (approved):
Thanks @nselliott -- looks great!

Approved assuming there are sufficient tests (as you commented above)

src/axom/sidre/core/Group.cpp
```diff
@@ -1340,19 +1340,33 @@ void Group::copyToConduitNode(Node& n) const
...
-
+    if(isUsingMap())
+    {
+      Node& v = n["views"].fetch(view->getName());
+      view->copyToConduitNode(v);
+    }
+    else
+    {
+      Node& v = n["views"].append();
+      view->copyToConduitNode(v);
```
kennyweiss, line 1351: Quick clarification -- how does this work?

My assumption is that it's appending an empty node `v` to `n["views"] and then `copyToConduitNode(v)` fills in `v`?
Is that right?

kennyweiss: Looks like it:
https://github.com/LLNL/conduit/blob/de3a873068ba971405cfc003d3286673f2cedfcd/src/libs/conduit/conduit_node.hpp#L3971-L3975

nselliott: Yes, your description is correct.

cyrush (approved):

agcapps (approved):
Thanks for the fix.

Issue #907, opened 2022-08-19
`detail::linear_winding_number` returns negation of true value


jcs15c:
The winding number for a linear segment can be computed by only a single arccosine calculation. Currently, the non-interface implementation file `in_curved_polygon_impl.hpp` contains a method to compute this value. Currently, it returns the negation of the true winding number. This is currently not causing any issues in execution, as this method is only called by other hidden implementation methods, which account for this incorrect behavior independently. This makes this "misnaming" of the function low priority. However, if this method is ever taken out of the implementation file, as such a method could be useful in practical applications, it should be corrected.

PR #934, opened 2022-09-13
Fixes misidentification for containment within `CurvedPolygon`s with nonconvex edges

jcs15c:
This PR is a bugfix that fixes the behavior of `winding_number` with respect to nonconvex Bezier curve objects. We now require that the control polygon of a Bezier curve be convex before we can ensure our winding number formulations will be correct, which is enforced by repeated recursion until such a condition is met. Previously, there existed cases where a query point could be external to the control polygon, but still contained in the shape formed by closing the Bezier curve. A new test has been added in `primal_winding_number.cpp` to detect this incorrect behavior, and verify the correct behavior in this update.

Additionally, this PR addresses issue #907 by fixing the sign in the output of `linear_winding_number`, and adjusting signs in other areas of the winding number functionality to ensure consistency. This change is nonfunctional, but adds significant mathematical clarity.

### DIFF
--- a/src/axom/primal/operators/detail/in_curved_polygon_impl.hpp
+++ b/src/axom/primal/operators/detail/in_curved_polygon_impl.hpp
@@ -188,23 +188,23 @@ double adaptive_winding_number(const Point<T, 2>& q,
 
   Polygon<T, 2> controlPolygon(c.getControlPoints());
 
-  // If q is outside the control polygon, for an open Bezier curve, the winding
-  //  number for the shape connected at the endpoints with straight lines is zero.
-  //  We then subtract the contribution of this line segment.
-  if(!in_polygon(q, controlPolygon, true, false, EPS))
-    return 0.0 - linear_winding_number(q, c[0], c[ord], edge_tol);
-
   // Check if our new curve is convex.
   //  If so, all subcurves will be convex as well
   if(!isConvexControlPolygon)
     isConvexControlPolygon = is_convex(controlPolygon, EPS);
+  else // Formulas for winding number only work if shape is convex
+  {
+    // If q is outside the control polygon, for an open Bezier curve, the winding
+    //  number for the shape connected at the endpoints with straight lines is zero.
+    //  We then subtract the contribution of this line segment.
+    if(!in_polygon(q, controlPolygon, true, false, EPS))
+      return 0.0 - linear_winding_number(q, c[0], c[ord], edge_tol);
 
-  // If the query point is at either endpoint, use direct formula
-  //   Can only use formula if the control polygon is convex
-  if(isConvexControlPolygon &&
-     ((squared_distance(q, c[0]) <= edge_tol * edge_tol) ||
-      (squared_distance(q, c[ord]) <= edge_tol * edge_tol)))
-    return convex_endpoint_winding_number(q, c, edge_tol, EPS);
+    // If the query point is at either endpoint, use direct formula
+    if((squared_distance(q, c[0]) <= edge_tol * edge_tol) ||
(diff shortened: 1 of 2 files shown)


src/axom/primal/operators/detail/in_curved_polygon_impl.hpp
```diff
@@ -183,28 +183,27 @@ double adaptive_winding_number(const Point<T, 2>& q,
...
-  // If q is outside the control polygon, for an open Bezier curve, the winding
-  //  number for the shape connected at the endpoints with straight lines is zero.
-  //  We then subtract the contribution of this line segment.
-  if(!in_polygon(q, controlPolygon, true, false, EPS))
-    return 0.0 - linear_winding_number(q, c[0], c[ord], edge_tol);
-
   // Check if our new curve is convex.
   //  If so, all subcurves will be convex as well
   if(!isConvexControlPolygon)
     isConvexControlPolygon = is_convex(controlPolygon, EPS);
```
kennyweiss, line 193: Minor:  (To more clearly distinguish between the `if` and `else` clauses) 

```suggestion
  if(!isConvexControlPolygon)
  {
    isConvexControlPolygon = is_convex(controlPolygon, EPS);
  }
```

src/axom/primal/operators/detail/in_curved_polygon_impl.hpp
```diff
@@ -183,28 +183,27 @@ double adaptive_winding_number(const Point<T, 2>& q,
...
-     ((squared_distance(q, c[0]) <= edge_tol * edge_tol) ||
-      (squared_distance(q, c[ord]) <= edge_tol * edge_tol)))
-    return convex_endpoint_winding_number(q, c, edge_tol, EPS);
+  else  // Formulas for winding number only work if shape is convex
+  {
+    // If q is outside the control polygon, for an open Bezier curve, the winding
+    //  number for the shape connected at the endpoints with straight lines is zero.
+    //  We then subtract the contribution of this line segment.
+    if(!in_polygon(q, controlPolygon, true, false, EPS))
+      return 0.0 - linear_winding_number(q, c[ord], c[0], edge_tol);
```
kennyweiss, line 200: Gut check, based on pattern matching with the changes above:

Should this be negated? E.g.:
```suggestion
    if(!in_polygon(q, controlPolygon, true, false, EPS))
      return linear_winding_number(q, c[ord], c[0], edge_tol);
```

I think the current way is correct since you reversed the order of `c[0]` and `c[ord]` compared to the one above (and thus reversed the orientation).

kennyweiss (approved):
Thanks @jcs15c !

src/axom/primal/operators/detail/in_curved_polygon_impl.hpp
jcs15c: I am very confident that the updated version is correct, as the minus sign in `linear_winding_number` was removed, which would flip the sign, but reversing the order of `c[0]` and `c[ord]` flips the sign again, keeping the behavior exactly as it was before (plus, it shouldn't pass the tests if we get rid of that minus sign as in your suggestion).

As it is, `0.0 - linear_winding_number(q, c[ord], c[0], edge_tol)` returns exactly the same number as `linear_winding_number(q, c[0], c[ord], edge_tol)`, but conceptually I believe writing it this second way is more clear. We should have that `winding_number + closure_winding_number = 0` for these exterior points, and the closure in this case is a line that connects the terminating endpoint of the Bezier curve to it's initial endpoint.

kennyweiss: Thanks -- I agree.

agcapps (approved):
Thank you for the clarifying fix.

PR #937, opened 2022-10-03
Remove obsolete comments about importing conduit lists

nselliott:
# Summary

- This PR is a comment-only change
- It does the following (modify list as needed):
  - Removes some obsolete comments about Group's importConduitTree not working for conduit lists. This was true for the original implementation, but later changes made it work. The original comments did not get removed at the time.

PR #938, opened 2022-10-03
Fix inconsistency in importConduitTreeExternal, and add test coverage

nselliott:
# Summary

- This PR is a fix for a code inconsistency, possibly a bug, and adds test coverage
  - Modifies Group::importConduitTreeExternal to make it handle lists with unnamed members the same way as importConduitTree. The two methods should construct their Sidre hierarchical representation of the Conduit input tree in the same way. I noticed this while preparing #937
  - Adds test coverage for importing Conduit lists using importConduitTreeExternal; nothing for this was existent.


rhornung67 (approved):
Thanks @nselliott

cyrush (approved):

agcapps (approved):
Thank you, Noah.

PR #945, opened 2022-10-20
Slam: Add polymorphic policies, other changes

publixsubfan:
# Summary

Adds a policy tag type for selecting virtual/concrete interfaces when instantiating Slam objects.
  - Two tag types are provided -- `slam::policies::VirtualInterface` and `slam::policies::ConcreteInterface`
  - Supported Slam objects inherit from a "selector" type instantiated with the provided tag, which aliases to the correct base class (this is to support selecting between a CRTP class and an abstract base class):
    - `OrderedSet/RangeSet`: `slam::Set` if virtual, `ConcreteInterface` if concrete
    -  `Map/SubMap/BivariateMap`: `slam::MapBase` if virtual, `ConcreteInterface` if concrete
    -  `ProductSet/RangeSet`: `slam::BivariateSet` if virtual, CRTP `ConcreteBivariateSet` if concrete
  - Adds conversions for Sets and BivariateSets between corresponding concrete and virtual interface types

Also:
- Access the underlying sets of a BivariateSet directly from the derived classes. For `ProductSet`, these are stored as member variables, and for `RelationSet`, these are accessed from the underlying relation.
- `PositionSet` and `RangeSet` now are just template aliases of `GenericRangeSet`
- Uses underlying Relation set types as default set template arguments in `RelationSet`
- Adds set template arguments to `DynamicVariableRelation` to correspond to `StaticRelation`
- Adds an allocator ID parameter to `BivariateMap` constructor


kennyweiss (approved):
Thanks @kanye-quest -- this is really nice -- it seems to have minimal changes for this new functionality.

Could you please compare the performance before/after these changes to ensure that they match those of the paper?
* The performance suite is in `multimat/examples/calculate.cpp` 
* The dataset is from https://github.com/lanl/MultiMatTest/blob/master/volfrac.dat.tgz
* The results are in Figures 5, 6 and 7 of https://zenodo.org/record/5559021 

Specifically, I'm concerned that some of the changes introduce virtual function calls that can no longer be inlined by the compiler. Since the "kernels" are so small, these add a significant overhead compared to the "native" performance.

If you'd prefer to deal with performance issues in a follow-up PR, that would be fine with me.

agcapps:
General comment: this good.  It's also in very strong need of documentation showing why you would choose one or the other, and how to use one or the other.

src/axom/slam/BivariateSet.hpp
```diff
@@ -183,18 +171,24 @@ class BivariateSet
...
+  virtual const FirstSetType* getFirstSet() const
+  {
+    return policies::EmptySetTraits<FirstSetType>::emptySet();
+  }
   /** \brief Returns pointer to the second set.   */
-  const SecondSetType* getSecondSet() const { return m_set2; }
+  virtual const SecondSetType* getSecondSet() const
+  {
+    return policies::EmptySetTraits<SecondSetType>::emptySet();
+  }
```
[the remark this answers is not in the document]
kanye-quest: I think I was trying to fix a bug related to [this workaround](https://github.com/LLNL/axom/blob/develop/src/axom/slam/BivariateMap.hpp#L136-L148). Let me see if there's a way to fix this without resorting to virtual functions -- I think we just need to ensure that the RelationSet updates the underlying sets when the relation is modified.

src/axom/slam/ProductSet.hpp
```diff
@@ -34,52 +36,102 @@ namespace slam
...
+  {
+    using Type = SetType;
+    RowSet(PositionType secondSetSize) : m_data(secondSetSize)
+    {
+      //fill in the row data now for getElements(i) function,
+      //since every row is the same, a call to getElements() returns the same set.
+      //
+      // HACK -- this should actually be returning a PositionSet since it always
+      //         goes from 0 to secondSetSize()
+      // This requires a change to the return type of BivariateSet::getElements()
```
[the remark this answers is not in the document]
kanye-quest: This workaround/comment was in `ProductSet` before; I think the issue is that we need a common type for the `BivariateSet::getElements()` virtual function declaration that makes sense for both `ProductSet` and `RelationSet`.

In the concrete case, this shouldn't be an issue; we can return a vector-indirection type `OrderedSet` for `RelationSet::getElements()`, and a `RangeSet` for `ProductSet::getElements()`.

But I'm not sure of the best way to deal with this in the virtual case:
- The common base class between the two is just `Set`, so we'd need to return a `const Set&` since `Set` is an abstract base class.
- We'd ideally like to refer to something concrete, so nothing temporarily-constructed. That might be fine for `RelationSet` since we only need to store one "row set", but for `RelationSet`, we need to store each "slice" of the `Relation` array for each index.

src/axom/slam/RelationSet.hpp
```diff
@@ -29,11 +30,15 @@ namespace slam
...
+          typename SetType2 = typename Relation::ToSetType,
+          typename InterfaceType = policies::VirtualInterface>
 class RelationSet final
-  : public OrderedSet<typename Relation::SetPosition, typename Relation::SetElement>,
-    public BivariateSet<SetType1, SetType2>
+  : public policies::BivariateSetInterface<
+      InterfaceType,
+      SetType1,
+      SetType2,
+      RelationSet<Relation, SetType1, SetType2, InterfaceType>>
```
agcapps, line 41: OK, maybe I'm just blind, but this looks like recursive templatizing.  RelationSet inherits from BivariateSetInterface whose final template is the same template class RelationSet.  Am I reading this right?  If so, where's the base case, so this doesn't go on recursing?

kanye-quest: Good question! I'm using the [curiously-recursive template pattern](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern) here. It's usually used to achieve similar results to polymorphism without the expense of virtual functions calls.

This will probably go away once I address this https://github.com/LLNL/axom/pull/945#discussion_r1000933708, though.

src/axom/slam/BivariateSet.hpp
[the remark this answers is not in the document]
kanye-quest: Reverted in [6813e0e](https://github.com/LLNL/axom/pull/945/commits/6813e0e95edc1662f5a4c4c3de30d18a36476ebf).

src/axom/slam/policies/SetInterfacePolicies.hpp
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2017-2022, Lawrence Livermore National Security, LLC and
```
white238, line 1: These copyrights will need to be updated. Sorry for letting this slip past the end of the year =/

src/axom/slam/policies/InterfacePolicies.hpp
```diff
@@ -0,0 +1,25 @@
...
+#ifndef SLAM_InterfacePolicies_HPP
+#define SLAM_InterfacePolicies_HPP
+
+namespace axom
+{
+namespace slam
+{
+namespace policies
+{
+struct VirtualInterface
```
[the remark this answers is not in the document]
white238: Bump.

src/axom/slam/policies/BivariateSetInterfacePolicies.hpp
```diff
@@ -0,0 +1,125 @@
...
+    return getSize<FirstSetType>(m_set1);
+  }
+  /** \brief Size of the second set.   */
+  inline PositionType secondSetSize() const
+  {
+    return getSize<SecondSetType>(m_set2);
+  }
+
+  /** \brief Returns pointer to the first set.   */
+  const FirstSetType* getFirstSet() const { return m_set1; }
```
white238, line 60: Does `getFirstSet()` follow the pattern in SLAM or should it be `firstSet()`?

src/axom/slam/SubMap.hpp
```diff
@@ -51,10 +51,11 @@ namespace slam
  */
 
 template <typename SuperMapType,
-          typename SubsetType  //= slam::RangeSet<SetPosition, SetElement>
-          >
-class SubMap : public MapBase<typename SubsetType::PositionType>,
-               public SuperMapType::StridePolicyType
+          typename SubsetType,  //= slam::RangeSet<SetPosition, SetElement>
+          typename InterfacePolicy = policies::VirtualInterface>
```
white238, line 55: Should `InterfacePolicy` be the first template parameter to be consistent with the new code below?

For example:

```
template <typename InterfaceTag, typename PosType, typename ElemType>
struct SetInterfaceSelector
```

Feel free to say no.

src/axom/slam/RelationSet.hpp
```diff
@@ -220,26 +231,16 @@ class RelationSet final
       s2 < m_relation->size(s1);
   }
 
-  void verifyPosition(PositionType sPos) const override
-  {  //override function from RangeSet, overloading to avoid warning in compiler
-    verifyPositionImpl(sPos);
-  }
-
-  void verifyPositionImpl(PositionType AXOM_DEBUG_PARAM(sPos)) const
+  void verifyPosition(PositionType AXOM_DEBUG_PARAM(sPos)) const
```
white238, line 234: 👍

src/axom/slam/RangeSet.hpp
```diff
@@ -19,51 +19,6 @@ namespace axom
 {
 namespace slam
 {
```
white238, line 21: Love the deleted code.

src/axom/slam/ProductSet.hpp
```diff
@@ -34,24 +36,81 @@ namespace slam
...
+    using Type = PositionSet<PositionType, ElementType>;
+
+    RowSet(PositionType) { }
+
+    Type get(PositionType secondSetSize) const { return Type(secondSetSize); }
+  };
+
+public:
+  using ConcreteSet = ProductSet<SetType1, SetType2, policies::ConcreteInterface>;
+  using PolymorphicSet =
```
white238, line 96: Can you explain the difference between Polymorphic vs Virtual here? Is there a reason its not VirtualSet?

src/axom/slam/Map.hpp
```diff
@@ -515,8 +522,8 @@ bool Map<T, S, IndPol, StrPol>::isValid(bool verboseOutput) const
   return bValid;
 }
 
-template <typename T, typename S, typename IndPol, typename StrPol>
-void Map<T, S, IndPol, StrPol>::print() const
+template <typename T, typename S, typename IndPol, typename StrPol, typename IfacePol>
+void Map<T, S, IndPol, StrPol, IfacePol>::print() const
```
white238, line 526: I can't say I am a fan of this level of shortening the name. It makes it hard to read. I would much prefer `InterfacePolicy`. I realize this is just following what is there but it also doesn't match the rest of the coding style.

src/axom/slam/policies/SetInterfacePolicies.hpp
kennyweiss: Hopefully, it should be sufficient to run this script: 
https://github.com/LLNL/axom/blob/develop/scripts/update_copyright_date.sh

src/axom/slam/policies/BivariateSetInterfacePolicies.hpp
kennyweiss: I'd be ok with either way (perhaps happier with the latter).
If we make a change, please be careful about consistency throughout slam.

src/axom/slam/SubMap.hpp
kennyweiss: The template order prioritizes the things that are most likely to change first and the ones that are most likely to use default values last. That allows users to specify fewer template parameters.

I'm not 100% sure about policy vs. tag. Tag makes me think about adding a type as a function parameter rather than a template parameter to the class.

src/axom/slam/RelationSet.hpp
```diff
@@ -154,14 +171,11 @@ class RelationSet final
...
-  {
-    return (*m_relation)[s1];
-  }
+  SubsetType getElements(PositionType s1) const { return (*m_relation)[s1]; }
 
-  ElementType at(PositionType pos) const override
+  ElementType at(PositionType pos) const
   {
-    verifyPositionImpl(pos);
+    verifyPosition(pos);
```
kennyweiss, line 178: @publixsubfan -- please verify that there's no performance cost associated with this change.

I intentionally added `verifyPositionImpl()` as a non-virtual function to allow the compiler to inline the call (especially in release builds, where this gets optimized away). `verifyPosition` was a virtual function and the compiler was not able to optimize it.

Perhaps this concern goes away with the new implementation. In which case, I'm strongly in favor of the cleaner code!

src/axom/slam/Map.hpp
kennyweiss: Thanks @white238. I agree that the brevity here doesn't match other parts of slam and would be in favor of more overall consistency throughout slam. It's ok with me to save it for a future PR though.

src/axom/slam/RelationSet.hpp
publixsubfan: I think this is non-virtual when used with a concrete policy. To address the virtual case, I think we can directly invoke the function nonvirtually, i.e. `this->RelationSet::verifyPosition(pos)`

src/axom/slam/Map.hpp
publixsubfan: Agreed in full -- I'll change it to be consistent with the rest of Slam in a future PR.

src/axom/slam/ProductSet.hpp
publixsubfan: Polymorphic sounds fancier :P

Point taken, I can change it to be consistent with Virtual[thing]Policy.

src/axom/slam/policies/BivariateSetInterfacePolicies.hpp
publixsubfan: These match the names in the virtual `BivariateSet` class -- I'll change them to `firstSet()`/`secondSet()` in a future PR

src/axom/slam/SubMap.hpp
publixsubfan: Renamed to `InterfacePolicy` for consistency.

white238 (approved):

PR #951, opened 2022-11-07
Update BlueOS compilers

white238:
This includes several bug fixes on the Spack side and moves us forward to the newest Spack develop.  This includes the new multi-build system support.

This upgrades our compiler versions and simplifies a lot of the compiler specs.  We should re-evaluate after the compiler survey to make sure we didn't miss any specs. But this does bring us closer to what our customers are using.

### DIFF
--- a/.gitlab/build_lassen.yml
+++ b/.gitlab/build_lassen.yml
@@ -39,80 +39,79 @@
 
 ####
 # PR Build jobs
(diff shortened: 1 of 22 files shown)


rhornung67 (approved):

kennyweiss (approved):
Thanks @white238 !

white238:
This is ready for review

PR #958, opened 2022-11-15
Document call-back functors for traversing BVH

gunney1:
Add documentations for public interfaces and mention that call-back functors should use memory available in their execution environment.

# Summary

- This PR is a source-code documentation update

  - Add documentation about the call-backs used in BVH tree traversals.
  - State that the call-backs should only access memory available in the execution space of the traversal code.

The BVH code is factored to several files.  We've mentioned a similar requirement for data to initialize the BVH class.  But we lack similar documentations on the call-back functions used in the tree traverser class.

### DIFF
--- a/src/axom/spin/policy/LinearBVH.hpp
+++ b/src/axom/spin/policy/LinearBVH.hpp
@@ -36,6 +36,24 @@ namespace policy
 {
 namespace lbvh = internal::linear_bvh;
 
+/*
+ * \brief Interface for traversing a BVH tree.
+ *
+ * \brief Traverse a tree to perform user-specified actions at the
+ * leaves, while limiting the search to branches satisfying a
+ * user-provided predicate.
+ *
+ * To a initiate traversals, use traverse_tree.  It requires
+ * -# an action functor to call at the leaves,
+ * -# a predicate functor to decide whether to descend a branch and
+ * -# some data to pass to the functors
+ *
+ * Both functors should only access memory that's available to the
+ * execution space.  For example, GPU execution should access only
+ * device and unified memory.
+ *
+ * \see internal::linear_bvh::bvh_traverse
+ */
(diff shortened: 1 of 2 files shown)


rhornung67 (approved):

src/axom/spin/policy/LinearBVH.hpp
```diff
@@ -36,6 +36,24 @@ namespace policy
...
 namespace lbvh = internal::linear_bvh;
 
+/*
+ * \brief Interface for traversing a BVH tree.
+ *
+ * \brief Traverse a tree to perform user-specified actions at the
+ * leaves, while limiting the search to branches satisfying a
+ * user-provided predicate.
+ *
+ * To a initiate traversals, use traverse_tree.  It requires
```
kennyweiss, line 46: Typo and suggested highlight:
```suggestion
 * To initiate traversals, use \a traverse_tree.  It requires
```

src/axom/spin/policy/LinearBVH.hpp
```diff
@@ -71,6 +89,11 @@ class LinearBVHTraverser
                        traversePref);
   }
 
+  /*
+   * Functors lf and predicate should access only memory compatible
```
kennyweiss, line 93: Suggestion:
```suggestion
   * Functors \a lf and \a predicate should access only memory compatible
```

kennyweiss (approved):
Thanks @gunney1 !

src/axom/spin/policy/LinearBVH.hpp
gunney1: done.  thanks.

src/axom/spin/policy/LinearBVH.hpp
gunney1: Done.  Thanks.

Issue #959, opened 2022-11-16
Conflict between existing fmt target and Axom's exported fmt target


andrewcorrigan:
Is it possible to qualify the `fmt` target exported by Axom, for example, as `axom::fmt`? 

I built and installed `axom` to a folder `third_party/llnl/install`.  Now, I'm trying to include Axom in my cmake project via:
```
find_dependency(Axom REQUIRED NO_DEFAULT_PATH PATHS "third_party/llnl/install/lib/cmake")
```

This triggers an error due to a conflict with an existing `fmt` target in my project:
```
CMake Error at third_party/fmt/CMakeLists.txt:237 (add_library):
  add_library cannot create target "fmt" because an imported target with the
  same name already exists.
```

As a quick workaround, I did a search-replace on `fmt` to `axom::fmt` in `third_party/llnl/install/lib/cmake/axom-targets.cmake` and that bypassed the error.   Therefore, I am wondering if there is a way to automatically qualify the exported `fmt` target as `axom::fmt`.

kennyweiss:
Thanks for reporting this @andrewcorrigan. 

We addressed the source code aspect of this in #802 and #805, but it looks like we missed the CMake export.

PR #960, opened 2022-11-16
sets EXPORT_NAME of the fmt target to axom::fmt to avoid conflicts in…

andrewcorrigan:
sets `EXPORT_NAME` of the `fmt` target to `axom::fmt` to avoid conflicts in external projects consuming both `axom` and `fmt`

Resolves https://github.com/LLNL/axom/issues/959


kennyweiss (approved):
Thanks for fixing this @andrewcorrigan !

white238 (approved):

PR #963, opened 2022-11-18
Local PR for #960

white238:
sets EXPORT_NAME of the fmt target to axom::fmt to avoid conflicts in external projects consuming both axom and fmt

resolves https://github.com/LLNL/axom/issues/959


white238:
I guarded the rest of the targets.

kennyweiss (approved):
Thanks @white238 

Should we also add `blt_export_tpl_targets(EXPORT axom-targets NAMESPACE axom)` somewhere in axom? Or is that covered in some other way?

white238:
> Thanks @white238
> 
> Should we also add `blt_export_tpl_targets(EXPORT axom-targets NAMESPACE axom)` somewhere in axom? Or is that covered in some other way?

It is there just fully expanded =)  We had the functionality before that function existed. I'll copy it at some point.

PR #966, opened 2022-11-24
Use libstdc++ 8 with XL compilers on blueos

publixsubfan:
Updates the XL compilers to use the GCC 8.3.1 toolchain on blueos.

Also:
- modifies the Axom Spack package to pass through C++ compiler flags to the host compiler when using NVCC
- fixes an issue with the MFEM Spack package when C++ compiler flags are present in the compiler spec

### DIFF
--- a/host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.2_cuda.cmake
+++ b/host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.2_cuda.cmake
@@ -11,11 +11,11 @@
 #------------------------------------------------------------------------------
 if(DEFINED ENV{SPACK_CC})
 
-  set(CMAKE_C_COMPILER "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_09_10_19_47/spack/lib/spack/env/clang/clang" CACHE PATH "")
+  set(CMAKE_C_COMPILER "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_12_01_14_30_45/spack/lib/spack/env/clang/clang" CACHE PATH "")
 
-  set(CMAKE_CXX_COMPILER "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_09_10_19_47/spack/lib/spack/env/clang/clang++" CACHE PATH "")
+  set(CMAKE_CXX_COMPILER "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_12_01_14_30_45/spack/lib/spack/env/clang/clang++" CACHE PATH "")
 
-  set(CMAKE_Fortran_COMPILER "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_09_10_19_47/spack/lib/spack/env/clang/gfortran" CACHE PATH "")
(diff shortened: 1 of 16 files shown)


rhornung67 (approved):

host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.2_cuda.cmake
```diff
@@ -109,7 +109,7 @@ set(MFEM_DIR "${TPL_ROOT}/mfem-4.4.0" CACHE PATH "")
 
 set(HDF5_DIR "${TPL_ROOT}/hdf5-1.8.22" CACHE PATH "")
 
-set(LUA_DIR "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_11_09_10_19_47/clang-10.0.1.1/lua-5.4.4" CACHE PATH "")
+set(LUA_DIR "/usr/WS1/axom/libs/blueos_3_ppc64le_ib_p9/2022_12_01_14_30_45/clang-10.0.1.1/lua-5.4.4" CACHE PATH "")
```
kennyweiss, line 112: @white238 -- is there a difference between `clang-10.0.1.1` and `clang-10.0.1.2` ?
I suppose this is another concretization error from spack?

We should probably switch over to the new `require` clause in our package specs -- https://spack.readthedocs.io/en/latest/build_settings.html#package-requirements

src/axom/core/ArrayBase.hpp
```diff
@@ -19,8 +19,8 @@
 #include <iostream>  // for std::cerr and std::ostream
 #include <numeric>   // for std::accumulate
 #if defined(__GLIBCXX__) && !defined(_GLIBCXX_USE_CXX11_ABI)
-  // Workaround for unimplemented C++11 type traits in libstdc++ versions <5
-  #include <tr1/type_traits>
+  #error \
+    "GNU libstdc++ versions less than 5 do not fully support C++11 and are unsupported by Axom."
```
kennyweiss, line 23: 👍 !

kennyweiss (approved):
Thanks @publixsubfan !

host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.2_cuda.cmake
white238: We just need to update uberenv and this will go away. They changed the default from "fresh" to "reuse" a while back and I fixed it in uberenv but apparently never updated our uberenv.  Basically Spack decided that the other clang's lua was good enough and used it instead of rebuilding from scratch.

white238: There is no difference between the two compilers. One is the CUDA build and one is not. It is so we can have two different directories and not conflict.

white238 (approved):

scripts/spack/packages/axom/package.py
```diff
@@ -242,6 +242,10 @@ def initconfig_hardware_entries(self):
             # CUDA_FLAGS
             cudaflags = "-restrict --expt-extended-lambda "
 
+            # Pass through any cxxflags to the host compiler via nvcc's Xcompiler flag
+            host_cxx_flags = spec.compiler_flags['cxxflags']
+            cudaflags += ' '.join(['-Xcompiler=%s ' % flag for flag in host_cxx_flags])
```
white238, line 247: This should be pushed up to Spack's CachedCMakePackage... I'll remember to do it next time I update the Axom package.

Issue #750, opened 2021-12-15
Clarify documentation of the path argument when creating Groups and Views held by ListCollection Groups


nselliott:
From discussion in pull request #749 

The path argument is ignored in methods such as `createView` for Groups that use ListCollection, as the child items are not accessed by name but only by iterator.  The recommended but not required usage is to pass an empty string.  This is mentioned in the comments in the ListCollection header, but not really anywhere in easily-accessed documentation.

nselliott:
As I work on this documentation, I notice that some things need to be cleaned up in the API for creating Groups that will be in a ListCollection.  There is `createUnnamedGroup` which works fine, but there is also `createGroup` which doesn't work correctly when given an empty string, and also does some unexpected things when given a path string. The string should be ignored, but it is not.

PR #969, opened 2022-12-02
Some clarification of usage of name/path strings in sidre Group's list format

nselliott:
# Summary

- This PR is a small feature modification, possibly a bug fix, with some documentation and testing added
  - Modifies the behavior when a path string is given to a Group or View held in the list format.  Specifically it assigns only the final name in the path and does not do the creation of intermediate Groups in the path.  e.g. if "foo/bar/baz" is given, the item will be named "baz" and "foo/bar" will be ignored.  This changes from previous behavior where the intermediate groups were created but were empty and inaccessible.
  - Adds documentation explaining the above and also adding more clarity to what happens in general with name/path arguments in the list format, as mentioned in #750 
  - Adds a test to check the results when various strings, or no strings, are used in the creation of Groups and Views in the list format.

The recommended usage is to not use names for the items stored in the list format.  Any name assigned is not useful for accessing items, unlike how the names are used in the default MapCollection format.  Nonetheless we allow names to be used without breaking anything.  The case where a path string is used gave some behavior that I think is unwanted when it creates intermediate groups that can't be used, so that is the substantive change here.


src/axom/sidre/core/Group.hpp
```diff
@@ -1627,6 +1643,10 @@ class Group
    * the method returns is the last entry in the path, either the string
    * following the last "/" in the input (if there is one) or the entire
    * input path string if it contains no "/".
+   *
+   * When this Group uses the list format ONLY:  this Group will always
+   * be returned, the path argement will be modified to the last entry in
```
rhornung67, line 1648:
```suggestion
   * be returned, the path argument will be modified to the last entry in
```

rhornung67 (approved):
Just one typo. Thanks @nselliott

agcapps (approved):
Thanks, @nselliott .

src/axom/sidre/core/Group.hpp
```diff
@@ -109,9 +109,25 @@ class MapCollection;
...
+ * It is recommended but not required that the items held in the list format
+ * be created without names.  String names may be assigned to these items,
+ * but the names will not be useful for accessing them from their parent
+ * Group, and none of the methods that access child items by name or path will
+ * return a valid pointer.  Unnamed Groups to be held in the list format
+ * should be created using the method createUnnamedGroup, while unnamed
+ * Views should be created by passing an empty string as the path argument
+ * to any of the createView methods.
+ *
+ * For the list format ONLY, if the method creating a Group or View is
```
white238, line 126: Is this documentation in the Sphinx docs as well?

src/axom/sidre/core/Group.hpp
```diff
@@ -109,9 +109,25 @@ class MapCollection;
...
+ * passed a string with path syntax (such as "foo/bar/baz"), only the
+ * name after the final delimiter is assigned to the new item, and the
```
white238, line 128: Is there a reason for this behavior because, as a user, this is not what I would have expected to happen. I would have expected either all intermediate groups to be created (like `mkdir -p foo/bar/baz`) or error out that you cannot create a group with the special character `/` in it.

nselliott: It's worthy of discussion what should happen here; my choice might not be the best.  Creating the intermediate groups is not good, as it will orphan the object at the end of the path, since the path cannot be used from a list-format Group.  Thinking of it now, it may be better to have create give a warning and return a null pointer, as there's no good use case to provide a path string.  A null pointer rather than an error would be consistent with the design choice across Sidre to return a null that the user can check for failed object creation rather than crashing a run.

src/axom/sidre/core/Group.hpp
nselliott: I need to add that.

src/axom/sidre/core/Group.cpp
```diff
@@ -2367,12 +2367,20 @@ Group* Group::walkPath(std::string& path, bool create_groups_in_path)
       }
       else if(create_groups_in_path)
       {
-        group_ptr = group_ptr->createGroup(*iter);
-
-        if(group_ptr == nullptr)
+        if(m_is_list)
```
kennyweiss, line 2370: Design question: Should the modification for lists happen here, inside of `walkPath`? or should the caller be aware of list paths, and possibly send in an empty string for `path`?

kennyweiss (changes requested):
Thanks @nselliott!

Looks like there are some outstanding design questions to discuss.

src/axom/sidre/core/Group.hpp
nselliott: Once the suggested change is implemented, created_group and found_group will be both be nullptr.  Test assertions can be added here to verify that.

src/axom/sidre/tests/sidre_group.cpp
```diff
@@ -574,6 +574,68 @@ TEST(sidre_group, child_lists)
...
+  Group* named_group = list_test->createGroup("named");
+
+  // With a path, the leading names are ignored and the final name is used.
+  Group* path_group_a = list_test->createGroup("testing/path");
+  Group* path_group_b = list_test->createGroup("testing/longer/path");
+
+  ASSERT_TRUE(blank_group == nullptr);
+  ASSERT_TRUE(named_group->getName() == "named");
+  ASSERT_TRUE(path_group_a->getName() == "path");
+  ASSERT_TRUE(path_group_b->getName() == "path");
```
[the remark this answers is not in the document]
nselliott: With a list, the names are meaningless to the parent group, so I don't see that there needs to be a restriction here.  In fact the recommended usage is that the child groups all have the same name (empty string).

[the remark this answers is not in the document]
kennyweiss: Ok, now I understand -- the group/view will have the provided name, but that name will not be registered with the parent and the group/view will not be accessible via that name.

Is this the desired behavior? Or would it be better to just ignore the name? I suppose it doesn't matter that much.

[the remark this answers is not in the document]
nselliott: When I first created `ListCollection` I considered forcing everything to be nameless, but given that there's an existing API to provide the name, and users may want to assign a name for their own reasons, I let it be allowed but not useful for the parent group.  The substantive API issue in this PR is what to do when a path string is provided, as that is not really addressed in the current `develop` implementation.  A path doesn't make sense in the context of a list, so the consensus here to disallow that with a `nullptr` return will clarify that part.

src/axom/sidre/core/Group.hpp
nselliott: The sphinx docs need a dedicated section on the list format, so I'll create an issue for that.

src/axom/sidre/core/Group.hpp
```diff
@@ -109,9 +109,23 @@ class MapCollection;
...
+ * be created without names.  String names may be assigned to these items,
+ * but the names will not be useful for accessing them from their parent
+ * Group, and none of the methods that access child items by name or path will
+ * return a valid pointer.  Additionally, strings with path syntax, (such
+ * as ("foo/bar/baz") will be considered invalid when creating items to be
+ * held in the list format, and the object creation methods will return
+ * a null pointer if such a string is provided.  Unnamed Groups to be held
+ * in the list format should be created using the method createUnnamedGroup,
+ * whlie unnamed Views should be created by passing an empty string as the
+ * path argument to any of the createView methods.
```
kennyweiss, line 127: 👍

src/axom/sidre/core/Group.hpp
```diff
@@ -109,9 +109,23 @@ class MapCollection;
...
+ * It is recommended but not required that the items held in the list format
+ * be created without names.  String names may be assigned to these items,
+ * but the names will not be useful for accessing them from their parent
+ * Group, and none of the methods that access child items by name or path will
+ * return a valid pointer.  Additionally, strings with path syntax, (such
+ * as ("foo/bar/baz") will be considered invalid when creating items to be
+ * held in the list format, and the object creation methods will return
+ * a null pointer if such a string is provided.  Unnamed Groups to be held
+ * in the list format should be created using the method createUnnamedGroup,
+ * whlie unnamed Views should be created by passing an empty string as the
```
kennyweiss, line 126: Minor(typo):
```suggestion
 * while unnamed Views should be created by passing an empty string as the
```

src/axom/sidre/tests/sidre_group.cpp
```diff
@@ -574,6 +574,76 @@ TEST(sidre_group, child_lists)
...
+  Group* path_group_b = list_test->createGroup("testing/longer/path");
+  Group* found_group_a = list_test->getGroup("testing/path");
+  Group* found_group_b = list_test->getGroup("testing/longer/path");
+
+  ASSERT_TRUE(blank_group == nullptr);
+  ASSERT_TRUE(named_group->getName() == "named");
+  ASSERT_TRUE(path_group_a == nullptr);
+  ASSERT_TRUE(path_group_b == nullptr);
+  ASSERT_TRUE(found_group_a == nullptr);
+  ASSERT_TRUE(found_group_b == nullptr);
```
kennyweiss, line 619: Minor:

Would these tests be better/clearer as `ASSERT_EQ`?

kennyweiss (approved):
Thanks @nselliott

Issue #982, opened 2022-12-23
Install Axom with RAJA on Windows


tlorentz77:
Hi there,

Could you please update the port file of Axom in vcpkg to integrate RAJA lib? 

It seems that RAJA can be build on Windows with capability of OPENMP now.

kennyweiss:
@tlorentz77 -- our `vcpkg` can now build `RAJA` with `OpenMP` on Windows through `uberenv`

tlorentz77:
Thank you @kennyweiss and sorry for the late feedback.  I successfully build axom based on vcpkg now, but failed to search the axom using `find_package`. It reports that there is no `camp-config.cmake` file in the direcotry. The reason is that the `PATHS` provided in file `axom-config.cmake.in` (as shown in the follow) is not compatible with `vcpkg`. In `vcpkg`, the config files of packages like `xxxx-config.cmake` are usually in directory like `xxx/share/camp`. So I think the file `axom-config.cmake.in` should be update as well if we want to call axom on Windows platform.

```
  if(AXOM_USE_CAMP)
    set(AXOM_CAMP_DIR      "@CAMP_DIR@")
    if(NOT CAMP_DIR) 
      set(CAMP_DIR ${AXOM_CAMP_DIR})
    endif()
    find_dependency(camp REQUIRED NO_DEFAULT_PATH PATHS "${CAMP_DIR}/lib/cmake/camp")
  endif()
```

Could you please provide an `axom-project-template` just like `RAJA-project-template` (https://github.com/LLNL/RAJA-project-template)? It would be helpful for users who are not familiar with axom. Thank you very much!

kennyweiss:
Thanks @tlorentz77 -- We currently build and test `axom+raja+umpire+openmp+mfem+conduit` using `vcpkg` (through `uberenv`), but have not yet taken the next step of importing that axom into another project.

We have similar examples to the `RAJA-project-template` in `axom/src/examples/using-with-{blt,cmake,make}` -- https://github.com/LLNL/axom/tree/develop/src/examples -- and describe them in our Quickstart guide -- https://axom.readthedocs.io/en/develop/docs/sphinx/quickstart_guide/zero_to_axom.html

Our unix-based CI exercises the `using-with-` examples, but our Windows CI does not. I'll create a new issue to fix this and to add it to our Windows CI.

tlorentz77:
Thanks very much. @kennyweiss

tlorentz77:
[...]

PR #983, opened 2022-12-28
Fixes build and tests when using RAJA without Umpire

kennyweiss:
# Summary

- This PR is a bugfix
- It allows the code to configure/build/test when using RAJA but not Umpire
- Discovered when working on #982

### DIFF
--- a/src/axom/core/CMakeLists.txt
+++ b/src/axom/core/CMakeLists.txt
@@ -94,6 +94,7 @@ set(core_sources
 #------------------------------------------------------------------------------
 
 set( core_depends fmt )
+blt_list_append( TO core_depends ELEMENTS camp IF CAMP_FOUND )
 blt_list_append( TO core_depends ELEMENTS umpire IF UMPIRE_FOUND )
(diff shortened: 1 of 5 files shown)


src/axom/core/CMakeLists.txt
```diff
@@ -94,6 +94,7 @@ set(core_sources
 #------------------------------------------------------------------------------
 
 set( core_depends fmt )
+blt_list_append( TO core_depends ELEMENTS camp IF CAMP_FOUND )
```
white238, line 97: Do we use camp w/o umpire/raja? Not that it super matters...

white238 (approved):

src/axom/core/CMakeLists.txt
kennyweiss: Not as far as I know. It looks like raja's import was not bringing in the camp include (and I suspect that Umpire's import was). Not sure if this has been fixed between `raja@2022.03` (our current version) and `raja@develop`

agcapps (approved):
Thanks, @kennyweiss .

PR #984, opened 2022-12-28
Multimat: more benchmark cleanup

publixsubfan:
# Summary

- Deletes unused methods from multimat/calculate.cpp (follow-up to #946)
(Note: for some reason, the overall diff is messed up, with both added and deleted lines -- underlying commits should show only the deletions)


kennyweiss (approved):
Thanks for the description @publixsubfan, and for the cleanup.

Going through the commits, I see that each commit consists of only deletions. The diffs might be cleaner if this branch were rebased against develop instead of merging develop in at the end.

Did we lose any comparisons to different implementations by consolidating the code in the previous PR? Or is it still testing the same things, but using a cleaned up implementation?  I think it's the latter, but wanted to double check.

publixsubfan:
> Did we lose any comparisons to different implementations by consolidating the code in the previous PR? Or is it still testing the same things, but using a cleaned up implementation? I think it's the latter, but wanted to double check.

Yeah, we switched to calling the cleaned-up implementations in a prior PR. This PR is just deleting the now-unused test implementations.

src/axom/multimat/examples/calculate.cpp
```diff
@@ -1986,24 +1571,39 @@ void calculate_pressure_mat_dom_full_mm_direct(MultiMat& mm)
...
 
-  DenseField Pressurefrac(mm, mm.getDense2dFieldSet(DataLayout::MAT_DOM));
+  auto Densityfrac = FieldGetterType::get(mm, "Densityfrac");
+  auto Volfrac = FieldGetterType::get(mm, "Volfrac");
+  auto Temperaturefrac = FieldGetterType::get(mm, "Tempfrac");
+
+  MultiMat::Field1D<double>& nmatconsts = mm.get1dField<double>("nmatconsts");
+
+  // Store result in a dense field
+  {
```
white238, line 1598: is this scope really needed?

src/axom/multimat/examples/calculate.cpp
```diff
@@ -2577,17 +2033,107 @@ void average_density_over_nbr_cell_dom_full_mm_direct(MultiMat& mm,
...
+
+      auto Volfrac_row = Volfrac(ic);
+      auto AvgMatDensity_row = MatDensity_average(ic);
+
+      for(int m = 0; m < nmats; ++m)
+      {
+        if(Volfrac_row(m) > 0.0)  //this check is not needed in sparse layout.
+        {
+          double den = 0.;
+          int nnm = 0;  // number of nbrs with this material
```
white238, line 2088: number of numbers? number of neighbors?

white238 (approved):

white238:
Yay for deleted code!

PR #1002, opened 2023-01-19
Bugfix for RAJA's optional RAJA::roctx dependency

kennyweiss:
# Summary

- This PR is a workaround for a potential bug in blt
- When we specify that `core` depends on `RAJA` in `blt_add_library`, it should transitively depend on RAJA's dependencies. However, the `RAJA::roctx` dependency was not being properly propagated. The latter is present when RAJA is configured with `RAJA_ENABLE_ROCTX`.
- Since `core` is a CMake  `OBJECT` library, we're internally setting it up as a `blt_register_library`
- As a temporary workaround, this PR explicitly adds `RAJA::roctx` when available/needed


rhornung67 (approved):

kennyweiss:
I described the underlying issue on https://github.com/LLNL/blt/issues/618

agcapps (approved):

PR #1005, opened 2023-01-20
Check if the polyhedron is still valid before being clipped by next plane

liyangrock:
# Summary

- This PR is a bugfix.
- It does the following:
  - Check if the polyhedron is still valid before being clipped by next plane. If the polyhedron is invalid, then clear the polyhedron and return. Otherwise, it may triger arithmetic exception due to the bounding box of polyhedron is invalid.


src/axom/primal/operators/detail/clip_impl.hpp
```diff
@@ -425,6 +425,13 @@ AXOM_HOST_DEVICE Polyhedron<T, NDIMS> clipPolyhedron(
   //Clip Polyhedron by each plane
   for(PlaneType plane : planes)
   {
+    // Check the obtained Polyhedron is still valid
+    if(!poly.isValid())
+    {
+      poly.clear();
+      break;
+    }
+
```
kennyweiss, line 434: Thank you for the contribution @liyangrock.

Can you please provide a reproducer for the problem? 
E.g. a polyhedron and set of clipping planes that generate the behavior that you were seeing?
It seems to me that the error could be elsewhere, e.g. in the `intersect()` algorithm, and we'd rather fix the problem at the source.  

Specifically, the solution in this PR would skip the current `plane`, but the state of the polyhedron, `poly`, has nothing to do with `plane`.

liyangrock: Thank you for your reply @kennyweiss. I failed to reproduce the problem. I create a reproducer, but it calculate `r` as `-inf` rather than throw an `arithmetic exception` in my FEM code, as shown in the following figs. I was fully confused because the values of all input variables in the reproducer are exactly the same as in my FEM code.




I believe this is a bug due to the function `intersect_plane_bbox` inappropriately deal with the infinite number (i.e., vector `e`, which results from invalid variable `poly` in function `clipPolyhedron` and consequently invalid bounding box `bb`).


### REPLY
> Specifically, the solution in this PR would skip the current `plane`, but the state of the polyhedron, `poly`, has nothing to do with `plane`.

I still think this PR can resolve the bug. The solution in this PR skips the current `plane` only if the `poly` is invalid. Note that it's meaningless to clip an invalid Polyhedron. Thus, clear and then return an empty Polyhedron `poly` is reasonable for the function `clipPolyhedron`. 


### REPRODUCER
```
#include "axom/core.hpp"
#include "axom/fmt.hpp"

#include "axom/primal/geometry/BoundingBox.hpp"
#include "axom/primal/geometry/Tetrahedron.hpp"
#include "axom/primal/geometry/Polyhedron.hpp"
#include "axom/primal/geometry/Polygon.hpp"
#include "axom/primal/geometry/Triangle.hpp"
#include "axom/primal/operators/compute_bounding_box.hpp"
#include "axom/primal/operators/clip.hpp"
#include "axom/primal/operators/intersect.hpp"

#include <iostream>

int main()
{
  constexpr int in3D = 3;
  using BoundingBoxType = axom::primal::BoundingBox<double, in3D>;
  using PointType = axom::primal::Point<double, in3D>;
  using VectorType = axom::primal::Vector<double, in3D>;
  using TetType = axom::primal::Tetrahedron<double, in3D>;
  using PolyhedronType = axom::primal::Polyhedron<double, in3D>;
  using PolygonType = axom::primal::Polygon<double, in3D>;
  using TriangleType = axom::primal::Triangle<double, in3D>;

  // Using fmt library exported by axom
  std::cout << axom::fmt::format(
                 "Example of using and installed version of Axom {}",
                 axom::getVersion())
            << std::endl
            << std::endl;

  // Uses installed axom library
  axom::about();

  PointType A = PointType::make_point(0, 1, 0);
  PointType B = PointType::make_point(-1, 0, 0);
  PointType C = PointType::make_point(0, -1, 0);
  PointType D = PointType::make_point(-0.25, 0, 0.25);

  PointType E = PointType::make_point(0.74899999999999922, 0, 1);
  PointType F = PointType::make_point(1.7489999999999992, 0, 0);
  PointType G = PointType::make_point(0.74899999999999922, -1, 0);
  PointType H = PointType::make_point(0.99899999999999922, 0, 0.25);

  TetType tet1 = TetType(A, B, C, D);
  TetType tet2 = TetType(E, F, G, H);

  PolyhedronType overlap;

  overlap = axom::primal::clip(tet1, tet2);

  tet1.print(std::cout) << std::endl;
  tet2.print(std::cout) << std::endl;

  overlap.print(std::cout) << std::endl;

  return 0;
}

```

kennyweiss: Thank you for the detailed explanation and simplified reproducer @liyangrock !  

Based on the above, the main problem was with `primal::intersect_plane_bbox()`. Specifically, we need to check that the provided bounding box is valid before using it's geometry.

I fixed this and some related things in https://github.com/LLNL/axom/pull/1013
I also added your test case to our unit tests and verified that `primal::clip()` returns an empty polyhedron with 0 volume.

When you get a chance, please let us know if this fixes the problem that you were seeing.

kennyweiss: Great!  Thanks for letting us know @liyangrock.

PR #1011, opened 2023-01-24
Use flux instead of srun on tioga

bmhan12:
This PR:
-  Includes fix from PR #1010
-  Uses flux instead of srun for MPI tests on tioga (srun hangs MPI tests)

### DIFF
--- a/host-configs/tioga-toss_4_x86_64_ib_cray-clang@14.0.0_hip.cmake
+++ b/host-configs/tioga-toss_4_x86_64_ib_cray-clang@14.0.0_hip.cmake
@@ -41,9 +41,9 @@ set(MPI_CXX_COMPILER "/usr/tce/packages/cray-mpich-tce/cray-mpich-8.1.16-rocmcc-
 
 set(MPI_Fortran_COMPILER "/usr/tce/packages/cray-mpich-tce/cray-mpich-8.1.16-rocmcc-5.2.3/bin/mpif90" CACHE PATH "")
 
-set(MPIEXEC_EXECUTABLE "/usr/bin/srun" CACHE PATH "")
+set(MPIEXEC_EXECUTABLE "/usr/bin/flux" CACHE PATH "")
 
-set(MPIEXEC_NUMPROC_FLAG "-n" CACHE STRING "")
+set(MPIEXEC_NUMPROC_FLAG "mini;run;-n" CACHE STRING "")
 
 set(ENABLE_MPI ON CACHE BOOL "")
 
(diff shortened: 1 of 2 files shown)


host-configs/tioga-toss_4_x86_64_ib_cray-clang@14.0.0_hip.cmake
```diff
@@ -41,9 +41,9 @@ set(MPI_CXX_COMPILER "/usr/tce/packages/cray-mpich-tce/cray-mpich-8.1.16-rocmcc-
 
 set(MPI_Fortran_COMPILER "/usr/tce/packages/cray-mpich-tce/cray-mpich-8.1.16-rocmcc-5.2.3/bin/mpif90" CACHE PATH "")
 
-set(MPIEXEC_EXECUTABLE "/usr/bin/srun" CACHE PATH "")
+set(MPIEXEC_EXECUTABLE "/usr/bin/flux" CACHE PATH "")
 
-set(MPIEXEC_NUMPROC_FLAG "-n" CACHE STRING "")
+set(MPIEXEC_NUMPROC_FLAG "mini;run;-n" CACHE STRING "")
```
kennyweiss, line 46: This will get lost the next time we run `build_tpls.py`

kennyweiss (approved):
Thanks @bmhan12 -- hopefully this will get us past the current CI problem, and we can find a better solution in the meantime.

host-configs/tioga-toss_4_x86_64_ib_cray-clang@14.0.0_hip.cmake
bmhan12: > This will get lost the next time we run build_tpls.py

I will do a follow-up PR with the spack fix for this host-config.
I think there are quite a few PR's that are ready to go, but just need this tioga fix for gitlab CI to pass.

kennyweiss: Thanks @bmhan12 -- I agree that we should get something working as soon as possible.

Longer term, I think the `MPIEXEC_NUMPROC_FLAG` should allow for spaces (i.e. `flux mini run`) and we should make changes to blt/axom to support that. 

@white238 pointed out a place yesterday where `MPIEXEC_NUMPROC_FLAG` likely needs to be quoted.
e.g.
https://github.com/LLNL/blt/blob/8c229991e65e7a9603c621b47cb3ba158bb7468c/cmake/BLTMacros.cmake#L904-L914

PR #1020, opened 2023-02-03
Polyhedron::getFaces() public function

### LINKED ISSUE #1014: Request public access to the function "getFaces" of a polyhedron object
Sometimes one may want to obtain the polygon faces of a polyhedron object. However, there is no direct way to get the connectivity of polygon faces because `getFaces` is a private function in current version of `Polyhedron.hpp` (line.391-484).

Is it possible to make funtion `getFaces` public or provide an alternative method to achieve this?

bmhan12:
This PR:
- Makes getFaces() function public
- Adds an example and fixes some documentation formatting
- Closes #1014


kennyweiss (changes requested):
Thanks for working on this @bmhan12.

Could you please add an example/unit test that shows how to iterate on the faces returned by `Polyhedron::getFaces()`?

Ideally, the example would have some faces that have different numbers of vertices.
E,g, a polygonal cone/pyramid and/or prism.

I recently added a pyramid with a polygonal base to the `polygonal_cone` test in this file recently, so that might be a good starting point.

src/axom/primal/tests/primal_polyhedron.cpp
```diff
@@ -53,6 +54,28 @@ TEST(primal_polyhedron, polyhedron_unit_cube)
   poly.addNeighbors(poly[7], {4, 6, 3});
 
   EXPECT_EQ(1, poly.volume());
+
+  // Example usage of experimental getFaces() function
+  // (input parameters and/or output may change in the future)
+  int faces[24];
```
kennyweiss, line 60: It might be worth checking that the provided sizes are appropriately sized.

src/axom/primal/geometry/Polyhedron.hpp
```diff
@@ -388,14 +392,18 @@ class Polyhedron
...
-   * \brief Finds the faces of the Polyhedron, assuming the vertex neighbors
-   *        are in counter-clockwise ordering.
+   * \brief Helper function to find the faces of the Polyhedron, assuming the
+   *        vertex neighbors are in counter-clockwise ordering.
+   *
+   * \param [out] faces is the vertex indices for faces
+   * \param [out] face_offset is the offset for each face
+   * \param [out] face_size is the number of vertices for each face
+   * \param [out] face_count is the number of faces
    *
```
kennyweiss, line 404: Please add a note that it is the user's responsibility to ensure that the provided buffers are sufficiently large and that the user must know these sizes ahead of time.  

`primal::Polyhedron` does not yet have a way to let the user know how big these arrays need to be.

kennyweiss: Technically, since we assume this is a convex polyhedron, Euler's formula tells us that $|v|-|e|+|f|=2$.
Since we know the number of vertices, $|v|$, and can calculate the number of edges, $|e|$, we can add a function to return the number of faces, $|f|$.

We should probably do that in a follow up PR.

rhornung67 (approved):
I approve assuming @kennyweiss comments are addressed. Please make an issue to add a function that returns the number of faces so we don't forget.

kennyweiss (approved):
Thanks @bmhan12 !

PR #1025, opened 2023-02-09
Bugfix for importing caliper-enabled mfem in cuda configs

kennyweiss:
# Summary

- This PR is a bugfix for importing a `caliper`-enabled `mfem` in a `cuda` configuration
    - The former depends on `cupti` which was not being exported, leading to compiler errors
    - As a hack, we patch mfem to depend on `CUDA::cupti`
- It also fixes some floating point comparisons in axom's unit tests

### DIFF
--- a/src/cmake/thirdparty/SetupAxomThirdParty.cmake
+++ b/src/cmake/thirdparty/SetupAxomThirdParty.cmake
@@ -196,6 +196,23 @@ else()
     message(STATUS "MFEM support is OFF")
 endif()
 
+# caliper-enabled mfem in cuda configs depends on cupti, which is not properly exported
+if(TARGET mfem AND ENABLE_CUDA)
+    # check if mfem depends on caliper; if so, patch it with cupti
+    get_target_property(_mfem_libs mfem INTERFACE_LINK_LIBRARIES)
+    if("${_mfem_libs}" MATCHES "caliper")
+        if(NOT TARGET CUDA::toolkit)
+            find_package(CUDAToolkit REQUIRED)
+        endif()
+
+        if(TARGET CUDA::toolkit)
+            blt_patch_target(NAME mfem DEPENDS_ON CUDA::toolkit)
+        endif()
+        if(TARGET CUDA::cupti)
+            blt_patch_target(NAME mfem DEPENDS_ON CUDA::cupti)
+        endif()
+    endif()
+endif()
 
 #------------------------------------------------------------------------------
 # Shroud - Generates C/Fortran/Python bindings
--- a/src/cmake/thirdparty/SetupAxomThirdParty.cmake
+++ b/src/cmake/thirdparty/SetupAxomThirdParty.cmake
@@ -198,16 +198,13 @@ endif()
 
 # caliper-enabled mfem in cuda configs depends on cupti, which is not properly exported
 if(TARGET mfem AND ENABLE_CUDA)
-    # check if mfem depends on caliper; if so, patch it with cupti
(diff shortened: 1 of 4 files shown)


src/cmake/thirdparty/SetupAxomThirdParty.cmake
```diff
@@ -196,6 +196,23 @@ else()
...
+        if(NOT TARGET CUDA::toolkit)
+            find_package(CUDAToolkit REQUIRED)
+        endif()
+
+        if(TARGET CUDA::toolkit)
+            blt_patch_target(NAME mfem DEPENDS_ON CUDA::toolkit)
+        endif()
+        if(TARGET CUDA::cupti)
+            blt_patch_target(NAME mfem DEPENDS_ON CUDA::cupti)
+        endif()
```
kennyweiss, line 213: For reference, here's what these two targets provide in my configuration:
```
-- [CUDA::toolkit property] 'CUDA::toolkit' is a CMake target
-- [CUDA::toolkit property] INTERFACE_INCLUDE_DIRECTORIES: /usr/tce/packages/cuda/cuda-11.1.1/nvidia/include
-- [CUDA::toolkit property] INTERFACE_LINK_DIRECTORIES: /usr/tce/packages/cuda/cuda-11.1.1/nvidia/lib64
-- [CUDA::toolkit property] INTERFACE_SYSTEM_INCLUDE_DIRECTORIES: /usr/tce/packages/cuda/cuda-11.1.1/nvidia/include
-- [CUDA::toolkit property] NAME: CUDA::toolkit
-- [CUDA::toolkit property] TYPE: INTERFACE_LIBRARY

-- [CUDA::cupti property] 'CUDA::cupti' is a CMake target
-- [CUDA::cupti property] INTERFACE_INCLUDE_DIRECTORIES: /usr/tce/packages/cuda/cuda-11.1.1/nvidia/include
-- [CUDA::cupti property] INTERFACE_LINK_LIBRARIES: /usr/tce/packages/cuda/cuda-11.1.1/nvidia/lib64/libcupti.so
-- [CUDA::cupti property] INTERFACE_SYSTEM_INCLUDE_DIRECTORIES: /usr/tce/packages/cuda/cuda-11.1.1/nvidia/include
-- [CUDA::cupti property] NAME: CUDA::cupti
-- [CUDA::cupti property] TYPE: INTERFACE_LIBRARY
```

white238 (approved):

agcapps (approved):
Thank you, @kennyweiss .

src/cmake/thirdparty/SetupAxomThirdParty.cmake
cyrush: We should also try to work to get logic like this into into caliper's cmake -- it doesn't handle forwarding deps (CUDA, HIP, gotcha, etc). That's ok for final link for an app that is already handling these deps, but for the case when one lib is linked with bells and whistles, makes things hard.

kennyweiss: I completely agree and was going to open an issue on caliper

Tag: @daboehme

kennyweiss: Having said that, I don't think it would help us in this case since we're getting this from mfem's `config.mk` which does not have CMake imports/exports, and axom does not (yet) use caliper.

cyrush: True, however it could also help mfem forward that info via a config.mk of its own :-)

PR #1017, opened 2023-02-01
Bugfixes in support of Windows build

kennyweiss:
# Summary

- This PR contains several bugfixes
- It improves the constraints on axom's `mfem` dependency in its spack package
- It adds an optional `lua` dependency to the Windows build
- It improves the import logic for lua (this manifested in a local WSL build, where `-dl` was required and missing from our previous formulation)
- It also fixes some failing tests when axom is configured without `hdf5`. 
    - An observed failure in the `spio_parallel` tests for this configuration will be resolved in
     #1018

PR #1034, opened 2023-02-17
Bugfixes for bugfixes

kennyweiss:
# Summary

- This PR contains some bugfixes for previous bugfix PRs
- It improves the import logic for lua (https://github.com/LLNL/axom/pull/1017)
    - CMake's `string(FIND )` returns an index, not a boolean
- It improves some floating point tests (https://github.com/LLNL/axom/pull/1025)
    - We need to use `EXPECT_NEAR` with a provided tolerance rather than `EXPECT_DOUBLE_EQ`
- I also slightly cleaned up the code in the test files

### DIFF
--- a/src/axom/spin/tests/spin_bvh.cpp
+++ b/src/axom/spin/tests/spin_bvh.cpp
@@ -15,9 +15,6 @@
 // gtest includes
 #include "gtest/gtest.h"
 
-using namespace axom;
-namespace xargs = mint::xargs;
-
 // Uncomment the following for debugging
 //#define VTK_DEBUG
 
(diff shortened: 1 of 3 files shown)


src/axom/spin/tests/spin_bvh.cpp
```diff
@@ -177,34 +184,28 @@ void generate_aabbs_and_centroids(const mint::Mesh* mesh,
...
+      sum.array() /= NUM_NODES_PER_CELL;
+
       c[cellIdx] = range.getCentroid();
 
 #ifndef AXOM_DEVICE_CODE
-      for(int dim = 0; dim < NDIMS; dim++)
-      {
-        EXPECT_DOUBLE_EQ(c[cellIdx][dim], sum[dim]);
-      }
+      EXPECT_NEAR(primal::squared_distance(c[cellIdx], sum), 0., EPS);
```
kennyweiss, line 208: Note: This function is called with `FloatType = {float,double}` but the coordinates are `double`, so there is some loss in precision in some cases when taking averages.

white238 (approved):

rhornung67 (approved):

agcapps (approved):
Thanks for the cleanup, @kennyweiss

PR #1052, opened 2023-03-17
Feature/whitlock/blueprint material output

### LINKED ISSUE #1050: Enhance Axom so the shaping driver can write data that includes matsets for visualization.
The shaping driver writes out materials as a set of scalar volume fraction arrays. Make it also write out the matset information needed so VisIt can interpret the material data as real materials so plots like FilledBoundary work.

BradWhitlock:
# Summary

- This PR is a small feature enhancement that makes the shaping driver and MFEMSidreDataCollection write Blueprint datasets that contain matset metadata needed for VisIt to treat volume fraction arrays as a material. This enables VisIt plots such as FilledBoundary.

- It does the following:
  - Adds a call to MFEMSidreDataCollection::AssociateMaterialSet in shaping driver program to drive the creation of matsets in the Blueprint output.
  - Enhances MFEMSidreDataCollection so it adds matset metadata into the Blueprint index file when necessary.
  - Changes the IntersectionShaper so the "completely free" material _(used to represent uncovered areas)_ is prefixed with the vol_frac naming prefix. This makes the material part of the matset so all volume fractions are accounted for. Without this, VisIt would attribute any unassigned volume fractions to the first real material.
  - Fixes #1050


src/axom/quest/IntersectionShaper.hpp
```diff
@@ -1075,11 +1077,25 @@ class IntersectionShaper : public Shaper
   // Switch back to public. This is done here because the CUDA compiler
   // does not like the following template functions to be private.
 public:
+  /*!
+   * \brief Set the name of the material used to account for free volume fractions.
+   * \param name The new name of the material.
+   * \note This should not be called once any shaping has occurred.
```
kennyweiss, line 1083: Should we add a check that shaping has not yet started?

src/axom/quest/IntersectionShaper.hpp
```diff
@@ -1075,11 +1077,25 @@ class IntersectionShaper : public Shaper
   // Switch back to public. This is done here because the CUDA compiler
   // does not like the following template functions to be private.
 public:
+  /*!
+   * \brief Set the name of the material used to account for free volume fractions.
+   * \param name The new name of the material.
```
kennyweiss, line 1082: Please add a note to the doxygen that `name` cannot contain an underscore

src/axom/quest/IntersectionShaper.hpp
```diff
@@ -276,7 +276,9 @@ class IntersectionShaper : public Shaper
   IntersectionShaper(const klee::ShapeSet& shapeSet,
                      sidre::MFEMSidreDataCollection* dc)
     : Shaper(shapeSet, dc)
-  { }
+  {
```
kennyweiss, line 279: Should we document the concept of the "free material" somewhere, e.g. in the class doxygen for `IntersectionShaper`, or possibly in the parent `Shaper` class ?

kennyweiss: I also created an issue for us to add user docs related to the shaping query -- https://github.com/LLNL/axom/issues/1053

kennyweiss (approved):
Thanks @BradWhitlock -- this is a really nice contribution that I've been looking forward to for a while!

src/axom/quest/IntersectionShaper.hpp
BradWhitlock: I added one using the number of cells member, which is set once the shaper has started its work.

src/axom/quest/IntersectionShaper.hpp
BradWhitlock: Done

src/axom/quest/IntersectionShaper.hpp
BradWhitlock: I added a class header that contains a bit about what IntersectionShaper does, how volume fractions are represented, and the purpose of the free material.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -993,6 +993,54 @@ void MFEMSidreDataCollection::UpdateStateToDS()
...
+      // Iterate over the fields in their group order.
+      sidre::Group* fields_grp = m_bp_grp->getGroup("fields");
+      for(auto& group : fields_grp->groups())
+      {
+        const std::string& field_name = group.getName();
+        const auto tokens = utilities::string::rsplitN(field_name, 2, '_');
+        // Expecting [base_field_name, material_id]
+        if(tokens.size() != 2)
+        {
+          continue;
```
white238, line 1018: Can you add an explanation why this is skipping?

white238 (approved):
Thanks!

PR #1072, opened 2023-04-07
Add methods to destroy Groups and Buffer data

### LINKED ISSUE #674: Add missing sidre group methods

nselliott:
# Summary

- This PR is a feature enhancement
- It does the following
  - Fixes #674 
  - Adds missing methods for destroying Groups which will also destroy Buffers if the destruction of a Group and the Views in its subtree cause the Buffer to become detached from all Views.

There already exist View destruction methods such as `destroyViewsAndData` that implement this destruction of Buffers for this use case.  This PR extends the API so that the same thing can be executed on the entire subtree under a Group or set of Groups.

### DIFF
--- a/src/axom/sidre/core/Group.cpp
+++ b/src/axom/sidre/core/Group.cpp
@@ -1033,6 +1033,31 @@ void Group::destroyGroup(const std::string& path)
 /*
  *************************************************************************
  *
+ * Detach child Group with given name or path and destroy it, and destroy
+ * any data that becomes disassociated from all Views
+ *
+ *************************************************************************
+ */
+void Group::destroyGroupAndData(const std::string& path)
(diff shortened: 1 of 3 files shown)


rhornung67 (approved):
Looks good. Thanks @nselliott

agcapps (approved):
Thank you, Noah.

kennyweiss:
Thanks for getting these in @nselliott !

PR #1073, opened 2023-04-17
Pro/E Mesh Reader

bmhan12:
This PR:

- Relates to #923 
- Adds a reader for Pro/E tetrahedral meshes
- Adds an example mesh file of a cup (LLNL/axom_data#15)

The cup, visualized as a vtk file:
cup

---------------------------------------------------------------------------------------------

A plot of the cup in 3 different places (translated, rotated):
multiple_cups


rhornung67 (approved):
Thanks @bmhan12 Just one orthogonal comment/suggestion for consideration.

agcapps (approved):
Thank you, Brian.  Sphinx docs in a separate PR?

bmhan12:
> Thank you, Brian. Sphinx docs in a separate PR?

Yes. Opened an issue here #1075, as the documentation needs to be expanded for c2c meshes as well.

src/axom/quest/readers/ProEReader.cpp
```diff
@@ -0,0 +1,149 @@
...
+    ifs >> id >> tet_nodes[0] >> tet_nodes[1] >> tet_nodes[2] >> tet_nodes[3];
+
+    for(int j = 0; j < NUM_NODES_PER_TET; j++)
+    {
+      for(int k = 0; k < NUM_COMPS_PER_NODE; k++)
+      {
+        m_nodes.push_back(nodes[tet_nodes[j]].comp[k]);
+      }
+    }
+  }
```
kennyweiss, line 93: The way this is implemented, we're losing track of the mesh connectivity.
I.e. every tet has 4 unique vertices rather than reusing the shared vertices from the input.

A more concise representation would be to store an array of coordinate positions, of size `m_num_unique_nodes` and an array of coordinate references (integers) of size `m_num_tets * 4`.  (I.e. as provided by the input mesh).

Since tet meshes can be large, it's probably worth implementing it that way and passing it to the mint mesh that way too. We can do this in a follow-up PR if you'd prefer.

src/axom/quest/tests/quest_pro_e_reader.cpp
```diff
@@ -0,0 +1,299 @@
...
+  EXPECT_TRUE(ofs.is_open());
+
+  ofs << "# Comment header to ignore" << std::endl;
+  ofs << "# Another comment" << std::endl;
+  ofs << "4 1" << std::endl;
+  ofs << "node1 -1.0 0.0 0.0" << std::endl;
+  ofs << "node2 1.0 0.0 0.0" << std::endl;
+  ofs << "node3 0.0 1.0 0.0" << std::endl;
+  ofs << "node4 0.0 0.0 1.0" << std::endl;
+  ofs << "tet1 node1 node2 node3 node4" << std::endl;
```
kennyweiss, line 40: Are you sure the vertex references can be strings? I'd imagine that they should be integers.

Clearly your implementation supports this, but it doesn't seem right to me.

src/axom/quest/tests/quest_pro_e_reader.cpp
```diff
@@ -0,0 +1,299 @@
...
+  EXPECT_NEAR(y[MID_OFFSET + 2], 117.715966, EPS);
+  EXPECT_NEAR(y[MID_OFFSET + 3], 78.384516, EPS);
+
+  EXPECT_NEAR(z[MID_OFFSET], -179.442905, EPS);
+  EXPECT_NEAR(z[MID_OFFSET + 1], -112.054428, EPS);
+  EXPECT_NEAR(z[MID_OFFSET + 2], -178.748301, EPS);
+  EXPECT_NEAR(z[MID_OFFSET + 3], -163.120569, EPS);
+
+  // Step 5: Dump mesh file
+  axom::mint::write_vtk(&mesh, "cup.vtk");
```
kennyweiss, line 288: Is writing the vtk sufficiently fast? Should it be optional/ commented out by default?

kennyweiss (approved):
Thanks @bmhan12 

Do we have a formal writeup of the spec for proe files? I don't think we have one from creo/proe, but perhaps we have one from J. White or A. Dayton?

One concern is about whether the vertex labels can be general strings as opposed to integer IDs. My assumption would be the latter.

src/axom/quest/tests/quest_pro_e_reader.cpp
bmhan12: > Are you sure the vertex references can be strings? I'd imagine that they should be integers.

Good catch! Yes, they are definitely integers. I'll change that.

src/axom/quest/readers/ProEReader.cpp
bmhan12: > A more concise representation would be to store an array of coordinate positions, of size m_num_unique_nodes and an array of coordinate references (integers) of size m_num_tets * 4. (I.e. as provided by the input mesh).

Refactored the storage so `m_nodes` are the coordinates and `m_tets` are the `int` IDs of the nodes.

src/axom/quest/tests/quest_pro_e_reader.cpp
bmhan12: > Is writing the vtk sufficiently fast? Should it be optional/ commented out by default?

It's fast in my experience (~0.30 seconds for all the unit tests), so I think it is fine if enabled by default.

I did notice there are a few Axom unit tests writing out meshes and other files, some optional and others not.
There does not look to be a standard way of enabling/disabling all these output options(commented out, #ifdef out, etc.).
Maybe something useful to standardize?

src/axom/quest/tests/quest_pro_e_reader.cpp
```diff
@@ -0,0 +1,327 @@
...
+    EXPECT_NEAR(y[inode],
+                y_expected[inode],
+                std::numeric_limits<double>::epsilon());
+    EXPECT_NEAR(z[inode],
+                z_expected[inode],
+                std::numeric_limits<double>::epsilon());
+  }  // END for all nodes
+
+  // STEP 4: remove temporary Pro/E file
+  std::remove(filename.c_str());
```
kennyweiss, line 204: > std::remove(filename.c_str());

Is this line necessary?

kennyweiss (approved):
Thanks for the updates @bmhan12 !

src/axom/quest/tests/quest_pro_e_reader.cpp
bmhan12: > Is this line necessary?

Yes, I think just to keep the build directory tidier.

PR #1077, opened 2023-04-26
axom::Array documentation additions

### LINKED ISSUE #1067: Investigate device errors using Primal primitive member functions with Axom::Array
Segfault when attempting to[ convert C-style arrays](https://github.com/LLNL/axom/pull/1049/files#r1138134652) of some Primal primitives to `axom::Array` or `axom::ArrayView`. 

Did not appear for all primitives, and only appeared when calling non-getter member functions.

Small failing example branch [here](https://github.com/LLNL/axom/compare/bugfix/han12/arrayGeo).

bmhan12:
This PR:
- Fixes gpu porting code example
- Switches to `ArrayView` for device memory in `hex_tet_volume`

closes #1067


rhornung67 (approved):
Looks good. Thank you @bmhan12

kennyweiss (approved):
Thanks @bmhan12!

publixsubfan (approved):
Thanks @bmhan12!

Did you see any compiler warnings with the failing case? Longer-term, I'm wondering if there's a way to guard against accidentally using axom::Array directly in a lambda; I thought that not having host/device-annotated copy constructors would be enough.

bmhan12:
> Did you see any compiler warnings with the failing case?

I reran the failing case, no compiler warnings.
It does throw an Umpire exception immediately at runtime, so it is clear something is wrong.
I think the documentation does make it clear to use `axom::ArrayView`, my own issue was getting the syntax wrong 😆

PR #1081, opened 2023-04-28
No more object libraries

white238:
This removes all of our CMake object library usage. This caused numerous problems.

Basically instead of having all the components being linked into a single `axom` library, we now have individual CMake targets and libraries for each component. There is also a new helper CMake target, named `axom`, created in the exported `axom-config.cmake`. Downstream users should not have to change anything in their CMake usage.

Important changes to point out:

- `axom` target is now an interface target that depends on all our underlying targets. It also only exists in the exported targets from `AxomConfig.cmake`. It does not exist in our internal build.
- All underlying targets are now namespaced with `axom::`, eg. `inlet` is now `axom::inlet`. Downstream users can now depend on these individual targets to minimize what they bring into their projects
- Our library file, `libaxom.a` or `libaxom.so`, no longer exists. There is a now a library for each component (except primal and spin, which are header only) and they are prefixed with `axom_`, eg. `libaxom_inlet.a`, `libaxom_core.a`.

### DIFF
--- a/.gitlab/build_lassen.yml
+++ b/.gitlab/build_lassen.yml
@@ -14,13 +14,9 @@
     - if: '$CI_COMMIT_BRANCH =~ /_lnone/ || $ON_LASSEN == "OFF"' #run except if ...
       when: never
     - when: on_success
-
-####
-# Load required CUDA module
-.with_cuda:
   before_script:
     - module load cuda/11.2.0
-    - module load cmake/3.21.1
+    - module load cmake/3.18.0
 
 ####
 # Template
@@ -50,7 +46,7 @@ lassen-clang_10_0_1_cuda-src:
   variables:
     COMPILER: "clang@10.0.1.2_cuda"
     HOST_CONFIG: "lassen-blueos_3_ppc64le_ib_p9-${COMPILER}.cmake"
-  extends: [.src_build_on_lassen, .with_cuda]
+  extends: [.src_build_on_lassen]
 
 lassen-gcc_8_3_1-src:
   variables:
@@ -62,7 +58,7 @@ lassen-gcc_8_3_1_cuda-src:
   variables:
     COMPILER: "gcc@8.3.1.2_cuda"
     HOST_CONFIG: "lassen-blueos_3_ppc64le_ib_p9-${COMPILER}.cmake"
-  extends: [.src_build_on_lassen, .with_cuda]
+  extends: [.src_build_on_lassen]
 
 lassen-xl_16_1_1-src:
   variables:
@@ -74,7 +70,7 @@ lassen-xl_16_1_1_cuda-src:
   variables:
     COMPILER: "xl@16.1.1.2_cuda"
     HOST_CONFIG: "lassen-blueos_3_ppc64le_ib_p9-${COMPILER}.cmake"
-  extends: [.src_build_on_lassen, .with_cuda]
+  extends: [.src_build_on_lassen]
 
 ####
 # Full Build jobs
@@ -89,7 +85,7 @@ lassen-clang_10_0_1_cuda-full:
     COMPILER: "clang@10.0.1.2"
     SPEC: "%${COMPILER}+mfem+cuda~openmp"
     EXTRA_SPEC: "cuda_arch=70"
-  extends: [.full_build_on_lassen, .with_cuda]
+  extends: [.full_build_on_lassen]
 
 lassen-gcc_8_3_1-full:
   variables:
@@ -102,7 +98,7 @@ lassen-gcc_8_3_1_cuda-full:
     COMPILER: "gcc@8.3.1.2"
     SPEC: "%${COMPILER}~mfem+cuda"
     EXTRA_SPEC: "cuda_arch=70"
-  extends: [.full_build_on_lassen, .with_cuda]
+  extends: [.full_build_on_lassen]
 
 lassen-xl_16_1_1-full:
   variables:
@@ -115,4 +111,4 @@ lassen-xl_16_1_1_cuda-full:
     COMPILER: "xl@16.1.1.2"
     SPEC: "%${COMPILER}+mfem+cuda~openmp~cpp14"
     EXTRA_SPEC: "cuda_arch=70"
-  extends: [.full_build_on_lassen, .with_cuda]
+  extends: [.full_build_on_lassen]
--- a/.gitlab/build_tioga.yml
+++ b/.gitlab/build_tioga.yml
@@ -13,6 +13,8 @@
     - if: '$CI_COMMIT_BRANCH =~ /_qnone/ || $ON_TIOGA == "OFF"' #run except if ...
       when: never
     - when: on_success
+  before_script:
+    - module load cmake/3.21.1
 
 ####
 # Template
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -48,8 +48,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 ### Changed
 - The Axom library has been broken down into its component libraries (prefixed with `axom_`).
   This change requires no change to downstream CMake users who import our targets.
-  The exported CMake target `axom` includes all components, but others will need to create the link line
-  themselves. The following replacement can be used:
+  The exported CMake target `axom` includes all components, but users who do not import our targets
+  will need to create the link line themselves. The following replacement can be used:
   `-laxom` -> `-laxom_quest -laxom_multimat -laxom_slam -laxom_mint -laxom_klee -laxom_inlet -laxom_sidre -laxom_slic -laxom_lumberjack -laxom_core`
   If you only need a subset of the components, you can now use those targets directly, ie. `axom::inlet`.
 - `IntersectionShaper` now implements material replacement rules.
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -46,6 +46,12 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   and by calling `Shaper::setPercentError()` to set a refinement error percentage.
 
 ### Changed
+- The Axom library has been broken down into its component libraries (prefixed with `axom_`).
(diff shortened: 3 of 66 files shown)


src/examples/using-with-cmake/CMakeLists.txt
```diff
@@ -53,27 +53,9 @@ if (ENABLE_HIP)
...
-# BEGIN FIXME:
-# Create fake empty target, this stops CMake from adding -lcuda_runtime to the link line
-add_library(cuda_runtime INTERFACE)
-
-# Create fake empty target, this stops CMake from adding -lblt_hip_runtime to the link line
-add_library(blt_hip_runtime INTERFACE)
-# END FIXME
+find_dependency(axom REQUIRED
+                NO_DEFAULT_PATH 
+                PATHS ${AXOM_DIR}/lib/cmake)
```
kennyweiss, line 60: Nice!

src/cmake/AxomMacros.cmake
```diff
@@ -115,6 +115,96 @@ macro(axom_add_component)
...
+    set(singleValueArgs NAME OUTPUT_DIR OUTPUT_NAME FOLDER)
+    set(multiValueArgs HEADERS SOURCES INCLUDES DEFINES DEPENDS_ON)
+
+    # Parse the arguments to the macro
+    cmake_parse_arguments(arg
+        "${options}" "${singleValueArgs}" "${multiValueArgs}" ${ARGN})
+
+    # Blanket add openmp as a dependency to get around not propagating
+    # openmp to fortran dependencies
+    blt_list_append(TO arg_DEPENDS_ON ELEMENTS openmp IF ENABLE_OPENMP)
```
kennyweiss, line 144: I guess (?)

Should this be
```suggestion
    # Blanket add openmp as a dependency to get around not propagating
    # openmp to fortran dependencies
    blt_list_append(TO arg_DEPENDS_ON ELEMENTS openmp IF AXOM_ENABLE_OPENMP)
```
e.g. in case the BLT variable is set, but the axom one is not? Or would we want it in both cases?
(we've had so many issues/bugs related to OpenMP...)

src/axom/spin/dummy.cpp
kennyweiss, line 1: I guess this means there are still issues w/ header-only libs, even when not using object libs?

src/axom/CMakeLists.txt
```diff
@@ -39,32 +38,31 @@ axom_add_component(COMPONENT_NAME klee   DEFAULT_STATE ${AXOM_ENABLE_ALL_COMPONE
...
-                SOURCES    Axom.cpp
-                DEPENDS_ON ${AXOM_COMPONENTS_ENABLED})
-
-install(TARGETS              axom
+install(TARGETS              ${AXOM_COMPONENTS_ENABLED}
         EXPORT               axom-targets
         DESTINATION          lib)
-install(EXPORT axom-targets DESTINATION lib/cmake)
+install(EXPORT               axom-targets 
+        NAMESPACE            axom::
```
kennyweiss, line 45: What's the effect of this `NAMESPACE axom::` ? Does it add an alias so everything is `axom::<foo>`?

kennyweiss (approved):
Thanks @white238  -- I'm also looking forward to faster builds for our tests and examples since they no longer depend on "axom"

What changes will make-based build systems need to make?
Should we generate a package-config file to make it easier for them to figure out which libraries they'll need? (Perhaps as a follow up PR).

Please update the RELEASE-NOTES about the change and any side effects.

src/axom/spin/dummy.cpp
white238: I honestly didn't try removing this. I'll try in a bit.

src/axom/CMakeLists.txt
white238: All targets added to the `axom-targets` export set will be namespaced by `axom::`. So `fmt`, `core`, and `mpi` will all be `axom::fmt`, etc...

When people include our targets they will end up with `axom`, which depends on everything, and the rest of the targets namespaced under that one. So Serac for example, can just depend on `axom::inlet`.

src/cmake/AxomMacros.cmake
white238: Ah yes.. i copied this from the previous attempt at this. Good catch.

src/cmake/AxomMacros.cmake
```diff
@@ -115,6 +115,96 @@ macro(axom_add_component)
...
+                    SOURCES     ${arg_SOURCES}
+                    HEADERS     ${arg_HEADERS}
+                    INCLUDES    ${arg_INCLUDES}
+                    DEFINES     ${arg_DEFINES}
+                    DEPENDS_ON  ${arg_DEPENDS_ON} ${axom_device_depends}
+                    OUTPUT_DIR  ${arg_OUTPUT_DIR}
+                    OUTPUT_NAME ${arg_OUTPUT_NAME}
+                    FOLDER      ${arg_FOLDER})
+
+    if(ENABLE_OPENMP AND (NOT "${arg_SOURCES}" STREQUAL ""))
```
kennyweiss, line 200: Should this change too?
```suggestion
    if(AXOM_ENABLE_OPENMP AND (NOT "${arg_SOURCES}" STREQUAL ""))
```

src/axom/spin/dummy.cpp
white238: fixed.

.gitlab/build_lassen.yml
```diff
@@ -14,13 +14,9 @@
...
       when: never
     - when: on_success
-
-####
-# Load required CUDA module
-.with_cuda:
   before_script:
     - module load cuda/11.2.0
-    - module load cmake/3.21.1
+    - module load cmake/3.18.0
```
rhornung67, line 19: FYI. RAJA and Umpire require cmake min version 3.20.

.gitlab/build_tioga.yml
```diff
@@ -13,6 +13,8 @@
     - if: '$CI_COMMIT_BRANCH =~ /_qnone/ || $ON_TIOGA == "OFF"' #run except if ...
       when: never
     - when: on_success
+  before_script:
+    - module load cmake/3.21.1
```
rhornung67, line 17: RAJA and Umpire require min cmake version 3.23 for HIP builds. 3.24.2 is the system default on tioga.

.gitlab/build_lassen.yml
white238: I was having some serious issues with an intermittent CMake issue that was only in CI. I wanted to limit the amount of variables so i picked 3.18.0 across the board (minus Tioga which doesn't have that version).

We aren't actually requiring this version at all and this ended up not being the issue.

RELEASE-NOTES.md
```diff
@@ -46,6 +46,12 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   and by calling `Shaper::setPercentError()` to set a refinement error percentage.
 
 ### Changed
+- The Axom library has been broken down into its component libraries (prefixed with `axom_`).
+  This change requires no change to downstream CMake users who import our targets.
+  The exported CMake target `axom` includes all components, but others will need to create the link line
```
rhornung67, line 51: what does 'others' refer to here?

white238: people who don't import our targets. I'll clarify...

src/axom/CMakeLists.txt
```diff
@@ -10,17 +10,16 @@
 #------------------------------------------------------------------------------
 
 # Core is an essential part of Axom and cannot be turned off
-add_subdirectory(core)
-set(AXOM_ENABLE_CORE ON CACHE BOOL "" FORCE)
-set(AXOM_COMPONENTS_FULL core CACHE STRING "List of all components in Axom" FORCE)
-set(AXOM_COMPONENTS_ENABLED core
-    CACHE STRING "List of all enabled components in Axom" FORCE)
+axom_add_component(COMPONENT_NAME core DEFAULT_STATE ON)
+if(NOT AXOM_ENABLE_CORE)
```
rhornung67, line 14: Curiosity question: if core is always required, why do we have the AXOM_ENABLE_CORE variable?

src/axom/core/examples/CMakeLists.txt
```diff
@@ -14,18 +14,16 @@ set(example_sources
...
 
-set( core_examples_depends axom ${axom_device_depends})
-
 #------------------------------------------------------------------------------
 # Add targets and tests for serial examples
 #------------------------------------------------------------------------------
 foreach(example_source ${example_sources})
     get_filename_component(exe_name ${example_source} NAME_WE)
-    blt_add_executable(
+    axom_add_executable(
```
rhornung67, line 22: 😮

rhornung67 (approved):
This looks like a lot of work. Thank you @white238

src/axom/CMakeLists.txt
white238: Because it automatically created it =) I'll remove it.

white238:
I may be the only one that cares about this. But this change HUGELY cleaned up a lot of duplication in our CMake target properties.

This whole section is removed:
axom-dependencies

src/axom/core/CMakeLists.txt
```diff
@@ -97,10 +97,7 @@ set( core_depends fmt )
 blt_list_append( TO core_depends ELEMENTS camp IF CAMP_FOUND )
 blt_list_append( TO core_depends ELEMENTS umpire IF UMPIRE_FOUND )
 blt_list_append( TO core_depends ELEMENTS RAJA IF RAJA_FOUND )
-list(APPEND core_depends ${axom_device_depends})
```
agcapps, line 100: Why is it that we can do away with this line?

agcapps: Oh, I see--- that is what `axom_add_executable` and `axom_add_library` do for us.

agcapps (approved):
Thank you, Chris!

gunney1 (approved):
Thanks @white238 I like the more abbreviated interface and not having to  bring in unneeded components to link.

PR #1083, opened 2023-05-03
Untie RAJA and Umpire

### LINKED ISSUE #679: Ensure Umpire usage is correctly guarded in Axom
@rhornung67 observed that there are several places in the code that are not properly guarded against Umpire not being present in configurations that have `RAJA` but not `Umpire`.

We can test this by using a configuration with `RAJA` that does not have `Umpire`.

agcapps:
Add guards for uses of members defined only if we have both RAJA and Umpire, as well as tests.

# Summary

- This PR is a bugfix.
- It does the following:
  - Adds preprocessor guards testing for AXOM_USE_RAJA and AXOM_USE_UMPIRE around uses of object members defined only if we have those libs
  - Adds tests on Azure knocking out those libraries
  - Fixes #679


src/axom/quest/tests/quest_intersection_shaper.cpp
kennyweiss, line 1: Note: 
The `IntersectionShaper` does not currently run without RAJA. 
Adding non-RAJA support is the subject of https://github.com/LLNL/axom/issues/973

(I don't think there are any implications for the current PR, just wanted to point this out).

azure-pipelines.yml
```diff
@@ -40,6 +40,30 @@ jobs:
...
+        TEST_TARGET: 'linux_clang10'
+        HOST_CONFIG: 'clang@10.0.0'
+      linux_clang10_noraja_noumpire:
+        VM_ImageName: 'ubuntu-20.04'
+        Compiler_ImageName: '$(CLANG10_IMAGENAME)'
+        CMAKE_EXTRA_FLAGS: '-DBUILD_SHARED_LIBS=ON -DAXOM_QUEST_ENABLE_EXTRA_REGRESSION_TESTS:BOOL=ON -U RAJA_DIR -U UMPIRE_DIR'
+        BUILD_TYPE: 'Debug'
+        COMPILER: 'clang++'
+        TEST_TARGET: 'linux_clang10'
+        HOST_CONFIG: 'clang@10.0.0'
```
agcapps, line 66: These three new checks are not showing up as Required.  Any ideas?  The whole reason we have them is to make sure they succeed!

src/axom/quest/IntersectionShaper.hpp
```diff
@@ -311,7 +311,9 @@ class IntersectionShaper : public Shaper
                      sidre::MFEMSidreDataCollection* dc)
     : Shaper(shapeSet, dc)
   {
+#if defined(AXOM_USE_RAJA) && defined(AXOM_USE_UMPIRE)
     m_free_mat_name = "free";
+#endif
```
kennyweiss, line 316: This change seems odd to me.

Is there a check elsewhere that requires RAJA/Umpire for this string being non-empty?

src/axom/quest/tests/quest_intersection_shaper.cpp
```diff
@@ -755,14 +755,15 @@ dimensions: 3
 
 //---------------------------------------------------------------------------
 // Define testing functions for different modes.
-#if defined(RUN_AXOM_SEQ_TESTS)
+#if defined(AXOM_USE_RAJA) && defined(AXOM_USE_UMPIRE)
```
kennyweiss, line 758: Minor: Do the SEQ and OpenMP tests really require Umpire? 
I would expect the GPU testes to require Umpire, but not the CPU-based tests.

azure-pipelines.yml
kennyweiss, line 66: 👍

kennyweiss (approved):
Thanks @agcapps !

azure-pipelines.yml
kennyweiss: I see them here: https://dev.azure.com/healy22/Status%20Check%20Testiing/_build/results?buildId=7847&view=results

image

white238: The Branch Protection rules controls this. The parent job "LLNL.axom" covers those jobs but I can add them explicitly like the rest of the jobs after it is merged and ran on the default branch ("develop").

agcapps: Thanks, Chris.  Yes, please, after it gets merged in.

kennyweiss: (Sorry -- I misread the "Required" part)

src/axom/quest/IntersectionShaper.hpp
agcapps: The definition of `m_free_mat_name`, at the bottom of this file, is within `#if defined(AXOM_USE_RAJA) && defined(AXOM_USE_UMPIRE)`.  Since `m_free_mat_name` is used extensively throughout code guarded on those symbols and nowhere else (except this point), it made more sense to me to guard the initialization rather than move the definition out of its guards.

src/axom/quest/tests/quest_intersection_shaper.cpp
agcapps: Yes.  The SEQ and OpenMP tests go into `replacementRuleTestSet()` and eventually through `prepareShapeQueryImpl()`, whose definition at `IntersectionShaper.hpp` line 362 is within `#if defined(AXOM_USE_RAJA) && defined(AXOM_USE_UMPIRE)`.

agcapps: If we wanted to run intersection shaping sequentially, we would need to pull that out and roll the reducers and for_all loops without RAJA.

PR #1084, opened 2023-05-04
Initial port of Multimat to the GPU

publixsubfan:
# Summary

- Changes to support Multimat on the GPU

## Public-facing changes

- `MultiMat::Field2D<T>` and similar types are now shallow, device-copyable views of a field instead of full copies of field data
- Added const overloads for `MultiMat::get1dField()`, `MultiMat::get2dField()`
- Added missing `MultiMat::getSparse2dField()` implementation
- Fixed instantiation of `MultiMat::getField()` functions with non-`double` types
- Added new functions `MultiMat::setSlamAllocatorID()` and `MultiMat::setFieldAllocatorID()` to control allocation of Slam-associated relation data and field storage, respectively
  - Right now, both allocators are required to point to unified memory in order to make use of transforms between different layouts -- device-side transforms will be the subject of a future PR
- Added `MultiMat::addExternalField()` function to add a field which is backed by external memory
  - Externally-backed fields  support conversion between cell-dominant and material-dominant layouts
  - Externally-backed fields don't support:
    - transposing between sparse and dense layouts
    - changing allocator IDs
  - Externally-backed fields are copied into internally-backed fields when the underlying MultiMat instance is copied
- Added `MultiMat::removeField()` function to remove fields from the object
- Use `axom::ArrayView<T>` instead of raw pointer in `MultiMat::addField()`, `MultiMat::setVolfracField()` interface
- Added an overload of `MultiMat::setCellMatRel` which accepts cell-material relations as a CSR array representation, i.e. array of number of materials/cell + flat array of material IDs

## Internal changes

- Slam types in Multimat use `ArrayView` indirection when possible, and underlying instances and fields are stored in `axom::Array`. This is controlled by the user-provided allocator ID to allow for accessing Slam data and fields on both the host and device
- Slam types now use the concrete interface, to avoid hazards with virtual tables crossing host/device boundary
- Added testing for `int` and `float` fields in addition to `double`, as well as non-unit striding
- Adds an example of using Multimat to access data on the GPU in `calculate_gpu.cpp`


src/axom/multimat/multimat.hpp
```diff
@@ -216,6 +208,29 @@ class MultiMat
...
+  void setAllocatorID(int alloc_id);
+
+  /**
+   * \brief Sets the allocator ID to use for Slam object allocations.
+   *  This should point to either a host Umpire pool if only being used on the
+   *  host, or a unified memory pool if also being used on the GPU.
+   *
+   * \param alloc_id the Umpire allocator ID to use
+   */
+  void setSlamAllocatorID(int alloc_id);
```
agcapps, line 225: Is there a reason that the allocator ID is not automatically the SLAM allocator?

src/axom/multimat/multimat.hpp
```diff
@@ -267,21 +305,52 @@ class MultiMat
...
+   * \param field_mapping
+   * \param data_layout
+   * \param sparsity_layout
+   * \param data_array The array containing data to the field. The length of the
+   *            array should be `num_mats * num_cells * ncomp` if the current
+   *            format is Dense, or `num_nonzero * ncomp` if the current format
+   *            is Sprase.
+   * \param (optional) ncomp The number of component of the field. Default is 1
+   * \return int the index of the field, can be used to retrieve the field later
+   */
```
agcapps, line 330: Please document what happens if a user tries to add a field and `field_name` already exists.

agcapps: I do realize that adding to the documentation expands the scope of the PR (since the Doxygen for (new) line 309 needs to match this comment).  But `addExternalField` is new work, and the behavior from (new) line 948-954 definitely needs to be documented.

src/axom/multimat/multimat.cpp
```diff
@@ -788,6 +983,12 @@ void MultiMat::convertFieldToSparse(int field_idx)
     return;
   }
 
+  if(!m_fieldBackingVec[field_idx]->isOwned())
+  {
+    throw std::runtime_error("Multimat error: cannot convert unowned field \"" +
+                             m_fieldNameVec[field_idx] + "\" to sparse layout.");
```
agcapps, line 989: Is this (and the `throw` at (new) line 1030) the reason for the try/catch at (new) lines 897--905 and 880--889?

agcapps (approved):
Looks like a thorough job, @publixsubfan .

src/axom/multimat/multimat.hpp
```diff
@@ -216,6 +208,29 @@ class MultiMat
...
+   * \brief Sets the allocator ID to use for allocations.
+   *
+   * \param The allocator ID to use
+   */
+  void setAllocatorID(int alloc_id);
+
+  /**
+   * \brief Sets the allocator ID to use for Slam object allocations.
+   *  This should point to either a host Umpire pool if only being used on the
+   *  host, or a unified memory pool if also being used on the GPU.
```
kennyweiss, line 221: Does it have to use unified memory on the GPU, or can it use a device memory pool there?
If it's the former, we'll likely need to add support for the latter.

src/axom/multimat/multimat.hpp
```diff
@@ -293,7 +362,13 @@ class MultiMat
    * \param data_array the array containing the volumn fraction information
```
kennyweiss, line 362: Typo: 
```suggestion
   * \param data_array the array containing the volume fraction information
```

src/axom/multimat/multimat.hpp
```diff
@@ -494,11 +607,26 @@ class MultiMat
 
 protected:
   //Return the Set pointer associalted with the given FieldMapping or field idx
```
kennyweiss, line 609: Typo:
```suggestion
  //Return the Set pointer associated with the given FieldMapping or field idx
```

src/axom/multimat/multimat.hpp
```diff
@@ -629,13 +783,82 @@ class MultiMat
   // dense layout bivariate sets
   axom::Array<ProductSetType> m_denseBivarSet;
 
+  struct FieldBacking
+  {
+  private:
+    bool m_isOwned {false};
+
+    axom::Array<unsigned char> m_ucharData;
+    axom::Array<int> m_intData;
```
kennyweiss, line 792: Should these fixed std types instead of `unsigned char` and `int`?
E.g.  `std::uint8_t` and `std::int32_t`

(The answer could be `no`).

src/axom/multimat/multimat.hpp
```diff
@@ -216,6 +208,29 @@ class MultiMat
    */
   void setNumberOfCells(int num_cells);
 
+  /**
+   * \brief Sets the allocator ID to use for allocations.
+   *
```
kennyweiss, line 213: Looking at the implementation, this function calls `setSlamAllocatorID(alloc_id)` and `setFieldAllocatorID(alloc_id)`.
Should we mention this in the doxycomment?

src/axom/multimat/multimat.cpp
```diff
@@ -260,58 +425,80 @@ void MultiMat::setCellMatRel(vector<bool>& vecarr, DataLayout layout)
...
+  }
+  else if(field_index < 0)
+  {
+    // Field does not exist.
+    SLIC_WARNING("Multimat Error: field with name \"" << field_name
+                                                      << "\" does not exist.");
+  }
+  else
+  {
+    m_fieldNameVec.erase(m_fieldNameVec.begin() + field_index);
```
kennyweiss, line 461: Would it make sense to use `sidre::IndexedCollection`s for these?
That is what `sidre::Group`s use for their underlying indexing of  children `View`s and `Group`s.
The advantage would be that it tracks indices and free spaces for us.


Perhaps we can move that class to core if we do not want `multimat` to depend on `sidre`.

If we want to do this, it should be done in a follow-up PR.

src/axom/multimat/multimat.cpp
```diff
@@ -701,15 +877,32 @@ void MultiMat::convertLayoutToSparse()
...
+    try
+    {
+      convertFieldToSparse(map_i);
+    }
+    catch(const std::runtime_error& ex)
+    {
+      SLIC_WARNING("Multimat: cannot convert field \""
+                   << m_fieldNameVec[map_i]
+                   << "\" to sparse layout, skipping...");
+    }
```
kennyweiss, line 889: Should we be introducing exceptions into the conversion operations? Are there downsides of doing this?

src/axom/multimat/multimat.cpp
kennyweiss: A different strategy would be to use `SLIC_WARNING` or `SLIC_ERROR` for invalid operations.

src/axom/multimat/examples/calculate_gpu.cpp
```diff
@@ -0,0 +1,387 @@
+// Copyright (c) 2017-2022, Lawrence Livermore National Security, LLC and
```
kennyweiss, line 1:
```suggestion
// Copyright (c) 2017-2023, Lawrence Livermore National Security, LLC and
```

src/axom/multimat/examples/calculate_gpu.cpp
kennyweiss, line 1: This is really nice!

Tag: @tomstitt

kennyweiss (approved):
Thanks @publixsubfan -- this is great!

Could you please post some performance numbers for this branch?
E.g. to compare to the results in https://zenodo.org/record/5559021, or perhaps more recent results.

It would also be great to compare the performance numbers in the new `calculate_gpu` examples for the different execution policies.

src/axom/multimat/multimat.hpp
publixsubfan: I haven't tested it with a device-only allocator, but I suspect there's a lot more work to be done before that can be supported. At a minimum, the conversions between sparse/dense and cell-dominant/material-dominant need to be GPU loops.

src/axom/multimat/multimat.hpp
publixsubfan: I've added documentation for the special-case behavior when adding a volume fraction field or an already-existing field.

src/axom/multimat/multimat.cpp
publixsubfan: Switched over to SLIC for error handling.

agcapps (approved):
Thanks for these fixes, @publixsubfan .

rhornung67 (approved):
Looks really good! Nice work @publixsubfan

src/axom/multimat/multimat.hpp
kennyweiss: Thanks @publixsubfan.

Unfortunately, device-only allocators is necessary for the intended application, so I think we'll need to address that in a follow-up PR.

PR #1090, opened 2023-05-17
Fix a bad cmake variable name and an error it was hiding.

gunney1:
# Fix bad cmake variable and error it was hiding

- This PR is a bugfix
  - Changes the cmake variable `AXOM_ENABLE_HDF5` to `FOUND_HDF5`.  Use of the wrong name left the distributed closest point search test disabled for the last couple of weeks.
  - Fix a compile error in the distributed closest point test.


white238 (approved):
Thanks for pushing this through, unfortunately CI is broken and spack is giving me serious issues. Working as fast as I can to fix it.

rhornung67 (approved):

kennyweiss (approved):
Thanks @gunney1 !

Issue #1057, opened 2023-03-20
Remove object libraries


white238:
Object libraries are becoming more trouble than they are worth and with exporting CMake targets under a single target it removes the original user level problem (the ordering of our component libraries).

white238:
Closed in #1081

Issue #1028, opened 2023-02-13
Fix `using-with-cmake` and `using-with-blt` examples on Windows


kennyweiss:
User @tlorentz77 reported trouble building the `using-with-cmake` and `using-with-blt` examples on Windows.

We should fix these and incorporate them into our Windows-based CI.

See #982 for more details.

tlorentz77:
Hello @kennyweiss , I've noticed that you have merged a branch (https://github.com/LLNL/axom/pull/1094) to fix this issue. Thanks very much for your hard work.

However, I found this issue have not been fixed yet. If you are trying to build one of the examples (https://github.com/LLNL/axom/tree/develop/src/examples/using-with-blt) after installing axom using vcpkg on Windows, you will find that lots of TPL pakcages cannot be found. I think the issue is caused by the package config files located in different folders on different platform (e.g. on Linux: "${CAMP_DIR}/lib/cmake/camp" and on Windows: "${CAMP_DIR}/share/camp". )

https://github.com/LLNL/axom/blob/a5b9b2239020df6283adb804532c691a5924904d/src/cmake/axom-config.cmake.in#L90

The above is just for example, and the paths of other packages have similar problem.

kennyweiss:
Thanks for your continued patience @tlorentz77.

I'm still working on the fix for the `using-with-cmake` problem (i.e. this issue).
In the process of looking into it, I noticed that the Windows build was broken, and fixed it in #1094.
I'm optimistic that the fix will be reasonably straightforward and will keep you posted on my progress.

tlorentz77:
Thanks @kennyweiss. I can hardly wait to run - Axom on Windows.

PR #1094, opened 2023-05-20
Fix windows build

kennyweiss:
# Summary

- This PR is a bugfix for a broken build on Windows
- When we split the axom object library target into separate components in #1057, it broke static builds on Windows. Specifically, we can no longer have a unified `AXOM_EXPORT` macro; we need one per component, e.g. `AXOM_CORE_EXPORT`, `AXOM_SLIC_EXPORT`, ...
- I also tried to get an MPI-enabled build on Windows.
   -  This led to a compiler issue w/ our `axom::mpi_traits` classes.
   - MSVC did not allow `static const` integer members; and wanted them to be `static constexpr`s defined inline
   - nvcc/ompi on the other hand did not like this b/c its `MPI_Datatype` is not an integer type
   - As a compromise, since there are only a handful of native types and it's only one line per type, I implemented it both ways.
   - If anyone has better ideas, I'd love to improve this.
- I didn't get MPI to work yet in this PR due to mfem's hypre requirement, which pulls in fortran (not part of msvc). We'll have to work around this in a future PR.
- Relates to #1028


src/axom/core/Types.hpp
```diff
@@ -69,97 +68,128 @@ using IndexType = std::int32_t;
...
-  //------------------------------------------------------------------------------
   #ifndef AXOM_NO_INT64_T
 template <>
 struct mpi_traits<std::int64_t>
 {
+    #ifdef _MSC_VER
+  static constexpr MPI_Datatype type = MPI_INT64_T;
+    #else
   static const MPI_Datatype type;
+    #endif
```
white238, line 176:
```suggestion
  #ifdef _MSC_VER
  static constexpr MPI_Datatype type = MPI_INT64_T;
  #else
  static const MPI_Datatype type;
  #endif
```

src/axom/core/Types.hpp
```diff
@@ -69,97 +68,128 @@ using IndexType = std::int32_t;
...
 
-//------------------------------------------------------------------------------
 template <>
 struct mpi_traits<std::uint64_t>
 {
+    #ifdef _MSC_VER
+  static constexpr MPI_Datatype type = MPI_UINT64_T;
+    #else
   static const MPI_Datatype type;
+    #endif
```
white238, line 186:
```suggestion
  #ifdef _MSC_VER
  static constexpr MPI_Datatype type = MPI_UINT64_T;
  #else
  static const MPI_Datatype type;
  #endif
```

src/axom/CMakeLists.txt
```diff
@@ -49,25 +49,21 @@ install(EXPORT               axom-targets
...
+        continue()
+    endif()
 
-        blt_add_target_definitions(
-            TO                 ${_component}
-            SCOPE              PRIVATE
-            TARGET_DEFINITIONS "axom_EXPORTS")
-    endforeach()
-endif()
+    set(_export_header ${PROJECT_BINARY_DIR}/include/axom/export/${_component}.h)
```
white238, line 60: Do these headers get included somewhere I can't find?

white238 (approved):

src/axom/mint/mesh/RectilinearMesh.hpp
```diff
@@ -7,8 +7,9 @@
 #define MINT_RECTILINEARMESH_HPP_
 
 #include "axom/config.hpp"
-#include "axom/mint/mesh/StructuredMesh.hpp"  // base class
-#include "axom/mint/config.hpp"               // for compile-time definitions
+#include "axom/mint/mesh/StructuredMesh.hpp"
+#include "axom/mint/config.hpp"
+#include "axom/export/mint.h"
```
kennyweiss, line 12: @white238 -- here's an example of including the exported header.

Here's what's generated for export/mint.h` on toss3
```

#ifndef AXOM_MINT_EXPORT_H
#define AXOM_MINT_EXPORT_H

#ifdef AXOM_MINT_STATIC_DEFINE
#  define AXOM_MINT_EXPORT
#  define AXOM_MINT_NO_EXPORT
#else
#  ifndef AXOM_MINT_EXPORT
#    ifdef mint_EXPORTS
        /* We are building this library */
#      define AXOM_MINT_EXPORT 
#    else
        /* We are using this library */
#      define AXOM_MINT_EXPORT 
#    endif
#  endif

#  ifndef AXOM_MINT_NO_EXPORT
#    define AXOM_MINT_NO_EXPORT 
#  endif
#endif

#ifndef AXOM_MINT_DEPRECATED
#  define AXOM_MINT_DEPRECATED __attribute__ ((__deprecated__))
#endif

#ifndef AXOM_MINT_DEPRECATED_EXPORT
#  define AXOM_MINT_DEPRECATED_EXPORT AXOM_MINT_EXPORT AXOM_MINT_DEPRECATED
#endif

#ifndef AXOM_MINT_DEPRECATED_NO_EXPORT
#  define AXOM_MINT_DEPRECATED_NO_EXPORT AXOM_MINT_NO_EXPORT AXOM_MINT_DEPRECATED
#endif

#if 0 /* DEFINE_NO_DEPRECATED */
#  ifndef AXOM_MINT_NO_DEPRECATED
#    define AXOM_MINT_NO_DEPRECATED
#  endif
#endif

#endif /* AXOM_MINT_EXPORT_H */
```

src/axom/sidre/spio/IOBaton.hpp
```diff
@@ -22,6 +22,7 @@
 #include "axom/core/Macros.hpp"
 #include "axom/core/Types.hpp"
 #include "axom/slic/interface/slic.hpp"
+#include "axom/export/sidre.h"
```
kennyweiss, line 25: @white238 -- here's another include for the new exports

src/axom/core/Types.hpp
kennyweiss: Unfortunately, the indentations are from clang-format.
We can disable formatting, or attempt to modify our `.clang-format` settings to achieve your recommendation (or live w/ the current indentation).

What's your preference?

white238: i say leave it as is. I expect we are going to be upgrading our clang-format soon so maybe it will be fixed?

kennyweiss:
Passed manual Windows TPL CI: https://github.com/LLNL/axom/actions/runs/5040134804

jcs15c (approved):
Looks good to me. Thanks for fixing this!

agcapps:
What version of MS Visual Studio did you use?  Was it with the MS compiler, or the Intel OneAPI compiler?

agcapps:
Are you using MSMPI  or something else (like Intel MPI)?

kennyweiss:
> What version of MS Visual Studio did you use? Was it with the MS compiler, or the Intel OneAPI compiler?

@agcapps -- I tested this locally w/ msvc 2022 and that's what our github actions CI is using as well 
(See: https://github.com/LLNL/axom/blob/develop/.github/workflows/test_windows_tpls.yml).

kennyweiss:
> Are you using MSMPI  or something else (like Intel MPI)?

@agcapps -- I encountered some errors when trying to enable mpi (specifically, MSMPI) w/ the rest of our TPLs on Windows and did not get it working in this PR.  (I alluded to this in the PR description).

The main problem was that `mfem+mpi` requires `hypre`, which requires a Fortran compiler, and MSVC does not supply one. There was an error when our vcpkg tried to use the mingw Fortran compiler that I did not have a chance to investigate.  

One possible solution would be to try to have an `axom+mpi` using a non-mpi version of `mfem`. This would likely involve some additional work w/ our `MFEMSidreDataCollection` class, which was not designed to use an `axom` w/ mpi and an `mfem` without it.

Another possibility might be to test this out w/ the intel compiler on Windows.

PR #1099, opened 2023-05-23
Multimat: GPU porting and acceleration

publixsubfan:
# Summary

## Slam
- Moves `RelationSet::flatToFirstIndex()` logic into CardinalityPolicy
- Adds a `MappedVariableCardinality` policy which holds an extra indirection buffer for mapping flat relation offsets to the first set index
  - This allows for constant-time `flatToFirstIndex()` lookups for variable cardinality, at the cost of extra space

## Multimat
- Use `MappedVariableCardinality` for static relations to accelerate transposition operations
- Ports transposition of field data between cell-dominant/material-dominant and sparse/dense layouts
- Ports `MultiMat::makeOtherRelation()` to GPU
- Ports `MultiMat::setCellMatRel(counts, indices)` overload to GPU
  - `counts` and `indices` must be in GPU-accessible space if `slamAllocatorID` is a GPU-accessible allocator

### DIFF
--- a/src/axom/multimat/multimat.cpp
+++ b/src/axom/multimat/multimat.cpp
@@ -10,6 +10,7 @@
  */
 
 #include "axom/multimat/multimat.hpp"
+#include "axom/core/execution/for_all.hpp"
 #include "axom/slic.hpp"
 
 #include <iostream>
@@ -1041,6 +1042,32 @@ void MultiMat::convertFieldToDense(int field_idx)
   m_fieldSparsityLayoutVec[field_idx] = SparsityLayout::DENSE;
 }
 
+template <typename ExecSpace, typename DataType>
+axom::Array<DataType> ConvertToSparseImpl(
+  const MultiMat::DenseField2D<DataType> oldField,
+  const MultiMat::RelationSetType* relationSet,
+  int allocatorId)
+{
+  int stride = oldField.stride();
+  int sparseSize = relationSet->totalSize() * stride;
+  axom::Array<DataType> sparseField(sparseSize, sparseSize, allocatorId);
+  const auto sparseFieldView = sparseField.view();
+
+  axom::for_all<ExecSpace>(
+    relationSet->totalSize() * stride,
+    AXOM_LAMBDA(int index) {
+      int flatIdx = index / stride;
+      int comp = index % stride;
+
+      auto firstIdx = relationSet->flatToFirstIndex(flatIdx);
+      auto secondIdx = relationSet->flatToSecondIndex(flatIdx);
+
+      sparseFieldView[index] = oldField(firstIdx, secondIdx, comp);
+    });
+
+  return sparseField;
+}
+
 template <typename DataType>
 void MultiMat::convertToSparse_helper(int map_i)
 {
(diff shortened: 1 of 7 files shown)


rhornung67 (approved):
LGTM. Thanks @publixsubfan

src/axom/multimat/multimat.cpp
```diff
@@ -382,16 +460,53 @@ void MultiMat::setCellMatRel(const vector<bool>& vecarr, DataLayout layout)
...
+  {
+    // Copy counts and indices to GPU memory.
+    counts = IndBufferType(counts, m_slamAllocatorId);
+    indices = IndBufferType(indices, m_slamAllocatorId);
+  }
+
   setCellMatRel(counts, indices, layout);
 }
 
+template <typename ExecSpace, typename IndexType>
```
kennyweiss, line 473: Please add a brief function comment (even though this won't get generated in the doxygen).

src/axom/multimat/multimat.cpp
```diff
@@ -382,16 +460,53 @@ void MultiMat::setCellMatRel(const vector<bool>& vecarr, DataLayout layout)
...
+void ScanRelationOffsetsRAJA(const axom::ArrayView<const IndexType> counts,
+                             const axom::ArrayView<IndexType> begins,
+                             const axom::ArrayView<IndexType> firstIndices)
```
kennyweiss, line 476: Note: The `firstIndices` view is marked as `const` but the underlying array is modified by this function

src/axom/multimat/multimat.cpp
```diff
@@ -799,65 +935,188 @@ bool MultiMat::removeEntry(int cell_id, int mat_id)
   return true;
 }
 
+#if defined(AXOM_USE_RAJA) && defined(AXOM_USE_UMPIRE)
+template <typename ExecSpace, typename IndexType>
+void TransposeRelationImplRAJA(const MultiMat::RelationSetType* oldRelationSet,
```
kennyweiss, line 940: Please add a brief function comment for the transpose functions too

src/axom/multimat/multimat.cpp
```diff
@@ -799,65 +935,188 @@ bool MultiMat::removeEntry(int cell_id, int mat_id)
...
-    }
-    newBeginVec[i] = newBeginVec[i - 1];
+    TransposeRelationImpl<axom::SEQ_EXEC>(oldRelationSet,
+                                          newBeginVec,
+                                          newIndicesVec,
+                                          newFirstIndicesVec,
+                                          m_flatMatToCellIndexMap,
+                                          m_flatCellToMatIndexMap,
+                                          m_slamAllocatorId);
   }
```
kennyweiss, line 1103: 👍

kennyweiss (approved):
This is really nice -- thanks for adding this @publixsubfan!

Please add a few brief function comments to clarify the expectations on the input/output sparse/dense and cell/material orderings, especially for the transpose functions.

src/axom/multimat/multimat.cpp
publixsubfan: `ArrayView<T>` is a shallow-const type, so `ArrayView<const IndexType>` is read-only while `ArrayView<IndexType>` is modifiable. The `const` in front of both is just to work around the GCC 8.1.0 issue with lambda-capture of `ArrayView`s.

PR #1105, opened 2023-06-01
Fix `ConnectivityArray` setting bad capacity in constructor.

gunney1:
# Fixed capacity set by `ConnectivityArray` constructors

- This PR is a bugfix
- It does the following:
  - Multiplies the tuple size (nodes per cell) with the number of tuples (cells) to get the capacity.  The number of tuples had been omitted.
  - Add this change to the `RELEASE-NOTES.md` file.
  - Updates some outdated doxygen comments.

The errors caused this behavior.  You want a capacity for N cells.  You get N/2 in 2D and N/3 in 3D.

The capacity set by the constructors are now correct and consistent with the capacity set using the `reserve` method.

There was probably some confusion because capacity meant different things to high-level `ConnectivityArray` vs low-level `Array` containers.  The `ConnectivityArray` only looks at the first dimension, because it considers the second dimension to be fixed even as the container grows.  `Array` looks at all dimensions.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -57,6 +57,9 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Quest: Adds ability to import volume fractions into `SamplingShaper` before processing `Klee` input
 
 ### Changed
+- Fixed bug in `mint::mesh::UnstructuredMesh` constructors, affecting capacity.
+  A mising factor was added.  If you worked around this by adding the factor yourself,
+  you may want to undo that work-around.
 - Updates blt submodule to HEAD of develop on 24Jan2023
 - Updates uberenv submodule to HEAD of main on 12May2023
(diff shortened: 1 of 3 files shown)


RELEASE-NOTES.md
```diff
@@ -57,6 +57,9 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Quest: Adds ability to import volume fractions into `SamplingShaper` before processing `Klee` input
 
 ### Changed
+- Fixed bug in `mint::mesh::UnstructuredMesh` constructors, affecting capacity.
+  A mising factor was added.  If you worked around this by adding the factor yourself,
```
agcapps, line 61: typo:
```suggestion
  A missing factor was added.  If you worked around this by adding the factor yourself,
```

agcapps (approved):
Thank you, @gunney1

kennyweiss (approved):
Great detective work!

white238 (approved):

PR #1115, opened 2023-06-07
Support replacement rules for preshaped geometry

### LINKED ISSUE #1104: In Klee, allow materials to participate in shaping and replacement rules without geometric data
Our Klee format currently supports input geometry in `stl` and `c2c` format.

We have recently added support for importing initial volume fractions into the sample-based shaper, but these shapes cannot participate in the replacement rules, which might lead to invalid results. (See #1069).

I'd like to add the ability for shapes/materials to participate in shaping without an input geometry.

My proposal is to allow a format of `none` so users can explicitly order the input materials and tag them with replacement rules.

e.g. note the `geometry` section in the below example
```yaml
dimensions: 2
units: cm

shapes:
  - name: input_rock
    material: rock
    geometry:
      format: none  # <-- new, has a material name, but no (new) geometric data

  - name: input_paper
    material: paper
    geometry:
      format: none  # <-- new
    does_not_replace: [scissor]  # <-- `paper` material will replace everything, including `rock`, but not `scissor`
```

I discussed this w/ @wnissen on 5/30/2023.

kennyweiss:
# Summary

- This PR is a feature related to quest's shaping application of klee input decks
- It allows "preshaped" materials (e.g. where the volume fractions are generated prior to invoking axom's shaping pipeline) to participate in replacement rules.
- It adds support in Klee for a shape to have no geometric data. This is specified with `geometry/format: none`. If so, it will not have a `geometry/path` entry.
- It also adds unit tests for quest's `SamplingShaper` class
- Resolves #1104 

#### TODO
- [x] Update `RELEASE-NOTES`
- Update klee/shaping files in `axom_data` repo, as necessary

_Update:_ I'll update the `axom_data` repo in a follow-up PR since it will involve non-trivial changes to the shaping example

### DIFF
--- a/src/axom/klee/tests/klee_io.cpp
+++ b/src/axom/klee/tests/klee_io.cpp
@@ -216,7 +216,7 @@ TEST(IOTest, readShapeSet_missingGeometryPath)
     }
     catch(const KleeError &err)
     {
-      FAIL() << "Should not have thrown";
+      FAIL() << "Should not have thrown. Error message: " << err.what();
     }
   }
 }
--- a/src/axom/klee/tests/klee_io.cpp
+++ b/src/axom/klee/tests/klee_io.cpp
@@ -173,24 +173,51 @@ TEST(IOTest, readShapeSet_missingMaterial)
 
 TEST(IOTest, readShapeSet_missingGeometryPath)
 {
-  auto input = R"(
-    dimensions: 2
-
-    shapes:
-      - name: wheel
-        material: steel
-        geometry:
-          format: test_format
-  )";
-
-  try
+  // Expect a validation error when 'geometry/path' is missing
+  // and 'geometry/format' is not "none"
   {
-    readShapeSetFromString(input);
-    FAIL() << "Should have thrown";
+    auto input = R"(
+      dimensions: 2
+
+      shapes:
+        - name: wheel
+          material: steel
+          geometry:
+            format: test_format
+    )";
+
+    try
+    {
+      readShapeSetFromString(input);
+      FAIL() << "Should have thrown";
+    }
+    catch(const KleeError &err)
+    {
+      EXPECT_THAT(err.what(), HasSubstr("Provided format"));
+    }
   }
-  catch(const KleeError &err)
+
+  // Valid when 'geometry/path' is missing and 'geometry/format' is none
   {
-    EXPECT_THAT(err.what(), HasSubstr("path"));
+    auto input = R"(
+      dimensions: 2
+
+      shapes:
+        - name: wheel
+          material: steel
+          geometry:
+            format: none
+    )";
+
+    try
+    {
+      readShapeSetFromString(input);
+      SUCCEED();
+    }
+    catch(const KleeError &err)
+    {
+      FAIL() << "Should not have thrown";
+    }
   }
 }
 
--- a/src/axom/quest/tests/quest_sampling_shaper.cpp
+++ b/src/axom/quest/tests/quest_sampling_shaper.cpp
@@ -255,8 +255,8 @@ public:
   {
     mfem::GridFunction* gf = m_dc.GetField(name);
 
-    mfem::LinearForm vol_form(gf->FESpace());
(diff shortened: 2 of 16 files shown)


src/axom/quest/tests/quest_sampling_shaper.cpp
```diff
@@ -0,0 +1,868 @@
...
+    std::ifstream ifs(m_filename.c_str(), std::ios::in);
+    std::stringstream buffer;
+    buffer << ifs.rdbuf();
+
+    return buffer.str();
+  }
+
+private:
+  const std::string m_filename;
+};
```
kennyweiss, line 78: This little utility class might be more generally useful in support of our internal testing

src/axom/quest/tests/quest_sampling_shaper.cpp
```diff
@@ -0,0 +1,868 @@
...
+  material: {1}
+  geometry:
+    format: none
+# Odd cells replace background void, but not left
+- name: odd_cells
+  material: {2}
+  geometry:
+    format: none
+  does_not_replace: [{1}]
+)";
```
kennyweiss, line 666: Note: We're using `geometry/format: none` to indicate that a "shape" in the list does not have geometric data, but that we want it to participate in the replacement rules.

Another possibility might be `geometry/format: preshaped` to indicate that an initial volume fraction has been supplied for the corresponding material. This terminology might be a bit more awkward (?), since even after a "shape" has been preshaped for a material, we can still supply further shapes for that material.

Thoughts and/or suggestions ?

src/axom/quest/tests/quest_sampling_shaper.cpp
```diff
@@ -0,0 +1,868 @@
...
+}
+
+TEST_F(SamplingShaperTest, disk_with_multiple_preshaped_materials)
+{
+  using Point2D = typename SamplingShaperTest::Point2D;
+
+  const std::string& testname =
+    ::testing::UnitTest::GetInstance()->current_test_info()->name();
+
+  const std::string shape_template = R"(
```
kennyweiss, line 735: This is a complicated example. I'll post some screenshots from VisIt to show what it does. See also below for a textual description.

rhornung67 (approved):

src/axom/quest/tests/quest_sampling_shaper.cpp
```diff
@@ -0,0 +1,868 @@
...
+
+      fes->GetElementDofs(idx, dofs);
+      vf->SetSubVector(dofs, res);
+    }
+  }
+
+  /// Helper to check integrated volume of a volume fraction grid fucntion
+  void checkExpectedVolumeFractions(const std::string& material_name,
+                                    double expected_volume,
+                                    double EPS = 1e-2)
```
agcapps, line 353: How did you determine this default value for `EPS`?

agcapps (approved):
Thanks, Kenny.  I appreciate the tests.

src/axom/quest/tests/quest_sampling_shaper.cpp
kennyweiss: For the testing, I'm using a really coarse mesh and relatively low linearization and quadrature order parameters, so I didn't expect the result to be that accurate compared to the analytic results (i.e. $\pi * (r^2_{outer} - r^2_{inner})$ for the annulus). I was only checking that the results are reasonably close.

src/axom/klee/tests/klee_io.cpp
```diff
@@ -173,24 +173,51 @@ TEST(IOTest, readShapeSet_missingMaterial)
...
+          geometry:
+            format: none
+    )";
+
+    try
+    {
+      readShapeSetFromString(input);
+      SUCCEED();
+    }
+    catch(const KleeError &err)
```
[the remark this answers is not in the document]
kennyweiss: Thanks -- I agree. 
I'll fix it.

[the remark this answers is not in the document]
kennyweiss: Done

PR #1124, opened 2023-06-23
Use srun flux wrapper path for toss 4 cray machines

bmhan12:
This PR:
- Changes unit tests to use srun flux wrapper instead of slurm srun.

This change will require modifying the spack recipe for axom to make it automated and not manual, which can be done in a follow-up PR (#1122 will modify the spack recipe)


kennyweiss (approved):

gunney1 (approved):
thank you @bmhan12 !

Issue #1117, opened 2023-06-09
Update automatic formatting to ensure one-liners are braced


kennyweiss:
We recently noticed that our automatic formatting (via `clang-format`) does not enforce braces around single-line control statements (e.g. `if`, `for`, ...).

Since braces are required by our coding guide, our automatic formatting should enforce this.

After a cursory glance, it looks like this feature was added to `clang-format` in clang@15 and we're currently using clang@10. The latest version of clang on LLNL's LC clusters is clang@14. It looks like we can use `clang-tidy` as a stopgap.

See: 
* Coding guideline:
> 9.1 Curly braces should be used in all conditionals, loops, etc. even when the content inside the braces is a “one-liner”.
https://axom.readthedocs.io/en/develop/docs/sphinx/coding_guide/sec09_format.html
* https://stackoverflow.com/questions/26111162/can-clang-format-add-braces-to-single-line-if-statements-etc

PR #1125, opened 2023-06-23
Uses clang-tidy to add braces around one line statements

kennyweiss:
# Summary

- This is a maintenance PR
- It uses `clang-tidy` and `clang-apply-replacements` to wrap one-line statements with curly brackets via the `readability-braces-around-statements` rule.
- It adds a simple clang-tidy config file to axom, which gets copied into our build directory
- It fixes https://github.com/LLNL/axom/issues/1117
  - This is a one time fix which we can apply sporadically, e.g. before releases
- Depends on BLT PR https://github.com/LLNL/blt/pull/644

#### Notes: 
* the changes from `clang-tidy` are not always safe and need to be verified before applying.
* Oddly enough, I had trouble getting this to work with a clang compiler on LLNL's LC, and had to use this with a GCC compiler. The issue related to incorrectly propagating flags for the underlying standard library.
* I locally added the following two CMake config variables to the toss3-gcc@8.3.1 host-config (but did not check these in):
```cmake
set(CLANGTIDY_EXECUTABLE "/usr/tce/packages/clang/clang-10.0.1/bin/clang-tidy" CACHE PATH "")
set(CLANGAPPLYREPLACEMENTS_EXECUTABLE "/usr/tce/packages/clang/clang-10.0.1/bin/clang-apply-replacements" CACHE PATH "")
```
* I didn't set up `clang-tidy` to format the code, so we have to run `clang-format` after applying `clang-tidy`

#### Discussion:
* This PR only applies the `readability-braces-around-statements` clang-tidy rule. Are there other rules that we want to apply, e.g. rules related to modernizing our C++ usage? 

#### TODO:
- [x] Update BLT submodule again after merging https://github.com/LLNL/blt/pull/644

### DIFF
--- a/src/axom/klee/tests/KleeMatchers.hpp
+++ b/src/axom/klee/tests/KleeMatchers.hpp
@@ -34,8 +34,7 @@ public:
     : m_mat(mat)
   { }
 
-  bool MatchAndExplain(const axom::numerics::Matrix<T>& other,
-                       std::ostream* /* listener */) const
+  bool MatchAndExplain(const axom::numerics::Matrix<T>& other, std::ostream*) const
   {
     if(other.getNumRows() != m_mat.getNumRows() ||
        other.getNumColumns() != m_mat.getNumColumns())
@@ -59,8 +58,8 @@ public:
     return true;
   }
 
-  void DescribeTo(std::ostream* os) const { }
-  void DescribeNegationTo(std::ostream* os) const { }
+  void DescribeTo(std::ostream*) const { }
+  void DescribeNegationTo(std::ostream*) const { }
 
 private:
   const axom::numerics::Matrix<T> m_mat;
@@ -85,7 +84,7 @@ public:
 
   explicit AlmostEqArrMatcher(const T& arr) : m_arr(arr) { }
 
-  bool MatchAndExplain(const T& other, std::ostream* /* listener */) const
+  bool MatchAndExplain(const T& other, std::ostream*) const
   {
     for(int i = 0; i < m_arr.dimension(); ++i)
     {
@@ -97,8 +96,8 @@ public:
     return true;
   }
 
-  void DescribeTo(std::ostream* os) const { }
-  void DescribeNegationTo(std::ostream* os) const { }
+  void DescribeTo(std::ostream*) const { }
+  void DescribeNegationTo(std::ostream*) const { }
 
 private:
   const T m_arr;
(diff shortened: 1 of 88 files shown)


src/axom/primal/operators/detail/winding_number_impl.hpp
```diff
@@ -109,11 +114,21 @@ double convex_endpoint_winding_number(const Point<T, 2>& q,
...
   Vector<T, 2> V1(q, c[idx]);
 
   for(idx = ord; idx >= 0; --idx)
-    if(squared_distance(q, c[idx]) > edge_tol_sq) break;
+  {
+    if(squared_distance(q, c[idx]) > edge_tol_sq)
+    {
+      break;
+    }
+  }
```
kennyweiss, line 131: Note: There is an bug in the `clang-tidy` when adding braces around nested if/for/while statements where it will add an opening brace for each item, but will only add a single closing brace. I had to manually add additional closing braces.

I was using `clang-tidy@10.0.1` -- not sure if this has been fixed in later versions.

src/axom/klee/tests/KleeMatchers.hpp
```diff
@@ -85,7 +84,7 @@ class AlmostEqArrMatcher
 
   explicit AlmostEqArrMatcher(const T& arr) : m_arr(arr) { }
 
-  bool MatchAndExplain(const T& other, std::ostream* /* listener */) const
```
bmhan12, line 88: The changes in this file are related to removing unused parameter names.
Just wanted to highlight the changes here are different from the other files, make sure the style changes in this file are intended.

kennyweiss: Thanks Brian -- I manually made those changes in a separate commit (https://github.com/LLNL/axom/pull/1125/commits/7a53c7f9ce7bb9187f0a57ee106eeed4d35ea341) to fix some warnings that I introduced in my previous PR (https://github.com/LLNL/axom/pull/1115).  I should have called it out explicitly with a PR comment.

white238 (approved):

rhornung67 (approved):
Looks good. Do we need a particular version (or min version) of clang to use this?

kennyweiss:
> Looks good. Do we need a particular version (or min version) of clang to use this?

I used `clang@10.0.1` after running into a problem w/ `clang@10.0.0`. 
I'm not sure if the latter could have been resolved without bumping to 10.0.1.
I posted more details about my setup in the PR description above.

PR #1144, opened 2023-07-21
Fix `MFEMSidreDataCollection` finite element space bug

jamiebramwell:
This short PR fixes two bugs when reading an `mfem::GridFunction` from an `MFEMSidreDataCollection`.

1) It constructs a new `mfem::FiniteElementSpace` when the vdim is different from what is currently stored in Sidre.
2) It correctly determines whether the stored grid function is ordered by nodes or VDIM.

### DIFF
--- a/src/axom/sidre/core/MFEMSidreDataCollection.cpp
+++ b/src/axom/sidre/core/MFEMSidreDataCollection.cpp
@@ -2636,7 +2636,7 @@ void MFEMSidreDataCollection::reconstructField(Group* field_grp)
     }
 
     // We cache the FESpaces to avoid reconstructing them when not needed
-    // An FESpace is uniquely identified by the basis of its FEColl, its ordering, and its vdim number
+    // An FESpace is uniquely identified by the basis of its FEColl, ordering, and vdim number
     const std::string fespace_id =
       axom::fmt::format("{0}_{1}_{2}",
                         basis_name,
--- a/src/axom/sidre/core/MFEMSidreDataCollection.cpp
+++ b/src/axom/sidre/core/MFEMSidreDataCollection.cpp
@@ -2640,7 +2640,7 @@ void MFEMSidreDataCollection::reconstructField(Group* field_grp)
     const std::string fespace_id =
       axom::fmt::format("{0}_{1}_{2}",
                         basis_name,
-                        (ordering == mfem::Ordering::byVDIM) ? "vdim" : "nodes", 
+                        (ordering == mfem::Ordering::byVDIM) ? "vdim" : "nodes",
                         vdim);
 
     // Only need to create a new FESpace if one doesn't already exist
--- a/src/axom/sidre/core/MFEMSidreDataCollection.cpp
+++ b/src/axom/sidre/core/MFEMSidreDataCollection.cpp
@@ -2636,11 +2636,11 @@ void MFEMSidreDataCollection::reconstructField(Group* field_grp)
     }
 
     // We cache the FESpaces to avoid reconstructing them when not needed
-    // An FESpace is uniquely identified by the basis of its FEColl and its ordering
+    // An FESpace is uniquely identified by the basis of its FEColl, its ordering, and its vdim number
     const std::string fespace_id =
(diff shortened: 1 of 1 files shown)


src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -2636,11 +2636,11 @@ void MFEMSidreDataCollection::reconstructField(Group* field_grp)
     }
 
     // We cache the FESpaces to avoid reconstructing them when not needed
-    // An FESpace is uniquely identified by the basis of its FEColl and its ordering
+    // An FESpace is uniquely identified by the basis of its FEColl, its ordering, and its vdim number
```
jamiebramwell, line 2639:
```suggestion
    // An FESpace is uniquely identified by the basis of its FEColl, ordering, and vdim number
```

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -2636,11 +2636,11 @@ void MFEMSidreDataCollection::reconstructField(Group* field_grp)
...
 
     // We cache the FESpaces to avoid reconstructing them when not needed
-    // An FESpace is uniquely identified by the basis of its FEColl and its ordering
+    // An FESpace is uniquely identified by the basis of its FEColl, ordering, and vdim number
     const std::string fespace_id =
-      axom::fmt::format("{0}_{1}",
+      axom::fmt::format("{0}_{1}_{2}",
                         basis_name,
-                        (ordering == mfem::Ordering::byVDIM) ? "nodes" : "vdim");
+                        (ordering == mfem::Ordering::byVDIM) ? "vdim" : "nodes", vdim);
```
kennyweiss, line 2643: I think you need to run `make style` in a config w/ clang@10

kennyweiss (approved):
Thanks @jamiebramwell !

Should we add some tests to distinguish between cases w/ different vdims?

src/axom/sidre/core/MFEMSidreDataCollection.cpp
kennyweiss, line 2643: Also -- good catch w/ the swapped vdim/nodes!

jamiebramwell:
> Thanks @jamiebramwell !
> 
> Should we add some tests to distinguish between cases w/ different vdims?

Yes - I haven't dug into how you're testing this yet. I just wanted to make sure this fix made sense.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
kennyweiss: As a shortcut -- this is what it wants to see:

```diff
     const std::string fespace_id =
       axom::fmt::format("{0}_{1}_{2}",
                         basis_name,
-                        (ordering == mfem::Ordering::byVDIM) ? "vdim" : "nodes", vdim);
+                        (ordering == mfem::Ordering::byVDIM) ? "vdim" : "nodes",
+                        vdim);
```
https://dev.azure.com/healy22/Status%20Check%20Testiing/_build/results?buildId=8442&view=logs&j=cb13b0c9-f406-547a-d755-9fe006f85343&t=435ca3cd-a81f-530c-87da-e79a43d1b3d6&l=511

kennyweiss:
> Yes - I haven't dug into how you're testing this yet. I just wanted to make sure this fix made sense.

I think so. Seems like it was doing the wrong thing before and the right thing now, so it's a step in the right direction.

We have some "sidre data collection" tests here:
* https://github.com/LLNL/axom/blob/develop/src/axom/sidre/tests/sidre_mfem_datacollection.cpp

and a few in the examples:
* https://github.com/LLNL/axom/blob/develop/src/axom/sidre/examples/sidre_mfem_datacollection_restart.cpp
* https://github.com/LLNL/axom/blob/develop/src/axom/sidre/examples/sidre_mfem_datacollection_materials.cpp
* https://github.com/LLNL/axom/blob/develop/src/axom/sidre/examples/sidre_mfem_datacollection_vis.cpp

white238 (approved):

PR #1152, opened 2023-08-04
Fill _IMPORT_ROOT in the CMake target.

agcapps:
# Summary

- This PR is a bugfix to the CMake target.  It defines a variable to allow us to refer to files with relative paths.

### DIFF
--- a/src/cmake/axom-config.cmake.in
+++ b/src/cmake/axom-config.cmake.in
@@ -156,6 +156,15 @@ if(NOT AXOM_FOUND)
   # Include targets exported by cmake
   #----------------------------------------------------------------------------
   get_filename_component(AXOM_CMAKE_CONFIG_DIR "${CMAKE_CURRENT_LIST_FILE}" PATH)
+  get_filename_component(_IMPORT_PREFIX "${AXOM_CMAKE_CONFIG_DIR}" PATH)
+
+  if(_IMPORT_PREFIX STREQUAL "/")
+    set(_IMPORT_PREFIX "")
+  endif()
+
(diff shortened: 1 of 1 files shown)


cyrush:
Can we use `_IMPORT_PREFIX` for the final var?

I have a script that identifies that as a special case to know relative imports are supported by a packages cmake logic?

cyrush:
I guess, let's discuss to select the right final plan.

It depends on how `_IMPORT_ROOT` and `_IMPORT_PREFIX` are used in other parts the cmake logic.

white238:
> Can we use `_IMPORT_PREFIX` for the final var?
> 
> I have a script that identifies that as a special case to know relative imports are supported by a packages cmake logic?



> I guess, let's discuss to select the right final plan.
> 
> It depends on how `_IMPORT_ROOT` and `_IMPORT_PREFIX` are used in other parts the cmake logic.

This comes from me copying/pasting from Conduit and not copying the whole solution. I am fine with whatever you want but this fixes the currently broken build in a downstream user. I am not sure why it works sometimes honestly. It seems quite broken.

This is how Conduit uses the variable though.

white238 (approved):

white238:
Made an issue to track this:  #1153

cyrush:
I think we are good. 

I panicked b/c there are two parts of the puzzle, and I confused the two.

The generated cmake exported targets use `_IMPORT_PREFIX`, we should be fine to use `_IMPORT_ROOT`, but I would even question if both are necessary.

rhornung67 (approved):

PR #1176, opened 2023-09-01
Add deep copy methods for Group and View

nselliott:
# Summary

- This PR is a feature
  - Adds deep copy methods for Group and View
  - Addresses #769 

The actual deep copy of buffer data occurs on the Views, with the deep copy of a Group copying the entire hierarchy with deep copies of all Views.

All deep copies of Views with Buffer or external data are copied into a new sidre Buffer attached to the new View.  If the source View's description sees only a part of its Buffer or external array due to offsets or strides, only the elements that are seen by the description will be copied, and the Buffer for the new View will be sized to receive the exact number of elements that are copied.

### DIFF
--- a/src/axom/sidre/core/Group.cpp
+++ b/src/axom/sidre/core/Group.cpp
@@ -966,6 +966,39 @@ View* Group::copyView(View* view)
   return copy;
 }
 
+/*
+ *************************************************************************
+ *
+ * Create a deep copy of given View and attach to this Group.
+ *
+ * The deep copy performs a copy of all described data.
+ *
+ *************************************************************************
+ */
+View* Group::deepCopyView(View* view)
+{
+  if(view == nullptr || hasChildView(view->getName()))
+  {
+    SLIC_CHECK_MSG(
+      view != nullptr,
+      SIDRE_GROUP_LOG_PREPEND << "Null pointer provided, no View to copy.");
+
+    if(view != nullptr)
+    {
+      SLIC_CHECK_MSG(!hasChildView(view->getName()),
+                     SIDRE_GROUP_LOG_PREPEND
+                       << "Group already has a View named '" << view->getName()
+                       << "' so View copy operation cannot happen");
+    }
+
+    return nullptr;
+  }
+
+  View* copy = createView(view->getName());
+  view->deepCopyView(copy);
+  return copy;
+}
+
 ////////////////////////////////////////////////////////////////////////
 //
 // Child Group query methods.
@@ -1367,6 +1400,56 @@ Group* Group::copyGroup(Group* group)
   while(indexIsValid(vidx))
   {
     res->copyView(group->getView(vidx));
+    vidx = group->getNextValidViewIndex(vidx);
+  }
+
+  return res;
+}
+
+/*
+ *************************************************************************
+ *
+ * Create a deep copy of given Group and make it a child of this Group.
+ *
+ * The deep copy of a Group will copy the group hierarchy and deep copy
+ * all Views within the hierarchy.
+ *
+ *************************************************************************
+ */
+Group* Group::deepCopyGroup(Group* group)
+{
+  if(group == nullptr || hasChildGroup(group->getName()))
+  {
+    SLIC_CHECK_MSG(
+      group != nullptr,
+      SIDRE_GROUP_LOG_PREPEND << "Null pointer provided, no Group to copy.");
+
+    if(group != nullptr)
+    {
+      SLIC_CHECK_MSG(!hasChildGroup(group->getName()),
+                     SIDRE_GROUP_LOG_PREPEND
+                       << "Invalid copy operation. Group already has "
+                       << "a child named '" << group->getName() << "'.");
+    }
+
+    return nullptr;
+  }
+
+  Group* res = createGroup(group->getName());
+
+  // copy child Groups to new Group
+  IndexType gidx = group->getFirstValidGroupIndex();
+  while(indexIsValid(gidx))
+  {
+    res->deepCopyGroup(group->getGroup(gidx));
+    gidx = group->getNextValidGroupIndex(gidx);
+  }
+
+  // copy Views to new Group
+  IndexType vidx = group->getFirstValidViewIndex();
+  while(indexIsValid(vidx))
+  {
+    res->deepCopyView(group->getView(vidx));
     vidx = group->getNextValidViewIndex(vidx);
   }
 
--- a/src/axom/sidre/core/Group.cpp
+++ b/src/axom/sidre/core/Group.cpp
@@ -975,8 +975,10 @@ View* Group::copyView(View* view)
  *
  *************************************************************************
  */
-View* Group::deepCopyView(View* view)
+View* Group::deepCopyView(View* view, int allocID)
 {
+  allocID = getValidAllocatorID(allocID);
+
   if(view == nullptr || hasChildView(view->getName()))
   {
     SLIC_CHECK_MSG(
@@ -995,7 +997,7 @@ View* Group::deepCopyView(View* view)
   }
 
   View* copy = createView(view->getName());
-  view->deepCopyView(copy);
+  view->deepCopyView(copy, allocID);
   return copy;
 }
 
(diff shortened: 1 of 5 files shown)


src/axom/sidre/core/Group.hpp
```diff
@@ -1242,10 +1262,29 @@ class Group
...
-   * is not moved into this Group.
+   * is not copied into this Group.
    */
   Group* copyGroup(Group* group);
 
+  /*!
+   * \brief Create a deep copy of Group hierarchy rooted at given Group and
+   *        make it a child of this Group.
+   *
+   * Note that all Views in the Group hierarchy are deep copied as well.
```
rhornung67, line 1273: Should we add a default bool arg to this; e.g., copy_child_groups = true?  This would have the same interface, but would allow users more control if they needed it.

rhornung67 (commented):
Will there be tests added for this?

src/axom/sidre/core/Group.hpp
```diff
@@ -896,14 +896,34 @@ class Group
...
+   * strides and or offsets into the data, only the values seen by the 
+   * description will be copied into the new View. The new View will have
+   * a Buffer that is the exact size of the number of values that are copied,
+   * with an offset of zero and a stride of one.
+   *
+   * If given View pointer is null or Group already has a View with
+   * same name as given View, method is a no-op.
+   *
+   * \return pointer to given argument Group object or nullptr if View
+   * is not copied into this Group.
```
agcapps, line 923: Returns a pointer to a View object (same fix as lines 902 and 903).

That said, it seems odd that `copyView` and `copyGroup` and their `deepCopy` variants return the pointer to the object given.  I think they should return the pointer to the newly created Group or View.  Perhaps that's a deeper design discussion, though.

rhornung67: That's an interesting point. Yes, we should discuss the implications of this choice.

src/axom/sidre/core/View.hpp
```diff
@@ -1370,6 +1370,18 @@ class View
    */
   void copyView(View* copy) const;
 
+  /*!
+   * \brief Deep copy view contents into an undescribed EMPTY view.
```
agcapps, line 1374: Please state "contents of this view".  The explicit "this view" makes things clearer as a user glances through the docs.

src/axom/sidre/core/View.cpp
```diff
@@ -1037,6 +1037,103 @@ void View::copyView(View* copy) const
...
+ *
+ *************************************************************************
+ */
+void View::deepCopyView(View* copy) const
+{
+  SLIC_ASSERT_MSG(copy->m_state == EMPTY && !copy->isDescribed(),
+                  SIDRE_VIEW_LOG_PREPEND
+                    << "deepCopyView can only copy into undescribed view "
+                    << "with empty state.");
+#if 1
```
agcapps, line 1053: why?

src/axom/sidre/core/Group.hpp
agcapps: Looking at Group.cpp (new) lines 966, 999, 1406, 1456, it appears we always return the new copy.

agcapps (approved):

nselliott:
> Will there be tests added for this?

There is a test in this PR in the sidre_group.cpp file.

src/axom/sidre/core/Group.hpp
nselliott: I tried to be consistent with the API used in the existing `copyGroup()` method.  A Group is a hierarchical container containing Groups and Views, and I think the expected behavior of copying a container is that you copy all of its contents.

rhornung67 (approved):

rhornung67:
> > Will there be tests added for this?
> 
> There is a test in this PR in the sidre_group.cpp file.

D'oh! You are correct, sir.

src/axom/sidre/core/Group.hpp
nselliott: Yes, the returned objects are the new copied objects, so I need to correct this text.

src/axom/sidre/core/View.cpp
nselliott: Oops, will remove.

kennyweiss (approved):
Thanks @nselliott !

I had some comments about improving the documentation.

We might also want to be more explicit about how the deep copies are removing any striding and offsetting, and perhaps think about if/how we can provide an option to retain the striding.

src/axom/sidre/core/Group.hpp
```diff
@@ -896,14 +896,34 @@ class Group
...
+   * \brief Create a deep copy of given View object and add it to this Group.
+   *
+   * Note that for Views that use sidre Buffers or external data, the
+   * deep copy performs a copy of the data described by the View
+   * into a new Buffer attached to the destination View. If the source View
+   * describes only part of a Buffer or part of an external array, due to
+   * strides and or offsets into the data, only the values seen by the 
+   * description will be copied into the new View. The new View will have
+   * a Buffer that is the exact size of the number of values that are copied,
+   * with an offset of zero and a stride of one.
```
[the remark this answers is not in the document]
nselliott: I think we'd need to define what we want for that case.  If we deep copy a strided View and give it a new Buffer with the same strides, then that Buffer will have the correct copied values in the strided locations as described by the View, and undefined data in the parts of the Buffer outside the View's description.  Is that something we want to do for the deep copy of a single View?

If your example is a single Buffer with three strided Views `x, y, z` attached to the Buffer, you may desire a use case where you deep copy the full Buffer and then create three new Views to attach to the new Buffer, but that isn't what I was attempting to implement here.  Based on previous discussions, the decision was to create a new Buffer for each deep copy of a View.

The tools to do that kind of single Buffer, multiple View copy are available now, but you'd have to do it step-by-step as there isn't a convenient single-call method for it.

src/axom/sidre/core/Group.hpp
```diff
@@ -1241,11 +1261,30 @@ class Group
...
+  /*!
+   * \brief Create a deep copy of Group hierarchy rooted at given Group and
+   *        make it a child of this Group.
+   *
+   * Note that all Views in the Group hierarchy are deep copied as well.
+   *
+   * The deep copy of the Group creates a duplicate of the entire Group
+   * hierarchy and performs a deep copy of the data described by the Views
+   * in the hierarchy. The Views in the new Group hierarchy will all allocate
+   * and use new Buffers to hold their copied data.
```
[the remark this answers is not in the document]
nselliott: I think this is a good suggestion, as this is available for the original allocation, it should be available for the allocation of the copies.

src/axom/sidre/core/Group.cpp
```diff
@@ -1373,6 +1406,56 @@ Group* Group::copyGroup(Group* group)
...
+    gidx = group->getNextValidGroupIndex(gidx);
+  }
+
+  // copy Views to new Group
+  IndexType vidx = group->getFirstValidViewIndex();
+  while(indexIsValid(vidx))
+  {
+    res->deepCopyView(group->getView(vidx));
+    vidx = group->getNextValidViewIndex(vidx);
+  }
```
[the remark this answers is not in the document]
nselliott: This was only because it was cut-and-paste from old code elsewhere in the class.  I'll update it here and in the old methods.

src/axom/sidre/core/Group.hpp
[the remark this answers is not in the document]
cyrush: Keeping strides and offsets for the general case can be very inefficient. 

For example, some codes allocate a single buffer, with many views that index into for all of their mesh and field data. 

If you try to deep copy a subgroup (say one field) and preserve the original striding, instead of copying just the data described by views in the group -- you will copy everything in the buffer (double mem usage) and most of it would be unused. 

For the specific case where you know you need an interleaved dataset, you may want a helper that gives you copies with those guarantees.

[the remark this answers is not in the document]
kennyweiss: I agree that this is an "advanced" use-case and does not need to be directly supported by Sidre. We might consider adding a similar case if/when we create a Sidre "cookbook".

In the meantime, it would probably suffice to just note this in the `deepCopyView` and `deepCopyGroup` functions.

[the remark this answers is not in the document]
kennyweiss: @nselliott (and others) -- would it make sense to add a function to `sidre::Buffer` to copy the data from a buffer to a new buffer and return a pointer to the new buffer -- e.g. `Buffer::copyBuffer() -> Buffer*` ?

On the other hand, perhaps this is intentional since the copy that I was proposing would be for the data but not its attached views?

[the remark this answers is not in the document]
nselliott: @kennyweiss We don't have that in Buffer because all Buffers must be owned by the DataStore, so a `DataStore::copyBuffer()` would be more appropriate.  This is another thing you could do now with two or three lines of code, and maybe a method would be useful, but it would have no Views.  You could do it and then manually attach Views according to the specific end result you desire.

PR #1199, opened 2023-10-10
Blueprint mesh topology vs coordset

gunney1:
# Prefer blueprint mesh topology over coordset

- This PR is a refactoring and guard against user error
- It does the following:
  - Refactors the distributed closest point test code to clarify that coordset is not independent of topology when reading in mesh.
  - Issues a warning if the user accidentally specifies a coordset when we want a topology.  This protects against codes that have not been updated since we switched to using topology.  (The reason for the change is that ghost layer info resides in topologies and not available when using coordsets.)


src/axom/quest/DistributedClosestPoint.hpp
```diff
@@ -1778,6 +1779,30 @@ class DistributedClosestPoint
...
+      {
+        auto errMsg = fmt::format("No such topology '{}' found.", topologyName);
+        if(domain.has_path("coordsets/" + topologyName))
+        {
+          errMsg += fmt::format(
+            "  You may have mistakenly specified a coordset name."
+            "  The interface has changed to use topology name"
+            " instead of coordset.");
+        }
+        SLIC_ERROR(errMsg);
```
kennyweiss, line 1801: Would it be helpful to list the domain ID or path to the `meshNode` to give users a better sense of where things went wrong?

kennyweiss (approved):
Thanks @gunney1 !

src/axom/quest/DistributedClosestPoint.hpp
gunney1: I don't think it's that helpful in practice, because the domains very typically have the same topology and coordset, even though blueprint allows for differences.

PR #1207, opened 2023-10-31
Build configs with +scr

bmhan12:
This PR:

- Adds back +scr to some toss4 configs (previous scr built only with some toss3 configs)
- ~~Edits `build_devtools` script to allow non-atk users to build to shared directory~~

### DIFF
--- a/scripts/llnl_scripts/build_devtools.py
+++ b/scripts/llnl_scripts/build_devtools.py
@@ -50,6 +50,9 @@ def main():
         if not os.path.exists(build_dir):
             os.makedirs(build_dir)
     else:
+        if getpass.getuser() != "atk":
+            print("ERROR: Only shared user 'atk' can install into shared directory. Use -d option.")
+            return 1
         build_dir = get_shared_devtool_dir()
     build_dir = os.path.abspath(build_dir)
 
--- a/scripts/llnl_scripts/build_devtools.py
+++ b/scripts/llnl_scripts/build_devtools.py
@@ -50,9 +50,6 @@ def main():
         if not os.path.exists(build_dir):
             os.makedirs(build_dir)
     else:
-        if getpass.getuser() != "atk":
-            print("ERROR: Only shared user 'atk' can install into shared directory. Use -d option.")
-            return 1
         build_dir = get_shared_devtool_dir()
     build_dir = os.path.abspath(build_dir)
 
--- a/scripts/spack/specs.json
+++ b/scripts/spack/specs.json
@@ -21,8 +21,8 @@
       "intel@19.0.4~openmp+devtools+mfem+c2c" ],
 
     "toss_4_x86_64_ib":
-    [ "gcc@10.3.1+devtools+hdf5+mfem+c2c",
-      "clang@14.0.6+devtools+hdf5+mfem+c2c",
+    [ "gcc@10.3.1+devtools+hdf5+mfem+c2c+scr",
+      "clang@14.0.6+devtools+hdf5+mfem+c2c+scr",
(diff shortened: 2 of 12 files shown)


src/cmake/thirdparty/SetupAxomThirdParty.cmake
```diff
@@ -234,6 +234,7 @@ if (SCR_DIR)
                         TREAT_INCLUDES_AS_SYSTEM ON
                         EXPORTABLE ON)
     blt_list_append(TO TPL_DEPS ELEMENTS scr)
+    set(SCR_FOUND ON CACHE BOOL "")
```
bmhan12, line 237: CMake variable is necessary to build `spio_IO_SCR_Checkpoint` test and example: https://github.com/LLNL/axom/blob/c35431a1ee71a7c46fe54805e8aa5278120d613e/src/axom/sidre/examples/spio/CMakeLists.txt#L33-L48

scripts/llnl_scripts/build_devtools.py
```diff
@@ -50,9 +50,6 @@ def main():
         if not os.path.exists(build_dir):
             os.makedirs(build_dir)
     else:
-        if getpass.getuser() != "atk":
-            print("ERROR: Only shared user 'atk' can install into shared directory. Use -d option.")
-            return 1
```
bmhan12, line 55: Removing this check expands who can build devtools script to the default shared directory (those with access to the shared directory). 

Are there downsides to removing this?

scripts/spack/specs.json
```diff
@@ -21,8 +21,8 @@
       "intel@19.0.4~openmp+devtools+mfem+c2c" ],
 
     "toss_4_x86_64_ib":
-    [ "gcc@10.3.1+devtools+hdf5+mfem+c2c",
-      "clang@14.0.6+devtools+hdf5+mfem+c2c",
+    [ "gcc@10.3.1+devtools+hdf5+mfem+c2c+scr",
+      "clang@14.0.6+devtools+hdf5+mfem+c2c+scr",
       "intel@2022.1.0+devtools+hdf5+mfem+c2c" ],
```
bmhan12, line 26: `+scr` was added based on the toss3 specs: https://github.com/LLNL/axom/blob/c35431a1ee71a7c46fe54805e8aa5278120d613e/scripts/spack/specs.json#L16-L21

On another note, are there any machines still on toss3? Should we keep or deprecate these toss 3 specs?

scripts/llnl_scripts/build_devtools.py
kennyweiss: Tag: @white238

kennyweiss (approved):
Thanks @bmhan12 !

scripts/llnl_scripts/build_devtools.py
white238: This guards us against having directories up there that have the wrong permissions and we have to contact LC to remove them

scripts/spack/specs.json
white238: There are no longer toss3 machines.

scripts/llnl_scripts/build_devtools.py
bmhan12: > This guards us against having directories up there that have the wrong permissions and we have to contact LC to remove them

Oh, I remember the issue now. I'll revert this change.

scripts/spack/specs.json
bmhan12: Toss 3 cleanup issue opened here: https://github.com/LLNL/axom/issues/1208

rhornung67 (approved):

nselliott (approved):

PR #1215, opened 2023-11-08
Add clip operator for a tet and plane

adayton1:
# Summary

- This PR is a feature
- It does the following:
  - Adds a clip operator for a tet and plane
  - Refactors internal code for clipping a polyhedron by a set of planes

### DIFF
--- a/src/axom/primal/operators/clip.hpp
+++ b/src/axom/primal/operators/clip.hpp
@@ -304,6 +304,82 @@ AXOM_HOST_DEVICE Polyhedron<T, 3> clip(const Tetrahedron<T, 3>& tet1,
   return detail::clipTetrahedron(tet1, tet2, eps, checkSign);
 }
 
+/*!
+ * \brief Clips a 3D tetrahedron against a plane in 3D, returning
+ *        the geometric intersection as a polyhedron
+ *
+ *  This function clips the tetrahedron against the given plane. This
+ *  involves finding new vertices at the intersection of the polyhedron
+ *  edges and the plane, removing vertices from the polyhedron that are
+ *  below the plane, and redefining the neighbors for each vertex (a
+ *  vertex is a neighbor of another vertex if there is an edge between
+ *  them).
+ *
+ * \param [in] tet The tetrahedron to clip
+ * \param [in] plane The plane used to clip the tetrahedron
+ * \param [in] eps The tolerance used when computing on which side of the
+ *             plane a point lies
+ * \param [in] checkSign If true (default is false), checks if the signed
+ *             volume of the tetrahedron is positive. If the signed volume
+ *             of the tetrahedron is negative, the order of some vertices
+ *             will be swapped in order to obtain a positive signed volume.
+ *
+ * \return The polyhedron obtained from clipping the tetrahedron against
+ *         the plane.
+ *
+ * \note Function is based off clipPolyhedron() in Mike Owen's PolyClipper.
+ *
+ * \note checkSign flag does not guarantee the shapes' vertex orders
+ *       will be valid. It is the responsiblity of the caller to pass
+ *       shapes with a valid vertex order.
+ */
+template <typename T>
+AXOM_HOST_DEVICE Polyhedron<T, 3> clip(const Tetrahedron<T, 3>& tet,
+                                       const Plane<T, 3>& plane,
+                                       double eps = 1.e-10,
+                                       bool checkSign = false)
+{
+  return detail::clipTetrahedron(tet, plane, eps, checkSign);
+}
+
+/*!
+ * \brief Clips a 3D tetrahedron against a plane in 3D, returning
+ *        the geometric intersection as a polyhedron
+ *
+ *  This function clips the tetrahedron against the given plane. This
+ *  involves finding new vertices at the intersection of the polyhedron
+ *  edges and the plane, removing vertices from the polyhedron that are
+ *  below the plane, and redefining the neighbors for each vertex (a
+ *  vertex is a neighbor of another vertex if there is an edge between
+ *  them).
+ *
+ * \param [in] plane The plane used to clip the tetrahedron
+ * \param [in] tet The tetrahedron to clip
+ * \param [in] eps The tolerance used when computing on which side of the
+ *             plane a point lies
+ * \param [in] checkSign If true (default is false), checks if the signed
+ *             volume of the tetrahedron is positive. If the signed volume
+ *             of the tetrahedron is negative, the order of some vertices
+ *             will be swapped in order to obtain a positive signed volume.
+ *
+ * \return The polyhedron obtained from clipping the tetrahedron against
+ *         the plane.
+ *
+ * \note Function is based off clipPolyhedron() in Mike Owen's PolyClipper.
+ *
+ * \note checkSign flag does not guarantee the shapes' vertex orders
+ *       will be valid. It is the responsiblity of the caller to pass
+ *       shapes with a valid vertex order.
+ */
+template <typename T>
+AXOM_HOST_DEVICE Polyhedron<T, 3> clip(const Plane<T, 3>& plane,
+                                       const Tetrahedron<T, 3>& tet,
+                                       double eps = 1.e-10,
+                                       bool checkSign = false)
+{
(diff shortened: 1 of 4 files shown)


rhornung67 (approved):

src/axom/primal/operators/detail/clip_impl.hpp
```diff
@@ -400,85 +400,95 @@ AXOM_HOST_DEVICE void poly_clip_reindex(Polyhedron<T, NDIMS>& poly,
...
- * \brief Finds the clipped intersection Polyhedron between Polyhedron
- *        poly and an array of Planes.
- *
- * \param [in] poly The polyhedron
- * \param [in] planes The array of planes
- * \param [in] eps The tolerance for plane point orientation.
+ * \brief Finds the polyhedron formed from clipping a polyhedron with a plane
  *
- * \return The Polyhedron formed from clipping the polyhedron with a set of planes.
+ * \param [in] poly The polyhedron to clip
```
kennyweiss, line 405: It looks like `poly` is an `[inout]` parameter since the implementation modifies it.
Please update the doxycomment.

Could this function instead return a bool indicating if the intersection is non-empty, and use the `[inout] poly` parameter without the extra copy for the return type?

If `clipPolyhedron` is used elsewhere, we'll have to modify those calls as well. The good news is that its an internal/detail function, so it's not meant to be used outside our internal implementation.

src/axom/primal/operators/clip.hpp
```diff
@@ -304,6 +304,82 @@ AXOM_HOST_DEVICE Polyhedron<T, 3> clip(const Tetrahedron<T, 3>& tet1,
...
+/*!
+ * \brief Clips a 3D tetrahedron against a plane in 3D, returning
+ *        the geometric intersection as a polyhedron
+ *
+ *  This function clips the tetrahedron against the given plane. This
+ *  involves finding new vertices at the intersection of the polyhedron
+ *  edges and the plane, removing vertices from the polyhedron that are
+ *  below the plane, and redefining the neighbors for each vertex (a
+ *  vertex is a neighbor of another vertex if there is an edge between
+ *  them).
```
kennyweiss, line 316: @adayton1 -- it might be clearer to say that this clips a 3D tetrahedron against the half-space defined by a plane

src/axom/primal/tests/primal_clip.cpp
```diff
@@ -1294,6 +1294,179 @@ TEST(primal_clip, tet_tet_clip_special_case_1)
...
+  TetrahedronType tet(PointType {0.0, 0.0, 0.0},
+                      PointType {1.0, 0.0, 0.0},
+                      PointType {0.0, 1.0, 0.0},
+                      PointType {0.0, 0.0, 1.0});
+
+  PlaneType plane(VectorType {0.0, 0.0, 1.0}, PointType {0.0, 0.0, -0.5});
+
+  PolyhedronType poly = axom::primal::clip(tet, plane, EPS, CHECK_SIGN);
+
+  EXPECT_NEAR(tet.volume(), poly.volume(), EPS);
```
kennyweiss, line 1332: Minor: should we compare `signedVolume()` just to make sure the orientations are preserved?

I don't see any reason it wouldn't be, but it can't hurt.

src/axom/primal/tests/primal_clip.cpp
```diff
@@ -1294,6 +1294,179 @@ TEST(primal_clip, tet_tet_clip_special_case_1)
...
+                      PointType {0.0, 0.0, 1.0});
+
+  PlaneType plane(VectorType {-1.0, 0.0, 1.0}, 0.0);
+
+  PolyhedronType poly = axom::primal::clip(plane, tet, EPS, CHECK_SIGN);
+
+  EXPECT_NEAR(tet.volume()/2.0, poly.volume(), EPS);
+}
+
+// TODO: Add a test for a plane that intersects four edges of a tet
```
kennyweiss, line 1468: 👍 -- that would be a nice additional test.

One possibility would be to define a tet w/ vertices at opposite corners of a unit cube, and a plane that bisects the cube.

I think something like:
```cpp
  TetrahedronType tet(PointType {0.0, 0.0, 0.0},
                      PointType {1.0, 1.0, 0.0},
                      PointType {0.0, 1.0, 1.0},
                      PointType {1.0, 0.0, 1.0});
  PlaneType plane(VectorType {0.0, 0.0, 1.0}, 0.5);
```
although you might need to play with the orientations.

By symmetry, I'd expect the clipped volume to be `tet.volume()/2`

kennyweiss (approved):
Thanks for this contribution @adayton1!

Please consider: 
* updating the doxygen for `clip(Tet, Plane)` to mention that it's clipping the tet against the half-space defined by the plane
* modifying the return type of the internal `clipPolyhedron` function to a bool to avoid an extra copy. 

Please also mention the new function `clip(Tet, Plane)` in the `RELEASE-NOTES`.

src/axom/primal/operators/detail/clip_impl.hpp
```diff
@@ -400,85 +400,95 @@ AXOM_HOST_DEVICE void poly_clip_reindex(Polyhedron<T, NDIMS>& poly,
...
   using BoxType = BoundingBox<T, NDIMS>;
-  using PlaneType = Plane<T, NDIMS>;
 
-  //Bounding Box of Polyhedron
-  BoxType polyBox(&poly[0], poly.numVertices());
-
-  //Clip Polyhedron by each plane
-  for(PlaneType plane : planes)
+  // Check that plane intersects Polyhedron
+  if (intersect(plane, BoxType(&poly[0], poly.numVertices()), true, eps))
```
bmhan12, line 420: Good to put a note that this `clipPolyhedron` overload functions slightly different from the others, that it only returns a Polyhedron if the polyhedron and plane intersect, else an empty Polyhedron.

bmhan12 (approved):

src/axom/primal/operators/detail/clip_impl.hpp
adayton1: I updated the doxygen to reflect it is an inout parameter and changed the return type to void. I experimented with returning a bool, but the following code pattern seems to say the clip failed rather than returning an empty polyhedron so I decided against it.

`if (!clipPolyhedron(poly, plane, eps))`

adayton1: Another option would be to return a reference to the original polyhedron to enable chaining.

`if (clipPolyhedron(poly, plane, eps).numVertices() == 0)`
`return clipPolyhedron(poly, plane, eps);`

src/axom/primal/operators/clip.hpp
adayton1: Done.

src/axom/primal/operators/detail/clip_impl.hpp
adayton1: If they don't intersect, but the polyhedron is above the plane, it will return the original polyhedron, so this is the normal clip operation I'll do an intersection method in a different merge request.

src/axom/primal/operators/detail/clip_impl.hpp
kennyweiss: That's an interesting idea!

adayton1: I think I'll leave the return type as void for now. But if anyone requests chaining in the future, I'd be happy to implement it.

src/axom/primal/tests/primal_clip.cpp
adayton1: Good thinking. Done.

src/axom/primal/tests/primal_clip.cpp
adayton1: That one worked! Thanks!

adayton1:
@kennyweiss , @bmhan12, @rhornung67, I think I've addressed all comments. Let me know if there's anything else you'd like me to change or if I can go ahead and merge when all required checks pass.

kennyweiss:
Thanks @adayton1  -- Other than adding a line to the RELEASE-NOTES, I think it's ready to merge once the branch is passing all CI tests

PR #1221, opened 2023-11-22
Add a FlatMap container

publixsubfan:
# Summary

Adds `axom::FlatMap`, an initial implementation of a hash map container as a mostly drop-in replacement for `std::unordered_map`. The eventual goal is to enable the use of hash maps on the GPU.

## Design

FlatMap is largely based on Boost's `unordered_flat_map` design, described [here](https://bannalia.blogspot.com/2022/11/inside-boostunorderedflatmap.html). Pertinent design points are:
- Open-addressing as opposed to `std::unordered_map`'s closed-addressing model.
- Non-relocating quadratic probing to resolve hash collisions.
- The hash map is split into groups of 15 "buckets":
  - Each group is represented by a 128-bit chunk of metadata, consisting of an "overflow" byte and one byte for each bucket in the group.
  - The "overflow" byte is used as a Bloom filter to test if the next group should be probed during lookups. This avoids the use of tombstones when performing quadratic probing.
  - Each "bucket" byte is 0 if the bucket is empty, 1 if it's at the end of the array, or a reduced hash of `[2,256)` for a filled bucket.

## API goals

Drop-in compatibility with `std::unordered_map` is aimed for as much as possible, with the exception of any methods that take in iterator "hints," e.g. for insertion.

### DIFF
--- a/src/axom/core/FlatMap.hpp
+++ b/src/axom/core/FlatMap.hpp
@@ -83,12 +83,22 @@ public:
 
   // Modifiers
   void clear();
-  std::pair<iterator, bool> insert(const value_type& value);
-  std::pair<iterator, bool> insert(value_type&& value);
+  std::pair<iterator, bool> insert(const value_type& value)
+  {
+    return emplace(value);
+  }
+  std::pair<iterator, bool> insert(value_type&& value)
+  {
+    return emplace(std::move(value));
+  }
   template <typename InputPair>
-  std::pair<iterator, bool> insert(InputPair&& pair);
-  template <typename InputPair>
-  std::pair<iterator, bool> emplace(InputPair&& pair);
+  std::pair<iterator, bool> insert(InputPair&& pair)
+  {
+    return emplace(std::forward<InputPair>(pair));
+  }
+  template <typename... InputArgs>
+  std::pair<iterator, bool> emplace(InputArgs&&... pair);
+
   template <typename InputIt>
   void insert(InputIt first, InputIt last);
 
@@ -207,6 +217,45 @@ auto FlatMap<KeyType, ValueType, Hash>::find(const KeyType& key) const
                      return true;
                    });
   return found_iter;
+}
+
+template <typename KeyType, typename ValueType, typename Hash>
+template <typename... InputArgs>
+auto FlatMap<KeyType, ValueType, Hash>::emplace(InputArgs&&... args)
+  -> std::pair<iterator, bool>
+{
+  KeyValuePair pair {std::forward<InputArgs>(args)...};
+  auto hash = Hash {}(pair.first);
+  // TODO: ensure that we have enough space to insert with given load factor
+
+  bool keyExistsAlready = false;
+  iterator keyIterator = end();
+  auto FindExistingElem = [&, this](IndexType bucket_index) -> bool {
+    if(this->m_buckets[bucket_index].first == pair.first)
+    {
+      keyExistsAlready = true;
+      keyIterator = iterator(this, bucket_index);
+      // Exit out of probing, we can't insert if the key already exists.
+      return false;
+    }
+    return true;
+  };
+
+  auto InsertEmptyBucket = [&, this](IndexType bucket_index) {
+    if(!keyExistsAlready)
+    {
+      keyIterator = iterator(this, bucket_index);
+      this->m_size++;
+      new(&(this->m_buckets[bucket_index])) KeyValuePair(std::move(pair));
+    }
+  };
+
+  this->probeEmptyIndex(m_numGroups2,
+                        m_metadata,
+                        hash,
+                        FindExistingElem,
+                        InsertEmptyBucket);
+  return {keyIterator, !keyExistsAlready};
 }
 
 }  // namespace axom
--- a/src/axom/core/FlatMap.hpp
+++ b/src/axom/core/FlatMap.hpp
@@ -203,6 +203,11 @@ class FlatMap<KeyType, ValueType, Hash>::IteratorImpl
 private:
   using MapType = FlatMap<KeyType, ValueType, Hash>;
 
+  template <bool OtherConst>
+  friend class IteratorImpl;
+
+  friend class FlatMap<KeyType, ValueType, Hash>;
+
 public:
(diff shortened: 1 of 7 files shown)


kennyweiss (approved):
Thanks @publixsubfan for this nice contribution! 

I realize that you're planning to extend this capability in the future, so feel free to address some of my comments in follow-up PRs.

src/axom/core/FlatMap.hpp
```diff
@@ -0,0 +1,771 @@
...
+  axom::Array<PairStorage> m_buckets;
+
+  // Boost flat_unordered_map uses a fixed load factor.
+  constexpr static double MAX_LOAD_FACTOR = 0.875;
+  std::uint64_t m_loadCount;
+};
+
+template <typename KeyType, typename ValueType, typename Hash>
+template <bool Const>
+class FlatMap<KeyType, ValueType, Hash>::IteratorImpl
```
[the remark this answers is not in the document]
publixsubfan: We can't implement `FlatMap::iterator::operator++()` in constant-time, since we have to iterate through an indeterminate number of buckets in order to find the next occupied bin. Since `IteratorBase` models a `RandomAccessIterator`, I think that precludes us from implementing `FlatMap::Iterator` in terms of `IteratorBase`.

src/axom/core/detail/FlatTable.hpp
```diff
@@ -0,0 +1,324 @@
...
+  /*!
+   * \brief Returns the next offset to jump given an iteration count.
+   *
+   *  Each probe point for a given iteration is given by the formula
+   *  H(i) = H_0 + i/2 + i^2/2 mod m. For m = 2^n, the sequence H(i) for
+   *  i in [0, m-1) is a permutation on [0, m-1).
+   *
+   *  We return the offset from H(i) to H(i+1), which is i+1.
+   */
+  int getNext(int iter) const { return iter + 1; }
```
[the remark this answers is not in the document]
publixsubfan: I believe this should give us the sequence `index, index + 1, index + 3, index + 6, ...` which should correspond to the quadratic sequence `H(k, i) = hash(k) + i/2 + (i^2)/2`.

I can add a unit test for this function.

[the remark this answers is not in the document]
kennyweiss: Thanks -- that'd be helpful since, based on the comment, I was expecting a more complicated implementation for the function.

src/axom/core/detail/FlatTable.hpp
```diff
@@ -0,0 +1,334 @@
...
+  {
+    struct
+    {
+      std::uint8_t ofw;
+      std::uint8_t buckets[Size];
+    } metadata;
+    std::uint64_t data[2];
+    static_assert(
+      sizeof(metadata) == sizeof(data),
+      "FlatMap::SwissTable::Bucket: sizeof(data_bytes) != sizeof(data)");
```
agcapps, line 160: FlatMap::SwissTable::Bucket?  I would think flat_map::GroupBucket, or something like that.  (Please check throughout file.)

src/axom/core/FlatMap.hpp
```diff
@@ -0,0 +1,818 @@
...
+                  "copy-constructible.");
+    static_assert(std::is_copy_constructible<ValueType>::value,
+                  "Cannot copy an axom::FlatMap when value type is not "
+                  "copy-constructible.");
+    // Copy all elements.
+    IndexType index = this->nextValidIndex(m_metadata, NO_MATCH);
+    while(index < bucket_count())
+    {
+      new(&m_buckets[index].data) KeyValuePair(other.m_buckets[index].get());
+      index = this->nextValidIndex(m_metadata, index);
```
gunney1, line 150: Should we be concerned about repeatedly constructing `ArrayView` for `m_metadata` in this inner loop?

src/axom/core/FlatMap.hpp
```diff
@@ -0,0 +1,818 @@
...
+  /*!
+   * \brief Erases all elements from the FlatMap.
+   */
+  void clear()
+  {
+    // Destroy all elements.
+    IndexType index = this->nextValidIndex(m_metadata, NO_MATCH);
+    while(index < bucket_count())
+    {
+      m_buckets[index].get().~KeyValuePair();
```
gunney1, line 344: What would happen when the key type's destructor is called again when `m_buckets` goes out of scope?  Do we need to state that the key type should allow multiple call to the destructor?

publixsubfan: Below this, the metadata also gets reset, which marks all the destroyed buckets as "empty." This should prevent the objects from getting double-destroyed.

src/axom/core/detail/FlatTable.hpp
```diff
@@ -0,0 +1,333 @@
...
+  /*!
+   * \brief Inserts a hash into the first empty bucket in an array of groups
+   *  for an open-addressing hash map.
+   *
+   * \param [in] ngroups_pow_2 the number of groups, expressed as a power of 2
+   * \param [in] groups the array of metadata for the groups in the hash map
+   * \param [in] hash the hash to insert
+   */
+  IndexType probeEmptyIndex(int ngroups_pow_2,
+                            ArrayView<GroupBucket> groups,
```
gunney1, line 184: Suggestion: Would `metadata` be a better name instead of `groups`?  That's what how `FlatMap` refers to this parameter.  Same for the `probeIndex` method below.

gunney1 (approved):
Thanks @publixsubfan   Just had a few suggestions.

src/axom/core/detail/FlatTable.hpp
```diff
@@ -0,0 +1,333 @@
...
+        // Stop probing.
+        break;
+      }
+      else if(empty_group == NO_MATCH)
+      {
+        // Set the overflow bit and continue probing.
+        groups[curr_group].setOverflow(hash_8);
+      }
+      curr_group =
+        (curr_group + ProbePolicy {}.getNext(iteration)) % groups.size();
```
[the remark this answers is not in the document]
publixsubfan: I ended up having `SequentialLookupPolicy` inherit from the ProbePolicy, which should have roughly the same effect,

PR #1227, opened 2023-12-04
Fix axom::Array reallocation for non-trivially relocatable types

publixsubfan:
# Summary

A type is considered "trivially relocatable" if a bitcopy of the object can be treated as a "move-and-destroy" of the original object. Types which are not trivially-relocatable need to call the move or copy constructor in order to correctly construct the destination object.

This is usually the case when a type maintains a pointer/reference to itself; one example of a common type which is not trivially-relocatable is `std::string` in the GCC STL (more examples [here](https://quuxplusone.github.io/blog/2019/02/20/p1144-what-types-are-relocatable/))

This PR improves the handling of these non-trivially relocatable types in `axom::Array`, by ensuring that the move/copy constructor is called on non-trivially copyable types when reallocating the underlying memory.

### DIFF
--- a/src/axom/core/tests/core_array.hpp
+++ b/src/axom/core/tests/core_array.hpp
@@ -2308,3 +2308,28 @@ TEST(core_array, reserve_nontrivial_reloc)
     EXPECT_EQ(&(array[i].m_member), array[i].m_localMemberPointer);
   }
 }
+
+TEST(core_array, reserve_nontrivial_reloc_2)
+{
+  const int NUM_ELEMS = 1024;
+  axom::Array<NonTriviallyRelocatable> array(NUM_ELEMS, NUM_ELEMS);
+
+  for(int i = 0; i < NUM_ELEMS; i++)
+  {
+    // Check initial values.
+    EXPECT_EQ(array[i].m_member, NONTRIVIAL_RELOC_MAGIC);
+    EXPECT_EQ(&(array[i].m_member), array[i].m_localMemberPointer);
+  }
+
+  EXPECT_EQ(array.capacity(), NUM_ELEMS);
+
+  // Emplace to trigger a resize.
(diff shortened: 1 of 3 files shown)


src/axom/core/tests/core_array.hpp
```diff
@@ -2257,3 +2257,79 @@ TEST(core_array, resize_stackarray)
...
+    return *this;
+  }
+
+  ~NonTriviallyRelocatable() = default;
+
+  int m_member;
+  int* m_localMemberPointer;
+};
+
+TEST(core_array, reserve_nontrivial_reloc)
```
kennyweiss, line 2290: Do these tests cover all the cases of different allocators (`HOST`, `DEVICE`, `DYNAMIC`)?

It looks like it only covers the CPU malloc allocator, but I might be missing something.

kennyweiss (approved):
Thanks @publixsubfan !

src/axom/core/tests/core_array.hpp
kennyweiss: Perhaps it doesn't matter since the copies are only allowed on the host (?), but there might be a case where we're using a custom allocator on the host? E.g. a pooled allocator, or a shared-memory allocator?

publixsubfan: I'll add a test for unified memory.

For device memory, we don't currently support non-trivially relocatable objects, since we'd need to be able to invoke the copy/move constructor in a device kernel. Instead of forcing users to annotate all their constructors as host/device, we pessimistically assume that non-trivial constructors are host-only.

In the future, we might consider allowing users to "declare" via a type trait specialization that their user-defined type is constructible/destructible on the host, which would allow us to support this use case for device memory.

agcapps (approved):
Thanks for this.

gunney1 (approved):
Thank you @publixsubfan !

PR #1232, opened 2023-12-07
Enable area computation for planar polygons with arbitrary orientation

simonradler:
# Summary

- This PR is a bugfix
- It does the following:
  - Fixes the area computation for planar 3d polygons that are not parallel to the xy-plane

### DIFF
--- a/src/axom/primal/geometry/Polygon.hpp
+++ b/src/axom/primal/geometry/Polygon.hpp
@@ -14,6 +14,7 @@
 
 #include "axom/core/Array.hpp"
 #include "axom/primal/geometry/Point.hpp"
+#include "axom/primal/geometry/Vector.hpp"
 
 #include <ostream>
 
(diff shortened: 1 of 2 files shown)


kennyweiss (approved):
Thanks @simonradler -- this is a nice fix. Clearly, we only previously considered polygons that were aligned with the XY plane.

I'd like to also add more tests for rotated/scaled polygons, but we can add this in a follow up PR.

**Edit:** See https://github.com/LLNL/axom/issues/1240

BradWhitlock (approved):
Thanks for fixing.

kennyweiss:
Thanks for the contribution @simonradler 

I branched off of this for some internal testing in https://github.com/LLNL/axom/pull/1249
I also slightly tweaked the formulation and added some additional tests for scaled, rotated and translated polygons.

PR #1239, opened 2023-12-11
Fix clipping of 3D triangle against axis-aligned bounding box

simonradler:
# Summary

- This PR is a bugfix
- It does the following:
  - Fixes clipping of 3D triangle against axis-aligned bounding box for cases without intersection (incorrectly returned non-empty polygons)


kennyweiss:
Thank you for the bugfix @simonradler -- I cherry-picked your changes into https://github.com/LLNL/axom/pull/1251 and added some additional tests

PR #1249, opened 2023-12-20
Pr from fork/1232

### LINKED ISSUE #1240: Improve testing of Polygon::area function on rotated/scaled polygons
PR #1232 fixed a bug in our `Polygon::area()` function, but still only tests the function on polygons that are aligned with the `XY-`, `YZ-` or `XZ-` planes. 

We should also test this algorithm against polygons that have arbitrary affine transforms (rotation, translation, scaling), after suitably accounting for the changes to area.

kennyweiss:
# Summary

- This PR is a fork from #1232 to allow for internal testing
- It also improves our testing of Polygon area computations. We now test against scaled, rotated and translated Polygons in 2D and 3D. Resolves #1240


white238 (approved):

agcapps (approved):

BradWhitlock (approved):
Good tests.

PR #1250, opened 2023-12-21
Log an explanatory message for quest_point_in_cell_mfem

### LINKED ISSUE #1222: Quest point-in-cell timing
The `quest_point_in_cell_mfem` unit test release is slower than debug.  There may be a good reason for this; noting here for possible investigation.  Is it possible the Debug build is doing less work?

|Platform|Debug|Release|
|---|---|---|
|Win10 MSVC|28.6s|50.5s|
|TOSS4 Clang 14|2.7s|19.1s|
|BlueOS XLClang + CUDA|3.5s|20.1s|

agcapps:
# Summary

- This PR is a feature.
- It does the following:
  - Modifies/refactors test quest_point_in_cell_mfem to print out the size of the test.
  - Fixes #1222 
  - Updates vcpkg to the current state of their master branch, fixing a problem with downloading a now-missing TPL of vcpkg

# Design review

On 18 Dec 2023, we reviewed the issue linked, which is that the quest_point_in_cell_mfem test takes much longer when running release than debug.  @kennyweiss pointed out that the release build tests 10x more points and twice as many refinement steps than the debug build.  That explained the difference in runtime.   The message in this PR is an attempt to alert future developers to the different problem size between debug and release.

### DIFF
--- a/.uberenv_config.json
+++ b/.uberenv_config.json
@@ -9,7 +9,7 @@
 "spack_packages_path": ["scripts/spack/radiuss-spack-configs/packages", "scripts/spack/packages"],
 "spack_concretizer": "clingo",
 "vcpkg_url": "https://github.com/microsoft/vcpkg",
-"vcpkg_commit": "52b52b4588b916152ff662f97842e03db83fa5d1",
+"vcpkg_commit": "c8696863d371ab7f46e213d8f5ca923c4aef2a00",
 "vcpkg_triplet": "x64-windows",
 "vcpkg_ports_path": "scripts/vcpkg_ports"
 }
--- a/.uberenv_config.json
+++ b/.uberenv_config.json
@@ -9,7 +9,7 @@
 "spack_packages_path": ["scripts/spack/radiuss-spack-configs/packages", "scripts/spack/packages"],
 "spack_concretizer": "clingo",
 "vcpkg_url": "https://github.com/microsoft/vcpkg",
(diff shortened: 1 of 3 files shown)


.uberenv_config.json
```diff
@@ -9,7 +9,7 @@
 "spack_packages_path": ["scripts/spack/radiuss-spack-configs/packages", "scripts/spack/packages"],
 "spack_concretizer": "clingo",
 "vcpkg_url": "https://github.com/microsoft/vcpkg",
-"vcpkg_commit": "6f7ffeb18f99796233b958aaaf14ec7bd4fb64b2",
+"vcpkg_commit": "52b52b4588b916152ff662f97842e03db83fa5d1",
```
kennyweiss, line 12: Please update the `RELEASE_NOTES` to mention the vcpkg update and to point to this version.

kennyweiss (approved):
Thanks for adding this @agcapps 

Thanks also for updating vcpkg! 
Can you please run the "Manual Windows TPLs" github action on this branch to check that it works with this update and that it resolves the issue you noticed w/ downloading `winpthread`?

rhornung67 (approved):

white238 (approved):

.uberenv_config.json
agcapps: done.

agcapps: Done.

PR #1261, opened 2024-01-27
Bugfix when loading group with attributes

kennyweiss:
# Summary

- This PR is a bugfix
- It allows loading a group with attributes into a datastore that already has that attribute
- When loading, only create the attribute if it doesn't already exist, but use the default value from the file
- Includes some minor code cleanup
- Fixes https://github.com/LLNL/axom/issues/67

### DIFF
--- a/src/axom/sidre/core/DataStore.cpp
+++ b/src/axom/sidre/core/DataStore.cpp
@@ -637,14 +637,9 @@ bool DataStore::saveAttributeLayout(Node& node) const
 
   bool hasAttributes = false;
 
-  IndexType aidx = getFirstValidAttributeIndex();
-  while(indexIsValid(aidx))
+  for(const auto& attr : attributes())
   {
-    const Attribute* attr = getAttribute(aidx);
-
-    node[attr->getName()] = attr->getDefaultNodeRef();
-
-    aidx = getNextValidAttributeIndex(aidx);
+    node[attr.getName()] = attr.getDefaultNodeRef();
     hasAttributes = true;
   }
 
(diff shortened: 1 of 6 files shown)


src/axom/sidre/core/DataStore.cpp
```diff
@@ -670,7 +665,9 @@ void DataStore::loadAttributeLayout(Node& node)
       Node& n_attr = attrs_itr.next();
       std::string attr_name = attrs_itr.name();
 
-      Attribute* attr = createAttributeEmpty(attr_name);
+      auto* attr = !hasAttribute(attr_name) ? createAttributeEmpty(attr_name)
+                                            : getAttribute(attr_name);
```
kennyweiss, line 669: This is the bugfix -- only create a new attribute if it doesn't already exist.

agcapps (approved):
Thank you for the fix and various clean-up, Kenny.

rhornung67 (approved):
@kennyweiss thanks for the cleanup and consolidation of conditional logic

src/axom/sidre/tests/sidre_attribute.cpp
```diff
@@ -56,11 +59,18 @@ const int g_nprotocols = 2;
 const std::string g_protocols[] = {"sidre_json", "json"};
 #endif
 
+const std::set<std::string> g_protocol_saves_attributes {"sidre_hdf5",
+                                                         "sidre_conduit_json",
+                                                         "sidre_json",
+                                                         "sidre_layout_json"};
```
kennyweiss, line 65: Apparently, only some of the Sidre protocols currently save/load attributes.
Not sure if that's by design or desire, but I left it alone for now.

See `saveAttributeLayout` calls in `Group::save()` implementation: 
https://github.com/LLNL/axom/blob/c5a31951a1d71ae574edd2c2ccfa5b42d1216fa2/src/axom/sidre/core/Group.cpp#L1908-L1971

PR #1262, opened 2024-01-30
Slam: Multi-dimensional Map indexing

publixsubfan:
# Summary

Adds support for multi-dimensional `slam::Map` and `slam::BivariateMap` instances.

A multi-dimensional `Map`/`BivariateMap` field is created by using the newly-added `policies::MultiDimStride` as the stride policy. An N-dimensional "shape" array can then be passed into the `Map` constructor, representing the number of components in each dimension.

Multi-dimensional indexing operators have been added to `Map`, `BivariateMap`, and `SubMap`:
- Indexing is always row-major for a multi-dimensional field, i.e. for a field with element shape (3, 4, 5), the component index (i, j, k) translates to `i * 20 + j * 5 + k`.
- Static asserts check that the number of indexes provided match the corresponding dimensionality of the stride policy.

Splits iterators for `Map`-like classes into two types, `FlatIterator` and `RangeIterator`:
- A `FlatIterator` iterates across all individual values in the `Map`. In the case of a non-unit stride policy (i.e. `Runtime/CompileTimeStride` > 1 or a `MultiDimStride`), the `FlatIterator` will iterate over each individual sub-component, before advancing to the next set element.
- A `RangeIterator` iterates across the individual set elements, returning a range of values associated with an individual set element. This range of values is returned as an `axom::ArrayView` with dimensionality corresponding to the StridePolicy.

`RangeIterator` most closely mimics the original behavior of iterators for N-component or multi-dimensional fields, while `FlatIterator` represents the original behavior for a unit-strided field.

Adds iterators to `BivariateSet`, `RelationSet`, and `ProductSet`.


kennyweiss (approved):
Thanks @publixsubfan -- this is a great addition!

Is it worth posting/updating a performance comparison for the various options, or should we hold off for other features?

Also, please update the RELEASE-NOTES.

src/axom/slam/policies/StridePolicies.hpp
```diff
@@ -51,12 +51,19 @@ struct RuntimeStride
...
+  using ShapeType = IntType;
+
+  static constexpr IntType DefaultSize() { return DEFAULT_VALUE; }
 
   AXOM_HOST_DEVICE RuntimeStride(IntType stride = DEFAULT_VALUE)
     : m_stride(stride)
   { }
 
   AXOM_HOST_DEVICE inline IntType stride() const { return m_stride; }
+  AXOM_HOST_DEVICE inline IntType shape() const { return m_stride; }
```
gunney1, line 66: Is this a typo?  Should it return `m_shape`?

kennyweiss: Not a typo. Perhaps a comment would help.

src/axom/core/ArrayView.hpp
```diff
@@ -371,7 +371,7 @@ ArrayView<T, DIM, SPACE>::ArrayView(
   , m_allocator_id(static_cast<const OtherArrayType&>(other).getAllocatorID())
 {
   static_assert(
-    std::is_const<T>::value,
+    std::is_const<T>::value || detail::ArrayTraits<OtherArrayType>::is_view,
```
gunney1, line 374: Why do we not allow `other` to be a view?  Do we want to exclude views of views?  Or does it have something to do with the const-ness of the data?

src/axom/slam/BivariateSet.hpp
```diff
@@ -278,6 +294,56 @@ bool BivariateSet<Set1, Set2>::isValid(bool verboseOutput) const
...
+  using iterator_category = std::forward_iterator_tag;
+
+  BivariateSetIterator(const BivariateSetType* bset, IndexType flatPos = 0)
+    : BaseType(flatPos)
+    , m_bset(bset)
+  { }
+
+  std::pair<IndexType, IndexType> operator*() const
+  {
+    // Going from flat-to-second index is always free for a StaticRelation.
```
gunney1, line 324: Something seems off here.  Maybe you meant
// Going from flat to second-index ...

src/axom/slam/Map.hpp
```diff
@@ -261,32 +270,98 @@ class Map : public StrPol,
...
+    return value(setIdx, 0);
+  }
+
   /**
    * \brief Access the value associated with the given position in the set and
    *        the component index.
    *
    * \pre `0 <= setIdx < size()`
-   * \pre `0 <= comp < numComp()`
+   * \pre `0 <= compIdx[i] < shape()[i]`
```
gunney1, line 293: Should `i` be `setIdx`?

src/axom/slam/Map.hpp
```diff
@@ -261,32 +270,98 @@ class Map : public StrPol,
...
+  {
+    return value(setIdx, 0);
+  }
+
+  /**
+   * \brief Access the value associated with the given position in the set and
+   *        the component index.
+   *
+   * \pre `0 <= setIdx < size()`
+   * \pre `0 <= compIdx[i] < shape()[i]`
```
gunney1, line 325: Same thing.  `i` -> `setIdx` ?

src/axom/slam/BivariateMap.hpp
```diff
@@ -750,6 +620,202 @@ template <typename T, typename BSet, typename IndPol, typename StrPol, typename
...
+    : IterBase(pos)
+    , m_map(sMap)
+    , m_mapIterator(m_map->getMap()->set_begin() + pos)
+    , m_bsetIterator(m_map->set(), pos)
+  { }
+
+  /**
+     * \brief Returns the current iterator value. If the BivariateMap has
+     *        multiple components, this will return the first component.
+     *        To access the other components, use iter(comp)
```
gunney1, line 753: I don't understand the last sentence.  I can't connect `comp` to this context.

gunney1:
Can you provide a couple sentences of motivation for this work?  I'm assuming it meets a current or future need, but what are those needs?

gunney1:
Thanks @publixsubfan  I just have some non-blocking comments and questions.

gunney1 (approved):
Thanks @publixsubfan  I just have some non-blocking comments.

rhornung67 (approved):

src/axom/slam/policies/StridePolicies.hpp
publixsubfan: Added some comments to clarify.

src/axom/core/ArrayView.hpp
publixsubfan: The latter -- this constructor matches with `const` instances of `Array`, so we want to prevent it from being called to construct a mutable `ArrayView<T>`. But we can relax this requirement for view-like types, such as `ArraySubslice`. This is used in the Slam multi-dimensional indexers to return sub-ranges.

src/axom/slam/BivariateSet.hpp
publixsubfan: Updated comment.

src/axom/slam/Map.hpp
publixsubfan: `i` represents the dimensionality of the stride. Updated docs to correspond to this.

src/axom/slam/Map.hpp
publixsubfan: Updated as per above.

src/axom/slam/Map.hpp
```diff
@@ -409,58 +550,148 @@ class Map : public StrPol,
...
+    using reference = const value_type&;
+    using pointer = const value_type*;
     using difference_type = SetPosition;
 
-    using IterBase = IteratorBase<MapIterator, SetPosition>;
-    using iter = MapIterator;
-    using PositionType = SetPosition;
-    using IterBase::m_pos;
+  private:
+    static StackArray<IndexType, Dims + 1> fetchDims(StrideIndexType stride)
```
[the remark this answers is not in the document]
publixsubfan: This overload only matches when `StridePolicy` is one of the 1D policies, `RuntimeStride` or `CompileTimeStride`. In that case, the value returned from `StridePolicyType::shape()` is a scalar int type, instead of a `StackArray` as with the `MultiDimStride` policy.

I'll rename the `stride` parameter to `shape` to correspond with its usage.

src/axom/slam/BivariateMap.hpp
publixsubfan: Updated documentation to clarify.

PR #1264, opened 2024-01-30
Slic tags for streams and messages

bmhan12:
This PR
- Adds the ability to create user-defined tags for tagged streams and messages.
  - Tagged streams only receive messages with the corresponding tag.
- Adds a `TextTagCombiner` Combiner for Lumberjack that combines messages based on message and tag content.

ReadTheDocs link for this branch: https://axom.readthedocs.io/en/feature-han12-slic_tags/axom/slic/docs/sphinx/sections/getting_started.html#step-5-1-tagged-log-streams

### DIFF
--- a/src/axom/lumberjack/TextEqualityCombiner.hpp
+++ b/src/axom/lumberjack/TextEqualityCombiner.hpp
@@ -34,6 +34,9 @@ namespace lumberjack
  *  you. If you want it removed call Lumberjack::removeCombiner with the string
  * "TextEqualityCombiner" as it's parameter.
  *
+ * \warning Using the TextEqualityCombiner with Message::tag has undefined
+ *          behavior.
+ *
  * \see Combiner Lumberjack
  *******************************************************************************
  */
--- a/src/axom/lumberjack/TextEqualityCombiner.hpp
+++ b/src/axom/lumberjack/TextEqualityCombiner.hpp
@@ -40,7 +40,7 @@ namespace lumberjack
 class TextEqualityCombiner : public Combiner
 {
 public:
-  TextEqualityCombiner() : m_id("TextEqualityCombiner") { }
+  TextEqualityCombiner() { }
 
   /*!
    *****************************************************************************
@@ -65,11 +65,7 @@ public:
   bool shouldMessagesBeCombined(const Message& leftMessage,
                                 const Message& rightMessage)
   {
-    if(leftMessage.text().compare(rightMessage.text()) == 0)
-    {
-      return true;
-    }
-    return false;
+    return (leftMessage.text().compare(rightMessage.text()) == 0);
   }
 
   /*!
@@ -83,6 +79,8 @@ public:
    * \param [in] combinee the Message that is combined into the other.
    * \param [in] ranksLimit The limit on how many individual ranks are tracked
    *  in the combined Message. Message::rankCount is always incremented.
+   *
+   * \pre shouldMessagesBeCombined(combined, combinee) must be true
    *****************************************************************************
    */
   void combine(Message& combined, const Message& combinee, const int ranksLimit)
@@ -91,7 +89,7 @@ public:
   }
 
 private:
-  std::string m_id;
+  const std::string m_id = "TextEqualityCombiner";
 };
 
 }  // end namespace lumberjack
--- a/src/axom/lumberjack/TextTagCombiner.hpp
+++ b/src/axom/lumberjack/TextTagCombiner.hpp
@@ -1,0 +1,102 @@
+// Copyright (c) 2017-2023, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level LICENSE file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/*!
+ *******************************************************************************
+ * \file TextTagCombiner.hpp
+ *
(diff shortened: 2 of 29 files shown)


src/axom/slic/interface/slic_macros.hpp
```diff
@@ -578,6 +578,35 @@ AXOM_HOST_DEVICE inline blackhole &operator<<(blackhole &bh, T)
                            __LINE__);                 \
   } while(axom::slic::detail::false_value)
 
+/*!
+ * \def SLIC_INFO_TAGGED( msg, tag )
```
bmhan12, line 582: For now, I have only added one tagged macro, `SLIC_INFO_TAGGED(msg, tag)` for unit testing. It's fairly straightforward to add more, but I figured Axom currently does not have a need for tagged macro variants?

src/axom/lumberjack/TextTagCombiner.hpp
```diff
@@ -0,0 +1,102 @@
...
+public:
+  TextTagCombiner() : m_id("TextTagCombiner") { }
+
+  /*!
+   *****************************************************************************
+   * \brief Returns the unique string identifier for this combiner. Used by
+   *  Lumberjack to differentiate between other combiners.
+   *****************************************************************************
+   */
+  const std::string id() { return m_id; }
```
kennyweiss, line 52: Since there's no way to change `m_id`, should it be a const static member variable?

kennyweiss (commented):
Thanks @bmhan12 -- overall, this looks like a nice contribution.

While reviewing the code, it wasn't entirely clear to me if one could have several logstreams at the same time where one of them didn't filter by tag and the other did.

I think the code could benefit from some additional tests/examples that demonstrate how this new functionality might be used. Perhaps that can come in a separate PR though.

bmhan12:
> While reviewing the code, it wasn't entirely clear to me if one could have several logstreams at the same time where one of them didn't filter by tag and the other did.

Good point! I've added a test case in commit c3ff7493f53d5de8be76dc6f8408329db0b8621d that has both a tag stream and message-level based stream at once.

> I think the code could benefit from some additional tests/examples that demonstrate how this new functionality might be used. Perhaps that can come in a separate PR though.

For now, I am thinking there will be more examples and tests in future PR's as more use cases/details get worked out/come up.

src/axom/lumberjack/TextTagCombiner.hpp
```diff
@@ -0,0 +1,100 @@
...
+   * triggered for combination if both Message::text and Message::tag are equal.
+   *
+   * \param [in] leftMessage One of the Messages to be compared.
+   * \param [in] rightMessage One of the Messages to be compared.
+   *****************************************************************************
+   */
+  bool shouldMessagesBeCombined(const Message& leftMessage,
+                                const Message& rightMessage)
+  {
+    return (leftMessage.text().compare(rightMessage.text()) == 0 &&
```
white238, line 69: Make this the default combiner because you would lose the tag in the current default

src/axom/lumberjack/TextEqualityCombiner.hpp
```diff
@@ -65,11 +65,7 @@ class TextEqualityCombiner : public Combiner
   bool shouldMessagesBeCombined(const Message& leftMessage,
                                 const Message& rightMessage)
   {
-    if(leftMessage.text().compare(rightMessage.text()) == 0)
-    {
-      return true;
-    }
-    return false;
+    return (leftMessage.text().compare(rightMessage.text()) == 0);
```
white238, line 68: This PR showed a bug in this combiner where depending on what is `leftMessage` and `rightMessage` the tag might be lost.

kennyweiss (approved):
Thanks @bmhan12

src/axom/lumberjack/TextTagCombiner.hpp
bmhan12: `TextTagCombiner` is the new default for Lumberjack, [ReadTheDocs ](https://axom.readthedocs.io/en/feature-han12-slic_tags/axom/lumberjack/docs/sphinx/combiner_class.html)was also updated with the new combiner default.

src/axom/lumberjack/TextEqualityCombiner.hpp
white238: Can you add a note that using this Combiner with tags has undetermined behavior?

src/axom/slic/docs/sphinx/sections/appendix.rst
```diff
@@ -87,6 +87,7 @@ The table below details which SLIC macros are collective:
 | | ``SLIC_INFO_IF``         | |                                                                          |
 | | ``SLIC_INFO_ROOT``       | |                                                                          |
 | | ``SLIC_INFO_ROOT_IF``    | |                                                                          |
+| | ``SLIC_INFO_TAGGED``     | |                                                                          |
```
white238, line 90: When will the rest of the macros be added?

src/axom/lumberjack/TextEqualityCombiner.hpp
bmhan12: Warning added to `TextEqualityCombiner` about undefined behavior with tags.

agcapps (approved):

PR #1269, opened 2024-02-06
Array: enable support for some device-side insertions

publixsubfan:
# Summary

- Enables using `Array::push_back()` and `Array::emplace_back()` inside a GPU kernel, with some restrictions and limitations:
  - The number of calls to `push_back()` plus the array size before the kernel must not exceed `Array::capacity()`, since we can't reallocate the array on the host.
  - The `Array` instance being inserted into must be allocated in a device-accessible memory space, i.e. should not be captured by-value in a lambda (although that isn't supported anyways).
  - A sequence of `push_back`/`emplace_back` in a GPU kernel will have an undefined order, since the underlying insertion is performed via a call to `RAJA::atomicAdd`.


src/axom/core/tests/core_array_for_all.hpp
```diff
@@ -1177,4 +1177,91 @@ AXOM_TYPED_TEST(core_array_for_all, nontrivial_emplace)
...
+        }
+      }
+      else
+      {
+        arr_v[0].emplace_back(3 * idx + 5);
+      }
+#else
+      arr_v[0].emplace_back(3 * idx + 5);
+#endif
+    });
```
kennyweiss, line 1234: We previously added a restriction to disallow use of `axom::Array` within a kernel  -- see https://github.com/LLNL/axom/pull/1255 and https://github.com/LLNL/axom/issues/724
It looks like this is getting around that by passing in a `axom::ArrayView` to an `axom::Array`.

* Does this feature imply that we should allow Arrays to be used in kernels?
* Alternatively, should we add `push_back` and/or `emplace_back` to an `axom::ArrayView` if/when the addition will be under capacity? Perhaps as a new `MutableArrayView` class derived from `ArrayBase` ?

kennyweiss (approved):
Interesting new feature, @publixsubfan 

Since we recently disallowed using `axom::Array` within a kernel, I had some questions about how we can support/expose this new functionality (see comments within my review).

Please also consider if/how we should update our documentation in light of this change.
(See https://axom.readthedocs.io/en/develop/axom/core/docs/sphinx/core_containers.html#using-arrays-in-gpu-code)

src/axom/core/tests/core_array_for_all.hpp
publixsubfan: There are a few things going on here. As I see it, the main limitation with `axom::Array` use in a lambda is primarily the fact that the value-capture would copy the underlying array, instead of providing view-like semantics; for the device, there is also the issue that this "array copy" is ill-defined, since we can't allocate in device code. Thus the requirement to use `axom::ArrayView`, and the checks in #1255 being implemented in the copy constructor.

However, this doesn't necessarily preclude all uses of `axom::Array` from device code. In a certain user code, we have large hierarchies of class types that are allocated in unified memory. Some of these classes have instances of an `std::vector`-like class, and the code is able to call `push_back` on the device, which is implemented as an atomic addition on the size variable. Since the vector-like class instance is itself allocated in unified memory, this code works as intended.

I think the second proposal is interesting, but it doesn't exactly match our user code's use case; we don't wrap an externally-provided buffer with our `std::vector`-like class; rather, it owns its own memory.

src/axom/core/Array.hpp
```diff
@@ -1484,6 +1509,37 @@ inline T* Array<T, DIM, SPACE>::reserveForInsert(IndexType n, IndexType pos)
...
+  // Device path: supports insertion while m_num_elements < m_capacity
+  // Does not support insertions which require reallocating the underlying
+  // buffer.
+  IndexType new_pos = RAJA::atomicAdd<RAJA::auto_atomic>(&m_num_elements, n);
+  if(new_pos >= m_capacity)
+  {
+  #ifdef AXOM_DEBUG
+    printf(
+      "Array::reserveForInsert: size() exceeded capacity() when inserting "
+      "on the device.\n");
```
agcapps, line 1531: Style note only: I'm a little surprised at the use of `printf()`, especially in device code, but I'm not sure what else there is.

agcapps (approved):
Thanks for the feature!

PR #1270, opened 2024-02-06
Guard install

white238:
This happened in another repo and I can't imagine this happening here, but you can't call install on a non-real target.

### DIFF
--- a/src/cmake/thirdparty/SetupAxomThirdParty.cmake
+++ b/src/cmake/thirdparty/SetupAxomThirdParty.cmake
@@ -148,10 +148,16 @@ if (TARGET mfem)
     # Note - white238: I can't seem to get this to pass install testing due to mfem being included
     # in multiple export sets
     message(STATUS "MFEM support is ON, using existing mfem target")
+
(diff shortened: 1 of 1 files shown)


src/cmake/thirdparty/SetupAxomThirdParty.cmake
```diff
@@ -148,10 +148,16 @@ if (TARGET mfem)
     # Note - white238: I can't seem to get this to pass install testing due to mfem being included
     # in multiple export sets
```
kennyweiss, line 149: Does the current PR address this comment?

kennyweiss (approved):
Thanks @white238

rhornung67 (approved):

src/cmake/thirdparty/SetupAxomThirdParty.cmake
white238: This was definitely true at one point. I might have fixed it in the various multiple fixes I've done. Soon I will get back to revisiting all these nightmares and do a bulk review.

PR #1285, opened 2024-02-26
Array: improve support for unified memory

publixsubfan:
# Summary

- Adds GPU-side `ArrayOps` operations for unified and pinned memory.

  As with the operations for device memory, unified and pinned memory operations try to use Umpire operations when the type is trivially-copyable, and instantiate a GPU kernel for fill operations when the type is trivially-constructible. This potentially avoids unwanted memory motion between the CPU and GPU.

  In the case where the type isn't trivially-copyable or trivially-constructible, the corresponding operation needs to be performed on the device in order to correctly call the required constructor. However, since unified and pinned memory are accessible on the device, we can just do that operation in-place instead of allocating a temporary buffer.

### DIFF
--- a/src/axom/core/ArrayBase.hpp
+++ b/src/axom/core/ArrayBase.hpp
@@ -1448,6 +1448,16 @@ struct MemSpaceToOpSpace<MemorySpace::Device>
 {
   static constexpr OperationSpace value = OperationSpace::Device;
 };
+template <>
+struct MemSpaceToOpSpace<MemorySpace::Pinned>
+{
+  static constexpr OperationSpace value = OperationSpace::Unified_Device;
+};
+template <>
+struct MemSpaceToOpSpace<MemorySpace::Unified>
+{
+  static constexpr OperationSpace value = OperationSpace::Unified_Device;
+};
 #endif
 
 template <typename T, MemorySpace SPACE>
@@ -1531,6 +1541,8 @@ private:
   using Base = ArrayOpsBase<T, OperationSpace::Host>;
 #if defined(AXOM_USE_GPU) && defined(AXOM_GPUCC) && defined(AXOM_USE_UMPIRE)
   using BaseDevice = ArrayOpsBase<T, OperationSpace::Device>;
+  // Works with unified and pinned memory.
+  using BaseUM = ArrayOpsBase<T, OperationSpace::Unified_Device>;
 #endif
 
 public:
@@ -1542,6 +1554,11 @@ public:
     if(space == MemorySpace::Device)
     {
       BaseDevice::init(array, begin, nelems);
+      return;
+    }
+    else if(space == MemorySpace::Unified || space == MemorySpace::Pinned)
+    {
+      BaseUM::init(array, begin, nelems);
       return;
     }
 #else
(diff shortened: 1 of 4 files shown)


src/axom/core/ArrayBase.hpp
```diff
@@ -1025,63 +1032,160 @@ struct ArrayOpsBase<T, false>
...
+  }
+
+  T* getStagingBuffer() const { return static_cast<T*>(m_staging_buf); }
+
+  void* m_staging_buf;
+  T* m_data;
+  IndexType m_begin;
+  IndexType m_num_elems;
+  bool m_is_destroying;
+};
```
kennyweiss, line 1127: Out of curiosity -- does new/delete always work in this context? 
Would people want to use Umpire, e.g. to use memory from an existing pool?

src/axom/core/ArrayBase.hpp
```diff
@@ -1025,63 +1032,160 @@ struct ArrayOpsBase<T, false>
...
+struct DeviceStagingBuffer;
+
 template <typename T>
-struct ArrayOpsBase<T, true>
+struct DeviceStagingBuffer<T, OperationSpace::Device>
+{
+  DeviceStagingBuffer(T* data,
+                      IndexType begin,
+                      IndexType nelems,
+                      bool destroying = false)
```
kennyweiss, line 1097: Please add doxygen for the parameters. 

In particular, that `begin` is the offset from the base pointer `data`, and what destroying does.

kennyweiss (approved):
Thanks @publixsubfan !

src/axom/core/ArrayBase.hpp
gunney1: A sentence or two of doxygen comment about the usage or purpose would help.

src/axom/core/ArrayBase.hpp
```diff
@@ -1261,14 +1350,26 @@ struct ArrayOpsBase<T, true>
...
+    {
+      // Type might not be trivially-relocatable, move the range on the host.
+      // Note that we only do this for objects in unified/pinned memory, since
+      // we assume that objects in device-only memory are trivially-relocatable.
+      HostOp::move(array, src_begin, src_end, dst);
+    }
+    else
+    {
+      // Since this memory is on the device-side, we copy it to a temporary buffer
+      // first.
```
gunney1, line 1364: Out of curiosity, why do we have to copy to a temporary buffer first?  Isn't `axom::copy` smart enough to do the right thing based on where src and dst are?

gunney1:
In the summary, can you briefly say if this this change lets us do something we couldn't do before?  Or maybe whether it only streamlines, slimplify, whatever..., something we can already do?

src/axom/core/ArrayBase.hpp
publixsubfan: My recollection is that there was some issue relating to memory overlap handling that required us to handle this case manually. But I could be wrong -- I'll take a closer look at this.

rhornung67 (approved):

publixsubfan:
> In the summary, can you briefly say if this this change lets us do something we couldn't do before? Or maybe whether it only streamlines, slimplify, whatever..., something we can already do?

This change mainly concerns performance for unified memory on Nvidia -- we generally want to avoid touching unified memory on the host when possible to avoid expensive page-faults in a subsequent device kernel.

src/axom/core/ArrayBase.hpp
publixsubfan: I modified this to use Umpire via `axom::allocate()`/`axom::deallocate()`.

PR #1288, opened 2024-03-14
Bugfix for intermittent failure of quest_inout_interface Fortran example

kennyweiss:
# Summary

- This PR is a bugfix for intermittent failure in our gcc@11 CI `quest_inout_interface` Fortran test
- It only calls the `finalize()` function for the dimension (2 or 3) that was initialized

### DIFF
--- a/src/axom/quest/interface/inout.cpp
+++ b/src/axom/quest/interface/inout.cpp
@@ -58,7 +58,7 @@ struct InOutHelper
     void setDefault() { *this = State {}; }
   };
 
-  InOutHelper() : m_surfaceMesh(nullptr), m_inoutTree(nullptr)
+  InOutHelper()
   {
     m_params.setDefault();
     m_state.setDefault();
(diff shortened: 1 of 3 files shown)


src/axom/quest/interface/inout.cpp
```diff
@@ -401,13 +401,21 @@ int inout_init(mint::Mesh*& mesh, MPI_Comm comm)
 int inout_finalize()
 {
   const int dim = inout_get_dimension();
+  SLIC_ASSERT(dim == 2 || dim == 3);
 
-  // Finalize the 2D and 3D structures and reset the params
-  int rc2 = s_inoutHelper2D.finalize();
-  int rc3 = s_inoutHelper3D.finalize();
```
kennyweiss, line 407: The problem might have been calling finalize in 2D \*and\* 3D when it was only initialized in 2D \*or\* 3D.

white238 (approved):
Thanks @kennyweiss !

rhornung67 (approved):

PR #1291, opened 2024-03-16
Add support for caliper and adiak dependencies

kennyweiss:
# Summary

- This PR is a feature in support of #142
- It adds support to `axom` for the `caliper` and `adiak` as optional third-party dependencies
- This PR adds the dependencies to axom's build system along with smoke tests
- A follow up PR will add macros to simplify caliper and adiak usage within axom and add support for caliper/adiak on Windows through uberenv/vcpkg

#### TODO:
- [x] Update RELEASE-NOTES
- [x] Update host-configs on LLNL's LC platforms (toss4, blueos, toss4_cray)
- [x] Update host-configs for docker containers

**Credits:** I modeled the build system integration after @white238's caliper and adiak integration on serac -- https://github.com/LLNL/serac


src/cmake/thirdparty/SetupAxomThirdParty.cmake
```diff
@@ -88,6 +88,8 @@ endif()
 # newer CMake versions
 if (CONDUIT_DIR)
     axom_assert_is_directory(VARIABLE_NAME CONDUIT_DIR)
+
+    unset(CONDUIT_FOUND CACHE)
```
kennyweiss, line 92: @white238 and @cyrush -- I had to add this for incremental reconfigs since conduit's config files explicitly check for `CONDUIT_FOUND`, but `conduit::conduit` is not otherwise a target

kennyweiss:
Note: To configure the installed examples (`using-with-cmake` and `using-with-blt`) on blueos, I had to load a newer version of cmake than the default (`3.14`). Specifically, I ran `module load cmake/3.21.1`. 

This was required to successfully generate the newer TPLs via the `./scripts/llnl_scripts/build_tpls` script.

Should we add a check for the CMake version to the beginning of the build_tpls's script to ensure we're using a sufficiently high version?

Note: This is not a problem for the `build_src` scripts since those use a specified CMake version from the host-config files.

src/cmake/thirdparty/SetupAxomThirdParty.cmake
cyrush: This is b/c axom is setting `CONDUIT_FOUND` a few lines down:

https://github.com/LLNL/axom/blob/75126d6a0b7d99541b017c48c628d24a571cdf6a/src/cmake/thirdparty/SetupAxomThirdParty.cmake#L102

Not sure 100% why, but maybe remove there b/c that seems to be what you are doing effectively by unsetting it before we call find again.

kennyweiss: Thanks -- I think it's there for consistency w/ how we treat other packages. 
@white238 mentioned offline that we might not need it to be a `CACHE` variable.

cyrush: Got it, but Conduit sets `CONDUIT_FOUND` itself when it is found -- so I think it could be dropped unless is some reason downstream in Axom's cmake that this needs to be a cached var?

white238: It shouldn't, my guess is I got a little excited to make everything "symmetrical" for lack of a better word.

host-configs/rzgenie-toss_4_x86_64_ib-clang@14.0.6.cmake
kennyweiss, line 1: Note: I removed the `rzgenie-toss_4` host-configs in favor of `rzwhippet-toss4` host-configs.
I confirmed that the latter can still be used on `rzgenie`.

src/cmake/thirdparty/SetupAxomThirdParty.cmake
kennyweiss: Are we ok w/ my proposed change? Or should I remove both (all) references to `CONDUIT_FOUND`  in `SetupAxomThirdParty.cmake`?

src/thirdparty/tests/CMakeLists.txt
```diff
@@ -122,6 +122,42 @@ if (MFEM_FOUND)
...
+
+  axom_add_test( NAME    caliper_smoke
+                 COMMAND caliper_smoke_test )
+
+endif()
+
+if ( ADIAK_FOUND )
+
+  set( adiak_smoke_dependencies adiak::adiak gtest )
+  blt_list_append( TO adiak_smoke_dependencies ELEMENTS mpi IF ENABLE_MPI)
```
white238, line 148:
```suggestion
  blt_list_append( TO adiak_smoke_dependencies ELEMENTS mpi IF AXOM_ENABLE_MPI)
```

scripts/spack/packages/axom/package.py
```diff
@@ -467,10 +487,10 @@ def initconfig_package_entries(self):
         entries.append(cmake_cache_path("CONDUIT_DIR", conduit_dir))
 
         # optional tpls
-        for dep in ("c2c", "mfem", "hdf5", "lua", "raja", "umpire"):
-            if "+%s" % dep in spec:
+        for dep in ("adiak", "caliper", "c2c", "mfem", "hdf5", "lua", "raja", "umpire"):
+            if spec.satisfies("^{0}".format(dep)):
```
white238, line 491: With this change, if Axom turns off the mfem variant but may still end up with it due to it being somewhere in the spec. I'm not sure that is desired.

kennyweiss: Fixed

kennyweiss:
I cleaned up the imports logic and removed workarounds for older versions of caliper/adiak.

scripts/spack/packages/axom/package.py
```diff
@@ -136,18 +139,35 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
...
         depends_on(raja_cuda, when="+{0}".format(raja_cuda))
         depends_on(umpire_cuda, when="+{0}".format(umpire_cuda))
+        depends_on("caliper cuda_arch={0}".format(val), when="+profiling cuda_arch={0}".format(val))
 
     for val in ROCmPackage.amdgpu_targets:
         raja_rocm = "raja +rocm amdgpu_target={0}".format(val)
         umpire_rocm = "umpire +rocm amdgpu_target={0}".format(val)
         depends_on(raja_rocm, when="+{0}".format(raja_rocm))
         depends_on(umpire_rocm, when="+{0}".format(umpire_rocm))
+        depends_on("caliper amdgpu_target={0}".format(val), when="+profiling amdgpu_target={0}".format(val))
```
white238, line 170: `+rocm` is not getting set for this target. Is this correct? Also it would be nice to follow the same pattern for all of these. The `+cuda` got mixed as well. The for loop for the `cuda_arch` and `amdgpu_target` handles `cuda` and  `rocm` respectively. This was probably changed since this was originally written.

src/cmake/thirdparty/SetupAxomThirdParty.cmake
white238: Either is fine.

scripts/spack/packages/axom/package.py
kennyweiss: Thanks -- I updated the recipe and am rebuilding the rocm and cuda TPLs

src/cmake/thirdparty/SetupAxomThirdParty.cmake
kennyweiss: Thanks -- in that case, I'll leave it as is.

rhornung67 (approved):

agcapps (approved):

scripts/llnl_scripts/move_unused_tpl_dirs_to_graveyard.py
```diff
@@ -40,23 +40,25 @@ def main():
...
+    print("[# of unused {}]".format(len(r["unused"])))
 
     success = True
     for d in r["unused"]:
-        cmd = "mv %s %s" % (d, graveyard_path(d)) 
-        res = sexe(cmd)
-        if res != 0:
-            success = False
+        if os.path.exists(d):
+            cmd = "mv %s %s" % (d, graveyard_path(d))
```
white238, line 57: what we definitely need is three different ways of formatting strings in ~20 lines...

white238 (approved):

scripts/llnl_scripts/move_unused_tpl_dirs_to_graveyard.py
kennyweiss: I'll fix it in a follow-up PR

PR #1299, opened 2024-03-21
Use `tee` instead of reading log files to the screen in Build scripts

white238:
This will hopefully help us track down what is causing the hang on Tioga


kennyweiss (approved):
Thanks!

rhornung67 (approved):

bmhan12 (approved):

PR #1300, opened 2024-03-22
Arbitrary stride ordering for Array

gunney1:
# Arbitrary stride ordering for `Array`

- This PR is a feature
- It does the following:
  - Move `ArrayIndexer` and its tests from `quest`, where it was initially needed, to `core`.
  - Move the index type template argument in `ArrayIndexer` to the end and make it optional.
  - Add array class constructors that take `ArrayIndexer`.
  - Replace the strides with an `ArrayIndexer` in `ArrayBase`.
  - Add array access example for checking performance degredation.
  - Move support for generic RAJA nested loops from `slic` to `core`.  This code is not specific to `slic` and is used in the indexer performance tests.

The motivations for this feature are:
- To support application native data, we need to support whatever stride ordering the application uses.
- Although we probably will only see row- and column-major ordering, the Blueprint convention allows arbitrary ordering, so we should support it.
- The `ArrayIndexer` is a bit more useful than array strides, because it simplifies laying out new data to match the stride order of existing data, for example, when creating ghost-padded data with the same ordering as ghost-free data.

## Performance checks
I added a performance example (`core_array_perf`) to verify that the changes didn't degrade the array performance.  There should be no changes, so this simply verifies that the compiler can optimize out the higher-level constructs.

The following results are from running the performance test on arrays with 1,200,000 elements each.  The shape in 2D is 1000x1200.  In 3D it is 100x100x120.  The data is arranged in row-major and column-major orders.  They are accessed multiple ways, represented by the bar colors:
- Flat loop, with access through C pointer
- Flat loop, with access through flat index (like C pointer, with high-level code that has to be optimized out)
- Row-major nested loop, with access through multidimensional indices
- Column-major nested loop, with access through multidimensional indices
- Dynamically-ordered nested loop that adjusts to data whose ordering is unknown at compile time.

The data presented is the average over 25 repetitions.  GPU results had little scatter.  Sequential and OpenMP policies seems to have more scatter, so I selected representative plots to show.

Comparing data before and after the chnage confirms that performance wasn't affected by replacing the strides with the `ArrayIndexer` object in `ArrayBase`.  1D results should ideally be the same, so differences could be attributed to the compiler optimization effectiveness or hiccups in the system.  As expected, when the looping order matches the data order, we get the best results.  When they don't match, the results are worse.  The dynamically ordered nesting didn't do well for the sequential policies.  For parallel policies, they generally perform much closer to the ideal ordering (row or column) than the wrong one.

Unexpectedly, flat indexing did better for sequential policies after the change, much closer to the performance of the simple C pointer.  I didn't investigate this, but there may have been something inhibitting full optimization before the change.

The performance test does no computation.  When significant computation are present, the differences would be less significant overall.

### RZAnsel results, with sequential and CUDA policies
Before changes:
ansel 240322_092130
After changes:
ansel 240326_080510

### RZVernal results, with sequential and HIP policies
Before changes:
vernal 240322_092541
After changes:
vernal 240326_074418

### RZGenie results, with sequential and OpenMP policies
Before changes: genie 240322_090635
After changes
genie 240326_074404


src/axom/core/ArrayBase.hpp
```diff
@@ -450,21 +470,24 @@ class ArrayBase
...
+
+    @pre indexer().getStrideOrder() & ArrayStrideOrder::ROW == true
+  */
   template <int UDim>
   AXOM_HOST_DEVICE IndexType offset(const StackArray<IndexType, UDim>& idx) const
   {
     static_assert(UDim <= DIM,
                   "Index dimensions cannot be larger than array dimensions");
-    return numerics::dot_product(idx.begin(), m_strides.begin(), UDim);
+    assert(indexer().getStrideOrder() & ArrayStrideOrder::ROW);
```
publixsubfan, line 489: If `getStrideOrder()` isn't a bitmask, you might wanna just use `==` to clarify intent.

src/axom/core/Array.hpp
```diff
@@ -933,6 +956,20 @@ Array<T, DIM, SPACE>::Array(const axom::StackArray<axom::IndexType, DIM>& shape,
...
 
+//------------------------------------------------------------------------------
+template <typename T, int DIM, MemorySpace SPACE>
+Array<T, DIM, SPACE>::Array(const axom::StackArray<axom::IndexType, DIM>& shape,
+                            const axom::ArrayIndexer<DIM>& indexer,
+                            int allocator_id)
+  : ArrayBase<T, DIM, Array<T, DIM, SPACE>>(shape, indexer)
+  , m_allocator_id(allocator_id)
+{
+  assert(indexer.fastestStrideLength() == 1);
```
publixsubfan, line 967: Is `Array` always expected to be initialized with a min-stride of 1? If so, you might want to document that requirement for the "arbitrary-strides" constructor.

publixsubfan (approved):
LGTM 👍 -- I just had a few comments/suggestions regarding API design.

white238:
Errors from `using` example because it's not easy to get:

```
[ 33%] Building CXX object CMakeFiles/example.dir/example.cpp.o
In file included from /usr/WS1/atk/gitlab-runner/builds/T5vctq-D/005/gitlab/axom/axom/_axom_build_and_test_2024_05_01_16_03_49_lassen-blueos_3_ppc64le_ib_p9-clang_10.0.1.2_cuda/install-lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.2_cuda/examples/axom/using-with-cmake/example.cpp:13:
In file included from /usr/WS1/atk/gitlab-runner/builds/T5vctq-D/005/gitlab/axom/axom/_axom_build_and_test_2024_05_01_16_03_49_lassen-blueos_3_ppc64le_ib_p9-clang_10.0.1.2_cuda/install-lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.2_cuda/include/axom/core.hpp:49:
/usr/WS1/atk/gitlab-runner/builds/T5vctq-D/005/gitlab/axom/axom/_axom_build_and_test_2024_05_01_16_03_49_lassen-blueos_3_ppc64le_ib_p9-clang_10.0.1.2_cuda/install-lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.2_cuda/include/axom/core/execution/nested_for_exec.hpp:138:46: error: use of undeclared identifier 'SYNCHRONOUS'
struct nested_for_exec<CUDA_EXEC<BLOCK_SIZE, SYNCHRONOUS>>
                                             ^
/usr/WS1/atk/gitlab-runner/builds/T5vctq-D/005/gitlab/axom/axom/_axom_build_and_test_2024_05_01_16_03_49_lassen-blueos_3_ppc64le_ib_p9-clang_10.0.1.2_cuda/install-lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.2_cuda/include/axom/core/execution/nested_for_exec.hpp:175:2: error: expected a type
};
 ^
/usr/WS1/atk/gitlab-runner/builds/T5vctq-D/005/gitlab/axom/axom/_axom_build_and_test_2024_05_01_16_03_49_lassen-blueos_3_ppc64le_ib_p9-clang_10.0.1.2_cuda/install-lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.2_cuda/include/axom/core/execution/nested_for_exec.hpp:178:46: error: use of undeclared identifier 'ASYNC'
struct nested_for_exec<CUDA_EXEC<BLOCK_SIZE, ASYNC>>
                                             ^
/usr/WS1/atk/gitlab-runner/builds/T5vctq-D/005/gitlab/axom/axom/_axom_build_and_test_2024_05_01_16_03_49_lassen-blueos_3_ppc64le_ib_p9-clang_10.0.1.2_cuda/install-lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.2_cuda/include/axom/core/execution/nested_for_exec.hpp:215:2: error: expected a type
};
 ^
4 errors generated.
make[2]: *** [CMakeFiles/example.dir/build.make:76: CMakeFiles/example.dir/example.cpp.o] Error 1
```

src/axom/core/ArrayBase.hpp
gunney1: It *is* a bitmask.

src/axom/core/ArrayIndexer.hpp
```diff
@@ -365,8 +396,12 @@ class ArrayIndexer
     os << ')';
     return os;
   }
+
+private:
+  axom::StackArray<T, DIM> m_strides;
+  axom::StackArray<std::uint16_t, DIM> m_slowestDirs;
```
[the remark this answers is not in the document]
gunney1: I wasn't expecting any.

src/axom/core/Array.hpp
gunney1: Yes, `Array` always has min-stride of 1.  Using `ArrayIndexer` or `strides` in constructors does let users specify inconsistent and non-sensical values.  We'd have to catch these errors with asserts.  Would it make sense to instead use constructors that make this kind of error impossible?  This means allowing users to specify only `ArrayStrideOrder`, like @publixsubfan suggested above, or `axom::StackArray<DirType, DIM>& slowestDirs` parameters, to handle all possible arbitrary cases.
```
Array(const axom::StackArray<axom::IndexType, DIM>& shape,
      int allocator_id = axom::detail::getAllocatorID<SPACE>(),
      ArrayStrideOrder strideOrder = ArrayStrideOrder::ROW);

template <typename DirType>
Array(const axom::StackArray<axom::IndexType, DIM>& shape,
      int allocator_id,
      const axom::StackArray<DirType, DIM>& slowestDirs);
```

src/axom/core/ArrayBase.hpp
```diff
@@ -334,26 +356,24 @@ class ArrayBase
     return m_shape;
   }
 
+  /// \brief Returns the indexer of the Array
+  AXOM_HOST_DEVICE const ArrayIndexer<DIM>& indexer() const
```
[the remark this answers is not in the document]
gunney1: Interesting idea.  I think that if we adopt the terminology from `mdspan`, maybe we should also rename `ArrayIndexer` to something like `MDMapping`.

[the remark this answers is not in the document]
gunney1: I'm going with `MDMapping`.

src/axom/core/Array.hpp
gunney1: Done.

src/axom/core/Array.hpp
```diff
@@ -161,6 +165,25 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM, SPACE>>
...
+          ArrayIndexer<DIM>{shape, ArrayStrideOrder::COLUMN});
+
+    Example of Array where j is fastest and k is slowest:
+        Array<3, int> ar(
+          shape,
+          ArrayIndexer<DIM>{shape, {2, 0, 1}});
+  */
+  Array(const axom::StackArray<axom::IndexType, DIM>& shape,
+        const axom::ArrayIndexer<DIM>& indexer,
+        int allocator_id = axom::detail::getAllocatorID<SPACE>());
```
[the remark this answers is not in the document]
gunney1: Done.

src/axom/core/MDMapping.hpp
```diff
@@ -106,23 +107,42 @@ class ArrayIndexer
...
+    initializeShape(shape, arrayStrideOrder);
+  }
 
   /*!
     @brief Initialize for row- or column-major indexing.
     @param [in] shape Shape of the array
-    @param [in] arrayStrideOrder An order indicator from
-                ArrayStrideOrder.
+    @param [in] arrayStrideOrder An order indicator,
+      not ArrayStrideOrder::ARBITRARY.
```
agcapps, line 133: Clarify to say "An order indicator other than ArrayStrideOrder::ARBITRARY"

src/axom/core/MDMapping.hpp
```diff
@@ -106,23 +107,42 @@ class ArrayIndexer
...
     @param [in] fastestStrideLength Stride in the fastest
-                direction.
+                chnaging direction.
```
agcapps, line 135: typo: correct to "fastest-changing"

agcapps: Please change Doxygen for all other occurrences of `fastestStrideLength` throughout this PR to say "fastest-changing direction"

src/axom/core/examples/CMakeLists.txt
```diff
@@ -61,3 +61,36 @@ if(AXOM_ENABLE_TESTS)
...
+
+axom_add_executable(
+    NAME        core_array_perf_ex
+    SOURCES     core_array_perf.cpp
+    OUTPUT_DIR  ${EXAMPLE_OUTPUT_DIRECTORY}
+    DEPENDS_ON  core
+    FOLDER      axom/core/examples )
+
+if(AXOM_ENABLE_TESTS)
+    # Run the marching cubes example on N ranks for each enabled policy
```
agcapps, line 73: You may want to change this comment, unless `core_array_perf` is the marching cubes example.

src/axom/core/examples/CMakeLists.txt
```diff
@@ -61,3 +61,36 @@ if(AXOM_ENABLE_TESTS)
...
+            NAME    "core_array_perf_2d_${_pol}"
+            COMMAND core_array_perf_ex --policy ${_pol} --shape 1000 1200 -r 10)
+        axom_add_test(
+            NAME    "core_array_perf_3d_${_pol}"
+            COMMAND core_array_perf_ex --policy ${_pol} --shape 100 100 120 -r 10)
+        if(NOT RAJA_FOUND)
+            axom_add_test(
+                NAME    "core_array_perf_4d_${_pol}"
+                COMMAND core_array_perf_ex --policy ${_pol} --shape 50 20 30 6 )
+        endif()
```
agcapps, line 94: Why only add the 4D test if RAJA is not found?

src/axom/core/execution/nested_for_exec.hpp
```diff
@@ -3,8 +3,8 @@
 //
 // SPDX-License-Identifier: (BSD-3-Clause)
 
-#ifndef AXOM_MINT_STRUCTURED_EXEC_HPP_
-#define AXOM_MINT_STRUCTURED_EXEC_HPP_
+#ifndef AXOM_CORE_STRUCTURED_EXEC_HPP_
+#define AXOM_CORE_STRUCTURED_EXEC_HPP_
```
agcapps, line 7: Since the file got renamed, please change to `AXOM_CORE_NESTED_FOR_EXEC_HPP_` (and of course at the file's end).

agcapps (approved):
Approved pending my comments (mainly documentation details).

src/axom/core/MDMapping.hpp
gunney1: Done.  Thanks.

src/axom/core/MDMapping.hpp
gunney1: Done.

src/axom/core/examples/CMakeLists.txt
gunney1: We require some RAJA tiling, and RAJA specifically supports up to 3D for it.  Without RAJA, the test uses C-style nested loops, so no there is no restriction.

src/axom/core/execution/nested_for_exec.hpp
gunney1: Done.  Thanks.

src/axom/core/examples/CMakeLists.txt
gunney1: Right.  Thanks.

src/axom/core/examples/core_array_perf.cpp
```diff
@@ -0,0 +1,975 @@
...
+#include "axom/core/MDMapping.hpp"
+#include "axom/core/execution/for_all.hpp"
+#include "axom/core/execution/nested_for_exec.hpp"
+#include "axom/core/execution/runtime_policy.hpp"
+#include "axom/core/memory_management.hpp"
+#include "axom/core/utilities/System.hpp"
+#include "axom/core/utilities/Timer.hpp"
+#include "axom/core/AnnotationMacros.hpp"
+
+#include "axom/fmt/format.h"
```
kennyweiss, line 22: Please include `fmt.hpp` instead of `fmt/format.h` -- it adds some necessary defines.

```suggestion
#include "axom/fmt.hpp"
```

src/axom/core/examples/core_array_perf.cpp
```diff
@@ -0,0 +1,975 @@
...
+      If dataSlowestDirections is specified, it must match length of shape.
+      If neither is specified, default to row-major ordering.
+    */
+    if(!dataSlowestDirections.empty())
+    {
+      if(dataSlowestDirections.size() != shape.size())
+      {
+        std::cerr << "slowestDimension size (" << dataSlowestDirections.size()
+                  << ") must match shape size (" << shape.size() << ")."
+                  << std::endl;
```
kennyweiss, line 113: Minor: Please consider increasing the slic and fmt usage in this example.

src/axom/core/examples/core_array_perf.cpp
kennyweiss, line 1: Thanks for adding this comprehensive performance example!

Now that we have caliper and adiak, please consider adding performance annotations to capture the performance.

kennyweiss (approved):
Thanks @gunney1

src/axom/core/examples/core_array_perf.cpp
gunney1: I'll use `fmt` more but I don't think I can use `slic` here because `core` can't depend on `slic`.

src/axom/core/examples/core_array_perf.cpp
gunney1: Done.

src/axom/core/examples/core_array_perf.cpp
kennyweiss: Good call.  This isn't required for the PR, so don't feel obligated to make the change.

`core` can't depend on `slic`, but I don't see any reason that `core`'s examples can't depend on `slic`
(obviously, the example should check that `slic` is enabled)

src/axom/core/examples/core_array_perf.cpp
gunney1: Done.

PR #1306, opened 2024-04-01
Revert usage of `tee`

white238:
I'd say this was an elaborate April Fool's Joke but it's just sad.

`tee` does not forward the error code of the previous command like I assumed and clearly didn't check. My bad.

```
$ ./error.sh; echo $?
~~~~~~~~ test about to error
1
$ ./error.sh | tee output; echo $?
~~~~~~~~ test about to error
0
```

There are some very specific to bash ways of handling this, which we can't rely on, or I can jump through some pipe'ing hoops in Python. I'll do the later later.


kennyweiss (approved):

agcapps (approved):
Thank you for fixing.

PR #1308, opened 2024-04-02
Fix a link issue with Clang 16 + CUDA

publixsubfan:
A user is seeing an issue with a "missing `__ppc_trap()` symbol" link error when using Clang 16 + CUDA 11.2. Replace this with `assert(false)`, which should serve a similar purpose in a debug build.


kennyweiss (approved):
Thanks @publixsubfan

rhornung67 (approved):

gunney1 (approved):
Thank you @publixsubfan

PR #1311, opened 2024-04-05
Add closest_point operator for Segment

adayton1:
# Summary

- This PR is a feature
- It does the following:
  - Adds closest_point operator for Segment at the request of myself

### DIFF
--- a/src/axom/primal/operators/closest_point.hpp
+++ b/src/axom/primal/operators/closest_point.hpp
@@ -49,7 +49,7 @@ AXOM_HOST_DEVICE inline Point<T, NDIMS> closest_point(const Point<T, NDIMS>& P,
   using detail::isLeq;
 
   constexpr T ZERO {0.};
-  constexpr T ONE  {0.};
+  constexpr T ONE  {1.};
 
   const PointType& A = seg[0];
   const PointType& B = seg[1];
--- a/src/axom/primal/operators/closest_point.hpp
+++ b/src/axom/primal/operators/closest_point.hpp
@@ -15,6 +15,7 @@
 #define AXOM_PRIMAL_CLOSEST_POINT_HPP_
 
 #include "axom/primal/geometry/Point.hpp"
+#include "axom/primal/geometry/Segment.hpp"
 #include "axom/primal/geometry/Triangle.hpp"
 #include "axom/primal/geometry/Sphere.hpp"
 #include "axom/primal/geometry/OrientedBoundingBox.hpp"
@@ -23,6 +24,50 @@ namespace axom
 {
 namespace primal
 {
+
+/*!
+ * \brief Computes the closest point from a point, P, to a given segment.
+ *
+ * \param [in] P the query point
+ * \param [in] seg user-supplied segment
+ * \param [in] EPS fuzz factor for equality comparisons
+ * \return cp the closest point from a point P and a segment
+ */
+template <typename T, int NDIMS>
+AXOM_HOST_DEVICE inline Point<T, NDIMS> closest_point(const Point<T, NDIMS>& P,
+                                                      const Segment<T, NDIMS>& seg,
+                                                      double EPS = 1E-8) {
+  using PointType = Point<T, NDIMS>;
+  using VectorType = Vector<T, NDIMS>;
+
+  using detail::isGeq;
+  using detail::isLeq;
+
+  const PointType& A = seg[0];
+  const PointType& B = seg[1];
+
+  const VectorType AB(A, B);
+  const T squaredNormAB = AB.squared_norm();
+
+  // Check if segment is degenerate
+  if (isLeq(squaredNormAB, T(0), EPS)) {
+    return A;
+  }
+
+  // Compute length of the projection of AP onto AB
+  const double t = VectorType(A, P).dot(AB) / squaredNormAB;
+
+  if (isLeq(t, T(0), EPS)) {
+    return A;
+  }
+  else if (isGeq(t, T(1), EPS)) {
+    return B;
+  }
+  else {
+    return seg.at(t);
+  }
+}
+
 /*!
  * \brief Computes the closest point from a point, P, to a given triangle.
  *
--- a/src/axom/primal/operators/closest_point.hpp
+++ b/src/axom/primal/operators/closest_point.hpp
@@ -30,12 +30,17 @@ namespace primal
  *
  * \param [in] P the query point
  * \param [in] seg user-supplied segment
+ * \param [out] t location along the line segment
  * \param [in] EPS fuzz factor for equality comparisons
  * \return cp the closest point from a point P and a segment
+ *
+ * \note t \f$ \in [0, 1] \f% represents the fraction of the way from A to B,
+ *  with 0 corresponding to A and 1 corresponding to B.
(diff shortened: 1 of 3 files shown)


src/axom/primal/operators/closest_point.hpp
```diff
@@ -23,6 +24,50 @@ namespace axom
...
+ *
+ * \param [in] P the query point
+ * \param [in] seg user-supplied segment
+ * \param [in] EPS fuzz factor for equality comparisons
+ * \return cp the closest point from a point P and a segment
+ */
+template <typename T, int NDIMS>
+AXOM_HOST_DEVICE inline Point<T, NDIMS> closest_point(const Point<T, NDIMS>& P,
+                                                      const Segment<T, NDIMS>& seg,
+                                                      double EPS = 1E-8) {
```
kennyweiss, line 39: It might also be useful to return the `t` value as an out parameter, which users could ignore if they don't need it. 

It shouldn't have any overhead since we're already computing it.

src/axom/primal/operators/closest_point.hpp
```diff
@@ -23,6 +24,50 @@ namespace axom
...
+
+  using detail::isGeq;
+  using detail::isLeq;
+
+  const PointType& A = seg[0];
+  const PointType& B = seg[1];
+
+  const VectorType AB(A, B);
+  const T squaredNormAB = AB.squared_norm();
+
```
kennyweiss, line 51: Suggestion (minor): 

It might be slightly clearer to use
```cpp
constexpr T ZERO {0.};
constexpr T ONE {1.};
```
for the comparisons to zero and one

or perhaps to change `T(0}` to either `T {0.}` or `static_cast<T>(0.)`

src/axom/primal/operators/closest_point.hpp
```diff
@@ -23,6 +24,50 @@ namespace axom
...
+  if (isLeq(t, T(0), EPS)) {
+    return A;
+  }
+  else if (isGeq(t, T(1), EPS)) {
+    return B;
+  }
+  else {
+    return seg.at(t);
+  }
```
kennyweiss, line 68: Minor: In case it matters, it might be possible to delay the division until the final `else` by comparing to the unnormalized dot product magnitude, e.g. something like:

```cpp
const auto AB = B-A;
const T ab_sqnorm = AB.squared_norm();
const double t_unnorm = (P-A).dot(AB);  // delay the normalization

if (isLeq(t_unnorm , ZERO), EPS)) 
{
  return A;
}
else if (isGeq(t_unnorm, ab_sqnorm, EPS))  // compare to t_unnorm rather than 1
{
  return B;
}
else 
{
  return seg.at( t_unnorm / ab_sqnorm );
}
```

(I think that should work, and the unit tests can help w/ getting it right)

kennyweiss (approved):
Thanks @adayton1

src/axom/primal/operators/closest_point.hpp
adayton1: Good idea. Done.

src/axom/primal/operators/closest_point.hpp
adayton1: Done.

src/axom/primal/operators/closest_point.hpp
adayton1: Done.

src/axom/primal/tests/primal_closest_point.cpp
```diff
@@ -11,6 +11,118 @@
...
+  using QPoint = primal::Point<CoordType, DIM>;
+  using QSegment = primal::Segment<CoordType, DIM>;
+
+  QPoint A({0.0, 1.0, 1.0});
+  QPoint B({1.0, 0.0, 0.0});
+  QSegment S(A, B);
+  double t;
+
+  // Query point is perpendicular to the midpoint of AB
+  EXPECT_TRUE(primal::closest_point(QPoint({0.0, 0.0, 0.5}), S, t, EPS) == QPoint::lerp(A, B, 0.5));
```
adayton1, line 118: Can someone sanity check this particular test case?

src/axom/primal/operators/closest_point.hpp
```diff
@@ -23,6 +24,84 @@ namespace axom
...
+ * \param [in] EPS fuzz factor for equality comparisons
+ * \return cp the closest point from a point P and a segment
+ */
+template <typename T, int NDIMS>
+AXOM_HOST_DEVICE inline Point<T, NDIMS> closest_point(const Point<T, NDIMS>& P,
+                                                      const Segment<T, NDIMS>& seg,
+                                                      double EPS = 1E-8)
+{
+  T t;
+  return closest_point(P, seg, t, EPS);
```
kennyweiss, line 102: 👍

src/axom/primal/tests/primal_closest_point.cpp
kennyweiss: Seems reasonable to me
$A := (0,1,1)$, $B := (1,0,0)$, $Q := (0,0,.5)$
$AB  := <1, -1, -1>$
$AQ :=  <0,-1,-.5>$
$AB \cdot AB = 3$
$AB \cdot AQ = 1.5$

$t = (AB \cdot AQ)/(AB \cdot AB) = .5$

kennyweiss: I suppose you can also try comparing the value against a segment with the reverse orientation, which should give you the same answer:

```
  QSegment S_reverse(B, A);
  EXPECT_TRUE(primal::closest_point(QPoint({0.0, 0.0, 0.5}), S_reverse, t, EPS) == QPoint::lerp(A, B, 0.5));
```

Edit: It might not be a bad idea to test the reversed segment in all cases.

src/axom/primal/operators/closest_point.hpp
```diff
@@ -23,6 +24,84 @@ namespace axom
...
+ * \return cp the closest point from a point P and a segment
+ *
+ * \note t \f$ \in [0, 1] \f% represents the fraction of the way from A to B,
+ *  with 0 corresponding to A and 1 corresponding to B.
+ */
+template <typename T, int NDIMS>
+AXOM_HOST_DEVICE inline Point<T, NDIMS> closest_point(const Point<T, NDIMS>& P,
+                                                      const Segment<T, NDIMS>& seg,
+                                                      T& t,
+                                                      double EPS = 1E-8)
```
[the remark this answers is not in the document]
kennyweiss: Good question!

We have wanted to centralize this in a common file (https://github.com/LLNL/axom/blob/develop/src/axom/primal/constants.hpp) for a long time, but have so far only added one constant there.

kennyweiss (approved):
Thanks for adding these @adayton1 and for optimizing/updating the algorithm following review suggestions!

BradWhitlock (approved):
Looks good.

rhornung67 (approved):
Nice work @adayton1 Thanks for the contribution.

src/axom/primal/tests/primal_closest_point.cpp
adayton1: I added tests for all the reversed segments.

Issue #1309, opened 2024-04-03
Request for additional Primal::Vector methods


adayton1:
- [ ] normalize (in place)
- [ ] cross and cross_product (non-static methods that use the current vector as the first operand)

PR #1313, opened 2024-04-08
Draft: Add missing functions in primal

adayton1:
# Summary

- This PR is a feature
- It does the following:
  - Adds missing AXOM_HOST_DEVICE specifiers to several methods in Point, Segment, and Vector classes
  - Adds length_squared method to the Segment class
  - Adds cross, normalize, and unitize methods to the Vector class
  - Adds squared_distance operator

Resolves https://github.com/LLNL/axom/issues/1309


src/axom/primal/geometry/Vector.hpp
```diff
@@ -347,6 +357,18 @@ class Vector
...
+   * \brief In-place normalization of the vector.
+   */
+  AXOM_HOST_DEVICE
+  void normalize();
+
+  /*!
+   * \brief In-place unitization (normalization) of the vector.
+   */
+  AXOM_HOST_DEVICE
+  void unitize();
```
kennyweiss, line 370: Suggestion: Is there a good reason to have two functions that do the same thing? 
If not, my preference would be to only have `normalize()` (which is the conventional way to refer to this operation).

src/axom/primal/geometry/Vector.hpp
```diff
@@ -516,6 +539,38 @@ AXOM_HOST_DEVICE inline void Vector<T, NDIMS>::negate()
...
+
+    // Set the first component to 1 and all others to 0
+    m_components[0] = static_cast<T>(1.);
+
+    for(int i = 1; i < NDIMS; ++i)
+    {
+      m_components[i] = static_cast<T>(0.);
+    }
+  }
+}
```
kennyweiss, line 566: Should `Vector::unitVector()` be implemented in terms of this one?

Here's the current implementation:
https://github.com/LLNL/axom/blob/379db95ceed3ce8e90145993a0d9c33f8103f330/src/axom/primal/geometry/Vector.hpp#L489-L507

src/axom/primal/geometry/Vector.hpp
```diff
@@ -537,6 +592,13 @@ AXOM_HOST_DEVICE inline T Vector<T, NDIMS>::dot(const Vector<T, NDIMS>& vec) con
   return dot_product(*this, vec);
 }
 
+//------------------------------------------------------------------------------
+template <typename T, int NDIMS>
+AXOM_HOST_DEVICE inline T Vector<T, NDIMS>::cross(const Vector<T, NDIMS>& vec) const
+{
+  return cross_product(*this, vec);
+}
```
kennyweiss, line 600: We might need to be careful about this one -- the current `cross_product` overloads are defined as:
* `cross_product(Vector3, Vector3) -> Vector3`
* `cross_product(Vector2, Vector2) -> Vector3`  // where the x- and y- components are 0

https://github.com/LLNL/axom/blob/379db95ceed3ce8e90145993a0d9c33f8103f330/src/axom/primal/geometry/Vector.hpp#L567-L586

src/axom/primal/geometry/Vector.hpp
```diff
@@ -700,6 +763,7 @@ std::ostream& operator<<(std::ostream& os, const Vector<T, NDIMS>& vec)
 
 //------------------------------------------------------------------------------
 template <typename T, int NDIMS>
+AXOM_HOST_DEVICE
 inline Vector<T, NDIMS> Vector<T, NDIMS>::make_vector(const T& x,
                                                       const T& y,
                                                       const T& z)
```
kennyweiss, line 769: Side note (outside the scope of this PR):

The `make_vector` utility function was introduced before we had an initializer list constructor for a `primal::Vector`. I'm not sure if we still need the `make_vector` helper function.  (and similarly for `Point::make_point`)

kennyweiss (changes requested):
Thanks for adding these @adayton1 -- looks good to me, but I'd have a bit more confidence if there were a few unit tests, especially for the new `Vector::cross` implementation in 2D and 3D.

I'd also like to hear more about the need for both `Vector::normalize()` and `Vector::unitize()`

adayton1:
The portion of changes that are ready are being pulled into this merge request: https://github.com/LLNL/axom/pull/1422

adayton1:
I'm going to pull out changes into different pull requests.

PR #1319, opened 2024-04-11
Improvements to annotations API

### LINKED ISSUE #1314: Add a CLI11 validator for caliper annotations
This would make it easier to use caliper in an axom example.

The following snippet has been used in a few places
```cpp
    app.add_option("--caliper", annotationMode)
      ->description(
        "caliper annotation mode. Valid options include 'none' and 'report'. "
        "Use 'help' to see full list.")
      ->capture_default_str()
      ->check([](const std::string& mode) -> std::string {
        if(mode == "help")
        {
          std::cerr << "Valid caliper modes are:\n"
                    << axom::utilities::annotations::detail::mode_help_string()
                    << std::endl;
        }
        return axom::utilities::annotations::detail::is_mode_valid(mode)
          ? ""
          : fmt::format(
              "'{}' invalid caliper mode. "
              "Run with '--caliper help' to see all valid options",
              mode);
      });
```

Per PR suggestion on #1312 

See: https://cliutils.github.io/CLI11/book/chapters/validators.html

### LINKED ISSUE #1315: Should the `AXOM_ANNOTATE_END` macro require a `name` parameter?
The `AXOM_ANNOTATE_BEGIN` and `AXOM_ANNOTATE_END` macros were added in https://github.com/LLNL/axom/issues/1312

During the review, @agcapps suggested that `AXOM_ANNOTATE_END` should not take a `name` parameter since it is not used in the implementation and can lead to code drift.

On the other hand, it can be useful to label the end using the same name as the beginning, in which case, we might prefer to keep the `name` parameter, rather than leaving it ambiguous or requiring a comment.

A third possibility might be to use the `name` parameter for internal bookkeeping within axom.

We should discuss this as a group.

kennyweiss:
# Summary

- This PR is a feature/enhancement
- It improves our usage of `caliper` and `adiak` in the annotations API based on an offline code review of #1312 from @daboehme (Thanks!)
- Fixes #1314 by adding a custom CLI Validator for caliper modes
- Fixes #1315 by using a caliper function call that uses the `name` passed to `axom::utility::annotations::end(name)`

### DIFF
--- a/src/axom/core/CMakeLists.txt
+++ b/src/axom/core/CMakeLists.txt
@@ -24,6 +24,7 @@ set(core_headers
     utilities/About.hpp
     utilities/Annotations.hpp
     utilities/BitUtilities.hpp
+    utilities/CommandLineUtilities.hpp
     utilities/FileUtilities.hpp
     utilities/RAII.hpp
     utilities/StringUtilities.hpp
--- a/src/axom/core/utilities/Annotations.cpp
+++ b/src/axom/core/utilities/Annotations.cpp
@@ -18,13 +18,6 @@
 
 #include <set>
 
-#ifdef AXOM_USE_CALIPER
-namespace cali
-{
-extern Attribute region_attr;
-}
-#endif
-
 namespace axom
 {
 namespace utilities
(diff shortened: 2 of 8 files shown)


agcapps (approved):
Thank you, Kenny.  Much nicer.

BradWhitlock (approved):
The new changes look cleaner.

PR #1325, opened 2024-04-20
Bugfix/gunney/marching cubes large mesh

gunney1:
# Use bigger integer type for number of facets per parent cell

- This PR is a bug fix
It changes the type of the number-of-facets-per-parent-cell from `uint16_t` to `IndexType`.

The 16-bit type was problematic because, the sum scan on it apparently uses 16-bit adds instead of promoting the items to the output `IndexType`.  The array doesn't need to be bigger than 16-bit, because it only has to store values in [0,5].  But this change is necessary to prevent the sum from rolling over at 2^16.  On rzgenie and rzansel, the data did get promoted to 32-bit before adding, so the bug didn't affect them.  RAJA issue https://github.com/LLNL/RAJA/issues/1631 should address this inconsistency.


gunney1:
Related tohttps://github.com/LLNL/RAJA/issues/1631

agcapps (approved):

white238 (approved):

PR #1326, opened 2024-04-22
Add tests for closest_point on triangle operator

### LINKED ISSUE #1307: Add test cases for finding the closest point on a triangle
My application is getting fairly different answers in some cases when using the closest_point query. I noticed there are no test cases for the triangle case. Having a few test cases would definitely increase my confidence that closest_point is actually doing the right thing.

Also, maybe label the vertices in the documentation for the function with A, B, and C (in addition to 0, 1, 2) to match comments in the source code like "Check if P in vertex region outside A"

adayton1:
# Summary

- This PR is a bugfix:
- It does the following:
  - Modifies closest_point on triangle query to handle degenerate triangles
  - Adds tests for closest_point on triangle query at the request of myself
  - Fixes #1307

### DIFF
--- a/src/axom/primal/operators/closest_point.hpp
+++ b/src/axom/primal/operators/closest_point.hpp
@@ -187,22 +187,6 @@ AXOM_HOST_DEVICE inline Point<T, NDIMS> closest_point(const Point<T, NDIMS>& P,
   }
 
   //----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


bmhan12 (approved):

agcapps (approved):
Thank you @adayton1 !

PR #1328, opened 2024-04-25
Fixes/improvements for closest_point to Segment operator

adayton1:
# Summary

- This PR is a refactoring and bugfix
- It does the following:
  - Removes ambiguous overload of closest_point to Segment operator
  - Updates default tolerance for closest_point to Segment operator (1e-12 -> PRIMAL_TINY = 1.0e-50)
  - Refactors closest_point to Segment operator to only punt if it needs to
  - Uses closest_point to Segment operator in squared_distance to Segment operator
  - Adds actual degenerate test case for closest_point to Segment operator


agcapps (approved):
Thanks, @adayton1 !

bmhan12 (approved):

PR #1330, opened 2024-05-02
Add a reflectPoint method to Plane

adayton1:
# Summary

- This PR is a feature
- It does the following:
  - Adds reflectPoint method to Plane at the request of myself
  - Makes sure all Plane methods (except I/O) are device accessible
  - Removes redundant (and potentially harmful) "inline" keyword on methods in Plane


rhornung67 (approved):

agcapps (approved):
Thanks.  Looks very useful.

PR #1331, opened 2024-05-03
Add default constructor and validity check to Plane

adayton1:
# Summary

- This PR is a feature
- It does the following:
  - Adds a default constructor to the Plane class at the request of myself
  - Since there is not an "ideal" default Plane, make it obviously invalid and add an isValid method
  - Make the normalize step optional (normalization is not essential for the plane to behave correctly)

A default constructor is needed for code like the following to compile:

```
Plane<double, 3>* planes = new Plane<double, 3>[10];

for (int i = 0; i < 10; ++i) {
   planes[i] = Plane<double, 3>(normal, point);
}
```

I'll note for future discussion that my application handles normalization of the zero vector differently than Axom. We return the zero vector while Axom returns {1.0, 0.0, 0.0}. I'm not sure what is "best", but making the normalization step in Plane construction optional allows us to apply our own normalization method, which avoids answer changes.


src/axom/primal/geometry/Plane.hpp
```diff
@@ -108,29 +108,42 @@ class Plane
...
    *
    * \param [in] normal the supplied plane normal, \f$ \mathcal{N} \f$
    * \param [in] x a specified point that the plane passes through, \f$ x \f$
+   * \param [in] normalize if true, normalize the given vector (default: true)
    *
-   * \note The supplied normal will be normalized, such that, the Plane's normal
-   *  will always be a unit normal.
+   * \note The Plane's normal should always be a unit normal, so the
+   *  normalize flag should only be set to false when the supplied
+   *  normal has already been normalized.
```
kennyweiss, line 128: > The Plane's normal should always be a unit normal, so the normalize flag should only be set to false when the supplied normal has already been normalized

This might be a good time to question whether a plane's normal should always be normalized. I don't recall why that was the case, but w/ the new parameter, I'd suggest the default should be `false` and a user can normalize if desired.

src/axom/primal/geometry/Plane.hpp
```diff
@@ -209,6 +222,19 @@ class Plane
   AXOM_HOST_DEVICE int getOrientation(const PointType& x,
                                       double TOL = 1.e-9) const;
 
+  /*!
+   * \brief Simple check for validity of a Plane
+   *
+   * Check that the normal is a unit vector (which also handles the zero
+   * vector case).
```
kennyweiss, line 229: See my comment above -- but I think we should revisit the notion that a plane must have a normalized unit vector to be valid.

Perhaps it should be valid if the normal is not the zero vector (perhaps up to some epsilon) ?

src/axom/primal/geometry/Plane.hpp
kennyweiss: Specifically, I'd think the if the normal vector for a plane is $({10,0,0})$, then normalizing it $({1,0,0})$ will change the `m_offset`, but both would be valid planes.

kennyweiss (commented):
Thanks for the improvement @adayton1
I like the new knob for normalizing the normal vector.

Before merging this change, I'd like the group to consider what makes a plane valid/invalid.
I'd suggest that a plane is invalid if it's normal is the zero vector (possibly within EPS), and that planes w/ non-unit normals are, in general, still valid.

src/axom/primal/geometry/Plane.hpp
adayton1: I'm fine with the notion of a valid plane being one that has a non-zero vector and removing the requirement that it be a unit vector.

kennyweiss: We'd have to be careful in a few places within axom and update the RELEASE_NOTES, 
but I don't think `primal::Plane` is that widely used (at least internally)

adayton1: Are you suggesting we also change the default behavior to be `normalize = false`?

src/axom/primal/geometry/Plane.hpp
bmhan12: The blame commit history shows the reasoning behind why normals were always normalized: 
https://github.com/LLNL/axom/commit/74d1cb26c9a9767272c9464f86120383e0a0fd4f

It looks like the intention back then was to make the normalization optional.
Sounds like the option was removed and made mandatory to resolve issues with ambiguous constructors in the past, back when the Plane constructors used to represent `Vector` and `Point` with pointers.

bmhan12 (approved):

src/axom/primal/geometry/Plane.hpp
kennyweiss: Thanks @bmhan12 for digging into the history. 
I'm reading it slightly differently though:

>    * \note The Plane constructor will always normalize the supplied normal
   *  unless the caller indicates that the supplied normal is a unit normal
   *  via the third (optional) argument to the constructor.

Sounds like the expectation was for the normal to be a unit vector, but to avoid the extra `normalize` call when the user claims the provided normal is already a unit vector.

My preference would be to not require the normal vector to be a unit vector -- i.e. the normalization should be opt-in. And a plane should be valid as long as the normal is not the zero vector.

src/axom/primal/geometry/Plane.hpp
kennyweiss: > Are you suggesting we also change the default behavior to be normalize = false?

Yes. This changes the default behavior, but I don't see a good reason to require unit vectors by default (which involves extra computational work and might affect answers).

kennyweiss: We could make this change in a follow-up PR if you feel uneasy about too many changes in a single PR, or if the team wants to discuss this further.

adayton1: I changed the isValid check to only make sure the normal vector is not zero. However, I left the default behavior the same (it will still be normalized by default). I'll change that in a future merge request.

src/axom/primal/geometry/Plane.hpp
adayton1: I updated the documentation to say that the normal is not required to be a unit vector.

adayton1:
I've made the following changes:
* isValid only checks that the normal vector is non-zero
* The requirement for the normal to be a unit vector has been removed and the doxygen has been updated accordingly
* The default behavior to normalize the given vector has NOT changed (but may change in a future pull request)

src/axom/primal/geometry/Plane.hpp
```diff
@@ -209,6 +214,17 @@ class Plane
...
+  /*!
+   * \brief Simple check for validity of a Plane
+   *
+   * Check that the normal is not the zero vector.
+   *
+   * \param [in] TOL user-supplied tolerance. Optional. Default is 1.0e-50.
+   *
+   * \return True, if the Plane is valid, False otherwise
+   */
+  AXOM_HOST_DEVICE bool isValid(T TOL = static_cast<T>(PRIMAL_TINY)) const;
```
kennyweiss, line 226: Should `T` be a `double`? 
We'd likely have a problem here if `T` were not floating point (`float` or `double`), and the tolerances in other functions, e.g. `getOrientation()`

kennyweiss (approved):
Thanks @adayton1 !

src/axom/primal/geometry/Plane.hpp
adayton1: isNearlyEqual expects all arguments (including the tolerance) to be the same type. Should they all be doubles or should they all be type T in Plane<T, NDIMS>? If Plane held integers in the current implementation, the tolerance would be 0, which would actually make sense. Maybe I'm thinking about it incorrectly, though.

adayton1: What would you like me to do?

kennyweiss: I was thinking the signature could be:
```cpp
AXOM_HOST_DEVICE bool isValid(double TOL = PRIMAL_TINY) const;
```
and if necessary, if could be `static_cast<T>` when calling `isNearlyEqual`.

adayton1: OK thanks. Done.

PR #1345, opened 2024-05-21
Convert Python scripts to use argparse

### LINKED ISSUE #1268: Replace optparse in llnl_scripts Python scripts
I went to make some changes to command line arguments in `build_devtools.py` . Looking at the documentation for `optparse` I find:

    Deprecated since version 3.2: The optparse module is deprecated and will not be developed further;
    development will continue with the argparse module.

I wanted to allow multiple uses of the `--spec` argument since `spack/specs.yaml` file can lists multiple specs. The `get_specs_for_current_machine` function prefixes each spec with `%` with the line

    specs = ['%' + spec for spec in specs]

This is not being done for the `--spec` command line argument and causes problems. It runs but is not building what I expected. It eventually ends but did not create the `.cmake` files.

    [ERROR: No host configs found at /g/g14/taylor/ws/axom-tpl/toss_4_x86_64_ib/2024_02_01_10_20_13]
    [ERROR: Build and test of src vs tpls test failed.]

I also ran into a problem where i used `-spec` instead of `--spec`. This did not report an error but ran. It ignored the spec I gave it and instead used `pec`.  A result of stacking single letter options I assume.

    [Building and testing tpls for specs: 
    pec
    ]

ltaylor16:
# Convert Python scripts to use argparse

- This PR is a Python module update
- It does the following:
  - Replaces deprecated module optparse with argparse.
  - Fixes #1268

### DIFF
--- a/scripts/llnl_scripts/archive_job.py
+++ b/scripts/llnl_scripts/archive_job.py
@@ -1,5 +1,5 @@
 #!/bin/sh
-"exec" "python" "-u" "-B" "$0" "$@"
+"exec" "python3" "-u" "-B" "$0" "$@"
 
 # Copyright (c) 2017-2024, Lawrence Livermore National Security, LLC and
 # other Axom Project Developers. See the top-level LICENSE file for details.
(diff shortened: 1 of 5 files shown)


agcapps (approved):
Thank you for the update.

kennyweiss (approved):
Thanks @ltaylor16 !

PR #1369, opened 2024-06-25
Polyhedron centroid

### LINKED ISSUE #1348: Feature require: polyhedron mass center
At present, the polyhedron's centroid is calculated as the mean position of its vertices. Would it be feasible to consider offering a feature that calculates the actual center of mass?

bmhan12:
This PR:
- Changes `Polyhedron::centroid()` function to return center of mass of the polyhedron.
- Adds `Polyhedron::vertexMean()`, returns average of polyhedron's vertices (previous `centroid` functionality).
- Adds `Polyhedron::moments()`, returns the volume and centroid of the polyhedron, the 0th and 1st moments respectively.
  - Based on PolyClipper's [moments ](https://github.com/LLNL/PolyClipper/blob/3091a15aba5ad685afdaf19016a79673b58308da/src/polyclipper3dImpl.hh#L183-L231)function. 
  - `centroid()` and `signedVolume()` delegate to `moments()`.

Resolves #1348

### DIFF
--- a/src/axom/primal/geometry/Polyhedron.hpp
+++ b/src/axom/primal/geometry/Polyhedron.hpp
@@ -521,6 +521,9 @@ public:
     VectorType centroid_vector;
     PointType centroid_point;
 
+    VectorType centroid_vector_squared;
+    PointType centroid_point_squared;
+
     if(!isValid() || hasDuplicateVertices())
     {
       return retVol;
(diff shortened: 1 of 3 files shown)


src/axom/primal/geometry/Polyhedron.hpp
```diff
@@ -503,33 +503,39 @@ class Polyhedron
...
    * \note Function is based off moments() in Mike Owen's PolyClipper.
    *
    * \pre polyhedron vertex neighbors are defined, and polyhedron is 3D
    *
    * \sa volume()
+   * \sa centroid()
    */
   AXOM_HOST_DEVICE
-  double signedVolume() const
+  void moments(double& volume, PointType& centroid) const
```
kennyweiss, line 520: Possible optimization:

It seems like this implementation will do extra computation in cases where we only want the volume but not the centroid
(Note that the centroid calculation always needs the volume).

 As an optimization, would it make sense to have an extra boolean flag `should_compute_centroid` (possibly defaulting to true) and guard the few operations that only relate to the centroid?

Then `signedVolume()` could be something like:
```cpp
  double signedVolume() const
  {
    double volume;
    PointType centroid;
    moments(volume, centroid, false);
    return volume;
  }
```

I'm raising this issue b/c we compute the polyhedron volume in our intersection-based shaping, and that call can be quite expensive, so we'd want to avoid making it unnecessarily more expensive.

src/axom/primal/tests/primal_polyhedron.cpp
```diff
@@ -284,24 +310,21 @@ void check_volume()
...
 
-  umpire::ResourceManager& rm = umpire::ResourceManager::getInstance();
-
-  // Save current/default allocator
-  const int current_allocator = axom::getDefaultAllocatorID();
-
-  // Determine new allocator (for CUDA or HIP policy, set to Device)
-  umpire::Allocator allocator =
-    rm.getAllocator(axom::execution_space<ExecSpace>::allocatorID());
+  constexpr double EPS = 1e-4;
```
kennyweiss, line 315: Is it possible to lower this `EPS` or was there some imprecision in the calculation that requires it to be so high?

kennyweiss (approved):
Thanks @bmhan12 !

Would it make sense to add some additional tests that translate, rotate and scale the polyhedra and verify that the volumes and centroids are still correct?

rhornung67 (approved):

agcapps (approved):
Approved.  Kenny, thanks for your comments; these should be addressed.

bmhan12:
> Would it make sense to add some additional tests that translate, rotate and scale the polyhedra and verify that the volumes and centroids are still correct?

Yes! I added additional tests that apply affine transformations.

PR #1377, opened 2024-07-04
Fix bug crashing MarchingCubes.

gunney1:
The bug was triggered when generating the contour from multiple meshes (such as SAMR levels), and the number of domains decrease.  When this happens, the memory-use optimization leaves all single-domain objects, but keeps track of the count separately.  Loops through domains should use the separately maintained count, not the number of items allocated.

### DIFF
--- a/src/axom/quest/MarchingCubes.cpp
+++ b/src/axom/quest/MarchingCubes.cpp
@@ -86,6 +86,10 @@ void MarchingCubes::setMesh(const conduit::Node& bpMesh,
     const auto& dom = bpMesh.child(d);
     m_singles[d]->setDomain(dom, m_topologyName, maskField);
   }
+  for(int d = newDomainCount; d < m_singles.size(); ++d)
+  {
+    m_singles[d]->getImpl().clearDomain();
+  }
 
   m_domainCount = newDomainCount;
 }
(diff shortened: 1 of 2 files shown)


rhornung67 (approved):

bmhan12 (approved):

agcapps (approved):

src/axom/quest/MarchingCubes.cpp
```diff
@@ -149,7 +153,7 @@ void MarchingCubes::computeIsocontour(double contourVal)
 axom::IndexType MarchingCubes::getContourNodeCount() const
 {
   axom::IndexType contourNodeCount =
-    m_singles.empty() ? 0 : m_facetCount * m_singles[0]->spatialDimension();
+    !m_domainCount ? 0 : m_facetCount * m_singles[0]->spatialDimension();
```
kennyweiss, line 156: This might read a bit clearer if it explicitly tested against 0 since `m_domainCount` is an int not a boolean:
```cpp
  axom::IndexType contourNodeCount = (m_domainCount == 0) 
      ? 0 
      : m_facetCount * m_singles[0]->spatialDimension();
```

kennyweiss (approved):

src/axom/quest/MarchingCubes.cpp
gunney1: Done.

PR #1385, opened 2024-07-16
Add ability to style a PR by commenting '/style'

white238:
This can't be used or tested until it is in the default branch. This was tested in Serac.


bmhan12 (approved):

rhornung67 (approved):
Where does this point to our clang format configuration file?

white238:
It runs CMake here:

https://github.com/LLNL/axom/blob/6ece465397f6f8ce522c708f9d0e971bc1f35b9a/.github/actions/apply-style/entrypoint.sh#L58-L61

which is just our normal build with very minimal CMake args.

PR #1387, opened 2024-07-22
Slic open before flush

bmhan12:
This PR:

- Explores an alternative way to create Slic log streams to files. With the current Slic approach, files must be opened/created when the stream object is constructed. If no messages are logged, the file remains empty. If there is an error and Slic tells MPI to abort, all ranks will attempt to flush, with files remaining empty if the rank could not flush before the non-collective abort call and/or no messages have been logged on that rank.

- Constructors were added to `GenericOutputStream`/`LumberjackStream`/`SynchronizedStream` that take in a `std::string` that is parsed and interpreted as `std::cout`, `std:cerr`, or a file name. In the case of a filename, the file is not opened until slic flushes and the stream has at least one message logged. This is done in `LumberjackStream`/`SynchronizedStream` by default constructing an `std::ofstream` object (no file has been opened yet), and delaying the `open(filename)` call until flushing. This is done in `GenericOutputStream` by initially constructing an `std::ostringstream` to buffer the messages in memory, and turning the `std::ostringstream` into a `std::ofstream` file when flushing.

~~- While these constructors work, I am not sure if it's the best way of addressing the current lack of file support in Slic. These `std::string` constructors take ownership of `std::ostream` from the user, which can be limiting if you want access to both the `std::ostream` object and the open-before-flush file behavior. In previous discussions, @gberg617 suggested adding getter methods to the in-memory messages storage (e.g. Lumberjack object in LumberjackStream, std::vector of messages in SynchronizedStream), so developers could take that information and create their own behavior. This would allow user-access to the `std::ostream` object and the open-before-flush file behavior with some effort.~~ (Resolved. Can add getter method later to `std::ostream` object if needed for access when using string constructors).


bmhan12:
This PR is ready for review, confirmed with @gberg617 that it is working for their use case.

rhornung67 (approved):
Thank you @bmhan12 Looks good!

src/axom/slic/streams/GenericOutputStream.hpp
```diff
@@ -49,6 +49,12 @@ class GenericOutputStream : public LogStream
    *   - "cerr" makes std::cerr the output stream
    *   - Any other input will construct a std::ofstream associated with input
    * \param [in] stream the string to control type of stream created
+   *
+   * \note The constructed std::ofstream will open and associate a file with
+   *       the stream the first time GenericOutputStream is flushed with at
+   *       least one buffered message. Use this constructor
+   *       to avoid empty file creation if GenericOutputStream is not appended
+   *       to.
```
agcapps, line 57: This `note` (and the one at new line 82) is hard to understand.  Please rephrase.  Consider dropping the first sentence and changing the second to "Use this constructor to avoid creating an empty file if this GenericOutputStream never gets a message."

bmhan12: Fixed!

agcapps (approved):
Thanks for working this through, Brian.

src/axom/slic/streams/GenericOutputStream.cpp
```diff
@@ -80,10 +104,38 @@ void GenericOutputStream::append(message::Level msgLevel,
...
+    {
+      std::string buffer = oss->str();
+      if(!buffer.empty())
+      {
+        delete m_stream;
+        m_stream = new std::ofstream(m_file_name);
+        (*m_stream) << buffer;
+        m_opened = true;
+      }
+    }
```
kennyweiss, line 122: Interesting approach! 

It took me a moment to figure out why the delete/new is necessary. 

Could you please add a brief comment that we're converting from the original `ostringstream` to an `ofstream` and writing the former's contents to the latter?

src/axom/slic/streams/SynchronizedStream.cpp
```diff
@@ -57,15 +61,70 @@ SynchronizedStream::SynchronizedStream(std::ostream* stream,
...
+  , m_file_name()
+  , m_isOstreamOwnedBySLIC(false)
+  , m_opened(false)
 {
   this->setFormatString(format);
 }
 
+//------------------------------------------------------------------------------
+SynchronizedStream::SynchronizedStream(const std::string stream, MPI_Comm comm)
+{
```
kennyweiss, line 73: (very) minor: The initialization of `m_cache` and `m_comm` are common to all three and can be initialized using a member initialization list, i.e.:
```cpp
SynchronizedStream::SynchronizedStream(const std::string stream, MPI_Comm comm) 
     : m_comm(comm)
     , m_cache(new MessageCache)  
{
 ...
}
```

src/axom/slic/tests/slic_macros.cpp
```diff
@@ -412,6 +413,67 @@ TEST(slic_macros, test_tagged_macros)
...
+  // do not create a a file until macros called, then flushed
+  std::string no_fmt = "file_no_fmt.txt";
+  std::string with_fmt = "file_with_fmt.txt";
+
+  slic::addStreamToAllMsgLevels(new slic::GenericOutputStream(no_fmt));
+
+  slic::addStreamToAllMsgLevels(new slic::GenericOutputStream(with_fmt, msgfmt));
+
+  EXPECT_EQ(axom::utilities::filesystem::pathExists(no_fmt), false);
+  EXPECT_EQ(axom::utilities::filesystem::pathExists(with_fmt), false);
```
kennyweiss, line 432: Minor: These might be slightly easier to read using `EXPECT_TRUE()` and `EXPECT_FALSE()` 
instead of `EXPECT_EQ(, {true,false}`

src/axom/slic/tests/slic_macros.cpp
```diff
@@ -412,6 +413,67 @@ TEST(slic_macros, test_tagged_macros)
...
+
+  EXPECT_EQ(no_fmt_buffer.str(), no_fmt_expected);
+
+  std::ifstream with_fmt_contents(with_fmt);
+  std::stringstream with_fmt_buffer;
+  with_fmt_buffer << with_fmt_contents.rdbuf();
+
+  EXPECT_EQ(with_fmt_buffer.str(), "Test");
+
+// Cleanup generated files (not working Windows)
```
kennyweiss, line 470: Is it known that this doesn't work on Windows? 
I didn't see an issue for this -- if we don't have one, can you please create one?

kennyweiss (approved):
Thanks @bmhan12 !  

Please also update the `RELEASE_NOTES` about this change.

src/axom/slic/tests/slic_macros.cpp
bmhan12: > I didn't see an issue for this -- if we don't have one, can you please create one?

Opened an issue here: https://github.com/LLNL/axom/issues/1415
I didn't dive too deep into why it wasn't working on Windows for this specific case.
More generally though, there are multiple places where we call `removeFile` in the code but don't check the return value/check if file is really deleted.

PR #1407, opened 2024-09-08
Update 2D GWN algorithm to match paper

jcs15c:
# Summary

This PR is an upgrade of the existing 2D generalized winding number (GWN) algorithm so that the implementation more closely matches the pseudocode in the paper ["Robust Containment Queries over Collections of Rational Parametric Curves via Generalized Winding Numbers"](https://dl.acm.org/doi/10.1145/3658228). The updated algorithm is more analytically tractable and computationally efficient, significantly reducing the number of trigonometric operations needed to compute the GWN for near-field points.
 
This upgrade does not change the external API for existing GWN methods, but does alter function calls in the `detail` namespace, i.e., replacing `curve_winding_number_recursive` with `construct_approximating_polygon`. 

Additionally, two additional functions are added to the interface provided by `winding_number.hpp`:
 - `winding_number<axom::Array<BezierCurve>>` 
 - A version of `winding_number<Polygon>` that has an additional reference parameter to report points located on the polygon boundary.

## Algorithm Details 

Specifically, the current implementation recursively discretizes a Bezier curve into (potentially many) linear segments and computes the GWN for each segment, evaluating an arccosine for each. In contrast, the updated algorithm only evaluates a single GWN value for each point. In this way, the new algorithm more directly decomposes GWN evaluation of Bezier curves into the following sequence of actions:
1. Check the point against an axis-aligned bounding-box for the curve. If outside, we can treat curve as a line between its endpoints, and exit the algorithm.
2. Close the curve with a straight line connecting the endpoints, and compute the *generalized* winding number of the line at the point.
3. Adaptively construct an `axom::Polygon` object that provably has the same *integer* winding number as the closed curve at the point, and evaluate the integer with a point-in-polygon algorithm.
4. Subtract the generalized winding number from the integer winding number from the resulting integer the *generalized* winding number of the closing line.

## Extra bugfix via refactor of `winding_number<polygon>`

During review, an additional bug was found relating to the evaluation of the winding number for points that are near to the closing line. Specifically, it was possible for a point to be considered as interior/exterior to the approximating polygon object, but coincident with the closing curve during the computation of it's generalized winding number, causing an incorrect value to be subtracted from the GWN. 

This "off-by-0.5" error, unacceptably high for points near the closing curve, is due to a different physical interpretation of numerical tolerances between `winding_number<polygon>` and `winding_number<line>`, where the former checks for colinearity of an edge's endpoints and the point using a determinant formula, and the latter checks for physical distance between the edge and the point.

To fix this, the various numerical tolerances in `winding_number<polygon>` (which also includes an Linf distances to vertices) have been replaced with checks for physical distances between the query point and edges. This approach is nominally more costly, but makes the method considerably more usable in conjunction with other `winding_number` methods. Beyond changing variable names, this change does not alter the interface of the existing method.

An additional test has been added to `primal_winding_number` to verify the corrected behavior.


src/axom/primal/operators/winding_number.hpp
```diff
@@ -109,16 +110,19 @@ int winding_number(const Point<T, 2>& q,
...
   const int nverts = P.numVertices();
+  isOnEdge = false;
 
   // If the query is a vertex, return a value interpreted
   //  as "inside" by evenodd or nonzero protocols
   if(axom::utilities::isNearlyEqual(P[0][0], R[0], EPS) &&
      axom::utilities::isNearlyEqual(P[0][1], R[1], EPS))
   {
+    isOnEdge = true;
     return includeBoundary;
```
kennyweiss, line 126: Should this return an integer, rather than a `bool`?
If so, would something like the following be more appropriate/clear? (here and elsewhere in this function).
```suggestion
    return includeBoundary ? 1 : 0;
```

... or am I misunderstanding something?

src/axom/primal/operators/winding_number.hpp
```diff
@@ -109,16 +110,19 @@ int winding_number(const Point<T, 2>& q,
 template <typename T>
 int winding_number(const Point<T, 2>& R,
                    const Polygon<T, 2>& P,
+                   bool& isOnEdge,
                    bool includeBoundary = false,
                    double EPS = 1e-8)
```
kennyweiss, line 115: Suggestion: To avoid accidental overload ambiguities, it might be a good idea for this version to be more explicit and not have any default parameters.
E.g.
```suggestion
int winding_number(const Point<T, 2>& R,
                   const Polygon<T, 2>& P,
                   bool& isOnEdge,
                   bool includeBoundary,
                   double EPS)
```

src/axom/primal/operators/winding_number.hpp
```diff
@@ -197,15 +205,40 @@ int winding_number(const Point<T, 2>& R,
...
+                   const Polygon<T, 2>& P,
+                   bool includeBoundary = false,
+                   double EPS = 1e-8)
+{
+  bool isOnEdge = false;
+  return winding_number(R, P, isOnEdge, includeBoundary, EPS);
+}
+
+/*!
+ * \brief Computes the generalized winding number (GWN) for a single 2D Bezier curve
```
kennyweiss, line 223: [Consistency / Somewhat pedantic, but hopefully not too much]
In some cases, we say that we're computing the GWN for a point w.r.t. a primitive, but this one doesn't mention the point. I think it'd be good to be consistent.

src/axom/primal/operators/winding_number.hpp
```diff
@@ -215,7 +248,64 @@ double winding_number(const Point<T, 2>& q,
...
+  approximating_polygon.addVertex(c[ord]);
+
+  // Compute the integer winding numberof the closed curve
+  bool isOnEdge = false;
+  double closed_curve_wn =
+    winding_number(q, approximating_polygon, isOnEdge, false, edge_tol);
+
+  // Compute the fractional value of the closed curve
+  int n = approximating_polygon.numVertices();
+  double closure_wn = detail::linear_winding_number(q,
```
kennyweiss, line 287: [nit] The code might be slightly easier to understand if these two variable were marked `const`:
```suggestion
  // Compute the fractional value of the closed curve
  const int n = approximating_polygon.numVertices();
  const double closure_wn = detail::linear_winding_number(q,
```

src/axom/primal/operators/detail/winding_number_impl.hpp
```diff
@@ -42,7 +42,7 @@ namespace detail
  * is the signed angle subtended by the query point to each endpoint.
  * Colinear points return 0 for their winding number.
  *
- * \return double The winding number
+ * \return double The GWN
```
kennyweiss, line 45: Minor: My initial read of this doxyline was that the return value is twice (double) that of the GWN
(not sure we should do anything about it, maybe remove the return type, i.e. the word `double`)

kennyweiss (approved):
Thanks @jcs15c !

Before merging, I'd like to check that our winding number example application still works on all the test shapes from our data repo -- https://github.com/LLNL/axom_data/tree/main/contours/svg

(I haven't gotten around to adding automated tests for these yet, but it's on my todo list).

src/axom/primal/operators/detail/winding_number_impl.hpp
agcapps: Hah.  I'd be fine with removing "double".

src/axom/primal/operators/winding_number.hpp
jcs15c: This is relying on the automatic conversion from bool to integer here, yes. Your version is certainly more clear semantically, probably even more than a cast from one type to the other would be.

src/axom/primal/operators/detail/winding_number_impl.hpp
jcs15c: This caught me off guard a couple of times too. Removing `double` it is!

rhornung67 (approved):
Thank you @jcs15c Very nice summary of the changes in the PR.

src/axom/primal/operators/winding_number.hpp
jcs15c: Good suggestion! I'm also going through and making much of this page's documentation more consistent with respect to acronyms, particularly when distinguishing (integer) winding numbers from generalized winding numbers, or the GWN.

agcapps (approved):

jcs15c:
> Before merging, I'd like to check that our winding number example application still works on all the test shapes from our data repo -- https://github.com/LLNL/axom_data/tree/main/contours/svg

@kennyweiss I checked against these shapes, and although the results are consistent with the `develop` branch in terms of performance and (mostly) classification accuracy, I did find through this check a handful of discrepancies. This led to the discovery of a much larger class of errors owing to some inconsistent numerical tolerances used by different winding number methods. I've fixed the issue by modifying the tolerances used by `winding_number<polygon>`, which I've described in more detail in the main PR text. 

Either way, it'd definitely be worth it to automate those tests somehow, these issues would have stuck out immediately.

kennyweiss:
Awesome -- thanks @jcs15c. Now that the issue's resolved, feel free to merge this PR. We'll automate those tests in a follow-up PR.

src/axom/primal/operators/winding_number.hpp
```diff
@@ -211,7 +182,7 @@ int winding_number(const Point<T, 2>& R,
  * \param [in] R The query point to test
  * \param [in] P The Polygon object to test for containment
  * \param [in] includeBoundary If true, points on the boundary are considered interior
- * \param [in] EPS The tolerance level for collinearity
+ * \param [in] edge_tol The distance at wich a point is considered on the boundary
```
kennyweiss, line 185: Typo:
```suggestion
 * \param [in] edge_tol The distance at which a point is considered on the boundary
```

src/axom/primal/tests/primal_winding_number.cpp
```diff
@@ -168,9 +168,16 @@ TEST(primal_winding_number, closure_edge_cases)
...
+  for(int i = 2; i < 15; ++i)
+  {
+    // In all cases, the winding number should be *near* 0.5.
+    //  If the tolerances don't match, we would get an "off-by-0.5" error
+
+    double diff = std::pow( 10, -i );
+    EXPECT_NEAR(winding_number(Point2D({0, diff}), quartic, 0.5 * diff, EPS), 0.5, 0.1);
+    EXPECT_NEAR(winding_number(Point2D({0, diff}), quartic, 1.0 * diff, EPS), 0.5, 0.1);
+    EXPECT_NEAR(winding_number(Point2D({0, diff}), quartic, 2.0 * diff, EPS), 0.5, 0.1);
+  }
```
kennyweiss: This is a great test!
Would it makes sense to also test `Point2D({0, -diff})` ?

jcs15c: Probably! The internal tolerance checks don't care which side of the closure the point is on, but might as well check it here too.

jcs15c: It actually made a lot of sense! There was the exact same issue with the `BoundingBox` check, since the closure is coincident with an edge. This means it would be counted as "outside" the bounding box but "on" the edge, and therefore give the wrong answer. I'm actually surprised that this didn't come up as an issue in the previous versions of `winding_number<Bezier>`, since nothing new to this PR has changed this behavior. Neat!

PR #1411, opened 2024-09-16
Refactor getValidIOProtocols

### LINKED ISSUE #1408: Invalid pointer error for static built Axom linked into a separate shared library
I get a `free(): invalid pointer` error on this static variable:
https://github.com/LLNL/axom/blob/5374375a1cb026642a774f43a109fb97ac87822e/src/axom/sidre/core/Group.hpp#L1933

This only occurs if our code is built as shared libraries but Axom is built as static libraries. If Axom is built as shared libraries or our code is built as a static library, we don't get this error. This was not an issue for version 0.7.0 when this static variable did not exist.

nselliott:
# Summary

- This PR is a refactoring
- It does the following:
  - Refactors a getValidIOProtocols method, moving it from Group to DataStore, making it a regular public member of DataStore that doesn't use static data.
  - Fixes #1408

The vector holding the valid protocols is filled on the first call and stored inside the DataStore object.

### DIFF
--- a/src/axom/sidre/core/DataStore.hpp
+++ b/src/axom/sidre/core/DataStore.hpp
@@ -501,6 +501,30 @@ public:
                               const std::string& index_path);
 #endif
 
+  /*!
+   * \brief Get valid protocols for I/O of objects in DataStore.
+   *
+   * Only protocols that work for both input and output are returned.
+   */
+  const std::vector<std::string>& getValidIOProtocols()
+  {
+    if(m_io_protocols.empty())
+    {
+      m_io_protocols = {
+#ifdef AXOM_USE_HDF5
+        "sidre_hdf5",
+        "conduit_hdf5",
+#endif
+        "sidre_json",
+        "sidre_conduit_json",
+        "conduit_bin",
+        "conduit_json",
+        "json"};
+    }
+
+    return m_io_protocols;
+  }
+
   //----------------
 
   /*!
@@ -555,6 +579,9 @@ private:
 
   /// Flag indicating whether SLIC logging environment was initialized in ctor.
   bool m_need_to_finalize_slic;
+
+  /// The valid I/O protocols for objects in the DataStore.
+  std::vector<std::string> m_io_protocols;
 
   /// Details of the most recent Conduit error.  Length > 0 indicates an error occurred.
   mutable std::string m_conduit_errors;
--- a/src/axom/sidre/core/Group.cpp
+++ b/src/axom/sidre/core/Group.cpp
@@ -37,18 +37,6 @@ namespace sidre
 // support path syntax.
 const char Group::s_path_delimiter = '/';
 
-// Initialization of static members holding I/O protocol strings
-const std::vector<std::string> Group::s_io_protocols = {
-#ifdef AXOM_USE_HDF5
-  "sidre_hdf5",
-  "conduit_hdf5",
-#endif
-  "sidre_json",
-  "sidre_conduit_json",
-  "conduit_bin",
-  "conduit_json",
-  "json"};
-
 ///////////////////////////////////////////////////////////////////////////////
 //
 // Private utility functions to cast ItemCollections to (named) MapCollections.
--- a/src/axom/sidre/core/Group.hpp
+++ b/src/axom/sidre/core/Group.hpp
@@ -150,16 +150,6 @@ public:
   char getPathDelimiter() const { return s_path_delimiter; }
 
   /*!
-   * \brief static method to get valid protocols for Group I/O methods.
-   *
-   * Only protocols that work for both input and output are returned.
-   */
-  static const std::vector<std::string>& getValidIOProtocols()
-  {
-    return s_io_protocols;
-  }
-
-  /*!
    * \brief static method to get the default I/O protocol.
    */
   static std::string getDefaultIOProtocol()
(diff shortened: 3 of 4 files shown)


agcapps (approved):

src/axom/sidre/core/DataStore.hpp
```diff
@@ -501,6 +501,30 @@ class DataStore
...
+#endif
+        "sidre_json",
+        "sidre_conduit_json",
+        "conduit_bin",
+        "conduit_json",
+        "json"};
+    }
+
+    return m_io_protocols;
+  }
```
kennyweiss, line 526: Thanks for working on this @nselliott 

I had something like the following in mind:

```cpp
 // a public static member of sidre::Group
 static const std::vector<std::string>& getValidIOProtocols()
  {
     static const std::vector<std::string>s_io_protocols = {
#ifdef AXOM_USE_HDF5
        "sidre_hdf5",
        "conduit_hdf5",
#endif
        "sidre_json",
        "sidre_conduit_json",
        "conduit_bin",
        "conduit_json",
        "json"};

    return s_io_protocols;
  }
```
This way, there is only one array of protocols, it doesn't increase the size of `Group` or the `Datastore`, and it solves the static initialization/destruction problem.

If this function stays in `sidre::Group`, there would be no API changes necessary for axom users.

Is there a reason to prefer the current solution from this PR to this one?

nselliott: @kennyweiss That is almost the same as what we have now, except for the location of the static vector s_io_protocols declaration and initialization.  As the error we are trying to solve occurs in the destruction of the static vector at program completion, I don't know if this suggestion would fix that.  I don't believe that our current implementation is incorrect by the standard, but a compiler isn't handling it well in the shared library context.  My attempt here is to avoid all of that by not using a static vector at all.

kennyweiss: Thanks @nselliott 

My assumption/understanding was that the problem in #1408 is related to the static member variable (`std::vector`) in the `Group` class, and that, if it lived in a static function, the problem would go away.

@ldowen -- would it be possible to check if my proposed solution resolves the problem for you?
Specifically:

1. Removing `sidre::Group::s_io_protocol`
2. Implementing `sidre::Group::getValidIOProtocols()` as in this comment?

Asking, b/c that solution would not require API changes for sidre users.

agcapps: Kenny, is the API change the main issue?  (I can understand that.)

While it is a change, I feel it is a small (relatively non-disruptive) change for the better.  Also, though we don't want to disrupt users, we have never claimed a stable API: that is one reason we're reluctant to go to Axom 1.0.

ldowen: @kennyweiss I will give this fix a shot and see if it works.

nselliott: > Thanks @nselliott
> 
> My assumption/understanding was that the problem in #1408 is related to the static member variable (`std::vector`) in the `Group` class, and that, if it lived in a static function, the problem would go away.
> 
> @ldowen -- would it be possible to check if my proposed solution resolves the problem for you? Specifically:
> 
> 1. Removing `sidre::Group::s_io_protocol`
> 2. Implementing `sidre::Group::getValidIOProtocols()` as in this comment?
> 
> Asking, b/c that solution would not require API changes for sidre users.

The thing about moving the static vector declaration into a static function is that it will likely succeed in hiding the bug from any application that never calls that function--the vector will never get instantiated and thus its destructor will never be called at the end--but it may not prevent the error from happening in the cases where the function is called.  Once the static vector exists, it will be destroyed during shutdown in an order that's outside of the programmer's control, and whatever is making it go bad could still happen.  I admit I don't understand why this error is happening to a vector of strings.

One alternative from @rhornung67 is still an API change but a smaller one.  Change the signature of the current static method to `static std::vector<std::string> getValidIOProtocols()`, removing the reference, build the vector on the fly each time, and return by value.  This is a more expensive operation, but it's likely a one-time call and leaves no static data waiting around.

BradWhitlock: I like returning the std::vector<std::string> approach since it eliminates any issues with static initialization.

ldowen:
I gave Kenny's fix a shot and it worked. Just to be sure, I added a call to the `getValidIOProtocols` into my code and it still worked. This fix seems consistent with recommended singleton implementations that we use in our own code and seems like an ideal candidate for being a static variable. If we really want to be consistent, I think the `std::vector` should be an `std::array` but that would break the API.

src/axom/sidre/core/DataStore.hpp
kennyweiss: I also like returning a `std::vector<std::string>` since there's no dangling references, or extra variables in `DataStore` or `Group`.

ldowen:
Also, thank you @nselliott @kennyweiss @rhornung67 @agcapps @BradWhitlock @cyrush for looking into this issue!

src/axom/sidre/core/DataStore.hpp
kennyweiss: @agcapps wrote:
> Kenny, is the API change the main issue? (I can understand that.)

That was one of my concerns. Another concern, which I didn't articulate is that this single change would leave sidre inconsistent -- `sidre::Group` also has `getDefaultIOProtocol()`, and the two functions should be defined in the same class. Since we're already making a change, it could be that `DataStore` is a better home for them.

Also, it seems like these functions should remain static since they're related to the configuration / class, and not the instance.

See: https://github.com/LLNL/axom/blob/efef4ec471e90cde65c6059eb0e6094c50430981/src/axom/sidre/core/Group.hpp#L152-L172

nselliott:
Given the confirmation from @ldowen , I'm going to scrap this PR and use the suggestion from @kennyweiss .

PR #1416, opened 2024-09-18
Move I/O protocol vector into static method

### LINKED ISSUE #1408: Invalid pointer error for static built Axom linked into a separate shared library

nselliott:
# Summary

- This PR is a refactoring
- It does the following:
  - Moves the declaration and initialization of the static vector containing all valid I/O protocol strings into the static method that uses it.
  - Fixes #1408

### DIFF
--- a/src/axom/sidre/core/Group.cpp
+++ b/src/axom/sidre/core/Group.cpp
@@ -37,18 +37,6 @@ namespace sidre
 // support path syntax.
 const char Group::s_path_delimiter = '/';
 
-// Initialization of static members holding I/O protocol strings
-const std::vector<std::string> Group::s_io_protocols = {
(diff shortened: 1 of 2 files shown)


kennyweiss (approved):
Thanks @nselliott !

agcapps (approved):

rhornung67 (approved):

PR #1423, opened 2024-09-24
Cleanup a bunch of GPU warnings

publixsubfan:
# Summary

Clears up the bulk of Slam/Core-associated host/device usage warnings.

- Most of these were associated with the Slam `IndirectionPolicy` accessing the array value by reference; the interface has been changed to return a pointer-to-value, which allows us to just return `nullptr` for `STLVectorIndirectionPolicy`.
- The remainder were associated with the construction of `Map`, `BivariateMap`, and `SubMap` iterators, and were resolved by adding a bunch of `AXOM_HOST_DEVICE` annotations, and using a concrete interface policy for `Map` types by default.

The other remaining warnings were:
- `FlatMap`: Improperly setting an unsigned hash value with a negative sentinel value. The assignment was extraneous and removed.
- `BVH`: A test passed in an `axom::Array` to `BVH::initialize()`, which is captured by a lambda. This was changed to pass in an `ArrayView` instead.

### DIFF
--- a/src/axom/spin/tests/spin_bvh.cpp
+++ b/src/axom/spin/tests/spin_bvh.cpp
@@ -1354,7 +1354,7 @@ void check_find_points_zip3d()
   // construct the BVH
   spin::BVH<NDIMS, ExecSpace, FloatType> bvh;
   bvh.setScaleFactor(1.0);  // i.e., no scaling
-  bvh.initialize(aabbs, ncells);
+  bvh.initialize(aabbs_view, ncells);
 
   BoxType bounds = bvh.getBounds();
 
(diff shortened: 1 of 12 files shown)


agcapps (approved):
Thanks for fixes!

src/axom/spin/tests/spin_bvh.cpp
```diff
@@ -1354,7 +1354,7 @@ void check_find_points_zip3d()
   // construct the BVH
   spin::BVH<NDIMS, ExecSpace, FloatType> bvh;
   bvh.setScaleFactor(1.0);  // i.e., no scaling
-  bvh.initialize(aabbs, ncells);
+  bvh.initialize(aabbs_view, ncells);
```
bmhan12, line 1357: Good catch!

bmhan12 (approved):

Issue #1415, opened 2024-09-18
Check removeFile() return value


bmhan12:
Issue is opened from this [discussion](https://github.com/LLNL/axom/pull/1387#discussion_r1764403743).

- [x] Various [parts of the code](https://github.com/search?q=repo%3ALLNL%2Faxom%20removeFIle&type=code) call `removeFile`, but do not verify that the file has actually been deleted through the return code/checking if file stills exist. Warning, error, or at least some indication should be given to user if file deletion has failed.
- [x] Fix `removeFile` usage on Windows in `slic_macros.cpp` test case (from original discussion)

kennyweiss:
Related: @agcapps noted in https://github.com/LLNL/axom/pull/1421 that `Unlink` on Windows only works when all file handles are closed.

bmhan12:
Closed by #1424

Issue #1410, opened 2024-09-16
SLIC: Populate rank and count earlier


white238:
From @gberg617:

I was noticing that the rank information (both list of ranks and count) are populated for Lumberjack streams only when pushing the messages.  Do you know if it would be possible at all to populate this information for ranks that are going call outputLocal and then abort?  No need to go out of your way if you think this will be difficult to change, but I was curious if it's something that could be a 
simple few lines of code changes.

bmhan12:
@gberg617 
My initial thought was I expected the rank information to be present for `outputLocal`, since we are testing for it in our unit tests. 

Looks like we are not checking the rank information yet in our [unit tests](https://github.com/LLNL/axom/blob/efef4ec471e90cde65c6059eb0e6094c50430981/src/axom/slic/tests/slic_macros.cpp#L50-L115).

I'll add unit tests, and see if I get the same missing-rank-information behavior.

bmhan12:
@gberg617 

So I am encountering errors in terms of unexpected rank/rank count information when I added unit tests (branch: [bugfix/han12/slic_rank_info](https://github.com/LLNL/axom/tree/bugfix/han12/slic_rank_info))
Still in the process of debugging what/where the issue is from...

gberg617:
> @gberg617
> 
> So I am encountering errors in terms of unexpected rank/rank count information when I added unit tests (branch: [bugfix/han12/slic_rank_info](https://github.com/LLNL/axom/tree/bugfix/han12/slic_rank_info)) Still in the process of debugging what/where the issue is from...

Thanks for the update!

bmhan12:
@gberg617 
I added rank and rank count unit tests in PR #1431.
The error I was running into was strictly Axom-related, involving `std::ends` (PR has more details).

In terms of rank and rank count, I did not notice the information missing on `outputLocalMessages()` calls.

I did want to iterate that for `outputLocalMessages()` the messages do not get combined, as that is a collective operation that does not bode well when trying to `MPI_Abort()`.
So its expected that the messages will only contain information for the current rank (e.g. Rank 0's message will have `rank: 0 rank_count 1` , Rank 1's message will have `rank: 1 rank_count: 1`, etc.)

PR #1424, opened 2024-09-24
removeFile() Usage

bmhan12:
This PR:

- Fixes `removeFile` failure on Windows in `slic_macros.cpp` unit test by closing the files associated with `ifstream` and the `ofstream`  member variables of `LogStream` instances.
- Checks the result of `removeFile` in unit tests to verify file has been deleted

Closes issue #1415


src/axom/slic/tests/slic_macros.cpp
```diff
@@ -467,11 +466,19 @@ TEST(slic_macros, test_macros_file_output)
...
 
-// Cleanup generated files (not working Windows)
-#ifndef WIN32
+  no_fmt_contents.close();
+  with_fmt_contents.close();
+
+// Closes open file streams associated with Slic streams when destructors
+// called during slic::finalize().
+// Windows _unlink file deletion fails if file is still in use.
+#ifdef WIN32
```
agcapps, line 475: Unless there's a reason non-Windows should not finalize(), let's please remove the guard on `WIN32` here and in the slic_macros_parallel.cpp test.  If there is a reason to test removeFile() on non-finaliz()ed slic on non-Windows platforms, I would appreciate knowing.

agcapps:
@bmhan12 , thank you for your attention to detail on this.

src/axom/slic/tests/slic_macros.cpp
bmhan12: Guard removed! `finalize()` call will work on Windows and non-Windows systems, guard was a relic of my testing.

rhornung67 (approved):

src/axom/slic/tests/slic_macros.cpp
```diff
@@ -457,7 +456,7 @@ TEST(slic_macros, test_macros_file_output)
   std::string no_fmt_expected;
   no_fmt_expected += "*****\n[INFO]\n\n Test \n\n ";
   no_fmt_expected += __FILE__;
-  no_fmt_expected += "\n441\n****\n";
+  no_fmt_expected += "\n440\n****\n";
```
kennyweiss, line 459: What is `440` here? Is it a line number? 
It might make sense to add a comment about this for future maintenance of these tests.

kennyweiss (approved):
Thanks @bmhan12 !

kennyweiss:
@bmhan12 -- could you please update the documentation for `removeFile` about it not working on Windows when there are open file handles?

src/axom/slic/tests/slic_macros.cpp
bmhan12: > What is 440 here? Is it a line number?

Yes. In a follow-up PR related to #1410, I'll be updating these line numbers to be like `slic_macros_parallel` to be `__LINE__ - offset` to make that clear.

PR #1431, opened 2024-09-30
Add unit testing for Slic ranks and rank counts

bmhan12:
This PR:
- Adds testing for ranks and rank counts information in `slic_macros_parallel` unit tests.
- Remove `std::ends` from `SLIC_ASSERT`/`SLIC_ASSERT_MSG`/`SLIC_CHECK`/`SLIC_CHECK_MSG` macros that prevented Lumberjack from combining messages. When packing/unpacking messages for combining, `std::ends` does not play nice in std::string --> C-style string conversions. Trailing `std::ends` null character does not get included in the conversion, so output/rank Node's message (with `std::ends` null character) will always be different from the other nodes (no`std::ends` null character), and as such no combining would take place.
  - Added check that messages are unique/combined for LumberjackStreams in unit tests.
  - Example Reproducer of behavior: https://godbolt.org/z/of9Y74sxz

Relates to #1410

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -29,6 +29,9 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 
 ###  Fixed
 - Added a guard for sidre-related mint API usage in a quest example
+- Removed `std::ends` usage from `SLIC_ASSERT`,`SLIC_ASSERT_MSG`,`SLIC_CHECK`,
+  and `SLIC_CHECK_MSG` macros that prevented Lumberjack from combining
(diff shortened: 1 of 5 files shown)


rhornung67 (approved):
Looks good. The only thought I have is that it may make the slic tests easier to understand if there were comments in the code indicating why the integer constants are chosen as they are; e.g., why `N` in something like `(__LINE__ - N)`

kennyweiss (approved):
Thanks @bmhan12 -- please don't forget to update the `RELEASE_NOTES`

PR #1433, opened 2024-09-30
Add sphinx jquery extension to fix read-the-docs search

bmhan12:
This PR:

- Fixes issue where search is broken with add-ons enabled in ReadTheDocs
  - Can try RTD search bar in develop documentation version to observe brokenness
- ReadTheDocs link to this branch: https://axom.readthedocs.io/en/bugfix-han12-rtd_search/
- Same as RAJA's LLNL/RAJA#1721

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -57,8 +57,7 @@ extensions = [
     'sphinxcontrib.jquery',
     'sphinx.ext.todo',
     'sphinx.ext.coverage',
-    'sphinx.ext.mathjax',
-    'sphinxcontrib.jquery'
+    'sphinx.ext.mathjax'
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/src/conf.py
+++ b/src/conf.py
@@ -57,7 +57,8 @@ extensions = [
     'sphinxcontrib.jquery',
     'sphinx.ext.todo',
     'sphinx.ext.coverage',
-    'sphinx.ext.mathjax'
+    'sphinx.ext.mathjax',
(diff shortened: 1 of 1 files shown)


BradWhitlock (approved):
Looks good.

rhornung67 (approved):

src/conf.py
```diff
@@ -57,7 +57,8 @@
     'sphinxcontrib.jquery',
     'sphinx.ext.todo',
     'sphinx.ext.coverage',
-    'sphinx.ext.mathjax'
+    'sphinx.ext.mathjax',
+    'sphinxcontrib.jquery'
```
white238, line 61: This is already on line 57

BradWhitlock: D'oh - I was too reliant on diff highlighting.

bmhan12: interesting... I'll play around with this a bit.

bmhan12: Looks to be only "develop" branch does not have the search working.

white238: Last time this happened, @kennyweiss helped by inspecting the html errors via the browser. This line specifically is what fixed it.

bmhan12:
As @white238 pointed out, we already have `jquery` added.
Closing this PR.
I think the issue was related to do with my web browser caching. Not observing the error in a private window or by switching browsers.

src/conf.py
bmhan12: Scratch that, works when I switch browsers/use a private tab. Probably just need to clear cache.

PR #1448, opened 2024-10-15
Add NURBS Curve Support to Primal

jcs15c:
# Summary

- This PR is a feature.
- It adds a `NURBSCurve` class to `primal`. `NURBSCurve` has the following features, along with relevant tests:
  - Constructors from arrays of points and either the degree or knot vector
  - Evaluation/Derivative methods
  - Knot Insertion
  - Curve Subdivision/Bezier Extraction
- It adds a `KnotVector` class to `primal` that maintains relevant methods and constraints on knot vector values.
  - Contains methods for span searching, basis function evaluation, basis function derivatives, etc.
  - It contains a `degree` parameter and enforces that the curve is
     - Monotone non-decreasing.
     - Open/Clamped, i.e. multiplicity of endpoints is `degree + 1`.
     - Continuous, i.e. maximum multiplicity of internal knots is `degree`.
   - These methods and checks will be re-used in a forthcoming NURBS patch class.
- It adds an interface to the `winding_number` method for NURBS curves.
- It also fixes a bug in the `NURBSInterpolator` class in the `C2CReader.cpp` file when evaluating derivatives of NURBS curves (See the NURBS Book pg. 127, Algorithm A4.2).


src/axom/primal/geometry/NURBSCurve.hpp
```diff
@@ -0,0 +1,1177 @@
...
+             NURBSCurve<T, NDIMS>& n2,
+             bool normalize = false) const
+  {
+    SLIC_ASSERT(t >= m_knotvec[0] && t <= m_knotvec[m_knotvec.getNumKnots() - 1]);
+
+    const bool isCurveRational = this->isRational();
+    const int p = getDegree();
+
+    // Handle the special case of splgitting at the endpoints
+    if(t == 0.0)
```
kennyweiss, line 709: Should this be `0` or the lowest knot vector value?
Similarly for the check in the `else if t == 1.0` branch

src/axom/primal/geometry/NURBSCurve.hpp
```diff
@@ -0,0 +1,1177 @@
...
+   * \return An array of Bezier curves
+   */
+  axom::Array<BezierCurve<T, NDIMS>> extractBezier() const
+  {
+    axom::Array<BezierCurve<T, NDIMS>> beziers;
+
+    const bool isCurveRational = this->isRational();
+
+    int p = getDegree();
+    if(p == 0)  // Handle this special case
```
kennyweiss, line 795: Is `p==0` a collection of disconnected points? Please add a comment.

src/axom/primal/geometry/NURBSCurve.hpp
```diff
@@ -0,0 +1,1177 @@
...
+   * \param [in] degree The target degree
+   * 
+   * \note Will clear any data already in these arrays.
+   */
+  void setParameters(int npts, int degree)
+  {
+    SLIC_ASSERT(npts >= degree + 1);
+    SLIC_ASSERT(degree >= 0);
+
+    m_controlPoints.resize(npts);
```
kennyweiss, line 870: This resizes the control points, but the data is still there. Should it explicitly set them to a value, e.g. `PointType::zero()`?

(Obviously, that would be extra work, so perhaps it's good as is.)

src/axom/primal/geometry/NURBSCurve.hpp
```diff
@@ -0,0 +1,1177 @@
...
+
+  /*!
+   * \brief Reset the degree and number of points in the curve
+   *
+   * \param [in] npts The target number of control points
+   * \param [in] degree The target degree
+   * 
+   * \note Will clear any data already in these arrays.
+   */
+  void setParameters(int npts, int degree)
```
kennyweiss, line 865: It might be worth noting that this will also make the knots uniform.

Is `setParameters` the best name for this function? (Full disclosure -- I didn't immediately think of a better name)

src/axom/primal/geometry/NURBSCurve.hpp
```diff
@@ -0,0 +1,1177 @@
...
+    m_controlPoints.clear();
+    m_knotvec.clear();
+    makeNonrational();
+  }
+
+  /// \brief Retrieve the control point at index \a idx
+  PointType& operator[](int idx) { return m_controlPoints[idx]; }
+
+  /// \brief Retrieve the control point at index \a idx
+  const PointType& operator[](int idx) const { return m_controlPoints[idx]; }
```
kennyweiss, line 962: Should this be an explicitly named function to access the control points (similar to `getWeight()`)

Alternatively, would it make sense to have the following API (w/ const overloads):
* `controlPoints() -> axom::Array<PointType>&`
* `weights() -> axom::Array<T>&`
* `knots() -> axom::Array<T>&`

Then users could use the underlying array functions to do things like:
```cpp
const& pt = nurbs.controlPoints()[3];
nurbs.weights()[2] = 2.3;
....
```

Note: This is different than the functions below that return a copy of the control points, knots and weights.

src/axom/primal/geometry/NURBSCurve.hpp
```diff
@@ -0,0 +1,1177 @@
...
+      {
+        axom::utilities::swap(m_weights[i], m_weights[npts - i - 1]);
+      }
+    }
+
+    // Reverse the orientation of the knots
+    m_knotvec.reverse();
+  }
+
+  /// \brief Returns an axis-aligned bounding box containing the Bezier curve
```
kennyweiss, line 1064:
```suggestion
  /// \brief Returns an axis-aligned bounding box containing the NURBS curve
```

src/axom/primal/geometry/NURBSCurve.hpp
```diff
@@ -0,0 +1,1177 @@
...
+  BoundingBoxType boundingBox() const
+  {
+    return BoundingBoxType(m_controlPoints.data(),
+                           static_cast<int>(m_controlPoints.size()));
+  }
+
+  /// \brief Returns an oriented bounding box containing the Bezier curve
```
kennyweiss, line 1071:
```suggestion
  /// \brief Returns an oriented bounding box containing the NURBS curve
```

src/axom/primal/tests/primal_nurbs_curve.cpp
kennyweiss, line 1: Really nice initial tests!

We should probably also add direct unit tests for the `KnotVector` class and more tests for 1D and 2D NURBS curves.
(These could also come in a future PR).

src/axom/primal/tests/primal_nurbs_curve.cpp
```diff
@@ -0,0 +1,831 @@
...
+  double weights[4] = {1.0, 2.0, 3.0, 4.0};
+
+  for(int deg = 2; deg <= 2; ++deg)
+  {
+    NURBSCurveType curve(data, weights, 4, deg);
+
+    // Do some knot insertion to make it interesting
+    curve.insertKnot(0.3, 2);
+    curve.insertKnot(0.6, 1);
+    curve.insertKnot(0.8, 1);
```
kennyweiss, line 505: Should there be some checks here that the curve and or/or knots have the right shapes after each knot insertion.
(These could also be tested in separate unit tests).

src/axom/primal/tests/primal_nurbs_curve.cpp
```diff
@@ -0,0 +1,831 @@
...
+  curve.insertKnot(0.33, 3);
+  curve.insertKnot(0.66, 1);
+  curve.insertKnot(0.77, 2);
+
+  NURBSCurveType bezier_curve;
+  auto bezier_list = curve.extractBezier();
+
+  EXPECT_EQ(bezier_list.size(), 4);
+
+  for(double t = 0.0; t < 1.0; t += 0.05)
```
kennyweiss, line 643: Do we have any tests where the knots vector is not normalized to the unit inverval $[0,1]$?

src/axom/primal/tests/primal_nurbs_curve.cpp
```diff
@@ -0,0 +1,831 @@
...
+      {
+        EXPECT_DOUBLE_EQ(std::lround(gwn), 0.0);
+      }
+      else
+      {
+        EXPECT_DOUBLE_EQ(std::lround(gwn), 1.0);
+      }
+    }
+  }
+}
```
kennyweiss, line 763: Would it better to move this test to primal winding number test file?

src/axom/primal/tests/primal_nurbs_curve.cpp
```diff
@@ -0,0 +1,831 @@
...
+  double knots[11] = {0.0, 0.0, 0.0, 0.0, 0.5, 0.5, 0.5, 1.0, 1.0, 1.0, 1.0};
+  double scaled_knots[11];
+
+  for(int i = 0; i < 11; ++i)
+  {
+    scaled_knots[i] = 2.0 * knots[i] - 5.0;
+  }
+
+  NURBSCurveType circle(data, weights, 7, knots, 11);
+  NURBSCurveType scaled_circle(data, weights, 7, scaled_knots, 11);
```
kennyweiss, line 792: 👍

kennyweiss (approved):
Thanks @jcs15c -- this is a really nice addition!

It would be good to have more testing of the KnotVector class, and of NURBS in 1D and 2D, but that can be added in future PRs.

Please don't forget to update the RELEASE_NOTES.

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -0,0 +1,501 @@
...
+      m_knots[i] = knots[i];
+    }
+
+    m_deg = degree;
+  }
+
+  /// \brief Constructor from a user-supplied knot vector (axom::Array)
+  KnotVector(const axom::Array<T>& knots, int degree)
+    : m_knots(knots)
+    , m_deg(degree)
```
kennyweiss, line 81: Clang warned about initializing these out of order -- please fix

```
axom/src/axom/primal/geometry/KnotVector.hpp:80:7: warning: field 'm_knots' will be initialized after field 'm_deg' [-Wreorder-ctor]
    : m_knots(knots)
      ^~~~~~~~~~~~~~
      m_deg(degree
```

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -0,0 +1,501 @@
+// Copyright (c) 2017-2024, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level LICENSE file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/*!
+ * \file NURBSCurve.hpp
+ *
+ * \brief A NURBS curve primitive
```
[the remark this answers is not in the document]
jcs15c: Looks like there were a lot of these I missed... Thanks for being so thorough in catching them!

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -0,0 +1,501 @@
...
+   * \param [in] npts the number of control points
+   * \param [in] deg the degree
+   * \pre degree is greater than or equal to -1, and deg < npts.
+   */
+  KnotVector(axom::IndexType npts, int deg) { makeUniform(npts, deg); }
+
+  /// \brief Constructor from a user-supplied knot vector (C-style)
+  KnotVector(const T* knots, axom::IndexType nkts, int degree)
+  {
+    m_knots.resize(nkts);
```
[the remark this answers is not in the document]
jcs15c: Good point! Most of these checks are done in the constructors for `NURBSCurve` objects, so I think they should be mirrored here too.

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -0,0 +1,501 @@
...
+    for(int i = 0; i < deg + 1; ++i)
+    {
+      m_knots[i] = 0.0;
+      m_knots[npts + deg - i] = 1.0;
+    }
+
+    // Interior knots (if any)
+    for(int i = 0; i < npts - deg - 1; ++i)
+    {
+      m_knots[deg + 1 + i] = (i + 1.0) / static_cast<T>(npts - deg);
```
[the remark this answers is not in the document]
jcs15c: It should always be positive, since if `npts - deg <= 0` then the for loop won't execute. Do you think it's still worth the check?

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -0,0 +1,501 @@
...
+  /// \brief Returns the index of the knot span and the multiplicity
+  axom::IndexType findSpan(double t, int& multiplicity)
+  {
+    const auto nkts = m_knots.size();
+    const auto span = findSpan(t);
+
+    multiplicity = 0;
+    for(auto i = (t == m_knots[nkts - 1]) ? nkts - 1 : span; i >= 0; --i)
+    {
+      if(m_knots[i] == t)
```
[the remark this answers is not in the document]
jcs15c: Yes, I think it should check strict equality here. Otherwise you open a whole can of worms about what constitutes a clamped curve, i.e. is a 2nd degree curve with a knot vector `{-1e-8, 0, 1e-8, 0.5, 1, 1, 1}` a valid curve?

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -0,0 +1,501 @@
...
+      }
+    }
+
+    return true;
+  }
+
+  /// \brief Check equality of two knot vectors
+  friend inline bool operator==(const KnotVector<T>& lhs, const KnotVector<T>& rhs)
+  {
+    return lhs.m_deg == rhs.m_deg && lhs.m_knots == rhs.m_knots;
```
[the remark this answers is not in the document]
jcs15c: Same as above, plus the added consideration that this strikes me more as a data structure comparison vs. a geometric comparison. This implementation is identical to the `==` operator for Bezier/NURBS curves, which checks for strict equality of arrays of coordinates.

src/axom/primal/geometry/NURBSCurve.hpp
jcs15c: Whoops! Should totally be the min/max knot value.

src/axom/primal/geometry/NURBSCurve.hpp
jcs15c: Yes, that's right. The degree 0 case is the only exception to the "curves must be continuous" rule, and is really only permitted because we allow order 0 Bezier curves that only have a single control point. I've added this exception to the `\brief` for `NURBSCurve`.

src/axom/primal/geometry/NURBSCurve.hpp
jcs15c: I'm not too attached to `setParameters`. but I think it's more clear than `setSize` and more concise than `setNumControlPointsAndDegree`.

src/axom/primal/geometry/NURBSCurve.hpp
jcs15c: It feels okay to me to leave the data there, since I think a function like this (along with the `setDegree` and `setNumControlPoints`) should be used more for initialization of a curve rather than in-place manipulation. Maybe changing the comments makes that expectation more clear?

src/axom/primal/geometry/NURBSCurve.hpp
jcs15c: I'm open to the idea of *adding* an explicitly named function, but the brackets are used in the `BezierCurve` class too, so this is just my way of keeping consistent with the old interface. On a more personal level, I think accessing control points via `[]` is fairly intuitive, and I think it's worth prioritizing the convenience of control point access vs weight/knot access.

More to your second point though, this might be the best time (or maybe last chance?) to update the old interface altogether while they aren't heavily used. For example, we access control nodes of patches with `(p, q)`, so there's already an inconsistency there. Similarly, I think adding `controlPoints()` as a type of accessor is a great idea, but I would also want to add those to `BezierCurve` and `BezierPatch`.

src/axom/primal/tests/primal_nurbs_curve.cpp
jcs15c: Makes sense. While I'm digging around in that test file, it probably makes sense to move the rational Bezier curve winding number test there too.

src/axom/primal/tests/primal_nurbs_curve.cpp
jcs15c: There's one that tests this specifically, but also the `split` test does this, since the knot spans aren't re-normalized after the split.

src/axom/primal/tests/primal_nurbs_curve.cpp
jcs15c: Some of this behavior is tested in the `insertKnot` test, but I've made it more thorough.

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -0,0 +1,501 @@
...
+    {
+      m_knots.insert(m_knots.begin() + span + 1, t);
+    }
+  }
+
+  /*!
+   * \brief Split a knot vector at the value at index `span`
+   * 
+   * \pre Assumes that the multiplicity of the specific knot is equal
+   *    to the degree of the knot vector
```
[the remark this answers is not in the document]
jcs15c: Looking into this revealed to me that this assumption wasn't even met by the `split` method of the `NURBSCurve` class, so I'm going to give this entire interface a second pass. Also a good reason to add tests for `KnotVector` directly, so expect that shortly...

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -0,0 +1,501 @@
...
+      }
+      r *= static_cast<T>(m_deg - k);
+    }
+  }
+
+  /// \brief Return the degree of the knot vector
+  int getDegree() const { return m_deg; }
+
+  /// \brief Return the number of knots in the knot vector
+  axom::IndexType getNumKnots() const { return m_knots.size(); }
```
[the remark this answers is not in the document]
jcs15c: Probably! I don't think I needed it for anything I implemented in the `NURBSCurve` class, but it seems worth the minimal effort to add.

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -0,0 +1,501 @@
...
+  KnotVector(const T* knots, axom::IndexType nkts, int degree)
+  {
+    m_knots.resize(nkts);
+    for(int i = 0; i < nkts; ++i)
+    {
+      m_knots[i] = knots[i];
+    }
+
+    m_deg = degree;
+  }
```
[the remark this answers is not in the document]
jcs15c: I'm open to changing it, as long as it's consistent across the different classes. I wonder if it would be worth a separate PR to handle this and the different array accessors, or just put them all here?

src/axom/primal/geometry/NURBSCurve.hpp
```diff
@@ -777,9 +783,12 @@ class NURBSCurve
     }
   }
 
-  /// \brief Normalize the parameterization of the knot vector
+  /// \brief Normalize the knot vector to the span of [0, 1]
   void normalize() { m_knotvec.normalize(); }
 
+  /// \brief Rescale the knot vector to the span of [a, b]
+  void rescale(T a, T b) { m_knotvec.rescale(a, b); }
```
kennyweiss, line 790: Does this handle the case where $a > b$ or $a == b$?
Please update the doxycomment

src/axom/primal/tests/primal_knot_vector.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+namespace primal = axom::primal;
+
+//------------------------------------------------------------------------------
+TEST(primal_knotvector, degree_constructor)
+{
+  const int degree = 2;
+  const int npts = 5;
+  primal::KnotVector<double> kvector(npts, degree);
+
```
kennyweiss, line 25: Should we also test that the number of spans matches expectations?

src/axom/primal/tests/primal_knot_vector.cpp
kennyweiss, line 1: These tests are really nice -- thanks for adding them!

src/axom/primal/tests/primal_winding_number.cpp
```diff
@@ -476,6 +476,101 @@ TEST(primal_winding_number, degenerate_cases)
     abs_tol);
 }
 
+TEST(primal_winding_number, rational_winding_number)
```
kennyweiss, line 479: Minor: Now that there NURBS tests, should this one be renamed to include Bezier?
```suggestion
TEST(primal_winding_number, rational_bezier_winding_number)
```

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -128,25 +258,88 @@ class KnotVector
...
+  {
+    T min_knot = m_knots[0];
+    T max_knot = m_knots[m_knots.size() - 1];
+    T span = max_knot - min_knot;
+
+    for(int i = 0; i < m_knots.size(); ++i)
+    {
+      m_knots[i] = a + (m_knots[i] - min_knot) * (b - a) / span;
+    }
+  }
```
kennyweiss, line 272: Gut check -- does this work when $a == b$ or when $a > b$?
Please add tests and/or assertions.

Also -- what happens if $min_knot >= max_knot$

kennyweiss (approved):
Thanks for the updates, bugfixes and improved tests!

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -0,0 +1,501 @@
...
+
+    if(t <= m_knots[0])
+    {
+      return m_deg;
+    }
+
+    if(t >= m_knots[nkts - 1])
+    {
+      return nkts - m_deg - 2;
+    }
```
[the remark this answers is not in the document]
jcs15c: I've thought about this a lot, and I think it's best to actually *not* do this kind of clamping. Generally, NURBS objects don't extrapolate as nicely as their Bezier counterparts (without extra checks and such), so evaluating these things outside the range of knots is pretty much always a bad idea. I've removed this behavior from `findSpan`, and added some much more thorough checks in the rest of the methods.

src/axom/primal/geometry/KnotVector.hpp
jcs15c: It shouldn't work, the knot vector wouldn't be monotonic in either case. I'm going to check this with asserts.

src/axom/primal/tests/primal_knot_vector.cpp
jcs15c: I've added some tests of this in the knot insertion test, thanks!

src/axom/primal/tests/primal_nurbs_curve.cpp
```diff
@@ -0,0 +1,831 @@
...
+  EXPECT_EQ(-1, nCurve.getDegree());
+
+  const int degree = 1;
+  const int npts = 3;
+  PointType controlPoints[3] = {PointType {0.6, 1.2, 1.0},
+                                PointType {0.0, 1.6, 1.8},
+                                PointType {0.2, 1.4, 2.0}};
+
+  nCurve.setNumControlPoints(npts);
+  nCurve.setDegree(degree + 1);
```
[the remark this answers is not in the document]
jcs15c: In the new `KnotVector` test, I think I've made more clear the kind of behavior I want to test: that you can decrease the degree of a curve as much as you want, but increasing it too much means you run out of control points to maintain a continuous curve.

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -0,0 +1,702 @@
...
+   * \brief Constructor for a normalized, uniform knot vector
+   *
+   * \param [in] npts the number of control points
+   * \param [in] deg the degree
+   * \pre degree is greater than or equal to -1, and deg < npts.
+   */
+  KnotVector(axom::IndexType npts, int deg) { makeUniform(npts, deg); }
+
+  /// \brief Constructor from a user-supplied knot vector (C-style)
+  KnotVector(const T* knots, axom::IndexType nkts, int degree)
```
BradWhitlock, line 67: A few of these functions are missing "param" documentation.

BradWhitlock (approved):
It will be super-useful to have this NURBSCurve class in primal. Thanks.

src/axom/primal/geometry/NURBSCurve.hpp
```diff
@@ -0,0 +1,1185 @@
...
+
+    makeNonrational();
+  }
+
+  /*!
+   * \brief Constructor for a NURBS curve from a Bezier curve
+   *
+   * \param [in] bezierCurve the Bezier curve to convert to a NURBS curve 
+   */
+  explicit NURBSCurve(BezierCurve<T, NDIMS> bezierCurve)
```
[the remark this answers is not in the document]
jcs15c: Nope! It should be const here and in the others, thanks for catching that!

src/axom/primal/geometry/NURBSCurve.hpp
jcs15c: I think I've come back on this point. I want to enforce that the data should always contain a valid NURBS curve, so I think explicitly resetting all the values is the easiest way to make sure that's the case.

src/axom/primal/operators/winding_number.hpp
```diff
@@ -211,7 +212,7 @@ int winding_number(const Point<T, 2>& R,
  * that constructs a polygon with the same *integer* WN as 
  * the curve closed with a linear segment. The *generalized* WN 
  * of the closing line is then subtracted from the integer WN to
- * return the GWN of the original curve. (See )
+ * return the GWN of the original curve.
```
bmhan12, line 215: nitpick: For clarity, maybe worth it to state the full "generalized winding number"  non-abbreviated form at the top of the file somewhere in the comments, as done in `winding_number.hpp`: https://github.com/LLNL/axom/blob/e5b5707ed09eeb46265b5740a810242970eaaf76/src/axom/primal/operators/winding_number.hpp#L9

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -0,0 +1,702 @@
...
+class KnotVector
+{
+public:
+  AXOM_STATIC_ASSERT_MSG(
+    std::is_arithmetic<T>::value,
+    "A knot vector must be defined using an arithmetic type");
+
+public:
+  /// \brief Default constructor
+  KnotVector() : m_deg(-1) { m_knots.resize(0); }
```
bmhan12, line 55: Might be worth mentioning that the default-constructed `KnotVector `is not valid.

src/axom/primal/geometry/NURBSCurve.hpp
```diff
@@ -0,0 +1,1185 @@
...
+
+    m_knotvec = KnotVectorType(knots, nkts, nkts - npts - 1);
+
+    SLIC_ASSERT(isValidNURBS());
+  }
+
+  /*!
+   * \brief Constructor for a NURBS Curve with axom arrays of data
+   *
+   * \param [in] controlPoints the control points of the curve
```
bmhan12, line 236:
```suggestion
   * \param [in] pts the control points of the curve
```

bmhan12 (approved):

src/axom/primal/geometry/KnotVector.hpp
jcs15c: Good idea, I've clarified this behavior a little bit by making a separate default constructor that always makes an "empty" curve, as distinct from a "degree -1" curve.

src/axom/primal/geometry/NURBSCurve.hpp
jcs15c: Fixed this and others!

PR #1459, opened 2024-10-28
Allow reshaping of `sidre::View` with array data

gunney1:
# Summary

- This PR is a feature
- It does the following:
  - Add `reshapeArray` to `sidre::View`.
  - Add unit tests for new method.
  - Update release notes

# Motivation
Sidre `View` objects don't always have shape that we need.  For example, importing a Blueprint mesh from a `conduit::Node`, we can get 1D arrays.  To then build a mint mesh with the imported `View`, mint insists that the array be a 2D (Nx1) array.

This change is required in https://github.com/LLNL/axom/pull/1455.  I'm separating it to help reduce the size of that PR.


RELEASE-NOTES.md
```diff
@@ -20,6 +20,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 ## [Unreleased] - Release date yyyy-mm-dd
 
 ###  Added
+- `sidre::View` holding array data may now be re-shaped.  See `sidre::View::reshapArray`.
```
nselliott, line 23:
```suggestion
- `sidre::View` holding array data may now be re-shaped.  See `sidre::View::reshapeArray`.
```

src/axom/sidre/core/View.cpp
```diff
@@ -295,6 +295,44 @@ View* View::reallocate(const DataType& dtype)
...
+/*
+ *************************************************************************
+ *
+ * Reshape an array View.
+ *
+ *************************************************************************
+ */
+View* View::reshapeArray(int ndims, const IndexType* shape)
+{
+  SLIC_ERROR_IF(m_state != BUFFER && m_state != EXTERNAL,
```
nselliott, line 307: Implementation choice:  The usual error-checking practice in Sidre for these cases where a required precondition is not met is to use SLIC_CHECK_MSG for a warning and return as a no-op, so that calling code has a chance to check the results and recover.

gunney1: Thanks @nselliott  Good idea to be consistent.  Fixed.

nselliott (approved):

BradWhitlock (approved):
This change looks fine.

src/axom/sidre/core/View.hpp
```diff
@@ -431,6 +431,21 @@ class View
 
   //@}
 
+  /*!
+   * \brief Reshape the array without changing its size.
+   *
+   * The View state must be either BUFFER or EXTERNAL.
+   * The new shape must have the same size as the current.
```
kennyweiss, line 438: Clarification -- do you think we should clarify what "same size" mean w.r.t. shapes (especially if the View originally did not have a shape)? Or do you think it will be clear to users?

Perhaps it would help to mention how "same size" is related to `View::getNumElements()`?

kennyweiss: Also, do have/need a way to remove/clear a shape from a View, i.e. to return it to a 1D array?

src/axom/sidre/tests/sidre_view.cpp
```diff
@@ -1852,6 +1852,119 @@ TEST(sidre_view, deep_copy_shape)
...
+    EXPECT_EQ(viewA->getNumElements(), 12);
+    EXPECT_EQ(nDim, 1);
+    EXPECT_EQ(shapeOutput[0], 12);
+  }
+
+  {
+    // Attempt to reshape empty View should be no-op.
+    EXPECT_TRUE(viewA->isEmpty());
+    IndexType badShape[] = {1, 2, 3};
+    viewA->reshapeArray(3, badShape);
```
kennyweiss, line 1880: Note: This is a "no-op", but it will emit a warning. 
Should we add a `SLIC_INFO` that the warning can be ignored?

src/axom/sidre/tests/sidre_view.cpp
```diff
@@ -1852,6 +1852,119 @@ TEST(sidre_view, deep_copy_shape)
...
+    IndexType shape2d[] = {2, 6};
+    viewA->reshapeArray(2, shape2d);
+    nDim = viewA->getShape(DMAX, shapeOutput);
+    EXPECT_EQ(viewA->getNumElements(), 12);
+    EXPECT_EQ(nDim, 2);
+    EXPECT_EQ(shapeOutput[0], 2);
+    EXPECT_EQ(shapeOutput[1], 6);
+  }
+
+  viewA->allocate();
```
kennyweiss, line 1924: Can you please add a brief comment about how the `deallocate`/`allocate` calls affect the test?
E.g. if this block is testing the operation of `reshape` w/ an unallocated array.

src/axom/sidre/core/View.cpp
```diff
@@ -295,6 +295,50 @@ View* View::reallocate(const DataType& dtype)
...
+    SLIC_WARNING(SIDRE_VIEW_LOG_PREPEND
+                 << "View can only reshape array states (BUFFER or EXTERNAL.");
+    return this;
+  }
+
+  IndexType oldSize = 1;
+  for(const auto& s : m_shape)
+  {
+    oldSize *= s;
+  }
```
kennyweiss, line 318: Suggestion (clarity): 
This might be clearer is `oldSize` was simply set to `this->getNumElements()`

src/axom/sidre/tests/sidre_view.cpp
```diff
@@ -1852,6 +1852,119 @@ TEST(sidre_view, deep_copy_shape)
...
+  auto* viewA =
+    root->createView("viewA", sidre::detail::SidreTT<int32_t>::id, 12);
+
+  {
+    nDim = viewA->getShape(DMAX, shapeOutput);
+    EXPECT_EQ(viewA->getNumElements(), 12);
+    EXPECT_EQ(nDim, 1);
+    EXPECT_EQ(shapeOutput[0], 12);
+  }
+
```
kennyweiss, line 1875: Would it make sense to add some tests w/ 1D versions of `reshapeArray` and `N x 1 x 1` versions
E.g.
```
reshapeArray(1,12);
reshapeArray(3, [12,1,1]);
```
in cases where the dimension was already 1 and in cases where it was different?

kennyweiss (approved):
Thanks @gunney1 

I added a few suggestions for extra tests, and also about indicating that some tests are expected to emit warnings.

src/axom/sidre/core/View.hpp
gunney1: Referenced `getNumElements()` as size.  There is a way to clear a shape from View:
`IndexType s = v->getNumElements(); v->reshapeArray(1, &s);`  I'll add it to the test.

src/axom/sidre/tests/sidre_view.cpp
gunney1: Good idea.  Done.

src/axom/sidre/tests/sidre_view.cpp
gunney1: Good idea.  Done.

src/axom/sidre/tests/sidre_view.cpp
gunney1: Sure.  Done.

src/axom/sidre/core/View.cpp
gunney1: Agreed.

Issue #1458, opened 2024-10-26
Deep copying `View` should use allocator id of parent


gunney1:
There are 2 ways to deep copy data into a `sidre::View`, and they should allocate array data the same way.

Deep-copying from a source `View` currently allocates data using the destination's parent's default allocator id.  However, deep-copying from a source `conduit::Node`, using `View::importArrayNode()`, currently allocates data on the host, ignoring the parent's default allocator id.  We should change the second version to behave like the first.  

This proposal constitutes a behavior change, which should be documented.  While it is possible to copy data to the desired memory space after the deep copy, to do so for a large hierarchy is messy and wasteful.

gunney1:
Resolved with MR https://github.com/LLNL/axom/pull/1460.

PR #1460, opened 2024-10-28
Feature/gunney/conduit import memory space

gunney1:
# Summary

- This PR is a feature
- It does the following:
  - Changes where destination data is allocated when importing array from Conduit.  Currently, data is allocated on host.  New, data is allocated using the allocator of the parent `Group`.
  - Fixes issue https://github.com/LLNL/axom/issues/1458
  - Updates RELEASE-NOTES.

Without this change, it is still possible to re-allocate data after the import, but that would require walking a `Group` hierarchy to look for `View`'s with data and re-allocating them individually.  The new behavior is consistent with the behavior when deep-copying from a Sidre source.

# Motivation
This change was needed in the work to support Blueprint meshes in shaping, https://github.com/LLNL/axom/issues/1452.  I wanted to separate it to keep that PR as clean as possible.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -22,6 +22,10 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 ###  Added
 
 ###  Changed
+- Importing Conduit array data into `sidre::View` now allocates destination
+  data using the `View`'s parent's allocator ID, insteading of always using
+  host memory.  This is consistent with the behavior of deep-copying data
+  from Sidre.
 
 ###  Deprecated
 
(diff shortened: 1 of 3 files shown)


RELEASE-NOTES.md
```diff
@@ -22,6 +22,10 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 ###  Added
 
 ###  Changed
+- Importing Conduit array data into `sidre::View` now allocates destination
+  data using the `View`'s parent's allocator ID, insteading of always using
```
kennyweiss, line 26:
```suggestion
  data using the `View`'s parent's allocator ID, instead of always using
```

kennyweiss (approved):
Thanks @gunney1

BradWhitlock (approved):
It looks like it will allocate using the parent allocator, as intended.

kennyweiss (approved):
Thanks for updating the API and the new tests!

PR #1470, opened 2024-11-12
Adds untrimmed NURBS patches to primal

jcs15c:
# Summary

- This PR is a feature
- It adds a `NURBSPatch` class to `primal`. `NURBSPatch` has the following features, along with relevant tests:
  - Construictors from arrays of points (optionally weights) and either the degree or knot vector
  - Evaluation/Derivative methods
  - Knot insertion
  - Patch subdivision/Bezier extraction
- It also enhances the Bezier extraction method of `NURBSCurve` to be more efficient, performing the knot insertion in place to avoid repeated calculations.
  - An extra test has been added to `primal_nurbs_curve.cpp` to test a currently unchecked edge case.
- Minor bug-fixes in `NURBSCurve` and `KnotVector`

### Additional Changes
- Rework the `KnotVector` and `NURBSCurve` classes to be more tolerant of parameters slightly outside the range of the knots.


src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -0,0 +1,2745 @@
...
+   * Adapted from Algorithm A3.5 on page 103 of "The NURBS Book"
+   * 
+   * \pre Requires \a u, v in the span of each knot vector
+   */
+  PointType evaluate(T u, T v) const
+  {
+    SLIC_ASSERT(u >= m_knotvec_u[0] &&
+                u <= m_knotvec_u[m_knotvec_u.getNumKnots() - 1]);
+    SLIC_ASSERT(v >= m_knotvec_v[0] &&
+                v <= m_knotvec_v[m_knotvec_v.getNumKnots() - 1]);
```
kennyweiss, line 736: This logic triggered an assert for me (in the NURBS curve class) when using a for loop with doubles, e.g.
```cpp
const double dt = 0.1
for(double u=0; u<=1.; u+=dt)
{
   ...
}
```

Would it make sense to check for fuzzy equals and clamp to the endpoints?

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -0,0 +1,2745 @@
...
+    setDegree_v(deg_v);
+  }
+
+  /*!
+   * \brief Set the number control points in u
+   *
+   * \param [in] npts The target number of control points
+   * 
+   * \warning This method does NOT maintain the patch shape,
+   *  i.e. is not performing knot insertion/removal.
```
kennyweiss, line 895: I think another warning is warranted that this will reset the knot vectors

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -0,0 +1,2745 @@
...
+    m_knotvec_v.makeUniform(npts, getDegree_v());
+  }
+
+  /*!
+   * \brief Set the knot value in the u vector at a specific index
+   *
+   * \param [in] idx The index of the knot
+   * \param [in] knot The updated value of the knot
+   */
+  void setKnot_u(int idx, T knot) { m_knotvec_u[idx] = knot; }
```
kennyweiss, line 963: Should these check that `idx` is in range? Perhaps w/ a SLIC_ASSERT if we don't want it to affect performance in release configs

kennyweiss: Might also be a good idea to note that users can/should run a consistency check, e.g. for monotonicity after modifying the knots.

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -0,0 +1,2745 @@
...
+  int getDegree_u() const { return m_knotvec_u.getDegree(); }
+
+  /// \brief Returns the degree of the NURBS Patch on the second axis
+  int getDegree_v() const { return m_knotvec_v.getDegree(); }
+
+  /// \brief Returns the order of the NURBS Patch on the first axis
+  int getOrder_u() const { return m_knotvec_u.getDegree() + 1; }
+
+  /// \brief Returns the order of the NURBS Patch on the second axis
+  int getOrder_v() const { return m_knotvec_v.getDegree() + 1; }
```
kennyweiss, line 1023: Might be good to reinforce the difference b/w degree and order in the doxygen comments. (and perhaps how each is used).

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -0,0 +1,2745 @@
...
+  KnotVectorType getKnots_u() const { return m_knotvec_u; }
+
+  /// \brief Return a copy of the knot vector on the first axis as an array
+  axom::Array<T> getKnotsArray_u() const { return m_knotvec_u.getArray(); }
+
+  /// \brief Return a copy of the knot vector on the second axis
+  KnotVectorType getKnots_v() const { return m_knotvec_v; }
+
+  /// \brief Return a copy of the knot vector on the second axis as an array
+  axom::Array<T> getKnotsArray_v() const { return m_knotvec_v.getArray(); }
```
kennyweiss, line 1035: The distinction here might be a bit confusing. The former is the KnotVector instance, while the latter is the underlying array.

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -0,0 +1,2745 @@
...
+   * 
+   * \pre We require evaluation of the patch at \a u and \a v between 0 and 1
+   */
+  VectorType normal(T u, T v) const
+  {
+    PointType eval;
+    VectorType Du, Dv;
+    evaluateFirstDerivatives(u, v, eval, Du, Dv);
+
+    return VectorType::cross_product(Du, Dv);
```
kennyweiss, line 1737: Are there cases where the derivative is not defined, e.g. near cusps? If so, what will the normal be?

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -0,0 +1,2745 @@
...
+  }
+
+private:
+  CoordsMat m_controlPoints;
+  WeightsMat m_weights;
+  KnotVectorType m_knotvec_u, m_knotvec_v;
+};
+
+//------------------------------------------------------------------------------
+/// Free functions related to BezierPatch
```
kennyweiss, line 2733: Typo:
```suggestion
/// Free functions related to NURBSPatch
```

kennyweiss (approved):
Thank you @jcs15c for this great new feature w/ comprehensive tests!

src/axom/primal/geometry/NURBSPatch.hpp
jcs15c: Fixed this and others!

src/axom/primal/geometry/NURBSPatch.hpp
jcs15c: The `KnotVector` class itself does these checks after calling the `[]` operator, since in my original inception the `KnotVector` class should be mostly responsible for making sure that it's valid. To your second point, I don't think there's any way (in debug mode, and assuming I didn't miss anything) to create an invalid knot vector in the first place, but we can be more clear in the documentation about what `isValidNURBS` is for.

src/axom/primal/geometry/NURBSPatch.hpp
jcs15c: Frankly my preference would be to just not have a `getOrder` method in this or the NURBS curve class, since it just adds an extra layer of confusion. But while "degree" is certainly the convention for NURBS objects, referencing the order isn't unheard of. I'll clarify the comments regardless.

src/axom/primal/geometry/NURBSPatch.hpp
jcs15c: Does this rework help clarify things?

src/axom/primal/geometry/NURBSPatch.hpp
jcs15c: It's possible that one or both derivatives are zero, in which case the normal would be zero too. That doesn't mean that the surface itself doesn't have a normal there (like for a corner of a degenerate triangular patch), but the normal with respect to this *parameterization* is zero. Accounting for all the relevant edge cases is tricky, and I haven't worked out a consistent system for that yet.

src/axom/primal/tests/primal_nurbs_patch.cpp
```diff
@@ -0,0 +1,1322 @@
...
+
+  NURBSPatchType nPatch(controlPoints, weights, npts_u, npts_v, degree_u, degree_v);
+
+  // Loop over the parameter space of the surface,
+  //  and check that `evalauteDerivatives` matches the results of the two isocurve methods
+  for(double u = 0; u <= 1; u += 0.1)
+  {
+    for(double v = 0; v <= 1; v += 0.1)
+    {
+      axom::Array<VectorType, 2> ders;
```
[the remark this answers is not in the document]
jcs15c: This `VectorType` refers to a geometric `primal::Vector`, not a `std::vector` array.

src/axom/primal/tests/primal_nurbs_patch.cpp
```diff
@@ -0,0 +1,1322 @@
...
+
+  double split_vals[3] = {0.3, 0.5, 0.7};
+
+  for(double val : split_vals)
+  {
+    nPatch.split_u(val, subpatch1, subpatch2);
+
+    for(double u = 0; u <= 1; u += 0.1)
+    {
+      for(double v = 0; v <= 1; v += 0.1)
```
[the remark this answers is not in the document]
jcs15c: There's a `numerics::linspace` somewhere that I could use for this purpose, should be pretty easy to incorporate.

kennyweiss (approved):
Thanks for the updates!

src/axom/primal/tests/primal_nurbs_patch.cpp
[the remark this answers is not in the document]
jcs15c: There is a `VectorType` defined within the NURBSPatch class, do you think it should be defined like this throughout the test? Or maybe instead of the `using` format, just define use it as `NURBSPatchType::VectorType` throughout.

```C++
const int DIM = 3;
using CoordType = double;
using NURBSPatchType = primal::NURBSPatch<CoordType, DIM>;
using VectorType = NURBSPatchType::VectorType;
```

src/axom/primal/tests/primal_nurbs_patch.cpp
```diff
@@ -609,19 +611,25 @@ TEST(primal_nurbspatch, surface_evaluate)
...
+
+  constexpr int npts = 11;
+  double u_pts[npts], v_pts[npts];
+  axom::numerics::linspace(0.0, 1.0, u_pts, npts);
+  axom::numerics::linspace(0.0, 1.0, v_pts, npts);
+
+  for(auto u : u_pts)
   {
-    for(double v = 0; v <= 1; v += 0.1)
+    for(auto v : v_pts)
```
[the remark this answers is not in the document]
jcs15c: Ah, I didn't realize `numerics::linspace` had the same issue! If there was such an epsilon to `KnotVector`, I think clamping to the actual range is the way to go, but it isn't clear to me what that `eps` should be, maybe an optional parameter? For `BezierCurve` objects, that little tolerance isn't super important, since they smoothly extrapolate beyond `[0, 1]`, but that isn't the case for NURBS classes.

Would it be worth making a small change to that `linspace` function to just manually set `v[N-1] = x1` at the end of the loop?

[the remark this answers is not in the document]
kennyweiss: In the short term, I think modifying `linspace` so that the last entry is always `x1` makes sense.

src/axom/primal/tests/primal_nurbs_patch.cpp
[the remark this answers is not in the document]
kennyweiss: I was thinking it might make sense to have something like 
```cpp
using Derivatives = axom::Array<VectorType>;  // or DerivativesType (?)
```
in `NURBSCurve` and/or `NURBSPatch`

It's not strictly necessary though.

src/axom/primal/geometry/NURBSPatch.hpp
kennyweiss: Thanks!

src/axom/primal/geometry/NURBSPatch.hpp
kennyweiss: Would it make sense to add a note? Or perhaps later when the edge cases are more worked out?

src/axom/primal/tests/primal_nurbs_patch.cpp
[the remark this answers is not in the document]
jcs15c: Ah, I see what you mean, I actually like the idea of having a specific `DerivativeType` in those classes, but I would want that to be reflected in all the other curved objects. Unless it's worth it to change all those classes now, I might save it for a later PR that also addresses some other things that I think would be good to rework among all of them (like array accessing).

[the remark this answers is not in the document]
kennyweiss: Thanks -- that works for me

src/axom/primal/geometry/NURBSPatch.hpp
jcs15c: Couldn't hurt to add a comment here, although it isn't like returning a zero vector is itself incorrect. For a simple cusp on a 2D curve, the tangent vector at the point of the cusp really is just continuously zero, and although I have less intuition for the 3D case, I'm fairly sure that's the case there too. Even though the direction of the tangent/normal vector changes depending on the direction of approach, the magnitude goes to zero regardless. In my experience, the more pressing issue is when you need to take a *unit* normal vector at a point like that, since that genuinely is always going to be discontinuous at the cusp. This is a bit troublesome with the current `primal::Vector` implementation, which returns something like `(1,0,0)` by default if the norm is zero.

As an example of how I've dealt with these geometric/unit vectors before, the 2D winding number classes need to know tangent vectors at endpoints, and in those cases I manually iterated over control points until I found a pair that correctly defined a geometric tangent vector. That doesn't strictly help with the cusp case, but in 2D you could imagine splitting the curve at the point of discontinuity and handling each half that way.

bmhan12 (approved):
LGTM!

One thing missing is `NURBSPatch.hpp` should be added to the list of headers in `primal/CMakeLists.txt`

jcs15c:
I made a few extra changes to the `KnotVector` class and the NURBS objects to make them more tolerant of input parameter values that are slightly outside of the range of the knot values, which happens surprisingly often with floating point arithmetic. These checks for "slightly outside the range" are handled by the `KnotVector::isValidParameter` method, which is the only place where the exact value of this numerical tolerance is kept.

In the future, it might be desirable to give NURBS curves and surfaces the ability to extrapolate beyond their knot spans slightly, as is naturally the case for the analogous Bezier objects, but that's a comparatively computationally expensive procedure compared to simple clamping to handle floating point errors. This is also why I don't think it's necessary to expose this tolerance parameter to a user, and to some extent, I prefer the idea of all instances of these objects using the same tolerance level.

@kennyweiss What do you think of this as a solution?

kennyweiss:
> I made a few extra changes to the `KnotVector` class and the NURBS objects to make them more tolerant of input parameter
> 
> @kennyweiss What do you think of this as a solution?

Thanks @jcs15c -- I think this is a reasonable and sufficient solution for now. If we need to extrapolate at parametric points near the boundary in the future, it should be easy to add that capability (at the expense of some additional computational work, as you mentioned).

PR #1482, opened 2024-12-26
Fix a memory leak in axom::Array copy constructor

publixsubfan:
# Summary

- Fixes leak in `axom::Array` copy constructor, along the lines of the fix in commit ae8e8c7
- Add unit test for leaky copy/move constructors and assignment operators


src/axom/core/tests/core_array.hpp
```diff
@@ -2389,102 +2389,52 @@ TEST(core_array, reserve_nontrivial_reloc_um)
...
-  }
-  arr2.shrink();
-
-  return {std::move(arr1), std::move(arr2)};
-}
+private:
+  axom::Array<int> m_value;
+};
 
-TEST(core_array, regression_array_move)
```
publixsubfan, line 2423: @kennyweiss -- I replaced the regression test you added in #1286 with a unit test that I think should capture all the leaky cases you were seeing. Let me know if you'd rather keep the regression test as well.

rhornung67 (approved):
Thanks @publixsubfan

gunney1 (approved):
Thank you @publixsubfan

src/axom/core/tests/core_array.hpp
kennyweiss: Thanks @publixsubfan -- it looks like it covers similar/more ground, so I'm ok w/ removing the old regression case.

Did you check the bugfix against lsan and/or valgrind or a similar tool?

kennyweiss (approved):
Thanks @publixsubfan -- please update the RELEASE-NOTES to reflect the bugfix

publixsubfan:
@kennyweiss - yep, I tested with an lsan build with and without the bugfix commit.

PR #1483, opened 2024-12-30
Fix HIP build when not using IntersectionShaper

gunney1:
Fix HIP build when not using IntersectionShaper.  IntersectionShaper is not always active, so we shouldn't rely on what it defines.


kennyweiss (approved):

rhornung67 (approved):
Not sure I like the conditional compilation blocks repeated throughout this code. But reworking that is outside the scope of the issue this PR is addressing.

PR #1491, opened 2025-01-15
Silence warnings

gunney1:
Silence warnings about unused variables and shadow parameters.

### DIFF
--- a/src/axom/primal/geometry/KnotVector.hpp
+++ b/src/axom/primal/geometry/KnotVector.hpp
@@ -546,7 +546,7 @@ public:
   {
     SLIC_ASSERT(isValidSpan(span, t));
 
-    const int m_deg = getDegree();
+    const int deg = getDegree();
 
     axom::Array<axom::Array<T>> ders(n + 1);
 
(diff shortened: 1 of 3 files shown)


src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -546,25 +546,25 @@ class KnotVector
   {
     SLIC_ASSERT(isValidSpan(span, t));
 
-    const int m_deg = getDegree();
+    const int deg = getDegree();
```
kennyweiss, line 549: It looks like all of the changes to this function can be undone if we remove this line since `getDegree()` just returns `m_deg`

@jcs15c -- could you please confirm?

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -546,25 +546,25 @@ class KnotVector
...
 
     axom::Array<axom::Array<T>> ders(n + 1);
 
     axom::Array<axom::Array<T>> ndu(m_deg + 1), a(2);
     axom::Array<T> left(m_deg + 1), right(m_deg + 1);
     for(int j = 0; j <= m_deg; j++)
```
kennyweiss, line 555: If `deg` is local, then these should probably used `deg` instead of `m_deg`

src/axom/primal/geometry/KnotVector.hpp
jcs15c: Yes, that's right, you can just remove the assignment. Sorry about that! Probably an artifact from when this was a function in `NURBSCurve` instead of in the dedicated `KnotVector` class.

rhornung67 (approved):
Thanks @gunney1

jcs15c (approved):
Thanks for catching this @gunney1!

src/axom/primal/geometry/KnotVector.hpp
gunney1: Since `deg` and `m_deg` are the same, I used `m_deg` throughout, as is done in other methods.  Removing the assignment as @jcs15c suggested.

kennyweiss (approved):
Thanks @gunney1!

src/axom/primal/geometry/KnotVector.hpp
kennyweiss: Thanks!

PR #1493, opened 2025-01-21
2D polygon robustness

### LINKED ISSUE #1481: axom::primal::clip() for polygons is unreliable.
I was trying to clip 2 polygons (a quad and a triangle) to determine their overlap and the clip() routine sometimes makes invalid polygons, which can lead to wrong results or a crash.

```
#include <axom/primal.hpp>

#include <iostream>

#if 1
#include <cmath>
// Don't use these in real test
#define EXPECT_TRUE(value) if(!(value)) {std::cout << #value << " was false but should be true" << std::endl; }
#define EXPECT_NEAR(A,B,EPS) if(fabs((A)-(B))>=(EPS)) {std::cout << "Difference of " << #A << " (" << (A) << ") and " << #B << " (" << (B) << ") greater than " << (EPS) << std::endl; }
#endif

void run_clip_test()
{
  using Polygon = axom::primal::Polygon<double, 2>;
  using Point = axom::primal::Point<double, 2>;

  // 3x3 grid of quads that occupies unit square.
  const double x[] = {0., 1./3., 2./3, 1.,
                      0., 1./3., 2./3, 1.,
                      0., 1./3., 2./3, 1.,
                      0., 1./3., 2./3, 1.};
  const double y[] = {0., 0., 0., 0.,
                      1./3., 1./3, 1./3., 1./3.,
                      2./3., 2./3, 2./3., 2./3.,
                      1., 1., 1., 1.};
  const int conn[][4] = {{0,1,5,4}, {1,2,6,5}, {2,3,7,6},
                         {4,5,9,8}, {5,6,10,9}, {6,7,11,10},
                         {8,9,13,12}, {9,10,14,13}, {10,11,15,14}};
  Polygon fine[9];
  for(int p = 0; p < 9; p++)
  {
    fine[p].addVertex(Point{x[conn[p][0]], y[conn[p][0]]});
    fine[p].addVertex(Point{x[conn[p][1]], y[conn[p][1]]});
    fine[p].addVertex(Point{x[conn[p][2]], y[conn[p][2]]});
    fine[p].addVertex(Point{x[conn[p][3]], y[conn[p][3]]});
  }

  // Overlap polygons
  Polygon clipLower;
  clipLower.addVertex(Point{0., 0.});
  clipLower.addVertex(Point{1., 0.});
  clipLower.addVertex(Point{0., 1.});

  Polygon clipUpper;
  clipUpper.addVertex(Point{1., 0.});
  clipUpper.addVertex(Point{1., 1.});
  clipUpper.addVertex(Point{0., 1.});

  // Expected VFs for overlaps.
  const double lower_vf[] = {1., 1., 0.5,
                             1., 0.5, 0.,
                             0.5, 0., 0.};
  const double upper_vf[] = {0., 0., 0.5,
                             0., 0.5, 1.,
                             0.5, 1., 1.};

  constexpr double eps = 1.e-8;
  constexpr double fine_area = (1./3.) * (1./3.);
  for(int p : std::vector<int>{0,1,2,3,4,6})
  {
    // Clip clipLower with fine quad
    const auto overlapPoly = axom::primal::clip(clipLower, fine[p], eps);
    std::cout << "p=" << p
              << ", subject=" << clipLower
              << ", clip=" << fine[p]
              << ", overlap=" << overlapPoly << std::endl;
    EXPECT_TRUE(overlapPoly.isValid());

    if(overlapPoly.isValid())
    {
      const double vf = overlapPoly.area() / fine_area;
      EXPECT_NEAR(vf, lower_vf[p], eps);
    }
  }
  for(int p : std::vector<int>{2,4,5,6,7,8})
  {
    // Clip clipUpper with fine quad
    const auto overlapPoly = axom::primal::clip(clipUpper, fine[p], eps);
    std::cout << "p=" << p
              << ", subject=" << clipUpper
              << ", clip=" << fine[p]
              << ", overlap=" << overlapPoly << std::endl;
    EXPECT_TRUE(overlapPoly.isValid());

    if(overlapPoly.isValid())
    {
      const double vf = overlapPoly.area() / fine_area;
      EXPECT_NEAR(vf, lower_vf[p], eps);
    }
  }
}

int main(int argc, char *argv[])
{
  run_clip_test();
  return 0;
}
```

p=0, subject={3-gon:(0,0),(1,0),(0,1)}, clip={4-gon:(0,0),(0.333333,0),(0.333333,0.333333),(0,0.333333)}, overlap={4-gon:(0,0.333333),(0,0),(0.333333,0),(0.333333,0.333333)}
p=1, subject={3-gon:(0,0),(1,0),(0,1)}, clip={4-gon:(0.333333,0),(0.666667,0),(0.666667,0.333333),(0.333333,0.333333)}, overlap={4-gon:(0.333333,0.333333),(0.333333,0),(0.666667,0),(0.666667,0.333333)}
p=2, subject={3-gon:(0,0),(1,0),(0,1)}, clip={4-gon:(0.666667,0),(1,0),(1,0.333333),(0.666667,0.333333)}, overlap=**{3-gon:(0.666667,0),(1,0),(0,0)}**
**Difference of vf (0) and lower_vf[p] (0.5) greater than 1e-08**
p=3, subject={3-gon:(0,0),(1,0),(0,1)}, clip={4-gon:(0,0.333333),(0.333333,0.333333),(0.333333,0.666667),(0,0.666667)}, overlap={4-gon:(0,0.666667),(0,0.333333),(0.333333,0.333333),(0.333333,0.666667)}
p=4, subject={3-gon:(0,0),(1,0),(0,1)}, clip={4-gon:(0.333333,0.333333),(0.666667,0.333333),(0.666667,0.666667),(0.333333,0.666667)}, overlap=**{4-gon:(0.333333,0.333333),(0.666667,0.333333),(0.666667,0.333333),(0,0)}**
cliptest: ArrayBase.hpp:663: T &axom::ArrayBase<axom::primal::Point<double, 2>, 1, axom::Array<axom::primal::Point<double, 2>, 1, axom::MemorySpace::Dynamic>>::operator[](const axom::IndexType) [T = axom::primal::Point<double, 2>, DIM = 1, ArrayType = axom::Array<axom::primal::Point<double, 2>, 1, axom::MemorySpace::Dynamic>]: Assertion `inBounds(idx)' failed.
Abort (core dumped)

bmhan12:
This PR:
- Adds robustness example to unit tests from #1481 
- Fixes robustness issue with 2D polygon clipping by:
  - Adding missing tolerance parameter to `intersect` overload   between plane and segment
  - Generalizing, simplifying logic to handle duplicate vertices in Sutherland–Hodgman algorithm by performing a pass at the end to remove duplicates.

Closes #1481

### DIFF
--- a/src/axom/primal/operators/detail/clip_impl.hpp
+++ b/src/axom/primal/operators/detail/clip_impl.hpp
@@ -763,7 +763,13 @@ AXOM_HOST_DEVICE Polygon<T, 2, ARRAY_TYPE, MAX_VERTS> clipPolygonPolygon(
   for(int i = 0; i < outputList.numVertices(); i++)
   {
     int prevIndex = ((i - 1) == -1) ? (outputList.numVertices() - 1) : (i - 1);
-    if(outputList[i] != outputList[prevIndex])
+
+    if(!axom::utilities::isNearlyEqual(outputList[i][0],
+                                       outputList[prevIndex][0],
+                                       eps) ||
+       !axom::utilities::isNearlyEqual(outputList[i][1],
+                                       outputList[prevIndex][1],
+                                       eps))
     {
       uniqueList.addVertex(outputList[i]);
     }
--- a/src/axom/primal/operators/detail/clip_impl.hpp
+++ b/src/axom/primal/operators/detail/clip_impl.hpp
@@ -732,7 +732,7 @@ AXOM_HOST_DEVICE Polygon<T, 2, ARRAY_TYPE, MAX_VERTS> clipPolygonPolygon(
       PointType intersecting_point;
       SegmentType subject_edge(prev_point, current_point);
 
-      if(intersect(plane, subject_edge, seg_param))
+      if(intersect(plane, subject_edge, seg_param, eps))
       {
         intersecting_point = subject_edge.at(seg_param);
       }
@@ -742,13 +742,7 @@ AXOM_HOST_DEVICE Polygon<T, 2, ARRAY_TYPE, MAX_VERTS> clipPolygonPolygon(
 
       if(cur_p_orientation == ON_POSITIVE_SIDE)
       {
-        // Handles the edge case of 3 consecutive vertices with orientations
-        // ON_POSITIVE_SIDE, ON_BOUNDARY, ON_POSITIVE. Default algorithm
-        // check (prev_p_orientation != ON_POSITIVE_SIDE) results in the
-        // vertex on the boundary being added twice.
-        if(prev_p_orientation == ON_NEGATIVE_SIDE ||
-           (prev_p_orientation == ON_BOUNDARY &&
-            intersecting_point != outputList[outputList.numVertices() - 1]))
+        if(prev_p_orientation != ON_POSITIVE_SIDE)
         {
           outputList.addVertex(intersecting_point);
         }
@@ -761,7 +755,22 @@ AXOM_HOST_DEVICE Polygon<T, 2, ARRAY_TYPE, MAX_VERTS> clipPolygonPolygon(
     }
   }  // end of iteration through edges of clip polygon
 
-  return outputList;
+  // Remove duplicate points.
+  // Handles edge cases such as 3 consecutive vertices with orientations
+  // ON_POSITIVE_SIDE, ON_BOUNDARY, ON_POSITIVE where point on the boundary
+  // is added twice.
+  PolygonType uniqueList;
+  for(int i = 0; i < outputList.numVertices(); i++)
+  {
+    int prevIndex = ((i - 1) == -1) ? (outputList.numVertices() - 1) : (i - 1);
+    if(outputList[i] == outputList[prevIndex])
+    {
+      continue;
+    }
+    uniqueList.addVertex(outputList[i]);
+  }
+
+  return uniqueList;
 }
 
 }  // namespace detail
--- a/src/axom/primal/operators/detail/clip_impl.hpp
+++ b/src/axom/primal/operators/detail/clip_impl.hpp
@@ -763,11 +763,10 @@ AXOM_HOST_DEVICE Polygon<T, 2, ARRAY_TYPE, MAX_VERTS> clipPolygonPolygon(
   for(int i = 0; i < outputList.numVertices(); i++)
   {
     int prevIndex = ((i - 1) == -1) ? (outputList.numVertices() - 1) : (i - 1);
-    if(outputList[i] == outputList[prevIndex])
+    if(outputList[i] != outputList[prevIndex])
     {
-      continue;
+      uniqueList.addVertex(outputList[i]);
     }
-    uniqueList.addVertex(outputList[i]);
   }
 
   return uniqueList;
--- a/src/axom/primal/operators/detail/intersect_impl.hpp
+++ b/src/axom/primal/operators/detail/intersect_impl.hpp
@@ -1102,8 +1102,7 @@ AXOM_HOST_DEVICE bool intersect_plane_bbox(const Plane<T, 3>& p,
  * \param [in] seg A segment
  * \param [out] t Intersection point of plane and seg, w.r.t. 
  *   parametrization of seg
- * \param [in] EPS tolerance parameter for determining if \a t
- *             is just within 0.0 <= t <= 1.0 .
+ * \param [in] EPS tolerance parameter for determining if 0.0 <= t <= 1.0
  * \return true iff plane intersects with segment, otherwise, false.
  */
 template <typename T, int DIM>
--- a/src/axom/primal/operators/detail/intersect_impl.hpp
+++ b/src/axom/primal/operators/detail/intersect_impl.hpp
@@ -1102,12 +1102,15 @@ AXOM_HOST_DEVICE bool intersect_plane_bbox(const Plane<T, 3>& p,
  * \param [in] seg A segment
  * \param [out] t Intersection point of plane and seg, w.r.t. 
  *   parametrization of seg
+ * \param [in] EPS tolerance parameter for determining if \a t
+ *             is just within 0.0 <= t <= 1.0 .
  * \return true iff plane intersects with segment, otherwise, false.
  */
 template <typename T, int DIM>
 AXOM_HOST_DEVICE bool intersect_plane_seg(const Plane<T, DIM>& plane,
                                           const Segment<T, DIM>& seg,
-                                          T& t)
+                                          T& t,
+                                          double EPS = 1E-12)
 {
   using VectorType = Vector<T, DIM>;
 
@@ -1117,7 +1120,7 @@ AXOM_HOST_DEVICE bool intersect_plane_seg(const Plane<T, DIM>& plane,
   t = (plane.getOffset() - normal.dot(VectorType(seg.source()))) /
     (normal.dot(ab));
 
-  if(t >= 0.0 && t <= 1.0)
+  if(isGeq(t, 0.0, EPS) && isLeq(t, 1.0, EPS))
   {
     return true;
   }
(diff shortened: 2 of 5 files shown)


src/axom/primal/operators/detail/clip_impl.hpp
```diff
@@ -761,7 +755,22 @@ AXOM_HOST_DEVICE Polygon<T, 2, ARRAY_TYPE, MAX_VERTS> clipPolygonPolygon(
...
-  return outputList;
+  // Remove duplicate points.
+  // Handles edge cases such as 3 consecutive vertices with orientations
+  // ON_POSITIVE_SIDE, ON_BOUNDARY, ON_POSITIVE where point on the boundary
+  // is added twice.
+  PolygonType uniqueList;
+  for(int i = 0; i < outputList.numVertices(); i++)
+  {
+    int prevIndex = ((i - 1) == -1) ? (outputList.numVertices() - 1) : (i - 1);
+    if(outputList[i] == outputList[prevIndex])
```
adayton1, line 766: I think I would change this to a != comparison, remove the continue, and add the vertex inside this conditional.

src/axom/primal/operators/detail/intersect_impl.hpp
```diff
@@ -1102,12 +1102,15 @@ AXOM_HOST_DEVICE bool intersect_plane_bbox(const Plane<T, 3>& p,
  * \param [in] seg A segment
  * \param [out] t Intersection point of plane and seg, w.r.t. 
  *   parametrization of seg
+ * \param [in] EPS tolerance parameter for determining if \a t
+ *             is just within 0.0 <= t <= 1.0 .
```
adayton1, line 1106: I would simplify to "EPS tolerance parameter for determining if 0.0 <= t <= 1.0"

src/axom/primal/operators/intersect.hpp
```diff
@@ -575,6 +575,8 @@ AXOM_HOST_DEVICE bool intersect(const Plane<T, 3>& p,
  * \param [in] seg A line segment
  * \param [out] t Intersection point of plane and seg, w.r.t. seg's
  *  parametrization
+ * \param [in] EPS tolerance parameter for determining if \a t
+ *             is just within 0.0 <= t <= 1.0
```
adayton1, line 579: Same simplification here.

src/axom/primal/operators/intersect.hpp
```diff
@@ -586,9 +588,10 @@ AXOM_HOST_DEVICE bool intersect(const Plane<T, 3>& p,
 template <typename T, int DIM>
 AXOM_HOST_DEVICE bool intersect(const Plane<T, DIM>& plane,
                                 const Segment<T, DIM>& seg,
-                                T& t)
+                                T& t,
+                                const double& EPS = 1e-8)
```
adayton1, line 592: 1e-12? I would match the default in intersect_impl.hpp.

src/axom/primal/tests/primal_clip.cpp
```diff
@@ -1954,6 +1954,121 @@ TEST(primal_clip, polygon_intersects_polygon)
...
+  Polygon2D clipUpper;
+  clipUpper.addVertex(Point2D {1., 0.});
+  clipUpper.addVertex(Point2D {1., 1.});
+  clipUpper.addVertex(Point2D {0., 1.});
+
+  // Expected VFs (volume fractions, areas) for overlaps.
+  const double lower_vf[] = {1., 1., 0.5, 1., 0.5, 0., 0.5, 0., 0.};
+  const double upper_vf[] = {0., 0., 0.5, 0., 0.5, 1., 0.5, 1., 1.};
+
+  constexpr double fine_area = (1. / 3.) * (1. / 3.);
```
adayton1, line 2043: Where does this come from?

src/axom/primal/tests/primal_clip.cpp
```diff
@@ -1954,6 +1954,121 @@ TEST(primal_clip, polygon_intersects_polygon)
...
+
+  for(int p : std::vector<int> {0, 1, 2, 3, 4, 6})
```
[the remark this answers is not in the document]
bmhan12: > What happens if you include the squares where the overlap is a single point? Or there is no overlap at all?

There is no overlap, and the result is an empty polygon with no vertices.

src/axom/primal/tests/primal_clip.cpp
bmhan12: > Where does this come from?
The arrays can be thought of as what fraction of the quad we expect the resulting area to be.
`fine_area` is to account for the fact that each quad is 1/9 the size of a unit cube. 
The area we then check for is the multiplication of the two values.

src/axom/primal/tests/primal_clip.cpp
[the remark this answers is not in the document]
bmhan12: Added unit tests to verify the non-overlap cases!

Arlie-Capps (approved):

src/axom/primal/tests/primal_clip.cpp
adayton1: Thanks for the explanation. Can you add this as a comment in the test?

BradWhitlock (approved):
Thanks for fixing this so soon @bmhan12 !

kennyweiss (approved):
Thanks @bmhan12!

Please note the bugfix in the RELEASE-NOTES before merging

src/axom/primal/tests/primal_clip.cpp
bmhan12: > Thanks for the explanation! Can you add this as a comment in the test?

Yes, added!

On a side note, I made a slight mistake with my explanation here in the PR comments.
To derive the volume fraction, we take the area of the overlapped polygon and *divide* (not multiply) it by 1/9, the original area of the quad.

adayton1 (approved):

src/axom/primal/operators/detail/clip_impl.hpp
```diff
@@ -761,7 +755,21 @@ AXOM_HOST_DEVICE Polygon<T, 2, ARRAY_TYPE, MAX_VERTS> clipPolygonPolygon(
...
-  return outputList;
+  // Remove duplicate points.
+  // Handles edge cases such as 3 consecutive vertices with orientations
+  // ON_POSITIVE_SIDE, ON_BOUNDARY, ON_POSITIVE where point on the boundary
+  // is added twice.
+  PolygonType uniqueList;
+  for(int i = 0; i < outputList.numVertices(); i++)
+  {
+    int prevIndex = ((i - 1) == -1) ? (outputList.numVertices() - 1) : (i - 1);
+    if(outputList[i] != outputList[prevIndex])
```
BradWhitlock, line 766: I have a late follow-up question about this point test. Might tolerances need to apply here too? I ask because I tried the fixed code with 2 quads where the resulting output ought to be a triangle but it outputs a quad where the first 2 points are essentially the same.

```
void test4()
{
  using Polygon = axom::primal::Polygon<double, 2>;
  using Point = axom::primal::Point<double, 2>;
  constexpr double eps = 1.e-6;

  Polygon p1;
  p1.addVertex(Point{0., 0.});
  p1.addVertex(Point{1., 0.});
  p1.addVertex(Point{1., 1.});
  p1.addVertex(Point{0., 1.});

  Polygon p2;
  p2.addVertex(Point{0., 0.});
  p2.addVertex(Point{0.3, 0.3});
  p2.addVertex(Point{0., 1.});
  p2.addVertex(Point{-0.3, 0.7});

  const auto overlapPoly = axom::primal::clip(p1, p2, eps);
  const auto area = overlapPoly.area();
  std::cout << "subject=" << p1
            << ", clip=" << p2
            << ", overlap=" << overlapPoly
            << ", area=" << area
            << std::endl;

// Leads to a polygon with an extra vertex... area is good.
// subject={4-gon:(0,0),(1,0),(1,1),(0,1)}, clip={4-gon:(0,0),(0.3,0.3),(0,1),(-0.3,0.7)}, overlap={4-gon:(-6.66134e-17,1),(0,1),(0,0),(0.3,0.3)}, area=0.15

}

```

image

bmhan12: Good catch! Added tolerance, and this example to the unit tests!

adayton1 (approved):

PR #1500, opened 2025-02-04
Move `primal::NumericArray` to core

gunney1:
# Move `primal::NumericalArray` class to `core`
The motivation for this:
1. `NumericalArray` doesn't depend on `primal` where it had lived.
2. It has general utilities like supporting multidimensional index arithmetic.

This PR implements issue https://github.com/LLNL/axom/issues/1488.


src/axom/core/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ set(core_headers
     utilities/BitUtilities.hpp
     utilities/CommandLineUtilities.hpp
     utilities/FileUtilities.hpp
+    utilities/NumericArray.hpp
```
BradWhitlock, line 29: NumericArray looks like StackArray plus some arithmetic operators. All of the other arrays are in the root level in the core directory. It doesn't match to have NumericArray one level down in utilities.

src/axom/core/utilities/NumericArray.hpp
```diff
@@ -383,14 +381,13 @@ class NumericArray  // NOLINT
   AXOM_HOST_DEVICE
   void verifyIndex(int AXOM_DEBUG_PARAM(idx)) const
```
BradWhitlock, line 382: Looks like the OSX build choked here, might need to remove the AXOM_DEBUG_PARAM macro or change how it plays with assert.

kennyweiss: It's actually a good catch!

`AXOM_DEBUG_PARAM` is guarded by `AXOM_DEBUG`, while `assert` is guarded by `NDEBUG` and the two can be independently controlled.

kennyweiss: We can use `AXOM_UNUSED_VAR` inside the function instead of `AXOM_DEBUG_PARAM`.

See: https://github.com/LLNL/axom/blob/c70f36f6fe0d38ba366e93e3737a606488fa0f8b/src/axom/core/Macros.hpp#L171-L212

src/axom/core/utilities/NumericArray.hpp
```diff
@@ -8,7 +8,7 @@
 
 #include "axom/core/Macros.hpp"
 #include "axom/core/utilities/Utilities.hpp"
-#include "axom/slic/interface/slic.hpp"
+#include <assert.h>
```
kennyweiss, line 11: Should this be:
```suggestion
#include <cassert>
```

src/axom/core/tests/core_numeric_array.cpp
```diff
@@ -276,7 +272,8 @@ TEST(primal_numeric_array, clamping)
   // NOTE: AXOM_DEBUG is disabled in release mode, so this test will only fail
   // in
   // debug mode
```
kennyweiss, line 274: Minor: I know you didn't change this, but would you mind cleaning up the awkward spacing in this comment?
```suggestion
  // NOTE: AXOM_DEBUG is disabled in release mode, so this test will only fail in debug mode
```

src/axom/core/CMakeLists.txt
kennyweiss: I agree that this should be in the same folder as the other array classes

src/axom/core/tests/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ set(core_serial_tests
     core_map.hpp
     core_flatmap.hpp
     core_memory_management.hpp
+    # core_numeric_array.cpp
```
kennyweiss, line 29: Why are the tests commented out?

kennyweiss (changes requested):
Thanks @gunney1 -- seems like a reasonable change. 

`NumericArray` is more of a (linear) algebra support class and does not include any geometry, although it contains the core functionality for `primal::Point` and `primal::Vector`.

#### Required changes:
* It looks like the core_numeric_array unit tests are commented out. I'd be happy to approve once those are added back in.
* I agree w/ @BradWhitlock that `NumericArray` should be in the same directory as the other array classes

#### Clarification:
Also, I'm a bit confused about this line in the PR description:
   ```
   It has general utilities like supporting multidimensional index arithmetic.
   ```
since I don't think it directly contains any functionality for multidimensional indexing. Am I missing something, or did you mean that it can be used to help w/ multidimensional indexing? If so, do we have any existing examples?

src/axom/core/tests/core_numeric_array.cpp
gunney1: Sure!

src/axom/core/tests/CMakeLists.txt
gunney1: a temporary change I didn't mean to commit.  Fixing.

src/axom/core/utilities/NumericArray.hpp
gunney1: Oh, thank you.

src/axom/core/utilities/NumericArray.hpp
gunney1: Yes.  Thanks.

src/axom/core/CMakeLists.txt
gunney1: Ok.

kennyweiss (approved):
Thanks @gunney1 !

BradWhitlock (approved):

PR #1503, opened 2025-02-17
Adds Trimming Curve Support for `primal::NURBSPatch`

jcs15c:
# Summary

- This PR is a feature, enhancing the `primal::NURBSPatch` class to maintain and manipulate an unstructured array of `primal::NURBSCurve` objects to be used for trimming the parameter space of the original patch. 
- If a point in the parameter space is outside the collection of trimming curves, it is not visible. Containment within the array of trimming curves is computed using 2D generalized winding numbers and an `EvenOdd` rule on the resulting decimal value. These features are tested in the `primal_trimming_curves.cpp` file.
- Trimmed surfaces are identified by a flag in the NURBS patch class, in order that trimmed patches can be defined with an empty array of trimming curves, i.e., the entire patch is trimmed out.
 - Additionally, this PR separates the implementation details of 2D and 3D generalized winding numbers to avoid circular dependencies resulting from visibility queries of trimmed surfaces. 
 - Currently, splitting a trimmed NURBS patch is not supported. This will be addressed in a future PR.


src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -1075,19 +1114,134 @@ class NURBSPatch
...
+
+    ParameterPointType uv = {u, v};
+
+    double gwn = 0.0;
+    for(const auto& curve : m_trimmingCurves)
+    {
+      gwn += detail::nurbs_winding_number(uv, curve);
+    }
+
+    return std::lround(gwn) % 2 != 0;
```
kennyweiss, line 1233: Should we have a (small) utility function like `primal::inside()` that takes a double and maybe an enum for the even-odd vs. non-zero rule and returns a bool?

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -1142,7 +1296,8 @@ class NURBSPatch
   {
     return (lhs.m_controlPoints == rhs.m_controlPoints) &&
       (lhs.m_weights == rhs.m_weights) && (lhs.m_knotvec_u == rhs.m_knotvec_u) &&
-      (lhs.m_knotvec_v == rhs.m_knotvec_v);
+      (lhs.m_knotvec_v == rhs.m_knotvec_v) &&
+      (lhs.m_trimmingCurves == rhs.m_trimmingCurves);
```
kennyweiss, line 1300: Does this also need to test the `m_isTrimmed`?
It seems that patches w/o curves but w/ different `m_isTrimmed` values should not be considered identical

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -1168,6 +1323,9 @@ class NURBSPatch
   /*!
    * \brief Reverses the order of one direction of the NURBS patch's control points and weights
    *
+   * This method does not change the position of the patch in space, 
+   *  including the trimming curves. It does reverse all normal vectors.
```
kennyweiss, line 1327: (Clarity)

Something like the following might be somewhat clearer:
```suggestion
   * This method does not affect the position of the patch in space, or its trimming curves,
   * but it does reverse the patch's normal vectors.
```

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -2324,7 +2518,7 @@ class NURBSPatch
   }
 
   /*!
-   * \brief Splits a NURBS surface (at each internal knot) into several Bezier patches
+   * \brief Splits the untrimmed NURBS surface (at each internal knot) into several Bezier patches
```
kennyweiss, line 2521: Please consider mentioning something about how this function deals with trimming curves

src/axom/primal/operators/intersect.hpp
```diff
@@ -1000,6 +1000,7 @@ bool intersect(const Ray<T, 3>& ray,
                axom::Array<T>& v,
                double tol = 1e-8,
                double EPS = 1e-8,
+               bool isTrimmed = true,
```
kennyweiss, line 1003: How does this parameter relate to `NURBSPatch::isTrimmed()`?

src/axom/primal/operators/winding_number.hpp
```diff
@@ -7,7 +7,7 @@
  * \file winding_number.hpp
  *
  * \brief Consists of methods to compute the generalized winding number (GWN) 
- *        for points with respect to various geometric objects.
+ *        for points with respect to various 2D geometric objects.
```
kennyweiss, line 10: Clarification: This file has both 2D and 3D functions, right?

kennyweiss (approved):
Thanks @jcs15c for another great PR!

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -1075,19 +1114,134 @@ class NURBSPatch
...
+
+    const double min_u = m_knotvec_u[0];
+    const double max_u = m_knotvec_u[m_knotvec_u.getNumKnots() - 1];
+
+    const double min_v = m_knotvec_v[0];
+    const double max_v = m_knotvec_v[m_knotvec_v.getNumKnots() - 1];
+
+    // For each min/max u/v, add a straight trimming curve along the boundary
+    TrimmingCurveType curve;
+    curve.setParameters(2, 1);
```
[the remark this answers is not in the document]
jcs15c: Good idea!

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -1075,19 +1114,134 @@ class NURBSPatch
...
+   * 
+   * Checks for containment of the parameter point in 
+   * the collection of trimming curves via an even-odd rule
+   */
+  bool isVisible(T u, T v) const
+  {
+    if(!isTrimmed())
+    {
+      return true;
+    }
```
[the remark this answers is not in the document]
jcs15c: Probably makes more sense than a `SLIC_ASSERT` that crashes if an evaluation point is outside the parameter space, but it might be easier to use those same methods than a full bounding box check.

src/axom/primal/geometry/NURBSPatch.hpp
jcs15c: I could see that being useful, but it should probably take an `int` as input, since that's what methods like `winding_number<polygon>` return. Do you think it should go in `Utilities.hpp`?

src/axom/primal/geometry/NURBSPatch.hpp
jcs15c: Good catch!

src/axom/primal/operators/intersect.hpp
jcs15c: There are some cases where a patch is trimmed, but you want to find intersections with the untrimmed patch, like recording "near-misses" with the visible portion. I should rename this variable to make that more clear, maybe to "treatAsTrimmed" or reverse the implementation and have it be "treatAsUntrimmed"/`considerUntrimmed`?

src/axom/primal/operators/winding_number.hpp
jcs15c: Whoops! Fixed that.

src/axom/primal/tests/primal_trimming_curves.cpp
```diff
@@ -0,0 +1,404 @@
...
+  EXPECT_EQ(nPatch.getNumTrimmingCurves(), 0);
+
+  // Remove all trimming curves
+  nPatch.makeUntrimmed();
+  EXPECT_FALSE(nPatch.isTrimmed());
+
+  // Mark as trimmed, but empty vector of trimming curves
+  nPatch.makeTrimmed();
+  EXPECT_TRUE(nPatch.isTrimmed());
+  EXPECT_EQ(nPatch.getNumTrimmingCurves(), 0);
```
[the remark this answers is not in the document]
jcs15c: `markAsTrimmed()` makes sense to me, sounds more intuitive than `setTrimmed()`

src/axom/primal/tests/primal_trimming_curves.cpp
```diff
@@ -0,0 +1,404 @@
...
+
+  // Add a simple trimming curve from a NURBS curve
+  axom::Array<ParameterPointType> trimmingCurveControlPoints {
+    ParameterPointType {0.25, 0.25},
+    ParameterPointType {0.75, 0.25},
+    ParameterPointType {0.75, 0.75},
+    ParameterPointType {0.25, 0.75},
+    ParameterPointType {0.25, 0.25}};
+
+  TrimmingCurveType trimmingCurve(trimmingCurveControlPoints, 2);
```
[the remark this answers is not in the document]
jcs15c: The method works the same way on open curves as closed curves, since it rounds the fractional GWN either way, but I should probably add a comment somewhere that that's what it does. I've also added a test that looks at the same patch but with an open curve.

src/axom/primal/tests/primal_trimming_curves.cpp
```diff
@@ -0,0 +1,404 @@
...
+      PointType pt1 = nPatch.evaluate(u, v);
+      PointType pt2 = nPatchCopy.evaluate(u, v);
+
+      for(int N = 0; N < DIM; ++N)
+      {
+        EXPECT_NEAR(pt1[N], pt2[N], 1e-10);
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
jcs15c: I don't know if this is how it should work, but I think that `NURBSPatch::evaluate()` itself shouldn't consider trimming curves at all, and should just evaluate the untrimmed patch. I've thought about adding a `PointType NURBSPatch::evaluate(T u, T v, bool isVisible)` method, but I don't know what the utility of a function like that is. It would have same return type and calls the exact same `isVisible` test on the point, so there's no removed redundancy there. I also really don't like the idea of `evaluate` being a no op for parameter points that aren't visible.

I've tried to make this behavior clear by adding "untrimmed" to the comments of all the evaluation routines, and I'll make it more clear here. I agree, it's probably a little more confusing than it needs to be.

src/axom/primal/tests/primal_trimming_curves.cpp
```diff
@@ -0,0 +1,404 @@
...
+  // Construct from 1D axom::Array
+  axom::Array<PointType> controlPointsArray({PointType {0.0, 0.0, 1.0},
+                                             PointType {0.0, 1.0, 0.0},
+                                             PointType {0.0, 2.0, 0.0},
+                                             PointType {1.0, 0.0, 0.0},
+                                             PointType {1.0, 1.0, -1.0},
+                                             PointType {1.0, 2.0, 0.0},
+                                             PointType {2.0, 0.0, 0.0},
+                                             PointType {2.0, 1.0, 0.0},
+                                             PointType {2.0, 2.0, 1.0}});
```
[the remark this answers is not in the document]
jcs15c: This is cool! I've updated the text to use the fixture.

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -1075,19 +1114,140 @@ class NURBSPatch
...
+
+    // Left
+    curve[0] = ParameterPointType({min_u, max_v});
+    curve[1] = ParameterPointType({min_u, min_v});
+    addTrimmingCurve(curve);
+
+    // Right
+    curve[0] = ParameterPointType({max_u, min_v});
+    curve[1] = ParameterPointType({max_u, max_v});
+    addTrimmingCurve(curve);
```
kennyweiss, line 1207: Should these be added in CCW order?
(I realize our WN algorithm doesn't require this, but it can't hurt)

src/axom/primal/operators/intersect.hpp
```diff
@@ -992,6 +992,24 @@ bool intersect(const Ray<T, 3>& ray,
...
+ * \param [in] EPS The tolerance for intersection (for parameter distances).
+ * \param [in] countUntrimmed True if intersections with the untrimmed patch should also be recorded.
+ * \param [in] isHalfOpen True if the patch is parameterized in [0,1)^2.
+ * 
+ * Perform Bezier extraction and record intersections with each patch.
+ * After intersections are recorded, parameter points located outside the trimming
+ *  curves are pruned from the list (unless specified by `countUntrimmed`).
+ *  
+ * \return true iff the ray intersects the patch, otherwise false.
+ */
```
kennyweiss, line 1012: Please fix the parameter order in the doxygen to match the function

src/axom/primal/operators/winding_number.hpp
```diff
@@ -197,142 +128,45 @@ int winding_number(const Point<T, 2>& R,
...
   bool isOnEdge = false;
-  return winding_number(R, P, isOnEdge, includeBoundary, edge_tol);
+  return detail::polygon_winding_number(R, P, isOnEdge, includeBoundary, edge_tol);
 }
 
 /*!
- * \brief Computes the GWN for a 2D point wrt a 2D Bezier curve
+ * \brief Computes the GWN for a 2D point wrt a 2D NURBS curve
  *
  * \param [in] query The query point to test
```
kennyweiss, line 137: Consistency w/ function API

```suggestion
 * \param [in] q The query point to test
```

src/axom/primal/operators/winding_number.hpp
```diff
@@ -197,142 +128,45 @@ int winding_number(const Point<T, 2>& R,
...
-  }
-
-  return gwn + closed_curve_wn - closure_wn;
+  return detail::nurbs_winding_number(q, n, edge_tol, EPS);
 }
 
 /*!
  * \brief Computes the GWN for a 2D point wrt a 2D NURBS curve
  *
  * \param [in] query The query point to test
```
kennyweiss, line 156: Consistency

```suggestion
 * \param [in] q The query point to test
```

kennyweiss (approved):
Some additional minor comments

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -1269,6 +1463,14 @@ class NURBSPatch
     }
 
     std::swap(m_knotvec_u, m_knotvec_v);
```
BradWhitlock, line 1465: Please use [axom::utilities::swap](https://github.com/LLNL/axom/blob/44562f92a400204e33915f48b848eb68e80a1bf1/src/axom/core/utilities/Utilities.hpp#L105) as a tiny step towards future device support.

BradWhitlock (approved):

src/axom/primal/geometry/NURBSPatch.hpp
jcs15c: Sure, might as well.

src/axom/primal/geometry/NURBSPatch.hpp
jcs15c: Changed it here, and elsewhere.

src/axom/primal/operators/intersect.hpp
jcs15c: Fixed it here and elsewhere in the file

src/axom/primal/operators/winding_number.hpp
jcs15c: Fixed here and elsewhere in the file!

PR #1504, opened 2025-02-19
Fixes build and tests when not using `hdf5`

kennyweiss:
# Summary

- This PR is a bugfix
- It fixes the build and tests in configurations without `hdf5`
- I also discovered and fixed an unknown dependency for `scr` on `zlib`

I noted in #1480 that Axom supports configurations without `hdf5`. We don't currently have CI for this, and when I configured the code without `hdf5`, I found several tests that did not work. This branch fixes them. 

Since non-`hdf5` configs require a `conduit` without `hdf5` (i.e. a new library set), I'm not sure that we need a CI that tests this -- it might be enough to just manually run the code and fix it every so often. Perhaps we should have a manually run github-action that does this, similar to our manual Windows TPL build?

### Details:
I configured the code with a host-config generated by running the following uberenv command:
```
> ./scripts/uberenv/uberenv.py --spec "%clang@14.0.6+devtools~hdf5+c2c+mfem+scr+profiling  ^conduit~hdf5"
```

### DIFF
--- a/src/axom/sidre/core/MFEMSidreDataCollection.cpp
+++ b/src/axom/sidre/core/MFEMSidreDataCollection.cpp
@@ -1220,7 +1220,7 @@ void MFEMSidreDataCollection::Save(const std::string& filename,
       // Root file support only available in hdf5.
       else
       {
-        writer.write(blueprint_indicies_grp, 1, file_path + ".root", protocol);
+        blueprint_indicies_grp->save(file_path + ".root", protocol);
       }
     }
   }
--- a/src/axom/sidre/core/MFEMSidreDataCollection.hpp
+++ b/src/axom/sidre/core/MFEMSidreDataCollection.hpp
@@ -472,6 +472,12 @@ public:
    */
   virtual void Load(int cycle_ = 0)
   {
+  #ifndef AXOM_USE_HDF5
+    SLIC_ERROR(
+      "MFEMSidreDataCollection::Load(<cycle>) is only implemented for the "
+      "'sidre_hdf5' protocol");
+  #endif
+
     SetCycle(cycle_);
     Load(get_file_path(name), "sidre_hdf5");
   }
--- a/src/axom/sidre/spio/IOManager.cpp
+++ b/src/axom/sidre/spio/IOManager.cpp
@@ -177,10 +177,13 @@ void IOManager::write(sidre::Group* datagroup,
 
   std::string test_file_base(broadcastString(file_base, m_mpi_comm, m_my_rank));
 
-  SLIC_WARNING_IF(test_file_base != file_base,
-                  "IOManager::write() file_base argument is not identical "
-                    << "on all ranks. This may cause the output files to be "
-                    << "incompatible with a call to IOManager::read().");
+  SLIC_WARNING_IF(
+    test_file_base != file_base,
+    "IOManager::write() file_base argument is not identical "
+      << "on all ranks. This may cause the output files to be "
+      << "incompatible with a call to IOManager::read()."
+      << axom::fmt::format("\n\tfile_base: '{}'", file_base)
+      << axom::fmt::format("\n\ttest_file_base: '{}'", test_file_base));
 
   std::string output_base =
     createRootFile(file_base, num_files, protocol, tree_pattern);
--- a/src/cmake/thirdparty/SetupAxomThirdParty.cmake
+++ b/src/cmake/thirdparty/SetupAxomThirdParty.cmake
@@ -257,12 +257,20 @@ endif()
 # SCR
 #------------------------------------------------------------------------------
 if (SCR_DIR)
+    # SCR depends on zlib, but this can be masked by hdf5's zlib dependency
+    if(NOT TARGET ZLIB::ZLIB)
+        if(ZLIB_DIR)
+            set(ZLIB_ROOT ${ZLIB_DIR})
+        endif()
+        find_package(ZLIB REQUIRED) # creates ZLIB::ZLIB target
+    endif()
+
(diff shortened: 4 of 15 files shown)


src/cmake/thirdparty/SetupAxomThirdParty.cmake
```diff
@@ -257,12 +257,20 @@ endif()
...
     axom_assert_is_directory(DIR_VARIABLE SCR_DIR)
 
     include(cmake/thirdparty/FindSCR.cmake)
     blt_import_library( NAME       scr
                         INCLUDES   ${SCR_INCLUDE_DIRS}
-                        LIBRARIES  ${SCR_LIBRARIES}
+                        LIBRARIES  ${SCR_LIBRARIES} ZLIB::ZLIB
                         TREAT_INCLUDES_AS_SYSTEM ON
                         EXPORTABLE ON)
     blt_list_append(TO TPL_DEPS ELEMENTS scr)
```
kennyweiss, line 276: @white238 -- does this seem reasonable? 

Do we need to export `zlib`, e.g. as `axom::zlib`
I checked the exported `axom-targets.cmake` and `ZLIB::ZLIB` was in the `INTERFACE_LINK_LIBRARIES`, but there was nothing explicit for `ZLIB::ZLIB`

```cmake
# Create imported target axom::scr
add_library(axom::scr INTERFACE IMPORTED)

set_target_properties(axom::scr PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "..."
  INTERFACE_LINK_LIBRARIES "...;ZLIB::ZLIB"
  INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "..."
```

kennyweiss: Looks like I need to fix this. The using-with-blt example failed:
```
CMake Error at .../axom/install-axom/lib/cmake/axom-targets.cmake:55 (set_target_properties):
  The link interface of target "axom::scr" contains:

    ZLIB::ZLIB

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

white238: I would recreate the ZLIB::ZLIB target in our exported config file:
https://github.com/LLNL/axom/blob/develop/src/cmake/axom-config.cmake.in#L189

Conduit has an example:

https://github.com/LLNL/conduit/blob/2eec1ad9548cf6d629e2145181ff7b8cef0a38fa/src/config/conduit_setup_deps.cmake#L114-L130

kennyweiss: Perfect!  Thanks.

Arlie-Capps (approved):
Thank you for this fix, @kennyweiss .

BradWhitlock (approved):
It looks good to me.

kennyweiss:
Note: My original spack config didn't include `mfem` and there were some additional cases that needed to be handled when we have `mfem` but not `hdf5`. I fixed these including an MPI hang in how we were using `spio` from `sidre::MFEMSidreDataCollection` for protocols other than `sidre_hdf5`.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -1220,7 +1220,7 @@ void MFEMSidreDataCollection::Save(const std::string& filename,
       // Root file support only available in hdf5.
       else
       {
-        writer.write(blueprint_indicies_grp, 1, file_path + ".root", protocol);
+        blueprint_indicies_grp->save(file_path + ".root", protocol);
```
kennyweiss, line 1223: @nselliott -- This call to `writer.write()` from only the root rank was hanging since `writer.write()` is collective. 
Switching it to directly use `Group::save()` resolved the issue.

src/axom/sidre/spio/IOManager.cpp
```diff
@@ -177,10 +177,13 @@ void IOManager::write(sidre::Group* datagroup,
...
-                  "IOManager::write() file_base argument is not identical "
-                    << "on all ranks. This may cause the output files to be "
-                    << "incompatible with a call to IOManager::read().");
+  SLIC_WARNING_IF(
+    test_file_base != file_base,
+    "IOManager::write() file_base argument is not identical "
+      << "on all ranks. This may cause the output files to be "
+      << "incompatible with a call to IOManager::read()."
+      << axom::fmt::format("\n\tfile_base: '{}'", file_base)
+      << axom::fmt::format("\n\ttest_file_base: '{}'", test_file_base));
```
kennyweiss, line 186: @nselliott -- I updated this log statement to show the filenames

src/axom/sidre/core/MFEMSidreDataCollection.hpp
```diff
@@ -472,6 +472,12 @@ class MFEMSidreDataCollection : public mfem::DataCollection
    */
   virtual void Load(int cycle_ = 0)
   {
+  #ifndef AXOM_USE_HDF5
+    SLIC_ERROR(
+      "MFEMSidreDataCollection::Load(<cycle>) is only implemented for the "
+      "'sidre_hdf5' protocol");
+  #endif
```
kennyweiss, line 479: It was somewhat surprising that the `MFEMSidreDataCollection::load(<cycle>)` didn't support other protocols. 
(Although the doxygen already noted this).

I added this error message so it's easier for people to see why this doesn't work. I don't think it would be too difficult to make this work with other protocols, but I also am not aware of any need/desire for this at the moment, so I'm planning to leave this as is.

PR #1505, opened 2025-02-20
Host execution space can use more than just its default and unified m…

gunney1:
Fix what memory spaces SEQ_EXEC and OMP_EXEC claim to be able to use.  They could use Dynamic _and_ Host, not just their default memory space.


kennyweiss (approved):
Thanks @gunney1 

Should we add any unit tests to exercise this?

rhornung67 (approved):
I echo @kennyweiss question about testing.

Arlie-Capps (approved):
Thank you, Brian.

gunney1:
@kennyweiss @rhornung67 I'm not sure what a non-trivial test would check.  The code just returns what it's hard-coded to return.  The error was my oversight when I wrote the original code, not a potential user input or logic error.

PR #1519, opened 2025-03-12
Fixes build and tests on Windows

kennyweiss:
# Summary

- This PR is a bugfix for the build on Windows
- It fixes:
   -  the MIR build, which required some exports for static members
    - MIR tests that depend on the compile time path to the data directory
    - A bug with the quicksort algorithm for some data orderings and adds a unit test
    - Some quest tests when mfem does not have MPI
    - Some quest lambdas that take a constexpr -- MSVC appears to require `static constexpr` w/ c++14. This might be relaxed in c++17
 - I also cleaned up/updated some sidre::Group implementations to use the iterator interface while working on sidre test bugfixes
 - **Update:** Passes all builds and tests in our Windows TPL CI -- https://github.com/LLNL/axom/actions/runs/13846008225
  
#### Status:
- The code + tests + examples currently build w/ our msvc configuration
- The following tests are still currently failing (checked when fixed):
  - [x] 85 - sina_Document (Failed)
  - [x] 86 - sina_ID (Failed)
  - [x] 87 - sina_Record (Failed)
  - [x] 88 - sina_Relationship (Failed)
  - [x] 89 - sina_Run (Failed)
  - [x] 170 - sidre_read_write_userdefined_data (Failed)
  - [x] 171 - sidre_mfem_datacollection (Failed)
  - Update: All fixed!
- The sina failures appear to be related to data that is moved. I will try to fix w/ `ubsan`
  - Update: Fixed. There was also an error related to trying to rename a tmp file. Windows does not allow this if the destination already exists.
- The sidre failures appear to be related to `MFEMSidreDataCollection::LoadExternalData()`. Perhaps related to changes from: 
    - https://github.com/LLNL/axom/pull/1473
    - https://github.com/LLNL/axom/issues/1479
    - https://github.com/LLNL/axom/pull/1494
    - Update: The sidre failures were due to loading external data into the wrong group when not going through spio

### DIFF
--- a/scripts/vcpkg_ports/umpire/v2024.07.0-chrono-include.patch
+++ b/scripts/vcpkg_ports/umpire/v2024.07.0-chrono-include.patch
@@ -1,0 +1,12 @@
+diff --git a/src/umpire/event/event.hpp b/src/umpire/event/event.hpp
+index 03f0d37..db84c1c 100644
+--- a/src/umpire/event/event.hpp
++++ b/src/umpire/event/event.hpp
+@@ -7,6 +7,7 @@
+ #ifndef UMPIRE_event_HPP
+ #define UMPIRE_event_HPP
+ 
++#include <chrono>
+ #include <cstdint>
+ #include <map>
+ #include <sstream>
--- a/src/axom/mir/clipping/ClipCases.h
+++ b/src/axom/mir/clipping/ClipCases.h
@@ -7,9 +7,11 @@
 //---------------------------------------------------------------------------
 // Axom modifications
 // NOTE: The values for EA-EL and N0-N3 were reduced.
+// NOTE: We're using AXOM_MIR_EXPORT instead of VISIT_VTK_LIGHT_API througout
 // clang-format off
-//#include <visit_vtk_light_exports.h>
-#define VISIT_VTK_LIGHT_API
+
+#include "axom/export/mir.h"
+
 #include <cstdlib>
 namespace axom {
 namespace mir {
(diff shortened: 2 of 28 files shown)


kennyweiss:
> /style

@white238 -- I am not sure that this worked. It seemed to trigger an action, but for some reason, the action was cancelled. 
See: https://github.com/LLNL/axom/actions/runs/13825505635

src/axom/mir/clipping/ClipCases.h
```diff
@@ -88,87 +90,87 @@ namespace visit {
 #define NOCOLOR 122
 
 // Tables
-extern VISIT_VTK_LIGHT_API int numClipCasesHex;
```
BradWhitlock, line 91: Thanks for fixing the export macro. I don't know how I overlooked that.

BradWhitlock (approved):
Looks good to me.

src/axom/mir/clipping/ClipCases.h
kennyweiss: (there were a lot of files :) )

scripts/vcpkg_ports/umpire/v2024.07.0-chrono-include.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/src/umpire/event/event.hpp b/src/umpire/event/event.hpp
+index 03f0d37..db84c1c 100644
+--- a/src/umpire/event/event.hpp
++++ b/src/umpire/event/event.hpp
+@@ -7,6 +7,7 @@
+ #ifndef UMPIRE_event_HPP
+ #define UMPIRE_event_HPP
+ 
++#include <chrono>
```
kennyweiss, line 9: @davidbeckingsale -- My local machine did not need this on Windows, but our CI needed this for umpire@2024.07.0.
I saw that it was recently added to umpire@develop -- https://github.com/LLNL/Umpire/pull/934

src/axom/sina/core/ID.cpp
```diff
@@ -65,10 +65,11 @@ IDField::IDField(ID value_, std::string localName_, std::string globalName_)
...
                  std::string globalName_)
-  : IDField {extractIDFromObject(parentObject, localName_, globalName_),
-             std::move(localName_),
-             std::move(globalName_)}
-{ }
+  : value(extractIDFromObject(parentObject, localName_, globalName_))
+{
+  std::swap(localName, localName_);
+  std::swap(globalName, globalName_);
+}
```
kennyweiss, line 72: @doutriaux1 @bgunnar5 @gwaegner -- this was tricky to track down. 

Presumably msvc has a different order of processing the arguments than clang and other compilers, or it wiped out the strings after moving them, rather than leaving them unchanged.

The call to `extractIDFromObject` was happening after `localName_` and `globalName_` were already moved.

src/axom/sidre/core/MFEMSidreDataCollection.cpp
```diff
@@ -997,7 +997,14 @@ void MFEMSidreDataCollection::LoadExternalData(const std::string& filename,
...
   {
-    grp->loadExternalData(path);
+    if(!group_name.empty())
+    {
+      grp->loadExternalData(path);
+    }
+    else
+    {
+      m_bp_grp->loadExternalData(path);
+    }
```
kennyweiss, line 1007: @white238 and @nselliott -- This was necessary to get `loadExternalData()` working in our Windows configuration, which has MPI for axom and a serial mfem.

I am not sure if we have test coverage for the `if` part (where the users provide a `group_name`)

bmhan12 (approved):

PR #1522, opened 2025-03-17
spack v1.0 support: `%foo +bar` -> `+bar %foo`

haampie:
# Summary

Spack v1.0 will parse `pkg %gcc +foo` as "pkg with direct dependency gcc
with variant foo enabled" instead of "pkg with variant foo enabled and compiler
gcc". Reorder as `pkg +foo %gcc` to be compatible with Spack v0.x and
v1.0.

### DIFF
--- a/scripts/spack/configs/toss_4_x86_64_ib/spack.yaml
+++ b/scripts/spack/configs/toss_4_x86_64_ib/spack.yaml
@@ -121,11 +121,11 @@ spack:
     mvapich2:
       buildable: false
       externals:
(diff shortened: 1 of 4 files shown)


kennyweiss (approved):
Thanks @haampie !

BradWhitlock (approved):

white238 (approved):

kennyweiss:
@haampie -- we're testing this PR in https://github.com/LLNL/axom/pull/1541

PR #1525, opened 2025-03-28
Work-around for ROCM overflowing ints when scan-summing arrays

gunney1:
# Summary

- This PR is a bugfix (work-around)
- It does the following:
  - Use a totally inefficient 32-bit int to represent a boolean, to work around a ROCM behavior.
  - Add comment about the same work-around for another array.
  - Add notes on when we can remove the work-around.

We have 2 arrays, one with booleans and with ints in [0,5], but we have to use more bits to represent them for ROCM.  We do sum scans on them and put the output in 32-bit ints.  However, ROCM is doing the sum in the **input** type without promoting to the output type.  This work-around bumps up the input type.

Related to https://github.com/LLNL/RAJA/issues/1631

### DIFF
--- a/src/axom/quest/MarchingCubes.hpp
+++ b/src/axom/quest/MarchingCubes.hpp
@@ -292,6 +292,15 @@ public:
   // Allow single-domain code to share common scratch space.
   friend detail::marching_cubes::MarchingCubesSingleDomain;
 
+  /*
+    TODO: CrossingFlagType can be a boolean value but is wastefully
+    stored in 32 bits because of a ROCM scan implementation that adds
+    them in the input type without promoting them to our 32-bit output
+    type.  When ROCM supports the promotion and RAJA uses it, we can
+    change this type to something more efficient.
+  */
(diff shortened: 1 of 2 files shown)


publixsubfan (approved):

kennyweiss (approved):
Thanks @gunney1 

Clearly, this is necessary. 
Other than more memory, are there performance implications?

kennyweiss:
Looks like yo missed an initialization:
```
/home/axom/axom/src/axom/core/../../axom/quest/detail/MarchingCubesImpl.hpp:77:7: error: invalid initialization of reference of type 'axom::Array<int>&' from expression of type 'axom::Array<long int>'
   77 |     , m_facetIncrs(facetIncrs)
      |       ^~~~~~~~~~~~~~~~~~~~~~~~
```
https://dev.azure.com/axom/axom/_build/results?buildId=13141&view=logs&j=0b50086d-0dae-5824-d20c-f0d441949587&t=ebe88c80-2dbb-5dd9-2006-9b9be4efc4c3

PR #1526, opened 2025-04-01
Improves c2c reader class

kennyweiss:
# Summary

- This PR is a refactoring of the C2CReader classes
- It uses axom's internal `NURBSCurve` class from primal to evaluate and discretize the C2C curve pieces.
   - We previously used a custom `NURBSInterpolator` class since axom did not yet have the `NURBSCurve` class

### Follow-up work:
I decided to keep this PR relatively small, and will work on several future features/improvements in follow up PRs:
- #1527 
- #1528 
- #1529 
- #1530

### DIFF
--- a/src/axom/quest/readers/C2CReader.cpp
+++ b/src/axom/quest/readers/C2CReader.cpp
@@ -14,15 +14,12 @@
 #include "axom/primal.hpp"
 #include "axom/fmt.hpp"
 
+#include <fstream>
+#include <string>
+
 // These macros enable some debugging utilities for linearization.
 // #define AXOM_DEBUG_LINEARIZE_VERBOSE
-// #define AXOM_DEBUG_WRITE_ERROR_CURVES
 // #define AXOM_DEBUG_WRITE_LINES
-
-/// Overload to format a c2c::Point using fmt
-template <>
-struct axom::fmt::formatter<c2c::Point> : ostream_formatter
-{ };
 
 namespace axom
 {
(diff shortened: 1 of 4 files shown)


src/axom/quest/readers/C2CReader.cpp
```diff
@@ -162,598 +166,47 @@ static void appendPoints(mint::UnstructuredMesh<mint::SINGLE_SHAPE>* mesh,
...
-      double F = 3. * (2. * xp * xpp + 2. * yp * ypp) *
-        (-yp * xppp + xp * yppp) / pow(xp2_plus_yp2, 5. / 2.);
-      double G = 3. * (-yp * xpp + xp * ypp) *
-        (2. * (xpp * xpp) + 2. * (ypp * ypp) + 2. * xp * xppp + 2. * yp * yppp) /
-        (2. * pow(xp2_plus_yp2, 5. / 2.));
-      double H = (-ypp * xppp + xpp * yppp) / pow(xp2_plus_yp2, 3. / 2.);
-
-      ders[1] = E - F - G + H;
-    }
-  }
```
kennyweiss, line 637: @BradWhitlock -- curvature is the only functionality that is being removed by this PR. It wasn't being used in the code in this file.  Should we add it to the `NURBSCurve` and `BezierCurve` classes?

src/axom/quest/readers/C2CReader.cpp
```diff
@@ -965,56 +457,39 @@ void C2CReader::getLinearMeshNonUniform(
...
-
-  FILE* fhcl = fopen("hicurvelen.curve", "wt");
-  fprintf(fhcl, "# hicurvelen\n");
-
-  FILE* fcl = fopen("curvelen.curve", "wt");
-  fprintf(fcl, "# curvelen\n");
-
-  FILE* fthresh = fopen("threshold.curve", "wt");
-  fprintf(fthresh, "# threshold\n");
-#endif
```
kennyweiss, line 980: @BradWhitlock -- as far as I could tell, these were only creating files with a single line, so I removed it and the associated compiler define.

Perhaps they were used more in previous debugging/development code and the functionality was removed?

src/axom/quest/readers/C2CReader.cpp
BradWhitlock: I wrote the curvature functions when I was looking for a sample non-uniformly and ended up using a different algorithm. Moving these to the curve classes sounds like a good idea. That's where they'd be if those classes existed at the time.

src/axom/quest/readers/C2CReader.cpp
BradWhitlock: I don't recall the current state of those but it may be that you'd only see anything interesting if dynamic sampling is turned on via the shaping driver.

src/axom/quest/readers/C2CReader.cpp
```diff
@@ -1130,26 +608,97 @@ void C2CReader::getLinearMeshNonUniform(
...
+                            eval[1],
+                            dt[0],
+                            dt[1]));
+#endif
+
+      // Accumulate weight times dx*r^2.
+      sum += W[i] * xp * (r * r);
+    }
+    // Guard against volumes being negative (if the curve went the wrong way)
+    vol += fabs(scale * sum);
```
BradWhitlock, line 686: Maybe fabs() to axom::utilities::abs()

BradWhitlock (approved):
Looks like a good cleanup PR for C2CReader.

src/axom/quest/readers/C2CReader.cpp
kennyweiss: I can add it back if you'd like.

A quick search seems to indicate that it's restricted to this file: 
https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/llnl/axom+AXOM_DEBUG_WRITE_ERROR_CURVES&patternType=keyword&sm=0

BradWhitlock: No, it looks like whatever else it was writing has been removed already. Deleting code is good, especially if it is not complete anymore.

src/axom/quest/readers/C2CReader.cpp
kennyweiss: #1530

src/axom/quest/readers/C2CReader.cpp
kennyweiss: Done.

PR #1544, opened 2025-04-08
FlatMap: add Umpire support

publixsubfan:
# Summary

- Adds an `Allocator` wrapper class to disambiguate between integer arguments and allocator ID
  
  This is a speculative change, and I'd like to hear feedback from others: my biggest complaint/regret about the `Array` constructor interface is that we aren't fully drop-in compatible with the `std::vector` constructor interface, as we've sprinkled in `allocator_id` function arguments haphazardly. I'm hoping this might provide an alternative path forward that we can test out in `FlatMap` for now, and maybe bring it over into `Array` proper once fleshed out.
- Adds initial Umpire allocation support for `FlatMap`, supporting the following operations:
  - Creating a `FlatMap` within a custom Umpire allocator
  - Copying `FlatMap` between two different allocators, including device-based allocators
  - Destroying `FlatMap` created within a Umpire device-based allocator.
- Fix move assignment operator by returning reference to current instance

### DIFF
--- a/src/axom/core/detail/FlatMapOps.hpp
+++ b/src/axom/core/detail/FlatMapOps.hpp
@@ -1,0 +1,134 @@
+// Copyright (c) 2017-2025, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#ifndef Axom_Core_Detail_FlatMapOps_Hpp
+#define Axom_Core_Detail_FlatMapOps_Hpp
+
+#include "axom/core/detail/FlatTable.hpp"
+
+namespace axom
+{
+namespace detail
+{
+namespace flat_map
+{
+
+inline void setSentinel(axom::ArrayView<GroupBucket> metadata)
+{
+#if defined(AXOM_USE_UMPIRE) && defined(AXOM_USE_CUDA)
+  MemorySpace space = getAllocatorSpace(metadata.getAllocatorID());
+  using DeviceExec = axom::CUDA_EXEC<256>;
+  if(space == MemorySpace::Device)
+  {
+    for_all<DeviceExec>(
+      1,
+      AXOM_LAMBDA(IndexType) { metadata[metadata.size() - 1].setSentinel(); });
+    return;
+  }
+#endif
+  metadata[metadata.size() - 1].setSentinel();
+}
+
+template <typename KVPair, typename LookupPolicy, typename StoragePair = TypeErasedStorage<KVPair>>
+inline void destroyBuckets(axom::ArrayView<GroupBucket> metadata, axom::ArrayView<StoragePair> buckets)
+{
+  if(std::is_trivially_destructible<KVPair>::value)
+  {
+    // Nothing to do.
+    return;
+  }
+
+#if defined(AXOM_USE_UMPIRE) && defined(AXOM_USE_CUDA)
+  MemorySpace space = getAllocatorSpace(metadata.getAllocatorID());
+  using DeviceExec = axom::CUDA_EXEC<256>;
+  // CUDA-only: buckets located in device-only memory and non-trivially
+  // destructible. We'll need to "relocate" the objects to the host to
+  // destroy.
+  //
+  // This presumes the objects are "trivially relocatable." (See a similar
+  // requirement in axom::Array for device allocations on CUDA)
+  axom::Array<StoragePair> buckets_host;
+  axom::Array<GroupBucket> metadata_host;
+  if(space == MemorySpace::Device)
+  {
+    int host_allocator_id = axom::execution_space<axom::SEQ_EXEC>::allocatorID();
+    metadata_host = axom::Array<GroupBucket>(metadata, host_allocator_id);
+    buckets_host = axom::Array<StoragePair>(buckets, host_allocator_id);
+    metadata = metadata_host.view();
+    buckets = buckets_host.view();
+  }
+#endif
+  IndexType index = LookupPolicy {}.nextValidIndex(metadata, LookupPolicy::NO_MATCH);
+  while(index < buckets.size())
+  {
+    buckets[index].get().~KVPair();
+    index = LookupPolicy {}.nextValidIndex(metadata, index);
+  }
+}
+
+template <typename KVPair, typename LookupPolicy, typename StoragePair = TypeErasedStorage<KVPair>>
+inline void copyBuckets(axom::ArrayView<const GroupBucket> metadata,
+                        axom::ArrayView<const StoragePair> from_buckets,
+                        axom::ArrayView<StoragePair> to_buckets)
+{
+  if(std::is_trivially_copyable<KVPair>::value)
+  {
+    axom::copy(to_buckets.data(), from_buckets.data(), sizeof(StoragePair) * from_buckets.size());
+    return;
+  }
+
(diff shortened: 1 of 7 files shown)


src/axom/core/tests/core_flatmap.hpp
```diff
@@ -670,3 +677,66 @@ AXOM_TYPED_TEST(core_flatmap, range_for_loop_write)
...
+  // Check that maps are equivalent with a different allocator ID.
+  EXPECT_EQ(test_map_device.getAllocatorID(), device_alloc_id);
+  EXPECT_EQ(test_map_device.bucket_count(), test_map.bucket_count());
+  EXPECT_EQ(test_map_device.size(), test_map.size());
+
+  // Copy back to the host.
+  MapType test_map_host = MapType(test_map_device, axom::Allocator {host_alloc_id});
+
+  // Check that maps are equivalent with the same allocator ID.
+  EXPECT_EQ(test_map_host.getAllocatorID(), host_alloc_id);
```
bmhan12, line 718: For consistency, I think it might make sense to make this `getAllocator()` to compare `Allocator`'s, and prevent the ability to access the `int` id's altogether.

bmhan12:
> Adds an Allocator wrapper class to disambiguate between integer arguments and allocator ID

I think this makes a lot of sense and helps guard against coding errors. 
Allocating an `axom::Array` using Umpire typically looks like some form of the following:

```
    axom::Array<int> my_array(
      size,
      capacity,
      allocator_id);
```
`size`, `capacity`, and `allocator_id` are all `IndexType`/`int`.

src/axom/core/memory_management.hpp
kennyweiss, line 1: I like this idea!

src/axom/core/memory_management.hpp
```diff
@@ -197,6 +197,24 @@ inline void copy(void* dst, const void* src, std::size_t numbytes) noexcept;
...
+struct Allocator
+{
+public:
+  explicit Allocator(int alloc_id = axom::getDefaultAllocatorID()) : m_id {alloc_id} { }
+
+  /// \brief Returns the allocator ID.
+  int get() const { return m_id; }
+
+private:
+  int m_id {0};
```
kennyweiss, line 215: Should this be `0` by default?  Or perhaps  `axom::INVALID_ALLOCATOR_ID`  
(which evaluates to `-1` and is not a valid umpire allocator ID).

src/axom/core/detail/FlatMapOps.hpp
```diff
@@ -0,0 +1,134 @@
...
+namespace axom
+{
+namespace detail
+{
+namespace flat_map
+{
+
+inline void setSentinel(axom::ArrayView<GroupBucket> metadata)
+{
+#if defined(AXOM_USE_UMPIRE) && defined(AXOM_USE_CUDA)
```
kennyweiss, line 20: Does/Will this also support hip? Or does hip not need extra handling?
If it's the latter, please add a comment.

src/axom/core/detail/FlatMapOps.hpp
```diff
@@ -0,0 +1,134 @@
...
+template <typename KVPair, typename LookupPolicy, typename StoragePair = TypeErasedStorage<KVPair>>
+inline void destroyBuckets(axom::ArrayView<GroupBucket> metadata, axom::ArrayView<StoragePair> buckets)
+{
+  if(std::is_trivially_destructible<KVPair>::value)
+  {
+    // Nothing to do.
+    return;
+  }
+
+#if defined(AXOM_USE_UMPIRE) && defined(AXOM_USE_CUDA)
```
kennyweiss, line 43: Similar comment about hip here (and presumably elsewhere in this file)

src/axom/core/tests/core_flatmap.hpp
```diff
@@ -670,3 +677,66 @@ AXOM_TYPED_TEST(core_flatmap, range_for_loop_write)
...
+  for(int i = 0; i < NUM_ELEMS; i++)
+  {
+    auto key = this->getKey(i);
+    auto value = this->getValue(i * 10.0 + 5.0);
+
+    test_map.insert({key, value});
+  }
+
+  // Copy from the host to the device.
+  MapType test_map_device = MapType(test_map, axom::Allocator {device_alloc_id});
```
kennyweiss, line 707: I suppose this would be our own idiomatic way to copy, but it's interesting!

src/axom/core/tests/core_flatmap.hpp
```diff
@@ -670,3 +677,66 @@ AXOM_TYPED_TEST(core_flatmap, range_for_loop_write)
...
+
+  // Check that maps are equivalent with a different allocator ID.
+  EXPECT_EQ(test_map_device.getAllocatorID(), device_alloc_id);
+  EXPECT_EQ(test_map_device.bucket_count(), test_map.bucket_count());
+  EXPECT_EQ(test_map_device.size(), test_map.size());
```
kennyweiss, line 712: Just to make sure I understand -- `test_map_device.size()` is on the host, right? 
Would we ever want the size to be on the device? If so, what would that look like?

kennyweiss (approved):
I like this idea @publixsubfan -- it feels a lot cleaner than our current approach.

I'm eager to see what this would look like when applied to `axom::Array` and `axom::ArrayView`

src/axom/core/detail/FlatMapOps.hpp
publixsubfan: Yes, it's the latter case since with HIP we can access device memory directly from the host. I'll add a comment.

src/axom/core/tests/core_flatmap.hpp
publixsubfan: In this case, yes -- after this PR, I'm planning on pushing out another follow-up with a `FlatMapView` type for device-side accesses.

It'll be similar to the `Array`/`ArrayView` dichotomy, where the device accesses the data through the "view" type that is value-captured in the lambda.

src/axom/core/detail/FlatMapOps.hpp
adayton1: That's true on some Hip platforms, but not all.

src/axom/core/tests/core_flatmap.hpp
```diff
@@ -670,3 +677,66 @@ AXOM_TYPED_TEST(core_flatmap, range_for_loop_write)
...
+
+AXOM_TYPED_TEST(core_flatmap, copy_host_device)
+{
+  using MapType = typename TestFixture::MapType;
+
+  using HostExec = typename TestFixture::HostExec;
+  using DeviceExec = typename TestFixture::DeviceExec;
+
+  const int host_alloc_id = axom::execution_space<HostExec>::allocatorID();
+  const int device_alloc_id = axom::execution_space<DeviceExec>::allocatorID();
```
BradWhitlock, line 690: Longer term, should we have axom::execution_space<ExecSpace>::allocator() that returns an Allocator? That would get rid of some of these lines where Allocator needs to be constructed from the integer id.

BradWhitlock (approved):
Looking forward to FlatMapView.

src/axom/core/memory_management.hpp
publixsubfan: The default value of `axom::getDefaultAllocatorID()` is passed through the constructor, so I just removed the inline initializer.

src/axom/core/detail/FlatMapOps.hpp
publixsubfan: Added a comment.

PR #1550, opened 2025-04-12
Add function for making regular polygons

BradWhitlock:
It was suggested in a different PR that it could be useful to have a function for generating regular polygons, taking a transformation matrix to affect how points are created. This PR adds a function for generating regular polygons and it takes an optional transformation matrix. I added tests for the function to make sure we can vary the transformation. In doing that, it became apparent that we have some scattered transformation routines for rotations, scaling, translation. Some of these were consolidated in ``axom/core/numerics/transforms.hpp``.


kennyweiss (approved):
Thanks @BradWhitlock !

rhornung67 (approved):
Looks good @BradWhitlock Thanks.

PR #1551, opened 2025-04-14
Fix error in loop upper bound in non-RAJA build.

gunney1:
Fix a loop bounds error in marching cubes for non-RAJA builds.

### DIFF
--- a/src/axom/quest/detail/MarchingCubesImpl.hpp
+++ b/src/axom/quest/detail/MarchingCubesImpl.hpp
@@ -471,7 +471,7 @@ public:
                                        RAJA::make_span(m_firstFacetIds.data() + 1, m_crossingCount),
                                        RAJA::operators::plus<axom::IndexType> {});
 #else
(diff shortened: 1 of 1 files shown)


src/axom/quest/detail/MarchingCubesImpl.hpp
```diff
@@ -471,7 +471,7 @@ class MarchingCubesImpl : public MarchingCubesSingleDomain::ImplBase
                                        RAJA::make_span(m_firstFacetIds.data() + 1, m_crossingCount),
                                        RAJA::operators::plus<axom::IndexType> {});
 #else
-      for(axom::IndexType n = 0; n < parentCellCount; ++n)
+      for(axom::IndexType n = 0; n < m_crossingCount; ++n)
```
Arlie-Capps, line 474: I see this change makes the non-RAJA loop match the RAJA scan at lines 470-471.  Thanks for the fix, Brian.

Arlie-Capps (approved):

PR #1552, opened 2025-04-15
Support multiple start/stop cycles in Timer

gunney1:
# Summary

- Currently, `Timer` may be started and stopped multiple times, but the elapsed time it reports ignores all but the last start/stop.  For timing repeated events, it would be nice for the timer to sum up the start/stop cycles.
- This PR allows `Timer` to sum up times from multiple start/stop cycles
- It does the following:
  - Stores a running sum of all start/stop cycles since the last `reset()` call.
  - Returns the running sum in new method `summed()`, analogous to `elapsed()` method that reports only the last start/stop cycle.
  - Adds a test for multiple start/stop cycles.

### DIFF
--- a/src/axom/core/tests/utils_Timer.hpp
+++ b/src/axom/core/tests/utils_Timer.hpp
@@ -71,13 +71,13 @@ TEST(utils_Timer, timer_check_sum)
     t2.stop();
   }
 
-  std::cout << "t1 measured: " << t1.summed() << "s in " << t1.periodCount() << " periods" << std::endl;
-  std::cout << "t2 measured: " << t2.summed() << "s in " << t2.periodCount() << " periods" << std::endl;
+  std::cout << "t1 measured: " << t1.summed() << "s in " << t1.cycleCount() << " cycles" << std::endl;
+  std::cout << "t2 measured: " << t2.summed() << "s in " << t2.cycleCount() << " cycles" << std::endl;
 
-  EXPECT_EQ(t1.periodCount(), N);
-  EXPECT_EQ(t2.periodCount(), N);
+  EXPECT_EQ(t1.cycleCount(), N);
+  EXPECT_EQ(t2.cycleCount(), N);
 
-  // Estimated inaccuracy per start-stop period.
+  // Estimated inaccuracy per start-stop cycle.
   // from the C++ code, function call, etc.
   const double tol = 1.4e-4;
 
--- a/src/axom/core/tests/utils_Timer.hpp
+++ b/src/axom/core/tests/utils_Timer.hpp
@@ -58,7 +58,7 @@ TEST(utils_Timer, timer_check_duration)
 
 TEST(utils_Timer, timer_check_sum)
 {
-  const int N = 3;
+  const int N = 10;
   axom::utilities::Timer t1(false);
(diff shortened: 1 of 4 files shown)


src/axom/core/tests/utils_Timer.hpp
```diff
@@ -55,3 +55,39 @@ TEST(utils_Timer, timer_check_duration)
...
+  EXPECT_EQ(t2.cycleCount(), N);
+
+  // Estimated inaccuracy per start-stop cycle.
+#ifdef _WIN32
+  const double tol = 0.01;
+#elif __OSX__
+  const double tol = 0.09;
+#else
+  const double tol = 0.0004;
+#endif
```
kennyweiss, line 87: where did these estimates come from?

src/axom/core/utilities/Timer.hpp
```diff
@@ -160,6 +193,8 @@ class Timer
   TimeStruct m_startTime;
   TimeStruct m_stopTime;
   bool m_running;
+  TimeDiff m_summedTime;
```
kennyweiss, line 196: Do you think people might prefer an array of the timings, which they could use for stats, e.g. to compute the mean and variance of the runtimes? Or the average of the N most recent times?

kennyweiss (commented):
This is an interesting addition, @gunney1 

I'm happy to approve as is, but thought it might be good to consider whether we might want to collect all times and store them in an array. This would allow us to compute better stats on the timings (e.g. variances, min/max)

Our of curiosity, for this specific application (the MC example), is there a reason not to use caliper to gather the total timings for each region/block of code, and then report it at the end?

src/axom/core/tests/utils_Timer.hpp
gunney1: The CI test failures logs.  They're not scientific, but this we're just trying to catch potential errors in the Timer class, not high-resolution timings.

src/axom/core/utilities/Timer.hpp
gunney1: Well, I'm the only user right now, and I just need the average.  Users who need more can certainly store an array of time values themselves and compute moments on those values.  I'm a bit concerned about any implementation that may have to allocate data in the middle of timing, since that could add excessive or unpredictable overheads.

gunney1:
> This is an interesting addition, @gunney1
> 
> I'm happy to approve as is, but thought it might be good to consider whether we might want to collect all times and store them in an array. This would allow us to compute better stats on the timings (e.g. variances, min/max)
> 
> Our of curiosity, for this specific application (the MC example), is there a reason not to use caliper to gather the total timings for each region/block of code, and then report it at the end?

I started looking at performance before we supported Caliper.  This may work just as well with Caliper.

gunney1:
> Thanks @gunney1 I thought that the timer would accumulate elapsed time during multiple start and stop calls.

It turns out that is what it should have done.  So this MR is now a bug fix.  I removed the new method and made the old method behave correctly.

Arlie-Capps (approved):
Thank you for the fix, @gunney1 .

kennyweiss (approved):
Thanks @gunney1 !

PR #1553, opened 2025-04-17
Add int64 variant to `package.py`

ebchin:
# Summary

- This PR is a small addition to the axom spack recipe.
- It does the following:
  - Enables setting `AXOM_USE_64BIT_INDEXTYPE` through the spack variant `int64`

# Context

Recently, we've been running large problems in [Tribol](https://github.com/LLNL/Tribol) which have needed this option.  Exposing it to the spack recipe will simplify enabling this in our TPL builds.

### DIFF
--- a/scripts/spack/packages/axom/package.py
+++ b/scripts/spack/packages/axom/package.py
@@ -109,6 +109,8 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
 
     variant("raja", default=True, description="Build with raja")
 
+    variant("int64", default=False, description="Use 64bit integers for IndexType")
+
     varmsg = "Build development tools (such as Sphinx, Doxygen, etc...)"
     variant("devtools", default=False, description=varmsg)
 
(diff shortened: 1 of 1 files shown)


kennyweiss (approved):
Thanks for adding this variant @ebchin 

I wonder if there's a better name we could use for this though (but nothing comes immediately to mind). I checked the `hypre` package, and they use `int64` for an equivalent option, so I suppose we should go with this.

See: https://github.com/spack/spack/blob/f0a7388496a5cb062d7ec6541273b6cc697ba051/var/spack/repos/builtin/packages/hypre/package.py#L65

ebchin:
Yeah, the `int64` name came from hypre.  Happy to change if there's a consensus on something better.

kennyweiss:
Thanks @rhornung67.

In that case, @ebchin -- could you please change the default for `int64` to `True`?

rhornung67 (approved):

src/axom/mir/tests/mir_coupled3d.cpp
```diff
@@ -462,7 +462,8 @@ class test_coupling
     using SrcCoordsetView = decltype(srcCoordsetView);
 
     // 3D Elvira makes polyhedral meshes
-    auto srcTopologyView = views::make_unstructured_polyhedral_topology<int>::view(n_src_topology);
+    auto srcTopologyView =
```
kennyweiss, line 465: @BradWhitlock -- this was failing w/ `axom::IndexType == std::int64_t`. Change the template to `axom::IndexType` fixed it and seems to be the right thing to do.

PR #1554, opened 2025-04-17
Fix some host device warnings.

BradWhitlock:
When building on blueos a lot of warnings are issued due to calling ``host`` functions from ``host device`` functions. This category of warnings were mainly coming from primal. This PR fixes many of these warnings.

* Replaced calls like min,max,fabs with equivalents from Axom utilities.
* Changed some functions to pass axom::StaticArray instead of axom::Array so they can be used in ``host device`` functions.
* Fixed an unrelated line of code that looked like a copy/paste error.

### DIFF
--- a/src/axom/primal/operators/detail/intersect_impl.hpp
+++ b/src/axom/primal/operators/detail/intersect_impl.hpp
@@ -1602,7 +1602,7 @@ AXOM_HOST_DEVICE bool intersect_plane_tet3d(const Plane<T, 3>& p,
  * 
  * \note Always returns false if the line is coplanar to a planar polygon
  * 
- * \return true iff the line intersects the bilinear patch, otherwise false.
+ * \return true if the line intersects the bilinear patch, otherwise false.
  */
 AXOM_HOST_DEVICE
 inline bool intersect_line_bilinear_patch(const Line<double, 3>& line,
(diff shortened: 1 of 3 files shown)


src/axom/primal/operators/detail/intersect_impl.hpp
```diff
@@ -1856,7 +1860,7 @@ inline bool intersect_line_bilinear_patch(const Line<double, 3>& line,
             if(t1 == t2)
             {
               t.push_back(0.5 * t1);
-              v.push_back(0.5);
+              u.push_back(0.5);
```
BradWhitlock, line 1863: This line looked like a cut/paste error considering the rest of the code block pushing values into the ``u`` vector.

kennyweiss: Based on the surrounding code, I agree.

@jcs15c -- please check this and confirm.
Also, can we come up with a test case that would catch this regression?

src/axom/primal/operators/intersect.hpp
```diff
@@ -881,16 +881,29 @@ bool intersect(const Ray<T, 3>& ray,
...
+      tc.push_back(stc[i]);
+    }
+    for(int i = 0; i < suc.size(); i++)
+    {
+      uc.push_back(suc[i]);
+    }
+    for(int i = 0; i < svc.size(); i++)
+    {
+      vc.push_back(svc[i]);
+    }
```
kennyweiss, line 906: Would we expect a (slight) performance hit from this extra copy, e.g. if we call this millions of times?
If so, would it be better for the function to directly take stack arrays?

Perhaps we can measure and improve this in a follow-up PR?

kennyweiss (approved):
Thanks for fixing these @BradWhitlock!

Approving, but please wait for confirmation from @jcs15c about the `u`->`v` change in the `Line`-`BilinearPatch` intersection algorithm.

src/axom/primal/operators/intersect.hpp
BradWhitlock: @kennyweiss - I changed the code structure so copying to axom::Array first is not needed. That eliminates these lines.

kennyweiss: Thanks @BradWhitlock !

src/axom/primal/operators/detail/intersect_impl.hpp
jcs15c: Sorry to leave you hanging! Yes, this should definitely be a push to the `u` vector. I'll look into finding these issues more systematically with testing.

BradWhitlock: @jcs15c - could I get an approval on this PR then?

Arlie-Capps (approved):

Issue #1549, opened 2025-04-11
Sample-based shaping from 3D tetrahedral mesh


kennyweiss:
A user has requested that we extend our sample-based shaping capability to include shaping from a 3D ProE tetrahedral mesh.

One strategy might be to use a spatial index like a BVH tree to find the candidate tetrahedra for a given point and then use barycentric coords to determine if a point is in the tetrahedron.

Related: #1393

PR #1560, opened 2025-04-28
Adds sample-based shaping for tetrahedral shapes

kennyweiss:
# Summary

- This PR is a feature
- It adds the ability to perform sample-based shaping on tetrahedral meshes 
(#1549)
- It currently works on the host. I will port it other policies in a follow-up PR
- Misc: 
    - It also fixes some inconsistent annotations and adds a script to help track down inconsistencies between `AXOM_ANNOTATE_BEGIN`/`AXOM_ANNOTATE_END` pairs
    - It adds some additional warnings in `sidre::View`
    - It adds a device compatible equivalent of `std::function`. I'm planning to generalize this and move it to `core` in a follow up PR, but have added this to quest in the short term.


src/axom/quest/IntersectionShaper.hpp
```diff
@@ -43,30 +46,6 @@
     #include "conduit_blueprint_mcarray.hpp"
   #endif
 
-// clang-format off
```
kennyweiss, line 46: @gunney1 and others -- I moved this block to `shaping_helpers` since it's needed for sample-based shaping as well.
I `ifdef`d out the MFEM dependency there, so `IntersectionShaper` does not require MFEM.

src/axom/quest/SamplingShaper.hpp
```diff
@@ -378,44 +145,77 @@ class SamplingShaper : public Shaper
...
-    SLIC_INFO(axom::fmt::format("{:-^80}", " Generating the octree "));
+    SLIC_INFO(axom::fmt::format("{:-^80}", " Generating the spatial index "));
 
     const auto& shapeName = shape.getName();
 
-    switch(shapeDimension)
+    // note: ignoring the global shapeDimension for now since it's causing problems
+    // reading c2c when the dimension is Three
+    AXOM_UNUSED_VAR(shapeDimension);
+    if(this->shapeFormat(shape) == "c2c")
```
kennyweiss, line 155: Note: This class is screaming for `std::variant` -- we have a bunch of implementations w/ a common API, but only one of them can be valid for the instance.  Really looking forward to when we can use `C++17` features...

src/axom/quest/SamplingShaper.hpp
kennyweiss, line 1: Huge thanks to @v-dobrev for helping me debug device memory issues w/ our mfem usage in this file!

src/axom/quest/detail/shaping/PrimitiveSampler.hpp
kennyweiss, line 1: Note for reviewers: This file along w/ the changes in `shaping_helpers.cpp` contains most of the new tet-based shaping capability for Pro-E meshes. Most of the other changes in this PR are in support of this.

src/axom/quest/detail/shaping/shaping_helpers.hpp
```diff
@@ -13,21 +13,158 @@
...
+  #include "mfem.hpp"
 #endif
 
-#include "mfem.hpp"
-
 namespace axom
 {
+
+template <typename Signature, size_t MaxSize = 16>
+class function;
```
kennyweiss, line 27: Here's the `axom::function` -- which is intended to be a device compatible drop-in for `std::function`.

I'm planning to generalize this and move it to axom's core component in a follow-up PR since I need it for the point projector lambdas in the sample-based shaping.

src/axom/quest/examples/CMakeLists.txt
```diff
@@ -621,7 +621,7 @@ if(AXOM_ENABLE_TESTS AND AXOM_DATA_DIR)
     endif()
 endif()
 
-if (ENABLE_FORTRAN)
+if (ENABLE_FORTRAN AND FALSE)
```
kennyweiss, line 624: I had to disable these Fortran tests (quest's signed distance and inout Fortran tests) due to a linking issue in a `clang+ifort` configuration.

Should I revert this change and locally patch in my application, or leave it as is and add an issue to address this in the future. (I'll take responsibility for fixing it).

src/axom/mir/examples/concentric_circles/runMIR.hpp
```diff
@@ -12,7 +12,7 @@
 template <typename ExecSpace, int NDIMS>
 int runMIR(const conduit::Node &hostMesh, const conduit::Node &options, conduit::Node &hostResult)
 {
-  AXOM_ANNOTATE_BEGIN("runMIR");
+  AXOM_ANNOTATE_SCOPE("runMIR");
```
BradWhitlock, line 15: I fixed this in one of my branches too. Your script works!

src/axom/multimat/examples/basic.cpp
```diff
@@ -265,10 +265,8 @@ void multimat_to_blueprint(axom::multimat::MultiMat &mm, conduit::Node &mesh)
     // Get field properties
     auto name = mm.getFieldName(i);
     auto mapping = mm.getFieldMapping(i);
-  #if defined(AXOM_DEBUG)
-    auto dataType = mm.getFieldDataType(i);
-    SLIC_ASSERT(dataType == axom::multimat::DataTypeSupported::TypeDouble);
-  #endif
+    SLIC_ASSERT(mm.getFieldDataType(i) == axom::multimat::DataTypeSupported::TypeDouble);
```
BradWhitlock, line 268: That's a better fix.

src/axom/core/NumericArray.hpp
```diff
@@ -349,27 +349,32 @@ class NumericArray  // NOLINT
    * \brief Find the max component.
    * \return The value of the largest component.
    */
+  AXOM_HOST_DEVICE
```
adayton1, line 352: Could this be on the same line to be consistent with the inline definition below?

src/axom/quest/SamplingShaper.hpp
BradWhitlock: Good ol' polymorphism still works too and would allow a single shaper implementation pointer, which sounds like what you want from this factory.

src/axom/quest/SamplingShaper.hpp
```diff
@@ -432,16 +232,45 @@ class SamplingShaper : public Shaper
...
 
     switch(getShapeDimension())
     {
     case klee::Dimensions::Two:
-      runShapeQueryImpl(m_inoutSampler2D);
+      runShapeQueryImpl(m_inoutSampler2D.get());
       break;
     case klee::Dimensions::Three:
-      runShapeQueryImpl(m_inoutSampler3D);
+      if(this->shapeFormat(shape) == "stl")
```
BradWhitlock, line 243: Are "stl" and "proe" checks done to determine whether there are triangles vs tets? If so, can this be queried from the shape so file format checks do not proliferate?

src/axom/quest/SamplingShaper.hpp
```diff
@@ -536,11 +365,19 @@ class SamplingShaper : public Shaper
   {
     AXOM_ANNOTATE_SCOPE("finalizeShapeQuery");
 
-    delete m_inoutSampler2D;
-    m_inoutSampler2D = nullptr;
+    m_inoutSampler2D.reset();
```
BradWhitlock, line 368: I see what you mean about having to keep track of various implementation pointers.

src/axom/quest/SamplingShaper.hpp
```diff
@@ -724,13 +553,263 @@ class SamplingShaper : public Shaper
...
+                                                                  vf_name,
+                                                                  m_volfracOrder,
+                                                                  dim,
+                                                                  mfem::BasisType::Positive);
+    const mfem::FiniteElementSpace* fes = vf->FESpace();
+    const int dofs = fes->GetFE(0)->GetDof();
+
+    // access or compute the mass matrix
+    mfem::DenseTensor* mass_mat {nullptr};
+    const std::string mass_matrix_name = "shaping_mass_matrix";
```
BradWhitlock, line 642: Do these names ever change? Should they be class members with these defaults, in other words?

src/axom/quest/Shaper.cpp
```diff
@@ -313,5 +312,25 @@ double Shaper::allReduceSum(double val) const
   return val;
 }
 
+double Shaper::allReduceMin(double val) const
```
BradWhitlock, line 315: Future candidates for ``axom::core``? Any chance the 2 returns result in a warning about unreachable code?

src/axom/quest/detail/shaping/PrimitiveSampler.hpp
```diff
@@ -0,0 +1,349 @@
...
+    : m_shapeName(shapeName)
+    , m_surfaceMesh(surfaceMesh)
+  { }
+
+  ~PrimitiveSampler() { }
+
+  std::shared_ptr<mint::Mesh> getSurfaceMesh() const { return m_surfaceMesh; }
+
+  /// Computes the bounding box of the surface mesh
+  void computeBounds()
```
BradWhitlock, line 69: Could this work to compute the bounds of the mint mesh be moved into the mint mesh itself and then just return a bbox? Seeing it here, I'm wondering why the mint mesh can't do this for itself.

src/axom/quest/detail/shaping/shaping_helpers.hpp
```diff
@@ -13,21 +13,158 @@
...
+/**
+ * \brief Basic implementation of a host/device compatible analogue to std::function
+ *  
+ * \tparam R The return type of the callable object
+ * \tparam Args The parameter types of the callable object
+ * \tparam MaxSize The maximum size of the callable (including its captured variables)
+ * 
+ * \note We will extend this and move it to the core component
+ */
+template <typename R, typename... Args, size_t MaxSize>
```
BradWhitlock, line 38: Good stuff.

src/axom/quest/examples/CMakeLists.txt
BradWhitlock: It would be ideal to fix before merging.

src/axom/quest/tests/quest_sampling_shaper.cpp
```diff
@@ -85,6 +94,48 @@ class ScopedTemporaryFile
 private:
   const std::string m_filename;
 };
+
+// Utility function to slice a tetrahedron along a plane
+primal::Polygon<double, 3> slice(const primal::Tetrahedron<double, 3>& tet,
```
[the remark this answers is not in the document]
kennyweiss: Good call. I wasn't sure what (if anything) you had in `mir` related to slicing.

I also hacked the "bowtie" thing together (using pieces we already had) since it was only a test. 
I'd fix it up if we move into primal proper.

src/axom/core/NumericArray.hpp
kennyweiss: I think the line breaks are entirely determined by the `clang-format` rules.
I'll manually move it and see if it reverts.

src/axom/quest/tests/quest_sampling_shaper.cpp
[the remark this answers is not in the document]
BradWhitlock: I can't think of any geometric slicing like this in mir. I was mostly using clipping.

src/axom/quest/SamplingShaper.hpp
kennyweiss: Since the templates are baked into the sampler type, 
I think this whole block will look like the following w/ `std::variant`:
```cpp
    std::visit([](auto& inst){ runShapeQueryImpl(inst.get(); }, m_sampler);
```

src/axom/quest/SamplingShaper.hpp
kennyweiss: I don't anticipate the names changing. 
These objects are created and used only in this function, but the function is called once for each material.

src/axom/quest/SamplingShaper.hpp
```diff
@@ -724,13 +553,263 @@ class SamplingShaper : public Shaper
...
+      mass_mat = m_inoutTensors.Get(mass_matrix_name);
+    }
+    else
+    {
+      AXOM_ANNOTATE_SCOPE("mass integrator assemble");
+
+      mass_mat = new mfem::DenseTensor(dofs, dofs, NE);
+      mass_mat->HostWrite();
+      (*mass_mat) = 0.;
+      mass_mat->ReadWrite();
```
kennyweiss, line 654: I should have mentioned -- there are likely too many copies (H<->D) in the current implementation. I'll reduce it to the minimum necessary after porting everything to cuda/hip/omp.

src/axom/quest/Shaper.cpp
kennyweiss: Oh -- good call -- that should have been in a `#else` -- I'll fix them.

src/axom/quest/detail/shaping/PrimitiveSampler.hpp
kennyweiss: Presumably, it already does.

I was planning to update this part of the class a bit so we filter out tets that are not in the bounding box of the current domain, which should help reduce the memory requirements.

src/axom/quest/detail/shaping/PrimitiveSampler.hpp
```diff
@@ -0,0 +1,349 @@
...
+    // axom::for_all<ExecSpace>(
+    //   nq,
+    //   AXOM_LAMBDA(axom::IndexType i) {
+    //     SLIC_INFO_IF(i < 100,
+    //                  axom::fmt::format("Query point {} of {} -- pt: {}; inout: {}",
+    //                                    i,
+    //                                    nq,
+    //                                    query_view[i],
+    //                                    inout_view[i]));
+    //   });
```
Arlie-Capps, line 259: Perhaps discard this commented-out sanity check

src/axom/quest/detail/shaping/shaping_helpers.cpp
```diff
@@ -128,244 +185,152 @@ void generatePositionsQFunction(mfem::Mesh* mesh, QFunctionCollection& inoutQFun
...
     // Some different options for beta:
-    //beta(i) = 1.0;
-    beta(i) = ML(i) * x(i);
-    //beta(i) = ML(i)*(x(i) + 1e-14);
-    //beta(i) = ML(i);
-    //beta(i) = Mxy(i);
+    //beta[i] = 1.0;
+    beta[i] = ML[i];
+    //beta[i] = ML[i]*(1. + 1e-14);
+    //beta[i] = ML[i];
```
Arlie-Capps, line 254: ?

If we need to keep different options for beta, which may well be the case, why not keep them all?  and why uncomment line 252 but comment out line 254?

src/axom/quest/detail/shaping/shaping_helpers.cpp
```diff
@@ -128,244 +185,152 @@ void generatePositionsQFunction(mfem::Mesh* mesh, QFunctionCollection& inoutQFun
...
 
     // The low order flux correction
-    z(i) = m(i) - ML(i) * x(i) * y_avg;
+    z[i] = m[i] - ML[i] * y_avg;
```
Arlie-Capps, line 257: Is this change a bug fix?

kennyweiss: No  -- the original FCT algorithm was more flexible than we need for volume fractions. The `x` mask allows for ignoring some quadrature points, but we were always passing in an array of 1s, so I removed it.

src/axom/quest/detail/shaping/shaping_helpers.cpp
kennyweiss: I'm not sure if we'll need the different `beta`s -- I can probably remove them and just use `ML`

Arlie-Capps (approved):
Thanks, @kennyweiss .

src/axom/quest/detail/shaping/shaping_helpers.cpp
Arlie-Capps: Thanks.  Whatever you feel is best.

src/axom/quest/examples/CMakeLists.txt
kennyweiss: I reverted the change -- I'll try to handle it separately or add a local patch for my config.

src/axom/core/NumericArray.hpp
kennyweiss: clang-format was agnostic about this change (`AXOM_HOST_DEVICE` on the same line or split across two lines).
I converted everything to be on the same line.

src/axom/quest/detail/shaping/PrimitiveSampler.hpp
kennyweiss: I'm going to leave this one as is since I'm planning to update the bounding box computation in a follow-up PR.

src/axom/quest/tests/quest_sampling_shaper.cpp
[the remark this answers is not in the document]
kennyweiss: I created an issue to track this: https://github.com/LLNL/axom/issues/1579

src/axom/quest/detail/shaping/shaping_helpers.cpp
kennyweiss: I left the beta choices, but removed the duplicate on line 254.

PR #1570, opened 2025-05-10
Enhance ElviraAlgorithm so it can output plane offset and point mesh.

### LINKED ISSUE #1565: Enhance ELVIRA MIR so it can return clipping plane offsets.
The ELVIRA MIR algorithm can return a _"normals"_ field on the output Blueprint mesh that contains the normal for the plane used to clip each fragment. This feature is activated by passing ``options["normal"]=1`` to the algorithm via the options node. This was originally for debugging/visualization. I have a request to be able to return both the normals and the plane offsets.

* Change the option to ``options["planes"]=1``
* Make it return the normals as it does now
* Make it return the clipping plane offset
* Explore returning a point mesh instead of a polygonal/polyhedral mesh since the user does not care about the clipped geometry. If we did this, use the plane origin as the coordinates for the point mesh. It might be best to control this via a different option, e.g. ``options["pointmesh"]=1``
* Update docs

BradWhitlock:
This PR resolves #1565.

* Enhanced the ``axom::mir::ElviraAlgorithm`` class so it can output plane normal and offset for the planes used to clip zone fragments. These fields are output when the _"plane"_ option is set to values other than zero.
* Enhanced the ``axom::mir::ElviraAlgorithm`` to accept a _"pointmesh"_ option that causes the algorithm to generate a mesh consisting of points located at the zone fragment clipping plane origins instead of making the usual polygonal or polyhedral meshes. This skips more costly work involved in generating the usual geometry when the user only wants the clipping planes.


src/axom/mir/docs/sphinx/mir_algorithms.rst
kennyweiss, line 1: Thanks for updating the documentation!

kennyweiss (approved):
Thanks @BradWhitlock

src/axom/mir/docs/sphinx/mir_algorithms.rst
```diff
@@ -133,14 +133,48 @@ data from neighbor zones. This means that each zone is cut multiple times using
 with different orientations, resulting in potentially jagged interfaces. ELVIRA prioritizes
 conservation of material volume fractions over the appearance of the resulting material
 interfaces so it is highly accurate but it can be less visually appealing. Since ELVIRA
-output is comprised of shapes that result from several cuts of the input zones, the resulting
-topology is not necessarily water-tight and it consists of polygons for 2D and polyhedra for 3D.
+output is typically comprised of shapes that result from several cuts of the input zones,
+the resulting topology is not necessarily water-tight and it consists of polygons for 2D
```
Arlie-Capps, line 137: Readability suggestion: break sentence so we have "...is not necessarily water-tight.  It consists of polygons..."

BradWhitlock: Fixed it.

src/axom/mir/ElviraAlgorithm.hpp
```diff
@@ -954,8 +995,15 @@ class ElviraAlgorithm : public axom::mir::MIRAlgorithm
         // Emit the last leftover fragment.
         const auto fragmentIndex = offset + matCount - 1;
         const auto matId = sortedMaterialIdsView[fragmentIndex];
-        const double *normalPtr = fragmentVectorsView.data() + (fragmentIndex * numVectorComponents);
-        buildView.addShape(zoneIndex, fragmentIndex, remaining, matId, normalPtr);
+        // The last fragment's normals are just (1,0,0).
+        //const double *normalPtr = fragmentVectorsView.data() + (fragmentIndex * numVectorComponents);
+        // It seems more useful to emit the opposite of the last fragment's normal instead.
```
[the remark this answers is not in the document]
BradWhitlock: I reworked that comment so it retains what I wanted without being an old line of code.

src/axom/mir/docs/sphinx/mir_algorithms.rst
```diff
@@ -133,14 +133,48 @@ data from neighbor zones. This means that each zone is cut multiple times using
...
 interfaces so it is highly accurate but it can be less visually appealing. Since ELVIRA
-output is comprised of shapes that result from several cuts of the input zones, the resulting
-topology is not necessarily water-tight and it consists of polygons for 2D and polyhedra for 3D.
+output is typically comprised of shapes that result from several cuts of the input zones,
+the resulting topology is not water-tight. The output topology consists of polygons for 2D
+and polyhedra for 3D.
+
+The ELVIRA algorithm also supports a mode where it instead creates a mesh consisting of
+points at the centers of the polygonal/polyhedral zone centers and emits the points in
+place of the other zones. This output mode is used to save work when the algorithm is
```
Arlie-Capps, line 142: I'm confused by this sentence: please clarify.  I think you mean something like

"The ELVIRA algorithm also supports a mode where it creates points instead of a polygonal/polyhedral mesh.  The points are located where the centroids of the zone faces would have been.  This output mode is used..."

Arlie-Capps: Or does `pointmesh` produce points only at the centroids of new fragment cut planes/facets?

src/axom/mir/docs/sphinx/mir_algorithms.rst
```diff
@@ -133,14 +133,48 @@ data from neighbor zones. This means that each zone is cut multiple times using
...
 material and are extracted into their own mesh while mixed zones are passed through the ELVIRA
 algorithm to produce polygonal or polyhedral output. In the end, the two meshes are merged.
 
+Axom's ELVIRA algorithm accepts a Conduit node containing MIR options that influence the
+algorithm. The MIR options in the table above are accepted, as well as the following options
+that are specific to ELVIRA.
+
++---------------------------------+------------------------------------------------------+
+| Option                          | Description                                          |
++=================================+======================================================+
```
Arlie-Capps, line 157: Please state the default behavior.  I am guessing that if `plane` or `pointmesh` is omitted, the behavior is the same as specifying `0`.  What is the default for `point_tolerance`?

src/axom/mir/docs/sphinx/mir_algorithms.rst
Arlie-Capps: Ack, I read it more carefully.  Do you mean "A point is created at the centroid of each zone."  ?

BradWhitlock: I tried again. ``pointmesh`` makes a point at the origin of every clipping plane. A zone with 3 materials would make 3 clipping planes (2 are at the same place with opposite normals). A zone with 1 material does not get clipped so the point goes at the zone centroid. This mode is for when wants the plane origin and normal and is not that interested in the clipped zone fragments - so we make points instead.

Arlie-Capps: Thank you!  Much clearer.

Arlie-Capps (approved):
Thank you, Brad!

Issue #1564, opened 2025-05-02
Improve `dimensions` specification for shapes in Klee


kennyweiss:
I was debugging a user problem and noticed that we set the **global** dimension for all shapes in a Klee input file using something like:
```
dimensions: 3
```
When I later loaded a c2c `.contour` file (2D spline), the code treated it as as a 3D input and then crashed when loading the geometry.

#### Context:
* We also have an optional `start_dimensions` field in each shape, but the assumption is that after transformations, the shape will match the global `dimensions` value.
* In the sample-based shaping workflow, we allow users to set point projection functions at runtime, which allows sampling using 2D or 3D query points on a 2D or 3D shape.

#### Proposed solution
I think we should change this behavior to assume the following:
* `c2c` files are 2D
* `proe` files are 3D
* `STL` files are 3D

We should also add a field like `shape_dimensions` to a shape (or perhaps `dimensions` is sufficient since it is within a shape), which will allow us to override the global dimension. This will be sufficient to support cases like a 2D STL mesh (as requested in #1393 )

PR #1571, opened 2025-05-12
Initial 2D intersection shaping for c2c

bmhan12:
This PR:
- Adds initial capability to accept c2c meshes as input to the 2D intersection shaper.
  - Runtime regex output test added in `src/axom/quest/examples/CMakeLists.txt` for `quest_shaping_driver_ex`
- Fixes a bug where `m_samplesPerKnotSpan` was not being propagated.
- Future work is to support 2D shape bounded by multiple c2c contours (current assumption is a single c2c contour is bounded by the z-axis (x-axis))

### DIFF
--- a/src/axom/quest/IntersectionShaper.hpp
+++ b/src/axom/quest/IntersectionShaper.hpp
@@ -898,11 +898,18 @@ private:
 
         // Check if any Octahedron are degenerate with all points {0,0,0}
         RAJA::ReduceSum<REDUCE_POL, int> num_degenerate(0);
+
+        const int device_allocator = m_allocatorId;
+        axom::Array<OctahedronType> degenerate_oct_host(1, 1, host_allocator);
+        degenerate_oct_host[0] = OctahedronType();
+        axom::Array<OctahedronType> degenerate_oct_device =
+          axom::Array<OctahedronType>(degenerate_oct_host, device_allocator);
+        auto degenerate_oct_device_view = degenerate_oct_device.view();
+
         axom::for_all<ExecSpace>(
           m_octcount,
           AXOM_LAMBDA(axom::IndexType i) {
-            OctahedronType degenerate_oct;
-            if(octs_device_view[i].equals(degenerate_oct))
+            if(octs_device_view[i].equals(degenerate_oct_device_view[0]))
             {
               num_degenerate += 1;
             }
--- a/src/axom/quest/IntersectionShaper.hpp
+++ b/src/axom/quest/IntersectionShaper.hpp
@@ -959,18 +959,11 @@ private:
           SLIC_ERROR("2D: Contour with non-negative r values expected");
         }
 
-        SLIC_INFO("Point 0 "
-                  << " is " << pts[0]);
-        SLIC_INFO("Point 1 "
-                  << " is " << pts[1]);
-
         // Right ear of quad
         tris_host[i * 2] = PolygonStaticType({pts[1], pts[0], Point2D {pts[0][0], 0}});
-        SLIC_INFO("Triangle " << i * 2 << " is " << tris_host[i * 2]);
 
         // Left ear of quad
         tris_host[(i * 2) + 1] = PolygonStaticType({pts[1], pts[0], Point2D {pts[1][0], 0}});
-        SLIC_INFO("Triangle " << (i * 2) + 1 << " is " << tris_host[(i * 2) + 1]);
       }
 
       // Copy triangles to device
--- a/src/axom/quest/IntersectionShaper.hpp
+++ b/src/axom/quest/IntersectionShaper.hpp
@@ -801,7 +801,9 @@ private:
     }  // end of verbose output for Pro/E
   }
 
-  // Prepares the C2C mesh cells for the spatial index
+  // Prepares the C2C mesh cells for the spatial index.
+  // Produced octahedra for the revolved contour in the 3D case,
+  // or triangles for the contour bounded by the x-axis in the 2D case.
   template <typename ExecSpace>
   void prepareC2CCells()
   {
@@ -811,9 +813,6 @@ private:
     int pointcount = getSurfaceMesh()->getNumberOfNodes();
 
     axom::Array<Point2D> polyline(pointcount, pointcount);
-
-    SLIC_INFO(
-      axom::fmt::format("{:-^80}", axom::fmt::format(" Refinement level set to {} ", m_level)));
 
     SLIC_INFO(axom::fmt::format(
       "{:-^80}",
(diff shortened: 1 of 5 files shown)


src/axom/quest/IntersectionShaper.hpp
```diff
@@ -829,91 +828,152 @@ class IntersectionShaper : public Shaper
...
+
+        SLIC_INFO(axom::fmt::format("DEBUG: Total volume of all generated octahedra is {}",
+                                    total_oct_vol.get()));
+
+        // Check if any Octahedron are degenerate with all points {0,0,0}
+        RAJA::ReduceSum<REDUCE_POL, int> num_degenerate(0);
+        axom::for_all<ExecSpace>(
+          m_octcount,
+          AXOM_LAMBDA(axom::IndexType i) {
+            OctahedronType degenerate_oct;
```
kennyweiss, line 904: Should the `degenerate_oct` construction be moved outside the kernel and passed in my reference?

src/axom/quest/IntersectionShaper.hpp
```diff
@@ -2906,6 +2990,38 @@ class IntersectionShaper : public Shaper
...
+      else if(mesh_type == "quad")
+      {
+        dim = 2;
+      }
+    }
+  #endif
+
+    SLIC_ERROR_IF(!(dim == 2 || dim == 3), "Invalid computational mesh dimension");
+    return dim;
+  }
```
kennyweiss, line 3023: Looks like this is keying of the computational mesh, rather than from Klee.

In https://github.com/LLNL/axom/issues/1564 I'm proposing a better way to control the shape dimension from Klee. 
Not sure if that will help here...

kennyweiss (approved):
Thanks @bmhan12 -- this is a nice first cut at 2D intersection-based shaping via c2c meshes.  Can you please document which cases you expect to work and which ones you don't?

From the PR description, it looks like this assumes the input contour can be projected onto the x-axis without overlaps, but it would not work if there were overlaps (e.g. a closed polygon).

Please also update the RELEASE-NOTES

src/axom/quest/IntersectionShaper.hpp
```diff
@@ -829,91 +828,152 @@ class IntersectionShaper : public Shaper
...
+                            "has dimensions:\n\t{}",
+                            all_oct_bb));
+
+        // Print out the total volume of all the octahedra
+        using REDUCE_POL = typename axom::execution_space<ExecSpace>::reduce_policy;
+        RAJA::ReduceSum<REDUCE_POL, double> total_oct_vol(0.0);
+        axom::for_all<ExecSpace>(
+          m_octcount,
+          AXOM_LAMBDA(axom::IndexType i) {
+            // Convert Octahedron into Polyhedrom
```
BradWhitlock, line 876: Typo "Polyhedrom"

src/axom/quest/IntersectionShaper.hpp
```diff
@@ -829,91 +828,152 @@ class IntersectionShaper : public Shaper
...
+            PolyhedronType octPoly;
+
+            octPoly.addVertex(octs_device_view[i][0]);
```
BradWhitlock, line 879: Was ``Polyhedron::from_primitive(octs_device_view[i])`` not suitable here to make the Polyhedron from the Octahedron?

src/axom/quest/IntersectionShaper.hpp
```diff
@@ -829,91 +828,152 @@ class IntersectionShaper : public Shaper
...
+            octPoly.addNeighbors(1, {0, 2, 3, 5});
+            octPoly.addNeighbors(2, {0, 4, 3, 1});
+            octPoly.addNeighbors(3, {1, 2, 4, 5});
+            octPoly.addNeighbors(4, {0, 5, 3, 2});
+            octPoly.addNeighbors(5, {0, 1, 3, 4});
+
+            total_oct_vol += octPoly.volume();
+          });
+
+        SLIC_INFO(axom::fmt::format("DEBUG: Total volume of all generated octahedra is {}",
```
BradWhitlock, line 896: Would SLIC_DEBUG() be appropriate here?

src/axom/quest/IntersectionShaper.hpp
```diff
@@ -2906,6 +2990,38 @@ class IntersectionShaper : public Shaper
     return isTri;
   }
 
+  /*!
+   * \brief Returns the dimension of the computational mesh
+   * \return dim The dimensions of the mesh (expected is 2 or 3, -1 in case of failure)
+   */
+  int getCompMeshDim()
```
BradWhitlock, line 2997: Do the objects used in here permit this method to be marked ``const``?

BradWhitlock (approved):
I left a few comments/questions for small things to address. It looks like a contour can't be closed or overlap on the horizontal axis but IIRC the 3D shaper has that limitation too.

src/axom/quest/IntersectionShaper.hpp
bmhan12: Yes, makes sense to just construct it once, changed to have it moved outside.

src/axom/quest/IntersectionShaper.hpp
bmhan12: Good catch, I think it's the case this was before the function was added. Changed to this block to use `from_primitive`

src/axom/quest/IntersectionShaper.hpp
bmhan12: These Slic outputs are enabled when verbose flag is passed, so not really debug statements. Changed the output labels to "VERBOSE" to better fit what is happening here.

src/axom/quest/IntersectionShaper.hpp
```diff
@@ -829,91 +828,152 @@ class IntersectionShaper : public Shaper
...
+        SLIC_INFO(axom::fmt::format("DEBUG: {} Octahedron found with all points (0,0,0)",
+                                    num_degenerate.get()));
+
+        // Dump discretized octs as a tet mesh
+        axom::mint::Mesh* tetmesh;
+        axom::quest::mesh_from_discretized_polyline(octs_host_view,
+                                                    m_octcount,
+                                                    polyline_size - 1,
+                                                    tetmesh);
+        axom::mint::write_vtk(tetmesh, "discretized_surface_of_revolution.vtk");
```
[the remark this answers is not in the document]
bmhan12: The output is guarded so that it is only done when verbose flag is passed, so I think it might be fine as-is.

src/axom/quest/IntersectionShaper.hpp
bmhan12: Gave this a shot, but there are mfem API function calls that are not currently `const` (the first error I received was related to `GetMesh()`, but there are possibly more).

src/axom/quest/IntersectionShaper.hpp
bmhan12: > Looks like this is keying of the computational mesh, rather than from Klee.

Yes, it's doing exactly that, since the dimension in the klee `yaml` file is currently always `dimensions: 2` for c2c contours.

> In https://github.com/LLNL/axom/issues/1564 I'm proposing a better way to control the shape dimension from Klee.
Not sure if that will help here...

I think this will help a case like this. Driving the 2D or 3D option from the computation mesh instead of the shapes specification feels weird. Open to adopting a better way to do things.

src/axom/quest/examples/CMakeLists.txt
```diff
@@ -231,6 +231,33 @@ if(AXOM_ENABLE_MPI AND MFEM_FOUND AND MFEM_USE_MPI
...
+
+                foreach(_policy ${_policies})
+                    set(_testname quest_shaping_driver_ex_semi_circle_2d_${_policy})
+
+                    axom_add_test(
+                        NAME ${_testname}
+                        COMMAND  quest_shaping_driver_ex
+                                -i ${shaping_data_dir}/revolved_circle.yaml
+                                --policy ${_policy}
+                                --segments-per-knot-span 1000
```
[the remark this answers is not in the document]
bmhan12: Here are the results for lower number of segments, for reference:

| # Segments | Volume |
| -------- | ------- |
| 25 | 1.569756 |
| 50 | 1.570536 |
| 250 | 1.570786 |
| 500 | 1.570794 |
| 1000 | 1.570794 |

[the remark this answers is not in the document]
kennyweiss: Thanks -- in that case, should we use 50 or 100 and test against something like `1.57`

PR #1576, opened 2025-05-19
Improve support for empty bounding boxes.

### LINKED ISSUE #957: Evaluate BoundingBox behavior for invalid bounding boxes.
BoundingBox lets the user specify a region of space. A bounding box is invalid if its coordinates have values such that m_min>m_max, which makes the isValid method return false.

Some operations performed in the code will actually modify the m_min, m_max values or worse, sort the values so m_min<m_max, making it into a valid bounding box. Some of these operations (e.g. addBox, scale) have been updated to preserve the "invalidness" of an invalid bounding box. Additional operations will likely also need to do this.

Some candidates:
* expand
* shift

Actions:
* Evaluate the rest of the BoundingBox functions for how they handle invalid bounding boxes.
* Discuss whether current method of indicating invalid bounding box is appropriate (probably, given memory representation).
* Enhance CI so it tests the invalid bounding box behavior in modified methods.
* Update Doxygen method comments if needed.

BradWhitlock:
This PR resolves #957.

These changes do the following:
* BoundingBox::expand does nothing on an empty bounding box.
* BoundingBox::shift does nothing on an empty bounding box.
* Tests for empty bounding boxes were improved.
* Doxygen docs comments were updated.


src/axom/primal/tests/primal_boundingbox.cpp
```diff
@@ -699,6 +730,37 @@ TEST(primal_boundingBox, bb_get_centroid)
   PointType centroid = bbox.getCentroid();
   EXPECT_DOUBLE_EQ(0.5, centroid[0]);
   EXPECT_DOUBLE_EQ(0.5, centroid[1]);
+
+  // The centroid of an invalid bbox is 0,0.
+  BoxType bbox2;
+  EXPECT_DOUBLE_EQ(bbox2.getCentroid()[0], 0.);
+  EXPECT_DOUBLE_EQ(bbox2.getCentroid()[1], 0.);
```
kennyweiss, line 737: Looks like this was the previous behavior as well (?)

kennyweiss (approved):
Thanks @BradWhitlock !

src/axom/primal/tests/primal_boundingbox.cpp
BradWhitlock: Yes, that's right. I thought that is what it would do but was checking.

bmhan12 (approved):

RELEASE-NOTES.md
```diff
@@ -51,6 +51,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Adds a new utility tool, `mesh_converter`, which converts between mesh formats. The first conversion
   is from a Pro-E tetrahedral mesh to an STL mesh of its boundary triangles.
 - Primal: Adds a method to determine if a point is contained within a Tetrahedron.
+- The `primal::BoundingBox` class' `expand()` and `shift()` methods were modified so they do
+  nothing when called on empty bounding boxes.
```
adayton1, line 55: I think "invalid" is clearer than "empty".

BradWhitlock: Fine. invalid is technically more correct. I'll change it.

src/axom/primal/tests/primal_boundingbox.cpp
adayton1: If an invalid bbox is centered at 0, then it's actually kind of surprising to me that expand and shift wouldn't move it around.

Issue #1440, opened 2024-10-08
Array crash using RelWithDebInfo with CUDA.


BradWhitlock:
I think I was able to make a reproducer for an axom::Array crash. Build the following example program on rzansel. It will build in _RelWithDebInfo_ mode and will crash in the axom::Array constructor trying to default-initialize memory with placement new on the host (but the memory should be on the GPU).

In thinking about this some more, having axom::Array constructors in a compiled .cpp file might be the real problem. In the original scenario, I did not do anything special in my CMakeLists.txt to indicate that the file has GPU code; I only added the source file to my library target. In this example, the bug.cpp file is marked as CUDA code so axom::CUDA_EXEC will be available. The execute.cpp file where the error happens is not marked as a GPU file, though some code in axom::Array might need it to be.

**bug.cpp**
```
#include <axom/config.hpp>
#include <axom/core.hpp>

// Forward declaration
namespace axom
{
namespace mir
{
namespace views
{
void buildShapeMap(axom::Array<axom::IndexType> &indices,
                   axom::Array<axom::IndexType> &values,
                   int allocatorID);
} // end namespace views
} // end namespace mir
} // end namespace axom

template <typename ExecSpace>
void execute()
{
  axom::Array<axom::IndexType> indices, values;
  int allocatorID = axom::execution_space<ExecSpace>::allocatorID();
  axom::mir::views::buildShapeMap(/*node,*/ indices, values, allocatorID);

  std::cout << "indices.size=" << indices.size() << std::endl;
  std::cout << "values.size=" << values.size() << std::endl;
}

int main(int argc, char *argv[])
{
#if defined(AXOM_USE_CUDA)
  constexpr int BLOCK_SIZE = 256;
  using cuda_exec = axom::CUDA_EXEC<BLOCK_SIZE>;
  execute<cuda_exec>();
#endif

  return 0;
}
```

**execute.cpp**
```
#include <axom/config.hpp>
#include <axom/core.hpp>

#include <vector>

// Forward declaration
namespace axom
{
namespace mir
{
namespace views
{
void buildShapeMap(axom::Array<axom::IndexType> &indices,
                   axom::Array<axom::IndexType> &values,
                   int allocatorID)
{
  std::vector<axom::IndexType> srcIndices{{0,1,2,3,4,5}};
  std::vector<axom::IndexType> srcValues{{5,4,3,2,1,0}};
  const axom::IndexType n = srcValues.size();

[...]

PR #1582, opened 2025-05-22
GPU-related ODR fixes for Array and FlatMap

publixsubfan:
# Summary

Resolves potential ODR-related issues in `axom::Array` and `axom::FlatMap` when compiling a host-only source file against a GPU-enabled Axom library.
  - For `axom::Array`, preprocessor guards on `AXOM_GPUCC` are removed; instead, a static assertion is added within the device-side specialization of `ArrayOpsBase`.
 
     This allows for host-compiled code which does not use or instantiate `axom::Array` to include `axom/Array.hpp` without issue, while blocking an inadvertent and potentially-conflicting instantiation of the device code path within a host-only compilation. [^1]

  - For `axom::FlatMap`, replaces a kernel launch with an `axom::copy`.

Resolves the issues https://github.com/LLNL/axom/issues/1182 and https://github.com/LLNL/axom/issues/1440

[^1]: I think the implementation of `ArrayOpsBase` for the device still technically violates ODR (different tokens between a host and device compile), but it only gets correctly instantiated in a device compile so I think it's okay.

### DIFF
--- a/src/axom/core/ArrayBase.hpp
+++ b/src/axom/core/ArrayBase.hpp
@@ -914,7 +914,7 @@ struct ArrayOpsBase<T, OperationSpace::Host>
    */
   static void fill_range(T* array, IndexType begin, IndexType nelems, const T* values, MemorySpace space)
   {
-#if defined(AXOM_USE_GPU) && defined(AXOM_GPUCC) && defined(AXOM_USE_UMPIRE)
+#if defined(AXOM_USE_GPU) && defined(AXOM_USE_UMPIRE)
     if(std::is_trivially_copyable<T>::value)
     {
       axom::copy(array + begin, values, sizeof(T) * nelems);
@@ -1020,7 +1020,7 @@ struct ArrayOpsBase<T, OperationSpace::Host>
   }
 };
 
-#if defined(AXOM_USE_GPU) && defined(AXOM_GPUCC) && defined(AXOM_USE_UMPIRE)
+#if defined(AXOM_USE_GPU) && defined(AXOM_USE_UMPIRE)
 /*!
  * \name Tag types for device initialization
  */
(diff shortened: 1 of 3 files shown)


src/axom/core/ArrayBase.hpp
```diff
@@ -1143,7 +1143,21 @@ struct DeviceStagingBuffer<T, OperationSpace::Unified_Device>
...
 {
-  #if defined(__CUDACC__)
+  #if !defined(AXOM_GPUCC)
+  // To avoid an ODR issue, we error out here to ensure that axom::Array
+  // device-aware operations are not instantiated within a host-compiler.
+  // See:
+  // - https://github.com/LLNL/axom/issues/1182
+  // - https://github.com/LLNL/axom/issues/1440
+  static_assert(SPACE != OperationSpace::Device,
+                "Cannot instantiate Array operations -- file is being compiled "
```
adayton1, line 1153: Would it be clearer to say "device Array operations"?

adayton1 (approved):
Thank you!

src/axom/core/ArrayBase.hpp
adayton1: Or "device specialization of Array operations"?

kennyweiss:
Does this fix @BradWhitlock's example from https://github.com/LLNL/axom/issues/1440?
If so, can we add it in this PR as a reproducer?

BradWhitlock:
@kennyweiss - these changes make #1440 not compile because of the static_assert.

src/axom/core/ArrayBase.hpp
publixsubfan: I tried to clarify the diagnostic along those lines.

adayton1: Looks good, thanks!

PR #1583, opened 2025-05-25
Add a DeviceHash class for GPU hashing

publixsubfan:
# Summary

- Adds `axom::DeviceHash` for supporting hashing certain primitive types (integers, floating points, enums) on the GPU
  - Mimics the libc++ behavior for `std::hash`, which is to return the identity of the hashed value (casted to `axom::IndexType`)
  - For unsupported types, just wraps a call to `std::hash::operator ()`

### DIFF
--- a/src/axom/core/DeviceHash.hpp
+++ b/src/axom/core/DeviceHash.hpp
@@ -65,7 +65,7 @@ struct DeviceHashHelper<T*, void>
   using result_type = axom::IndexType;
   AXOM_HOST_DEVICE axom::IndexType operator()(T* ptr) const
   {
-    return static_cast<axom::IndexType>(ptr);
+    return reinterpret_cast<axom::IndexType>(ptr);
   }
 };
 
--- a/src/axom/core/DeviceHash.hpp
+++ b/src/axom/core/DeviceHash.hpp
@@ -1,0 +1,100 @@
+// Copyright (c) 2017-2024, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#ifndef Axom_Core_DeviceHash_Hpp
+#define Axom_Core_DeviceHash_Hpp
+
+#include "axom/config.hpp"
+#include "axom/core/Types.hpp"
+
+#include <type_traits>
+
+namespace axom
+{
+namespace detail
+{
+template <typename T, typename Enable = void>
+struct DeviceHashHelper;
+
+/// \brief Specialization for integral types
+template <typename T>
+struct DeviceHashHelper<T, std::enable_if_t<std::is_integral<T>::value>>
+{
+  using argument_type = T;
+  using result_type = axom::IndexType;
+  axom::IndexType operator()(T value) const { return value; }
+};
+
+/// \brief Specialization for floating-point types
+template <typename T>
+struct DeviceHashHelper<T, std::enable_if_t<std::is_floating_point<T>::value>>
+{
+  using argument_type = T;
+  using result_type = axom::IndexType;
+  axom::IndexType operator()(T value) const
+  {
+    // Special case: -0.0 and 0.0 compare equal but have different byte representations.
+    if(value == T {0.})
+    {
+      return 0;
+    }
+    return value;
+  }
+};
+
+/// \brief SFINAE specialization for enum types
+template <typename T>
+struct DeviceHashHelper<T, std::enable_if_t<std::is_enum<T>::value>>
+{
+  using argument_type = T;
+  using result_type = axom::IndexType;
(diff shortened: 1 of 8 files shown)


src/axom/core/tests/core_device_hash.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+// Axom includes
+#include "axom/config.hpp"
+#include "axom/core/Macros.hpp"
+#include "axom/core/DeviceHash.hpp"
+
+// gtest includes
+#include "gtest/gtest.h"
+
+TEST(core_device_hash, hash_int)
```
BradWhitlock, line 14: DeviceHash is being tested here, but not on various devices. Is that demonstrated elsewhere such as in FlatMap tests?

publixsubfan: I think that was originally the idea. We could add additional testing to this PR as well.

src/axom/core/tests/core_device_hash.hpp
kennyweiss, line 1: Can you please add a test to show how one might extend this to new types, e.g. to a Point struct w/ `x`-, `y`- and `z` fields, or to an `axom::NumericalArray` ?

kennyweiss (approved):
Thanks @publixsubfan 

It's be helpful to see how `DeviceHash` could be used w/ a composite type, e.g. a simple 3D-point type.

Please also update the `RELEASE-NOTES`

src/axom/core/tests/core_device_hash.hpp
publixsubfan: My initial thought is that for user-provided types, we could ask users to create a template specialization of `axom:::DeviceHash`; if we wanted to provide a "default" hash for our internal types, we could specialize the helper function `axom::detail::DeviceHashHelper`.

kennyweiss: Thanks -- I don't think we need to add any for our internal types to this PR, just thought it'd be helpful to add a simple example to show how one might do it.

src/axom/core/DeviceHash.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+/// \brief Specialization for floating-point types
+template <typename T>
+struct DeviceHashHelper<T, std::enable_if_t<std::is_floating_point<T>::value>>
+{
+  using argument_type = T;
+  using result_type = axom::IndexType;
+  AXOM_HOST_DEVICE axom::IndexType operator()(T value) const
+  {
+    // Special case: -0.0 and 0.0 compare equal but have different byte representations.
```
rhornung67, line 39: 👍

src/axom/core/DeviceHash.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+/// \brief Specialization for pointer types
+template <typename T>
+struct DeviceHashHelper<T*, void>
+{
+  using argument_type = T*;
+  using result_type = axom::IndexType;
+  AXOM_HOST_DEVICE axom::IndexType operator()(T* ptr) const
+  {
+    return reinterpret_cast<axom::IndexType>(ptr);
```
rhornung67, line 68: Why not static_cast here?

src/axom/core/tests/core_device_hash.hpp
```diff
@@ -0,0 +1,142 @@
...
+{
+  axom::DeviceHash<int> device_hasher;
```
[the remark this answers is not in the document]
publixsubfan: I think the conversions between different types, as well as enum handling, might be a little ugly. Also, there is a special case for floating points that we have to test (`DeviceHash(-0.0) == DeviceHash(0.0)`)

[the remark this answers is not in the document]
rhornung67: understood. I didn't look at it too closely.

rhornung67 (approved):

src/axom/core/tests/core_device_hash.hpp
publixsubfan: I added a small test as an example of a user-defined hash specialization. The main caveat is that the `DeviceHash` specialization must be defined within the axom namespace.

src/axom/core/tests/core_device_hash.hpp
publixsubfan: I added some tests on the GPU.

src/axom/core/tests/core_device_hash.hpp
```diff
@@ -0,0 +1,267 @@
...
+
+    axom::IndexType hash_result {};
+    for(int i = 0; i < NWORDS; i++)
+    {
+      hash_result ^= (bytes[i] + 0x853c49e6);
+    }
+    return hash_result;
+  }
+};
+}  // namespace axom
```
kennyweiss, line 219: Thanks @publixsubfan !

PR #1587, opened 2025-06-09
Add early return in combineMessages if no combiners exist

gberg617:
# Summary

This PR is a bugfix that adds an early return in `Lumberjack::combineMessages()` if the combiners size is zero.  This improves performance for cases where no combiners exist, particularly when there are many messages.


src/axom/lumberjack/Lumberjack.cpp
```diff
@@ -211,6 +211,12 @@ void Lumberjack::combineMessages()
   std::vector<Message*> finalMessages;
   std::vector<int> indexesToBeDeleted;
   int combinersSize = (int)m_combiners.size();
+
+  if(combinersSize == 0)
+  {
+    return;
+  }
```
kennyweiss, line 218: What does it mean for lumberjack to not any registered combiners?

Would this be considered to be a bad configuration? Or does it just mean that messages will be passed through without being combined?

kennyweiss: The docs don't appear to say anything about the case where no combiners are registered -- https://axom.readthedocs.io/en/develop/axom/lumberjack/docs/sphinx/core_concepts.html#combining

(unless it's mentioned somewhere else)

gberg617: From my understanding, it seems that it is valid to have a Lumberjack object with no combiners.  I would like to get feedback from the Axom team on this, because I'm not sure if my assumption is correct.

The main reason I have a Lumberjack object with no combiners is that I need to call `getMessages()` for certain reasons in my application.  I couldn't find an equivalent "get" method for other types of MPI-aware streams such as `SynchronizedStream`, but adding such a method to other streams could also be a potential solution.  Any recommendations would be greatly appreciated!

bmhan12: > Or does it just mean that messages will be passed through without being combined?

This is the behavior I have observed. So it acts like a `SynchronizedStream` when there are no combiners, with the difference being the choice of `Communicator` for the MPI communication pattern.

bmhan12 (approved):

src/axom/lumberjack/Lumberjack.cpp
bmhan12: > This is the behavior I have observed. So it acts like a SynchronizedStream when there are no combiners, with the difference being the choice of Communicator for the MPI communication pattern.

A unit test was added to confirm the messages will pass through without being combined: https://github.com/LLNL/axom/pull/1587/commits/b0c6d2c0ae0f6daabbbe85e35f682d27af2403c5

PR #1592, opened 2025-06-25
Enhance mesh_tester application so it can write STL.

### LINKED ISSUE #262: Export triangle meshes in STL format
Several queries in axom are based on `STL` triangle meshes (e.g. the In/Out and Signed distance query).

We also have a `mesh_tester` application that validates geometric and topological properties of an input `STL` mesh and performs some simple mesh repair.

However, we currently only have the ability to export triangle meshes in `VTK` format.

We should add the ability to write out `STL` meshes in ascii or binary format.

BradWhitlock:
This PR resolves #262 by enhancing the mesh_tester application so it can write STL files.

* Adds `STLWriter` class - in quest since the STLReader is there.
  * The `STLWriter` class can write ascii or binary STL files.
  * A `write_stl` function was added that calls the `STLWriter` class for those who prefer a function interface.
* Quest `readers` directory renamed to `io` since there is a writer in there now.
* Added tests for writing STL files and reading them back in.
* Enhanced mesh_tester with `--format` and `--binary` command line arguments allow STL output (optionally binary).


RELEASE-NOTES.md
```diff
@@ -53,6 +54,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - The `primal::BoundingBox` class' `expand()` and `shift()` methods were modified so they do
   nothing when called on invalid bounding boxes.
 - Updates to [MFEM version 4.8.0][https://github.com/mfem/mfem/releases/tag/v4.8]
+- Readers in Quest were moved from a `quest/readers` directory to `quest/io`.
```
kennyweiss, line 57: 👍

src/axom/quest/io/STLWriter.hpp
```diff
@@ -0,0 +1,116 @@
...
+#include <string>
+
+namespace axom
+{
+namespace quest
+{
+/*!
+ * \class STLWriter
+ *
+ * \brief A simple STL writer for the faces of Mint meshes.
```
kennyweiss: Future leaning suggestion: I think we'll want to update the readers/writers to work with blueprint/BUMP. 

Should we emphasize that this works on collections of triangles (or perhaps planar polygons) rather than the input representation (mint)?

```suggestion
 * \brief A simple STL writer for a collection of triangular faces with limited support for planar polygons
```

Obviously, we can change this in the future, so feel free to ignore this suggestion in this PR.

src/axom/quest/io/STLWriter.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+  /*!
+   * \brief Determines whether the input mesh looks topologically 2D since
+   *        the mesh dimension seems to reflect the number of coordinate components.
+   *
+   * \param mesh The mesh to test.
+   *
+   * \return True if the mesh seems 2D, false otherwise.
+   */
+  bool isTopologically2D(const mint::Mesh* mesh) const;
```
kennyweiss: Should the `mesh` be a class member? I'm assuming you did it this way for consistency w/ the STLReader.

src/axom/quest/io/STLWriter.hpp
```diff
@@ -0,0 +1,116 @@
...
+   *
+   * \param mesh The input mesh.
+   *
+   * \return The number of triangles.
+   */
+  IndexType getNumberOfTriangles(const mint::Mesh* mesh) const;
+
+  /*!
+   * \brief Determines whether the input mesh looks topologically 2D since
+   *        the mesh dimension seems to reflect the number of coordinate components.
```
kennyweiss: Minor: I'm confused by this comment. Could you please update it?

Specifically, the word "looks" and the reference to "coordinate components" made me think we were going to check if the mesh was (geometrically) planar.

The check in the implementation is about topological cell types (triangle and quad), not about geometric coordinates, which is in line with the function name.

src/axom/quest/io/STLWriter.cpp
```diff
@@ -0,0 +1,232 @@
...
+ * \param coords The triangle coordinates.
+ * \param N The triangle normal.
+ */
+template <typename NormalType>
+void writeTriangle(std::ofstream &out, bool binary, double coords[3][3], const NormalType &N)
+{
+  if(binary)
+  {
+    using float32 = float;
+    using uint16 = unsigned short;
```
kennyweiss: Any reason not to use `std::uint32_t` here?

src/axom/quest/io/STLWriter.cpp
```diff
@@ -0,0 +1,232 @@
...
+    float32 n32[3], coords32[3][3];
+    for(int comp = 0; comp < 3; comp++)
+    {
+      n32[comp] = static_cast<float32>(N[comp]);
+      coords32[0][comp] = static_cast<float32>(coords[0][comp]);
+      coords32[1][comp] = static_cast<float32>(coords[1][comp]);
+      coords32[2][comp] = static_cast<float32>(coords[2][comp]);
+    }
+    const uint16 attr = 0x7fff;
```
kennyweiss: Is this value arbitrary? Perhaps add a comment?

src/axom/quest/io/STLWriter.cpp
```diff
@@ -0,0 +1,232 @@
...
+{
+  if(binary)
+  {
+    using float32 = float;
+    using uint16 = unsigned short;
+    float32 n32[3], coords32[3][3];
+    for(int comp = 0; comp < 3; comp++)
+    {
+      n32[comp] = static_cast<float32>(N[comp]);
+      coords32[0][comp] = static_cast<float32>(coords[0][comp]);
```
kennyweiss, line 41: Aside: I didn't appreciate until now that binary STLs use floats for the geometry and normals

src/axom/quest/tests/quest_stl_writer.cpp
kennyweiss, line 1: Great tests!

kennyweiss (approved):
Thanks @BradWhitlock for this great addition!

src/axom/quest/io/STLWriter.hpp
BradWhitlock: I agree that the readers/writers can eventually provide Conduit interfaces and use Bump internally for accessing those meshes. Future PR.

```
// Future
struct STLWriter {
//...
#if defined(AXOM_USE_CONDUIT)
   void write(const conduit::Node &mesh);
#endif
//...
};
```

src/axom/quest/io/STLWriter.hpp
BradWhitlock: I changed `mesh` to a class member. Someday we can include a `conduit::Node*` member too.

src/axom/quest/io/STLWriter.hpp
BradWhitlock: I changed the comments. This functionality was needed when I started using the writer in the mesh_tester. The surface mesh has 3D coordinates. `mint::Mesh::getDimension()` returns 3 (like the coordinates) but there are no faces, just cells, so we have to iterate over the cells rather than using `mint::for_all_faces`.

src/axom/quest/io/STLWriter.cpp
BradWhitlock: I changed it to `std::uint16_t`.

src/axom/quest/io/STLWriter.cpp
BradWhitlock: I added a comment. Some packages use it for surface color.

PR #1593, opened 2025-06-25
Add MPI to candidates example

bmhan12:
This PR:

- Adds MPI support to `quest_candidates_examples.cpp`
  - Input mesh (donor) to the spatial index can either be single domain or partitioned with one domain per rank.
  - Query (target) to the spatial index must be single domain.
- Fix issue with potentially loading `int64` or `int32` connectivity data.

### DIFF
--- a/src/axom/quest/examples/quest_candidates_example.cpp
+++ b/src/axom/quest/examples/quest_candidates_example.cpp
@@ -269,16 +269,6 @@ HexMesh loadBlueprintHexMesh(const std::string& mesh_path, bool verboseOutput = 
   int connectivity_size =
     (n_load[0]["topologies/topo/elements/connectivity"]).dtype().number_of_elements();
 
-  // Sanity check for number of cells
-  int cell_calc_from_nodes = std::round(std::pow(std::pow(num_nodes, 1.0 / 3.0) - 1, 3));
-  int cell_calc_from_connectivity = connectivity_size / HEX_OFFSET;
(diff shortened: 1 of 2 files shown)


rhornung67 (approved):

src/axom/quest/examples/quest_candidates_example.cpp
```diff
@@ -298,9 +351,32 @@ HexMesh loadBlueprintHexMesh(const std::string& mesh_path, bool verboseOutput =
...
+  else if(conn_id == conduit::DataType::INT64_ID)
+  {
+    temp_conn_data.resize(connectivity_size);
+    auto conn_int64_data = n_load[0]["topologies/topo/elements/connectivity"].as_int64_ptr();
+    for(int i = 0; i < connectivity_size; ++i)
+    {
+      temp_conn_data[i] = static_cast<int>(conn_int64_data[i]);
+    }
+    conn_data = temp_conn_data.data();
+  }
```
nselliott, line 366: I think there are conduit utilities that can do the cast-copy from 64 bit to 32 in the "else if" block without needing to write the for loop or introduce axom::Array as an intermediate container type.

```
Node temp_conn_data;  // declared before if-else block
n_load[0]["topologies/topo/elements/connectivity"].to_int32_array(temp_conn_data);
conn_data = temp_conn_data.as_int32_ptr();
```

bmhan12: Fixed!

nselliott (approved):

PR #1601, opened 2025-06-30
Add Message creation time and sort by creation time

gberg617:
# Summary

- This PR is a feature that adds a creation time to each message and sorts log messages in Lumberjack by creation time.  This allows for outputting log messages coming from separate ranks in the same order that they were created after being combined and filtered.


src/axom/lumberjack/LineFileTagCombiner.hpp
```diff
@@ -88,6 +88,9 @@ class LineFileTagCombiner : public axom::lumberjack::Combiner
                const int ranksLimit)
   {
     combined.addRanks(combinee.ranks(), combinee.count(), ranksLimit);
+    if (combinee.creationTime() < combined.creationTime()) {
```
gberg617: When adding a creation time, we have to make some sort of assumptions on what happens when duplicate messages are combined.  I'm assuming here that the first message's creation time will be kept when combined.

src/axom/lumberjack/Lumberjack.cpp
```diff
@@ -149,6 +166,9 @@ void Lumberjack::pushMessagesOnce()
   receivedPackedMessages.clear();
 
   combineMessages();
+
+  std::sort(m_messages.begin(), m_messages.end(),
```
gberg617: I'm assuming here that messages will always be sorted by creation time.  Should we have the ability to disable this?

src/axom/lumberjack/Lumberjack.hpp
```diff
@@ -194,6 +195,34 @@ class Lumberjack
...
+   *
+   * \param [in] text Text of the Message
+   * \param [in] text Ranks list
+   * \param [in] fileName File name of Message
+   * \param [in] lineNumber Line number of Message
+   * \param [in] level The level of the severity of the Message.
+   * \param [in] tag The tag of where the Message originated.
+   *****************************************************************************
+   */
+  void queueMessage(const std::string& text,
```
gberg617, line 218: This version of `queueMessage` was added because it is useful for transferring messages between Lumberjack objects while preserving the ranks list and count (I've found that transferring messages between Lumberjack objects is useful for accomplishing various things)

src/axom/lumberjack/tests/lumberjack_Lumberjack.hpp
```diff
@@ -116,12 +116,12 @@ TEST(lumberjack_Lumberjack, combineMessagesPushOnce02)
   axom::lumberjack::Lumberjack lumberjack;
   lumberjack.initialize(&communicator, ranksLimit);
 
-  lumberjack.queueMessage("");
-  lumberjack.queueMessage("Should be combined.");
-  lumberjack.queueMessage("Should be combined.");
-  lumberjack.queueMessage("Should be combined.");
-  lumberjack.queueMessage("Should be combined.");
-  lumberjack.queueMessage("Should be combined.");
+  lumberjack.queueMessage("", 0.0);
```
gberg617, line 126: Adding a creation time is needed for some tests in order to preserve the ordering of the messages when they are sorted (note: I'm assuming that messages are always sorted by creation time)

src/axom/slic/streams/LumberjackStream.cpp
```diff
@@ -184,7 +185,7 @@ void LumberjackStream::append(message::Level msgLevel,
     return;
   }
 
-  m_lj->queueMessage(message, fileName, line, msgLevel, tagName);
+  m_lj->queueMessage(message, fileName, line, msgLevel, MPI_Wtime(), tagName);
```
gberg617, line 188: The actual creation time is fed into `queueMessage` in the `LumberjackStream` in slic.  I'm not sure this is the best place to put the actual time.  It's possible perhaps to move this into `Lumberjack`.  I've had better luck with `MPI_Wtime()` than other tools such as `std::chrono::time_point` which seemed to produce inconsistent results when running problems on multiple ranks.

src/cmake/blt
bmhan12, line 1: Is the blt update necessary?

src/axom/lumberjack/Message.cpp
```diff
@@ -37,6 +37,8 @@ int Message::lineNumber() const { return m_lineNumber; }
 
 int Message::level() const { return m_level; }
 
+double Message::creationTime() const { return m_creationTime; }
```
bmhan12, line 45: Turns out, `slic` already has the notion of a `timestamp`: https://github.com/LLNL/axom/blob/db9d01d6eef190bccb23cf4e609f2d7b729973b3/src/axom/slic/core/LogStream.hpp#L63
It should be possible to extend that to lumberjack, instead of creating a new `creationTime` field. Likely overlap with what you have so far.

src/axom/lumberjack/LineFileTagCombiner.hpp
bmhan12: If creation time is necessary to combine messages, I think the name of the Combiner should reflect that (`LineFileTagCreationTimeCombiner`) given the current convention.

src/axom/lumberjack/Lumberjack.cpp
bmhan12: > Should we have the ability to disable this?

Yes, I think we should be able toggle it on/off, depending on if the Combiner cares about creation time/timestamps or not. 

Can the sorting not be done in the Combiner? This feels like something that should be specified in the Combiner class, invoked by `combineMessages()`.

src/axom/lumberjack/Message.cpp
bmhan12: I also prefer `timestamp` over `creationTime` in terms of naming.

gberg617: From my understanding, it looked like the slic timestamp is based on the time that `getFormatedMessage` is called, which is not the same as the creation time of the actual message.  Is this correct?

src/cmake/blt
gberg617: No, I'll remove this.  Thanks for catching that!

src/axom/lumberjack/LineFileTagCombiner.hpp
gberg617: this wasn't strictly necessary in the combiners, I just added it in to enforce some consistency.  I can consider removing this if you don't think it's useful to include in every combiner.

src/axom/lumberjack/Message.cpp
bmhan12: It looks like the specific time is specified here in `getTimeStamp()`: https://github.com/LLNL/axom/blob/db9d01d6eef190bccb23cf4e609f2d7b729973b3/src/axom/slic/core/LogStream.cpp#L47-L64

You should be able to override that implementation in `LumberjackStream` with the MPI API call.

src/axom/lumberjack/Lumberjack.cpp
gberg617: I'm not sure how this could be done in the Combiner class.  The Combiner is mostly used to combine two duplicate messages, whereas the sorting re-orders a fixed set of messages based on creation time.  Did you have some thoughts or ideas on how to do this?

bmhan12: That's a good point, Combiner class is not really cut out for that... then the sorting is likely fine as-is.

src/axom/lumberjack/Message.cpp
gberg617: Oh right, good call.  I'll give that a try!

gberg617: Another subtle distinction is that `MPI_Wtime()` measures a wall time relative to some fixed point and returns a double that represents some elapsed time.  Because of the way it works, it's not very easy to convert this into a date/time format.  In that sense, its more of a relative time than an actual timestamp.

bmhan12: > In that sense, its more of a relative time than an actual timestamp.

Interesting, sounds like timestamp != creationTime like I originally thought, and that it would not make sense to use Slic/LogStream's timestamp in place of Lumberjack's creationTime.

src/axom/lumberjack/LineFileTagCombiner.hpp
```diff
@@ -88,6 +88,10 @@ class LineFileTagCombiner : public axom::lumberjack::Combiner
                const int ranksLimit)
   {
     combined.addRanks(combinee.ranks(), combinee.count(), ranksLimit);
+    if(combinee.creationTime() < combined.creationTime())
+    {
+      combined.creationTime(combinee.creationTime());
+    }
```
bmhan12, line 96: Maybe useful to note that the combined message is using the earliest creation time across the messages.

src/axom/lumberjack/tests/lumberjack_Lumberjack.hpp
```diff
@@ -370,12 +370,12 @@ TEST(lumberjack_Lumberjack, combineMixedMessages03)
...
-  lumberjack.queueMessage("Should not be combined 3.");
-  lumberjack.queueMessage("Should be combined.");
-  lumberjack.queueMessage("Should be combined.");
-  lumberjack.queueMessage("Should be combined.");
+  lumberjack.queueMessage("", 1.0);
+  lumberjack.queueMessage("Should not be combined 2.", 2.0);
+  lumberjack.queueMessage("Should not be combined 3.", 3.0);
+  lumberjack.queueMessage("Should be combined.", 4.0);
+  lumberjack.queueMessage("Should be combined.", 5.0);
+  lumberjack.queueMessage("Should be combined.", 6.0);
```
bmhan12, line 385: Does the test here check if the combined creationTime is as expected (e.g. I think it would be 4.0 for "Should be combined.")? That would be a useful check to add to the unit test here.

src/axom/slic/streams/LumberjackStream.cpp
bmhan12: > I'm not sure this is the best place to put the actual time. It's possible perhaps to move this into Lumberjack

Here is likely fine. 
I'm not too familiar with the different timing tools, so having the default creationTime in `queueMessage` overloads be `0.0` (essentially a do nothing/no time sorting) or an API users must specify makes sense to me.

Just curious - does `MPI_Wtime() guarantee the value is consistent across all ranks, or just the calling rank (the documentation seems to suggest only for the calling rank: https://www.open-mpi.org/doc/v3.0/man3/MPI_Wtime.3.php) ?  

When used in conjunction with the Combiner behavior of always using the earliest creationTime, does that always give the ordering you are looking for?

Regardless, it's a good idea to document here any implications of using `MPI_Wtime()` here.

src/axom/lumberjack/LineFileTagCombiner.hpp
gberg617: I went ahead and changed the name of all the combiners.  But thinking about this a bit more, perhaps it's better to keep the original names, since the combination criteria aren't changing (we're only adding the creation time after we decide whether to combine).  Any thoughts on this?

src/axom/slic/streams/LumberjackStream.cpp
gberg617: That's a good point.  I added a new member/method in the communicator classes to compute a start time that is consistent across ranks.  The time passed to `queueMessage` is now the elapased time (`MPI_Wtime() - startTime`).  It seemed to work before without this common start time, but it's good to add this anyways.

src/axom/lumberjack/tests/lumberjack_Lumberjack.hpp
gberg617: I added a check for this!

src/axom/lumberjack/LineFileTagCombiner.hpp
gberg617: I added a comment on this.

src/axom/lumberjack/LineFileTagCombiner.hpp
bmhan12: > But thinking about this a bit more, perhaps it's better to keep the original names, since the combination criteria aren't changing (we're only adding the creation time after we decide whether to combine). 

🤦 . You're right, definitely makes more sense to keep the original names. I was confusing the combination criteria and the actual contents being combined in regards to the naming scheme. 

We definitely want to update the documentation section here on combining: https://axom.readthedocs.io/en/develop/axom/lumberjack/docs/sphinx/core_concepts.html#combining, which only mentions the ranks and message count being adjusted, no mention of creation time yet.

src/axom/lumberjack/BinaryTreeCommunicator.cpp
```diff
@@ -29,6 +29,10 @@ namespace lumberjack
 void BinaryTreeCommunicator::initialize(MPI_Comm comm, int ranksLimit)
 {
   m_mpiComm = comm;
+
+  MPI_Barrier(m_mpiComm);
+  m_startTime = MPI_Wtime();
```
gberg617, line 34: I am assuming here that the call to initialize is done at about the same time on all ranks.  If we want to be more accurate, we can broadcast `m_startTime` from the root rank to all others.

src/axom/lumberjack/LineFileTagCombiner.hpp
gberg617: OK sounds good, I added the documentation!

bmhan12 (approved):

src/axom/lumberjack/LineFileTagCombiner.hpp
bmhan12: Thanks! ReadTheDocs build for this PR can be found here: https://axom.readthedocs.io/en/feature-bergel1-lumberjack_creation_time_sorting/axom/lumberjack/docs/sphinx/index.html

PR #1607, opened 2025-07-07
FlatMap: Add FlatMapView type for GPU accesses

publixsubfan:
# Summary

- Adds `axom::FlatMapView` to expose view-like access to an `axom::FlatMap` from within a GPU kernel.
- Adds `FlatMap::view()` to return a view type from a given `FlatMap`.

## Details/Limitations
 
`FlatMapView` exposes all access operations which do not require mutating the underlying `FlatMap`; we consider a mutating operation one which modifies the `m_metadata` array within the parent `FlatMap`. With this constraint, we implement the following interface for `FlatMapView`:
 - Container-wide read-only functions: `empty()`, `size()`, `bucket_count()`
 - Key-based read-only queries: `find(key)`, `begin/end()`, `contains(key)`, `count(key)`, `operator []` (with some modifications)

The behavior of `FlatMapView::operator[]` diverges somewhat from that of `FlatMap::operator[]`:
- If a key-value pair matches, a reference to the existing value is returned.
- If a key is not found, a default value is returned in the reference - but *not inserted into the map*.
- The value reference returned is always `const` - assignment of a value to a key is not supported through this interface.

Modifying the value of a key-value pair already present in the parent `FlatMap` is possible through the `FlatMapView::find()` interface. This requires a "mutable" `FlatMapView`. No guarantees on thread safety are provided using this functionality.

`FlatMapView` accesses are supported from GPU code only if the underlying `FlatMap` is allocated from within a GPU-accessible memory space. (see Allocator interface)


Arlie-Capps (approved):
Looks like a solid contribution, Max.  Thank you.  In a subsequent PR, please write up a few examples of how to use FlatMapView and other "views."

src/axom/core/tests/core_flatmap_for_all.hpp
```diff
@@ -0,0 +1,152 @@
...
+      {
+        keys_out[idx] = it->first;
+        values_out[idx] = it->second;
+        valid_out[idx] = true;
+      }
+      values_out_bracket[idx] = test_map_view[idx];
+    });
+
+  // Check contents on the host
+  for(int i = 0; i < NUM_ELEMS; i++)
```
kennyweiss, line 143: It looks like ExecSpace is hard coded above to `axom::SEQ_EXEC`. 

Are you planning to add device tests for this in a follow-up PR?

src/axom/core/tests/core_flatmap_for_all.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+    test_map.insert({key, value});
+  }
+
+  MapViewConstType test_map_view(test_map);
+
+  axom::Array<int> valid_vec(NUM_ELEMS + EXTRA_THREADS, 0);
+  axom::Array<int> keys_vec(NUM_ELEMS);
+  axom::Array<double> values_vec(NUM_ELEMS);
+  axom::Array<double> values_vec_bracket(NUM_ELEMS + EXTRA_THREADS);
```
kennyweiss: Out of curiosity -- what does the `_bracket` part of the name signify?

src/axom/core/FlatMapView.hpp
```diff
@@ -0,0 +1,266 @@
...
+namespace axom
+{
+
+/*!
+ * \class FlatMap
+ *
+ * \brief Provides a read-only view of a key-value container.
+ *
+ * \see FlatMapView
+ */
```
kennyweiss: Should this be

```suggestion
/*!
 * \class FlatMapView
 *
 * \brief Provides a read-only view of a key-value container.
 *
 * \see FlatMap
 */
```

src/axom/core/detail/FlatTable.hpp
```diff
@@ -120,20 +120,73 @@ struct GroupBucket
...
+      pack_data.metadata.ofw = hashOfwBit;
+      RAJA::atomicOr<RAJA::auto_atomic>(&data[0], pack_data.data[0]);
+  #else
+      RAJA::atomicOr<RAJA::auto_atomic>(&(metadata.ofw), hashOfwBit);
+  #endif
+      return;
+    }
+#endif
     metadata.ofw |= hashOfwBit;
   }
```
kennyweiss: Can this be reimplemented in terms of axom's atomics (which use raja, when available)?

https://github.com/LLNL/axom/blob/develop/src/axom/core/execution/atomics.hpp

src/axom/core/detail/FlatTable.hpp
```diff
@@ -120,20 +120,73 @@ struct GroupBucket
...
+    std::uint8_t curr_ofw;
+#if defined(AXOM_USE_RAJA)
+    if(Atomic)  // TODO: should be constexpr
+    {
+  #if defined(AXOM_USE_CUDA)
+      // CUDA workaround for lack of 8-bit atomicLoad.
+      curr_ofw = *const_cast<const volatile std::uint8_t*>(&(metadata.ofw));
+  #else
+      // TODO: why is the const_cast required? (RAJA issue?)
+      curr_ofw = RAJA::atomicLoad<RAJA::auto_atomic>(const_cast<std::uint8_t*>(&(metadata.ofw)));
```
kennyweiss: Can we add an `axom::atomicLoad` to `axom/core/execution/atomics.hpp` and use it here?

kennyweiss (approved):
Thanks @publixsubfan !

Please don't forget to update the RELEASE_NOTES
* I had some comments about implementing the RAJA atomics in terms of the axom atomics to simplify the code
* Also, whenever you think it's ready, it would be nice to have a short example in `core` that uses a hashmap on the host and device. E.g. computing the histogram of a distribution, or something simple like that.

src/axom/core/tests/core_flatmap_for_all.hpp
publixsubfan: Ah, good catch -- that should be in this PR. Let me add the test for `CUDA_EXEC`/`HIP_EXEC`.

src/axom/core/tests/core_flatmap_for_all.hpp
publixsubfan: That is for testing the `FlatMapView::operator[]` return behavior; specifically, we're just testing that for any keys that aren't present in the FlatMap, `operator[]` returns the default value of `ValueType`.

src/axom/core/detail/FlatTable.hpp
publixsubfan: I'll take a look.

src/axom/core/tests/core_flatmap_for_all.hpp
publixsubfan: Added testing for CUDA and HIP.

PR #1610, opened 2025-07-09
FlatMap: add method for batched GPU construction

publixsubfan:
# Summary

Adds a method `FlatMap::create()` for constructing a flat hash map over a batch of corresponding key-value pairs.
  - A template parameter `ExecSpace` is used to specify whether the batched construction happens on the CPU, GPU, or over multiple threads via OpenMP. The passed-in `allocator` argument must be accessible from the specified execution space.
  - If two equivalent keys are inserted, the key-value pair with the higher index is selected.

Also adds a benchmark example for `FlatMap`, which tests the performance of both batched insertion and lookups.

### DIFF
--- a/src/axom/core/FlatMap.hpp
+++ b/src/axom/core/FlatMap.hpp
@@ -726,7 +726,7 @@ FlatMap<KeyType, ValueType, Hash>::FlatMap(IndexType bucket_count, Allocator all
   // TODO: we should add a countl_zero overload for 64-bit integers
   {
     std::int32_t numGroups = std::ceil((bucket_count + 1) / (double)BucketsPerGroup);
-    m_numGroups2 = 31 - (axom::utilities::countl_zero(numGroups));
+    m_numGroups2 = 32 - (axom::utilities::countl_zero(numGroups));
   }
 
   IndexType numGroupsRounded = 1 << m_numGroups2;
--- a/src/axom/core/FlatMap.hpp
+++ b/src/axom/core/FlatMap.hpp
@@ -613,6 +613,11 @@ public:
   ConstView view() const;
   /// }@
 
+  template <typename ExecSpace>
+  static FlatMap create(axom::ArrayView<KeyType> keys,
+                        axom::ArrayView<ValueType> values,
+                        Allocator allocator = Allocator {});
+
 private:
   friend class FlatMapView<KeyType, ValueType, false, Hash>;
   friend class FlatMapView<KeyType, ValueType, true, Hash>;
@@ -859,5 +864,7 @@ auto FlatMap<KeyType, ValueType, Hash>::erase(const_iterator pos) -> iterator
 }
 
 }  // namespace axom
+
+#include "FlatMapUtil.hpp"
 
 #endif  // Axom_Core_FlatMap_HPP
--- a/src/axom/core/FlatMap.hpp
+++ b/src/axom/core/FlatMap.hpp
@@ -613,6 +613,25 @@ public:
   ConstView view() const;
   /// }@
 
+  /*!
+   * \brief Constructs and returns a FlatMap given a set of key-value pairs.
+   *
+   *  Duplicate keys are handled by selecting the last value in the values
+   *  array corresponding to the equivalent key.
+   *
+   * \param keys   [in] array of keys for the pairs to insert
+   * \param values [in] array of values for the pairs to insert
+   * \param allocator [in] allocator to use for the constructed FlatMap
+   *
+   * \tparam ExecSpace the execution space in which to perform the batched
+   *                   construction
+   *
+   * \return the constructed FlatMap
+   *
+   * \pre keys.size() == values.size()
+   * \pre {keys, values}.getAllocatorID() is accessible from ExecSpace
+   * \pre allocator is accessible from ExecSpace
+   */
   template <typename ExecSpace>
   static FlatMap create(axom::ArrayView<KeyType> keys,
                         axom::ArrayView<ValueType> values,
--- a/src/axom/core/FlatMap.hpp
+++ b/src/axom/core/FlatMap.hpp
@@ -602,7 +602,7 @@ public:
    *
    * \param count the number of elements to fit without a rehash
    */
-  void reserve(IndexType count) { rehash(std::ceil(count / MAX_LOAD_FACTOR)); }
+  void reserve(IndexType count) { rehash(count); }
 
   /*!
    * \brief Returns a read-only view of the FlatMap.
@@ -739,13 +739,13 @@ FlatMap<KeyType, ValueType, Hash>::FlatMap(IndexType bucket_count, Allocator all
   , m_loadCount(0)
 {
   IndexType minBuckets = MIN_NUM_BUCKETS;
-  bucket_count = axom::utilities::max(minBuckets, bucket_count);
+  bucket_count = axom::utilities::max<IndexType>(minBuckets, bucket_count / MAX_LOAD_FACTOR);
   // Get the smallest power-of-two number of groups satisfying:
   // N * GroupSize - 1 >= minBuckets
   // TODO: we should add a countl_zero overload for 64-bit integers
   {
     std::int32_t numGroups = std::ceil((bucket_count + 1) / (double)BucketsPerGroup);
-    m_numGroups2 = 32 - (axom::utilities::countl_zero(numGroups));
+    m_numGroups2 = 32 - (axom::utilities::countl_zero(numGroups - 1));
   }
 
   IndexType numGroupsRounded = 1 << m_numGroups2;
--- a/src/axom/core/FlatMapUtil.hpp
+++ b/src/axom/core/FlatMapUtil.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2024, Lawrence Livermore National Security, LLC and
+// Copyright (c) 2017-2025, Lawrence Livermore National Security, LLC and
 // other Axom Project Developers. See the top-level COPYRIGHT file for details.
 //
 // SPDX-License-Identifier: (BSD-3-Clause)
--- a/src/axom/core/FlatMapUtil.hpp
+++ b/src/axom/core/FlatMapUtil.hpp
@@ -1,0 +1,189 @@
+// Copyright (c) 2017-2024, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#ifndef Axom_Core_FlatMap_Util_HPP
+#define Axom_Core_FlatMap_Util_HPP
+
+#include "axom/config.hpp"
+#include "axom/core/FlatMap.hpp"
+
+namespace axom
+{
+namespace detail
+{
+
+struct SpinLock
+{
+  int value {0};
+
+  AXOM_HOST_DEVICE bool tryLock()
+  {
+    int still_locked = 0;
+#if defined(__HIP_DEVICE_COMPILE__)
+    still_locked = __hip_atomic_exchange(&value, 1, __ATOMIC_ACQUIRE, __HIP_MEMORY_SCOPE_AGENT);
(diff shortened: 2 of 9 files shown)


src/axom/core/FlatMap.hpp
```diff
@@ -721,7 +745,7 @@ FlatMap<KeyType, ValueType, Hash>::FlatMap(IndexType bucket_count, Allocator all
   // TODO: we should add a countl_zero overload for 64-bit integers
```
kennyweiss, line 745: 👍

src/axom/core/FlatMap.hpp
```diff
@@ -721,7 +745,7 @@ FlatMap<KeyType, ValueType, Hash>::FlatMap(IndexType bucket_count, Allocator all
...
   {
     std::int32_t numGroups = std::ceil((bucket_count + 1) / (double)BucketsPerGroup);
-    m_numGroups2 = 31 - (axom::utilities::countl_zero(numGroups));
+    m_numGroups2 = 32 - (axom::utilities::countl_zero(numGroups));
```
kennyweiss: This change seems subtle/hard won. Do we have a unit test targeting this line?

src/axom/core/FlatMapUtil.hpp
```diff
@@ -0,0 +1,210 @@
...
+namespace detail
+{
+
+struct SpinLock
+{
+  int value {0};
+
+  AXOM_HOST_DEVICE bool tryLock()
+  {
+    int still_locked = 0;
```
kennyweiss, line 24: Any chance the axom atomics can be used/updated to handle/help with this logic?
(Mostly b/c that could harden the axom atomics. If you think this is a one-off and not useful elsewhere, it's fine as is)

src/axom/core/FlatMapUtil.hpp
```diff
@@ -0,0 +1,210 @@
...
+
+template <typename KeyType, typename ValueType, typename Hash>
+template <typename ExecSpace>
+auto FlatMap<KeyType, ValueType, Hash>::create(ArrayView<KeyType> keys,
+                                               ArrayView<ValueType> values,
+                                               Allocator allocator) -> FlatMap
+{
+  assert(keys.size() == values.size());
+
+  IndexType num_elems = keys.size();
```
kennyweiss: Thanks for all the comments!

Suggestion:

This code is fairly complex. Marking const variables as `const` would help a bit w/ readability/understandability/correctness:

```suggestion
  const IndexType num_elems = keys.size();
```

src/axom/core/FlatMapUtil.hpp
```diff
@@ -0,0 +1,210 @@
...
+  key_index_dedup_vec.resize(num_groups * GroupBucket::Size, -1);
+  const auto key_index_dedup = key_index_dedup_vec.view();
+
+  // Map k-v pair indices to bucket slots. This is essentially the inverse of
+  // the above mapping.
+  Array<IndexType> key_index_to_bucket_vec(num_elems, num_elems, allocator.getID());
+  const auto key_index_to_bucket = key_index_to_bucket_vec.view();
+
+  for_all<ExecSpace>(
+    keys.size(),
```
kennyweiss: Should this be `num_elems`?

src/axom/core/FlatMapUtil.hpp
```diff
@@ -0,0 +1,210 @@
...
+auto FlatMap<KeyType, ValueType, Hash>::create(ArrayView<KeyType> keys,
+                                               ArrayView<ValueType> values,
+                                               Allocator allocator) -> FlatMap
+{
+  assert(keys.size() == values.size());
+
+  IndexType num_elems = keys.size();
+
+  FlatMap new_map(allocator);
+  new_map.reserve(num_elems);
```
kennyweiss, line 68: Would the hashmap be more efficient w/ some extra room (e.g. a load factor of 1.25 or 1.5)?

src/axom/core/detail/FlatTable.hpp
```diff
@@ -120,6 +120,24 @@ struct GroupBucket
     return InvalidSlot;
   }
 
+  template <typename Func>
+  AXOM_HOST_DEVICE int visitHashOrEmptyBucket(std::uint8_t hash, Func&& visitor) const
```
kennyweiss, line 192: Please document this function

In particular, what sort of functor is this function expecting? It seems like it calls `visitor` on all buckets until it finds an empty one, and then eventually either returns the id of an empty bucket or `InvalidSlot`

src/axom/core/tests/core_flatmap_for_all.hpp
```diff
@@ -206,3 +206,186 @@ AXOM_TYPED_TEST(core_flatmap_forall, insert_and_modify)
...
+
+  // Check that every element we inserted is in the map
+  for(int i = 0; i < NUM_ELEMS; i++)
+  {
+    auto expected_key = this->getKey(i);
+    auto expected_val1 = this->getValue(i * 10.0 + 5.0);
+    auto expected_val2 = this->getValue(i * 10.0 + 7.0);
+    EXPECT_EQ(1, test_map.count(expected_key));
+    EXPECT_TRUE((test_map.at(expected_key) == expected_val1) ||
+                (test_map.at(expected_key) == expected_val2));
```
kennyweiss: The docs for `FlatMap::create()` say that 
```
   *  Duplicate keys are handled by selecting the last value in the values
   *  array corresponding to the equivalent key.
```

So, based on the way the data is initialized, should it always be `expected_val2`?

src/axom/core/tests/core_flatmap_for_all.hpp
```diff
@@ -206,3 +206,186 @@ AXOM_TYPED_TEST(core_flatmap_forall, insert_and_modify)
...
+
+  const int NUM_ELEMS = 100;
+
+  axom::Array<int> keys_vec(NUM_ELEMS * 2);
+  axom::Array<double> values_vec(NUM_ELEMS * 2);
+  // Create batch of array elements
+  for(int i = 0; i < NUM_ELEMS; i++)
+  {
+    auto key = this->getKey(i);
+    auto value = this->getValue(i * 10.0 + 5.0);
```
kennyweiss, line 268: Would it make sense to have a test that has repeated value entries?

I'd expect it to be handled properly, but might be good to add a test for it anyway.

src/axom/core/tests/core_flatmap_for_all.hpp
```diff
@@ -206,3 +206,186 @@ AXOM_TYPED_TEST(core_flatmap_forall, insert_and_modify)
...
+  }
+}
+
+template <typename KeyType>
+struct ConstantHash
+{
+  using argument_type = KeyType;
+  using result_type = axom::IndexType;
+
+  AXOM_HOST_DEVICE axom::IndexType operator()(KeyType) const { return 0; }
```
kennyweiss, line 345: Please add a comment about this test.

If the hash is always zero, I suppose that means there will be collisions for every element, so it is testing the worst case performance (?)

src/axom/core/examples/core_flatmap_perf.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include "axom/core/Array.hpp"
+#include "axom/core/FlatMap.hpp"
+#include "axom/core/FlatMapView.hpp"
+#include "axom/core/execution/for_all.hpp"
+#include "axom/core/execution/runtime_policy.hpp"
+#include "axom/core/utilities/Timer.hpp"
+
+#include "axom/fmt.hpp"
+#include "axom/CLI11.hpp"
+
```
kennyweiss, line 20: Please add a brief note about what this example demonstrates.

src/axom/core/examples/core_flatmap_perf.cpp
kennyweiss, line 1: Nice example!

kennyweiss (approved):
Thanks @publixsubfan !

Could you please post some performance results?

Please don't forget to update the RELEASE-NOTES.

src/axom/core/FlatMapUtil.hpp
publixsubfan: I think adding this to the axom atomics would be dependent on support from within RAJA for atomics with memory ordering. Otherwise the logic to implement that might get a little nasty.

src/axom/core/FlatMapUtil.hpp
publixsubfan: Let me take a look, but the interface for `FlatMap::reserve(num_elems)` *should* be "reserve enough space for num_elems to be inserted without rehashing," so if it doesn't already allocate extra space, it's a bug and should be fixed. On insertions, we enforce a maximum load factor of 0.875.

src/axom/core/FlatMap.hpp
publixsubfan: Yeah, I added a new test here: https://github.com/LLNL/axom/blob/d64fa39c4229f7c21b1c246cab78ef9ca0284dc2/src/axom/core/tests/core_flatmap.hpp#L117-L129

src/axom/core/tests/core_flatmap_for_all.hpp
publixsubfan: That's it -- even with worst-case hash performance, we should still be able to construct a valid hash map and query it.

src/axom/core/FlatMapUtil.hpp
rhornung67: IIRC, RAJA default atomics don't support memory ordering. RAJA can be configured to use desul atomics, which do support memory ordering. Unfortunately, we only support using those through the original RAJA atomic interface and so we only provide a default we define: https://github.com/LLNL/RAJA/blob/develop/include/RAJA/policy/desul/atomic.hpp#L22.

We should revisit whether we want to switch to desul atomics by default in RAJA. I think the last time we discussed this, there were still some cases where RAJA atomics were faster than desul. If we did switch to desul by default (which is what Kokkos uses), then we could support the full desul interface. 

@publixsubfan let me know if you think we should go this route.

publixsubfan:
@kennyweiss -- here's some performance graphs for construction and querying. These were "scaled" to be node-to-node comparisons, meaning we multiplied the ATS-2/ATS-4 numbers for each run by 4 to account for the 4 sockets. For CTS-2, we ran 2 MPI ranks with `--cpu-bind=socket` and 112 threads each, and summed the results for each run.

image
image

image

src/axom/core/FlatMapUtil.hpp
publixsubfan: Maybe we could play around with a partial desul default? Something like "default for ordered atomics, but use the original backend for unordered"

publixsubfan: I did have a PR for the ordered atomics here: https://github.com/LLNL/RAJA/pull/1616, if we wanted to try and clean that up.

src/axom/core/detail/FlatTable.hpp
publixsubfan: Done.

src/axom/core/examples/core_flatmap_perf.cpp
publixsubfan: Done.

src/axom/core/FlatMapUtil.hpp
publixsubfan: Done

src/axom/core/FlatMapUtil.hpp
publixsubfan: Done

src/axom/core/FlatMapUtil.hpp
kennyweiss: Thanks -- since this is somewhat of a one-off and it's not super easy to consolidate it into `axom::atomics`, I think it's fine as is.

src/axom/core/tests/core_flatmap_for_all.hpp
publixsubfan: That's right. I changed it to test that the second value wins out.

PR #1611, opened 2025-07-10
Fixes several bugs related to undefined behavior

kennyweiss:
# Summary

- This PR is a bugfix associated with undefined behavior as identified by ubsan in our clang configuration
- In particular:
  - It fixes invalid shifts in the BVH construction. The Morton indexing assumed that the `MortonIndexType` was unsigned, but was being passed in a signed integer. I changed this to an unsigned type and added a `static_assert` to `spin::MortonBase`
  - It fixes possible invalid `spin:UniformGrid` initialization when passed geometry with invalid bounding boxes. We now skip those cases
  - It fixes a validity check in `slam::OrderedSet` when presented with a stride of zero. There was previously a divide by zero.
  - It fixes checks on empty `axom::MapCollection` instances when using sparsehash, which expected a sentinel key (even for empty hashtables). We now check if the table is empty before checking for items. This is used heavily by `sidre` and was reporting undefined behavior in user applications.

### DIFF
--- a/src/axom/core/MapCollection.hpp
+++ b/src/axom/core/MapCollection.hpp
@@ -179,8 +179,15 @@ public:
   ///
   bool hasItem(const std::string& name) const
   {
-    typename MapType::const_iterator mit = m_name2idx_map.find(name);
-    return (mit != m_name2idx_map.end() ? true : false);
+    if(m_name2idx_map.empty())
+    {
+      return false;
+    }
+    else
+    {
+      typename MapType::const_iterator mit = m_name2idx_map.find(name);
+      return (mit != m_name2idx_map.end() ? true : false);
+    }
   }
 
   ///
@@ -193,15 +200,29 @@ public:
   ///
   T* getItem(const std::string& name)
   {
-    typename MapType::iterator mit = m_name2idx_map.find(name);
-    return (mit != m_name2idx_map.end() ? m_items[mit->second] : nullptr);
+    if(m_name2idx_map.empty())
+    {
+      return nullptr;
+    }
+    else
+    {
+      typename MapType::iterator mit = m_name2idx_map.find(name);
+      return (mit != m_name2idx_map.end() ? m_items[mit->second] : nullptr);
+    }
   }
 
   ///
   T const* getItem(const std::string& name) const
   {
-    typename MapType::const_iterator mit = m_name2idx_map.find(name);
-    return (mit != m_name2idx_map.end() ? m_items[mit->second] : nullptr);
+    if(m_name2idx_map.empty())
+    {
+      return nullptr;
+    }
+    else
+    {
+      typename MapType::const_iterator mit = m_name2idx_map.find(name);
+      return (mit != m_name2idx_map.end() ? m_items[mit->second] : nullptr);
+    }
   }
 
   ///
@@ -226,8 +247,15 @@ public:
   ///
   IndexType getItemIndex(const std::string& name) const
   {
-    typename MapType::const_iterator mit = m_name2idx_map.find(name);
-    return (mit != m_name2idx_map.end() ? mit->second : InvalidIndex);
+    if(m_name2idx_map.empty())
+    {
+      return InvalidIndex;
+    }
+    else
+    {
+      typename MapType::const_iterator mit = m_name2idx_map.find(name);
+      return (mit != m_name2idx_map.end() ? mit->second : InvalidIndex);
+    }
   }
 
   ///
--- a/src/axom/core/MapCollection.hpp
+++ b/src/axom/core/MapCollection.hpp
@@ -185,7 +185,7 @@ public:
     }
     else
     {
-      typename MapType::const_iterator mit = m_name2idx_map.find(name);
+      const auto mit = m_name2idx_map.find(name);
       return (mit != m_name2idx_map.end() ? true : false);
     }
   }
@@ -193,8 +193,8 @@ public:
   ///
   bool hasItem(IndexType idx) const
   {
-    return (idx >= 0 && static_cast<unsigned>(idx) < m_items.size() &&
-            m_items[static_cast<unsigned>(idx)]);
+    return (idx >= 0 && static_cast<std::size_t>(idx) < m_items.size() &&
+            m_items[static_cast<std::size_t>(idx)]);
   }
 
   ///
@@ -206,7 +206,7 @@ public:
     }
     else
     {
-      typename MapType::iterator mit = m_name2idx_map.find(name);
+      const auto mit = m_name2idx_map.find(name);
       return (mit != m_name2idx_map.end() ? m_items[mit->second] : nullptr);
     }
   }
@@ -220,7 +220,7 @@ public:
     }
     else
     {
-      typename MapType::const_iterator mit = m_name2idx_map.find(name);
+      const auto mit = m_name2idx_map.find(name);
       return (mit != m_name2idx_map.end() ? m_items[mit->second] : nullptr);
     }
   }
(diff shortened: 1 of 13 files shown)


src/axom/core/MapCollection.hpp
```diff
@@ -179,8 +179,15 @@ class MapCollection : public ItemCollection<T>
...
   {
-    typename MapType::const_iterator mit = m_name2idx_map.find(name);
-    return (mit != m_name2idx_map.end() ? true : false);
+    if(m_name2idx_map.empty())
+    {
+      return false;
+    }
+    else
+    {
+      typename MapType::const_iterator mit = m_name2idx_map.find(name);
```
BradWhitlock: Not too important but it would read better as "const auto mit".

BradWhitlock (approved):
Looks good.

src/axom/core/MapCollection.hpp
```diff
@@ -179,8 +179,15 @@ class MapCollection : public ItemCollection<T>
   ///
   bool hasItem(const std::string& name) const
   {
-    typename MapType::const_iterator mit = m_name2idx_map.find(name);
-    return (mit != m_name2idx_map.end() ? true : false);
+    if(m_name2idx_map.empty())
```
kennyweiss, line 186: These `empty()` checks prevent calling `m_name2idx_map.find()` and `m_name2idx_map.end()` on an empty hash table. Sparsehash assumes that we have set sentinel values for empty and deleted entries, and `MapCollection` only (currently) does this when inserting or deleting entries.

src/axom/slam/OrderedSet.hpp
```diff
@@ -262,8 +262,8 @@ struct OrderedSet : public policies::SetInterface<InterfacePolicy, PosType, Elem
         const double str = m_stride.stride();
         const auto diff = (m_rangeUpper - m_rangeLower);
 
-        // size is 0 if upper==lower, or signs of diff and stride differ
-        return (diff == 0 || ((diff > 0) != (str > 0))) ? 0 : ceil(diff / str);
+        // size is 0 if upper==lower, or stride is 0 or signs of diff and stride differ
+        return (diff == 0 || str == 0 || ((diff > 0) != (str > 0))) ? 0 : ceil(diff / str);
```
kennyweiss, line 266: This is a check for whether the stride is valid and previously triggered a divide by zero error in cases where the stride was zero. It now correctly returns that the stride is not valid.

src/axom/spin/MortonIndex.hpp
```diff
@@ -126,10 +126,14 @@ template <typename CoordType, typename MortonIndexType, typename Derived>
...
-  AXOM_STATIC_ASSERT_MSG(std::is_integral<MortonIndexType>::value,
-                         "MortonIndexType must be integral for Morton indexing");
+  static_assert(std::is_integral<CoordType>::value,
+                "Coordtype must be integral for Morton indexing");
+  static_assert(std::is_integral<MortonIndexType>::value,
+                "MortonIndexType must be integral for Morton indexing");
+
+  // we get undefined behavior w/ our bit shifts if using signed types
+  static_assert(std::is_unsigned<MortonIndexType>::value,
+                "MortonIndexType must be an unsigned type");
```
kennyweiss, line 136: This static_assert ensures that the `MortonIndexType` is unsigned. The only case within axom where it was a signed type was the BVH construction, which was triggering undefined behavior in a user code.

src/axom/spin/SparseOctreeLevel.hpp
```diff
@@ -40,13 +40,13 @@ template <typename CoordType, int DIM, typename BroodDataType, typename Represen
 struct BroodRepresentationTraits
 {
   using GridPt = primal::Point<CoordType, DIM>;
-  using PointRepresenationType = RepresentationType;
+  using PointRepresentationType = RepresentationType;
```
kennyweiss, line 43: Minor: Fixes a typo in an internal type

src/axom/spin/UniformGrid.hpp
```diff
@@ -515,9 +514,15 @@ void UniformGrid<T, NDIMS, ExecSpace, StoragePolicy>::initialize(axom::ArrayView
   axom::for_all<ExecSpace>(
     bboxes.size(),
     AXOM_LAMBDA(IndexType idx) {
-      const BoxType bbox = bboxes[idx];
+      const BoxType& bbox = bboxes[idx];
+      if(!bbox.isValid())
+      {
+        return;
+      }
```
kennyweiss, line 521: This check ensures that we do not try to insert geometry associated with invalid bounding boxes into our `UniformGrid`

src/axom/spin/internal/linear_bvh/build_radix_tree.hpp
kennyweiss, line 1: @publixsubfan -- please double check that these changes are ok.
We're now using an unsigned Morton index type, and then casting to the expected integer.

It's possible that BVH should switch its underlying index to unsigned, but at the very least, this should not be worse than what we had before.

src/axom/spin/MortonIndex.hpp
```diff
@@ -147,7 +151,6 @@ struct MortonBase
    * so, e.g. in 2D, 6 == 0b0110 becomes 0b*0*1*1*0 == 0b00010100 == 20
    */
   AXOM_HOST_DEVICE
-  AXOM_SUPPRESS_UBSAN
```
kennyweiss, line 150: With the new static_assert (a few lines up) and the changes to the BVH generation, I was able to remove this attribute.

src/axom/core/MapCollection.hpp
kennyweiss: Thanks -- I updated all of them

src/axom/core/tests/core_device_hash.hpp
```diff
@@ -183,30 +183,30 @@ AXOM_TYPED_TEST(core_device_hash, hash_enum)
   }
 }
 
-namespace
+namespace axom_testing
```
kennyweiss, line 186: @publixsubfan -- gcc was warning about using a type from an anonymous namespace in the type that uses this lambda

src/axom/spin/internal/linear_bvh/build_radix_tree.hpp
publixsubfan: I'm fine with these changes, but just surprised that this was an issue. Do you know what the undefined behavior was?

src/axom/core/MapCollection.hpp
Arlie-Capps: Please consider a comment like 
`The check against empty() avoids calls to find() and end().  In July 2025 this avoided undefined behavior with empty MapCollection.`

I just know that when I have come across a construction like this, it looks like a road was carefully built around an innocent patch of dirt.  I find myself filled with doubt and I have to wonder what (minor?) monster lies beneath.

Arlie-Capps: We use this pattern several times in this file.  I think a comment here is sufficient, rather than flagging all usages---or  you could put a comment at subsequent occurrences pointing back to this one.

src/axom/spin/SparseOctreeLevel.hpp
Arlie-Capps: thank yuo

rhornung67 (approved):
Thank you @kennyweiss Very nice improvements!

src/axom/spin/internal/linear_bvh/build_radix_tree.hpp
kennyweiss: The message was:
> axom/spin/MortonIndex.hpp:155:19: runtime error: left shift of 16777433 by 8 places cannot be represented in type 'int'

Essentially, shifting into the sign bit is undefined behavior.

src/axom/core/MapCollection.hpp
kennyweiss: Done

PR #1614, opened 2025-07-15
Use axom::FlatMap in spin's Octree implementation

kennyweiss:
# Summary

- This PR is progress towards a GPU-capable InOutOctree implementation (#993)
- This first PR uses `axom::FlatMap` in the implementation of `spin::SparseOctreeLevel`
- It also cleans up and modernizes the Octree implementation a bit

#### Performance

Here is the performance difference I saw using `axom::FlatMap` instead of sparsehash in the implementation
in the InOutOctree `containment_driver` example.

|                   | Sphere | (1K tris)                     |                      | BoxedSphere     | (65K tris)                       |                      | Plane      | (373K tris)                     |                      |    Golf ball   | (246K tris)                     |                     |
|-------------------|:----------------------------------:|:--------------------:|:--------------------:|:-------------------------------:|:--------------------:|:--------------------:|:--------------------------:|:--------------------:|:--------------------:|:------------------------------:|:--------------------:|:--------------------:|
|                   | Construct                         | Query $64^3$         | Query $256^3$        | Construct                       | Query $64^3$         | Query $256^3$        | Construct                | Query $64^3$         | Query $256^3$        | Construct                      | Query $64^3$         | Query $256^3$        |
| **Using FlatMap** |                                   |                      |                      |                                 |                      |                      |                          |                      |                      |                                |                      |                      |
| Debug             | 0.153                             | 0.856                | --                   | 25.145                          | 0.700                | --                   | 116.298                  | 0.394                | --                   | --                             | --                   | --                   |
| Release           | 0.012                             | 0.155                | 9.187                | 1.500                           | 0.112                | 6.412                | 7.301                    | 0.027                | 1.608                | 3.181                          | 0.041                | 2.405                |
| **Using sparsehash** |                                |                      |                      |                                 |                      |                      |                          |                      |                      |                                |                      |                      |
| Debug             | 0.145                             | 0.763                | --                   | 25.833                          | 0.639                | --                   | 111.526                  | 0.412                | --                   | 41.878                         | 0.396                | --                   |
| Release           | 0.009                             | 0.126                | 7.405                | 1.163                           | 0.086                | 4.903                | 5.608                    | 0.027                | 1.364                | 2.536                          | 0.032                | 1.818                |
|                             |                        |                      |                      |                                 |                      |                      |                          |                      |                      |                                |                      |                      |
| **Diff FlatMap/Sparsehash** |                        |                      |                      |                                 |                      |                      |                          |                      |                      |                                |                      |                      |
[...]

PR #1616, opened 2025-07-16
FlatMap: Use SSE2 intrinstics

publixsubfan:
# Summary

When available, uses SSE2 operations for `GroupBucket::getEmptyBucket()` and `GroupBucket::visitHashBucket()`. This should accelerate performance of lookup and non-batched insertion operations on the CPU.

## Performance

image
image

We see a roughly 3x performance bump at small numbers of elements, which drops to 2x at the 100k-900k element count, and to 1.3-1.5x at 1M-9M elements.

### DIFF
--- a/src/axom/core/detail/FlatTable.hpp
+++ b/src/axom/core/detail/FlatTable.hpp
@@ -12,6 +12,21 @@
 #include "axom/core/ArrayView.hpp"
 #include "axom/core/utilities/BitUtilities.hpp"
 
+#if defined(_MSC_VER)
+  // MSVC does *not* define __SSE2__ to indicate SSE2 intrinsic support.
+  #if defined(_M_X64) || (defined(_M_IX86_FP) && _M_IX86_FP >= 2)
+    // Either:
+    //  - We're building for x86_64 (SSE2 is always available), or...
+    //  - We're building for a 32-bit platform with SSE2 support.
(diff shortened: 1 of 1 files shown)


src/axom/core/detail/FlatTable.hpp
```diff
@@ -12,6 +12,21 @@
 #include "axom/core/ArrayView.hpp"
 #include "axom/core/utilities/BitUtilities.hpp"
 
+#if defined(_MSC_VER)
```
kennyweiss, line 15: Do we need to do anything special to get the intrinsics, e.g. compile with the `-march` flag?

publixsubfan: My understanding is that nothing special is required if Axom is compiled as a 64-bit library, since support for up to SSE2 is a part of the x86-64 spec. For 32-bit x86 systems, you would need to specify `-march=...`.

kennyweiss: I confirmed with my changes from https://github.com/LLNL/axom/pull/1614 that I am using the new SSE2 intrinsics w/ the rzwhippet-clang host-config

kennyweiss (approved):
Thanks @publixsubfan

bmhan12 (approved):

rhornung67 (approved):

Issue #1393, opened 2024-08-07
Sample-based shaping from 2D triangle mesh


kennyweiss:
A user has requested that we extend our sample-based shaping capability to include shaping from a 2D STL triangle mesh, i.e. a triangle mesh where all the triangles lie in the XY plane.

One strategy might be to use a spatial index like a BVH tree to find the candidate triangles for a given point and then check `Triangle::checkInTriangle(Point)`.  We can use a similar strategy for In/Out queries on a Tetrahedral mesh in 3D.

A different strategy might be to extrude the 2D triangle mesh in the z-direction and use our existing In/Out capability for watertight 3D STL meshes.

https://github.com/LLNL/axom/blob/b5fa2217ed7a694952b415d1940d636cc366bb93/src/axom/primal/geometry/Triangle.hpp#L340-L356

Example meshes might look like:
```stl
solid STL generated by MeshLab
  facet normal  0.000000e+00  0.000000e+00  1.000000e+00
    outer loop
      vertex   4.832074e-01  1.920779e-01  0.000000e+00
      vertex   4.533584e-01  2.114626e-01  0.000000e+00
      vertex   4.585903e-01  1.868454e-01  0.000000e+00
    endloop
  endfacet
  facet normal  0.000000e+00  0.000000e+00  1.000000e+00
    outer loop
      vertex   4.832074e-01  1.920779e-01  0.000000e+00
      vertex   4.779753e-01  2.166951e-01  0.000000e+00
      vertex   4.533584e-01  2.114626e-01  0.000000e+00
    endloop
  endfacet
  facet normal  0.000000e+00 -0.000000e+00  1.000000e+00
    outer loop
      vertex   4.832074e-01  1.920779e-01  0.000000e+00
      vertex   5.000000e-01  1.956473e-01  0.000000e+00
      vertex   4.779753e-01  2.166951e-01  0.000000e+00
    endloop
  endfacet
  facet normal -0.000000e+00  0.000000e+00  1.000000e+00
    outer loop
      vertex   4.779753e-01  2.166951e-01  0.000000e+00
      vertex   5.000000e-01  1.956473e-01  0.000000e+00
      vertex   5.000000e-01  2.213766e-01  0.000000e+00
    endloop
  endfacet
  facet normal  0.000000e+00 -0.000000e+00  1.000000e+00
    outer loop
      vertex   4.533584e-01  2.114626e-01  0.000000e+00
      vertex   4.779753e-01  2.166951e-01  0.000000e+00
[...]

PR #1627, opened 2025-07-23
Klee: Adds an optional per-shape `dimensions` field to a shape's geometry specification

kennyweiss:
# Summary

- This feature PR adds an optional per-shape `dimensions` specification to the geometry of a shape in a Klee file
  - This can be used to override the global dimension
  - Among other use-cases, I'm planing to use it to enhance the flexibility of our STL mesh loader, to allow user bring in STL meshes modeling planar geometry (see #1393) in addition to current 3D boundary representations
- While working through this PR, I reorganized the Klee source directory -- moving files related to IO into the `klee/io` directory -- and cleaned up some doxygen comments in the Klee source files
- Minor: I also fixed one of the badges in the README file to point to our github actions instead of Azure CI

#### API changes
* Klee's `Dimensions` enum now has an `Unspecified` value
* I added helper functions `getInputDimensions()` and `getOuputDimensions()` to Klee's `Geometry` class to get the dimension of the shape before and after applying operators, respectively.

### DIFF
--- a/src/axom/klee/Geometry.hpp
+++ b/src/axom/klee/Geometry.hpp
@@ -3,16 +3,17 @@
 //
 // SPDX-License-Identifier: (BSD-3-Clause)
 
-#ifndef AXOM_KLEE_GEOMETRY_HPP
-#define AXOM_KLEE_GEOMETRY_HPP
-
-#include <memory>
-#include <string>
+#ifndef AXOM_KLEE_GEOMETRY_HPP_
+#define AXOM_KLEE_GEOMETRY_HPP_
 
 #include "axom/klee/Dimensions.hpp"
 #include "axom/klee/Units.hpp"
+
 #include "axom/primal.hpp"
 #include "axom/sidre.hpp"
+
+#include <memory>
+#include <string>
 
 namespace axom
 {
@@ -343,4 +344,4 @@ private:
 }  // namespace klee
 }  // namespace axom
 
-#endif  //AXOM_KLEE_GEOMETRY_HPP
+#endif  // AXOM_KLEE_GEOMETRY_HPP_
--- a/src/axom/klee/Geometry.hpp
+++ b/src/axom/klee/Geometry.hpp
@@ -21,9 +21,7 @@ namespace klee
 {
 class GeometryOperator;
 
-/**
- * Properties of a geometric object which can be transformed by operators
- */
+/// Properties of a geometric object which can be transformed by operators
 struct TransformableGeometryProperties
 {
   Dimensions dimensions;
@@ -51,9 +49,7 @@ inline bool operator!=(const TransformableGeometryProperties &lhs,
   return !(lhs == rhs);
 }
 
-/**
- * Represents the geometry specified in a Shape.
- */
+/// Represents the geometry specified in a Shape.
 class Geometry
 {
 public:
@@ -67,8 +63,7 @@ public:
   /**
    * Create a Geometry object based on a file representation.
    *
-   * \param startProperties the transformable properties before any
-   * operators are applied
+   * \param startProperties the transformable properties before any operators are applied
    * \param format the format of the file
    * \param path the path of the file
    * \param operator_ a possibly null operator to apply to the geometry.
@@ -81,8 +76,7 @@ public:
   /**
    * Create a Geometry object based on a blueprint tetrahedral mesh.
    *
-   * \param startProperties the transformable properties before any
-   * operators are applied
+   * \param startProperties the transformable properties before any operators are applied
    * \param simplexMeshGroup the geometry in blueprint format.
    *   The elements should be segments, triangles or tetrahedra.
    * \param topology The blueprint topology to use.
@@ -96,8 +90,7 @@ public:
   /**
    * Create a tetrahedron Geometry object.
    *
-   * \param startProperties the transformable properties before any
-   * operators are applied
+   * \param startProperties the transformable properties before any operators are applied
    * \param tet Tetrahedron
    * \param operator_ a possibly null operator to apply to the geometry.
    */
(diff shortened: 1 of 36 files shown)


src/axom/quest/examples/shaping_driver.cpp
```diff
@@ -642,6 +650,8 @@ int main(int argc, char** argv)
                                           shape.getMaterial(),
                                           shapeFormat)));
 
+    const klee::Dimensions shapeDim = shape.getGeometry().getInputDimensions();
```
kennyweiss, line 651: A key contribution of this PR is that, as we loop through the shapes in a Klee input, we can get the per-shape dimension, rather than being limited to the global dimension of the shapeset (`ShapeSet::getDimensions()`)

src/axom/quest/tests/quest_sampling_shaper.cpp
```diff
@@ -233,12 +233,13 @@ class SamplingShaperTest : public ::testing::Test
...
-    auto getShapeDim = [defaultDim = m_shapeSet->getDimensions()](const auto& shape) {
+    auto getShapeDim = [](const auto& shape) {
       static std::map<std::string, klee::Dimensions> format_dim = {{"c2c", klee::Dimensions::Two},
                                                                    {"stl", klee::Dimensions::Three}};
 
+      const auto& shape_dim = shape.getGeometry().getInputDimensions();
       const auto& format_str = shape.getGeometry().getFormat();
-      return format_dim.find(format_str) != format_dim.end() ? format_dim[format_str] : defaultDim;
+      return format_dim.find(format_str) != format_dim.end() ? format_dim[format_str] : shape_dim;
     };
```
kennyweiss, line 211: Some of our testing infrastructure (and user codes) hard-code the dimension of STL meshes as 3D and of C2C contours as 2D.  I intend to loosen this restriction when I work through #1393.

src/axom/klee/io/IO.cpp
```diff
@@ -26,16 +27,16 @@ namespace
...
 struct GeometryData
 {
   std::string format;
   std::string path;
-  LengthUnit startUnits;
-  LengthUnit endUnits;
-  Dimensions startDimensions;
-  bool dimensionsSet;
+  LengthUnit startUnits {LengthUnit::unspecified};
+  LengthUnit endUnits {LengthUnit::unspecified};
```
BradWhitlock, line 37: These 2 enums (LengthUnit, Dimensions) have different capitalizations. Maybe go with upper case in both.

src/axom/klee/io/IO.cpp
```diff
@@ -217,20 +225,25 @@ Geometry convert(GeometryData const &data,
...
+  const auto computed_end_dims = geometry.getEndProperties().dimensions;
+  const auto expected_end_dims = has_explicit_dims ? data.explicitDimensions : fileDimensions;
+  if(computed_end_dims != expected_end_dims)
   {
-    throw KleeError(
-      {data.pathInFile, "Did not end up in the number of dimensions specified by the file"});
+    throw KleeError({data.pathInFile,
+                     axom::fmt::format("Did not end up with the expected number of dimensions. "
+                                       "Expected: {}, but got: {}",
+                                       expected_end_dims == Dimensions::Two ? 2 : 3,
```
BradWhitlock: Is there any way the dimensions could still be `Unspecified`? Would an fmt formatter be appropriate?

src/axom/klee/Geometry.hpp
```diff
@@ -252,95 +243,77 @@ class Geometry
...
-   represents a hexahedron.
-  */
+  /// \brief Return the hex geometry, when the Geometry represents a hexahedron
   const axom::primal::Hexahedron<double, 3> &getHex() const { return m_hex; }
 
-  /**
-   @brief Return the sphere geometry, when the Geometry
-   represents an alalytical sphere.
-  */
+  /// \brief Return the sphere geometry, when the Geometry represents an alalytical sphere.
```
BradWhitlock: Typo "alalytical"

BradWhitlock (approved):
Looks good to me.

wnissen:
Thinking about the purpose of this PR (rather than the implementation), the original goal of the Klee format was to be as declarative, human-readable, and error-resistant as possible. Part of that was the `dimension` specification, so one couldn't accidentally include 2D and 3D shapes in the same file without explicitly changing the dimension, as shown here:
https://axom.readthedocs.io/en/develop/axom/klee/docs/sphinx/specifying_shapes.html#changing-dimensions-on-a-per-shape-basis

So is the main goal here to make it possible to specify all the shapes for both a 2D and 3D setup in a single file? I agree that it's time-consuming and error-prone to specify the name, region, file, etc., for all the shapes twice. Just trying to understand the tradeoffs here.

src/axom/klee/io/IO.cpp
kennyweiss: I agree that it's not ideal, but I wonder if fixing this is more trouble than it's worth
- since they're both `enum class`, we'd have to explicitly spell out `LengthUnit::unspecified` or `Dimensions::Unspecified` and misspelling it would cause a compiler error
- changing the spelling would cause any users to have to fix theirs
- a different option might be to add both cases, e.g. `Dimensions::Unspecified` and `Dimension::unspecified` and set their values equal to each other, but that might be even messier

The least disruptive change would be to make the new `Dimensions` value be lower case (which would be internally inconsistent, but perhaps not horrible, e.g.
```diff
enum class Dimensions : int
{
  Two = 2,
  Three = 3
- Unspecified = 0,
+ unspecified = 0,
};
```

Note: The `LengthUnit` values are all lower case
https://github.com/LLNL/axom/blob/daba41b2d862d633d6c2908765f2ff075de9c3ba/src/axom/klee/Units.hpp#L22-L37

src/axom/klee/io/IO.cpp
kennyweiss: Great suggestion!

I added a custom formatter for the enum.

In the process, I realized that `klee::Dimensions dim {}` would get a value of `0`, which would not match the current value for `Dimensions::Unspecified /* ==-1 */`, so I modified the value of `Dimensions::Unspecified` to be `0` instead of `-1`

src/axom/klee/Geometry.hpp
kennyweiss: done

rhornung67 (approved):

PR #1630, opened 2025-07-29
bugfix/capps2/fmtoperatorquotenospace

Arlie-Capps:
# Summary

- This PR is a bugfix
- It does the following:
  - Removes a space following `operator""` in our bundled copy of fmt
  - The presence of a space causes recent versions of clang to throw a warning of a deprecated feature (the space).  This PR cuts down on the volume of warnings in codes using Axom.

### DIFF
--- a/src/thirdparty/axom/fmt/xchar.h
+++ b/src/thirdparty/axom/fmt/xchar.h
@@ -49,7 +49,7 @@ constexpr format_arg_store<wformat_context, Args...> make_wformat_args(
 
 inline namespace literals {
 #if AXOM_FMT_USE_USER_DEFINED_LITERALS && !AXOM_FMT_USE_NONTYPE_TEMPLATE_ARGS
-constexpr detail::udl_arg<wchar_t> operator"" _a(const wchar_t* s, size_t) {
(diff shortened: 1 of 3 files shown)


kennyweiss (approved):
Thanks @Arlie-Capps 

Please add these changes to a/the patch files (in this directory) so it's easier for us to migrate these changes when we upgrade fmt.

Would it make sense to try to upstream this change?

src/thirdparty/axom/fmt/xchar.h
```diff
@@ -49,7 +49,7 @@ constexpr format_arg_store<wformat_context, Args...> make_wformat_args(
 
 inline namespace literals {
 #if AXOM_FMT_USE_USER_DEFINED_LITERALS && !AXOM_FMT_USE_NONTYPE_TEMPLATE_ARGS
-constexpr detail::udl_arg<wchar_t> operator"" _a(const wchar_t* s, size_t) {
+constexpr detail::udl_arg<wchar_t> operator""_a(const wchar_t* s, size_t) {
```
kennyweiss, line 52: Looks like this these are already fixed in upstream fmt:
https://github.com/search?q=repo%3Afmtlib%2Ffmt%20%22%22_a&type=code

PR #1632, opened 2025-08-01
Cleanup: Use std::variant in SamplingShaper

kennyweiss:
# Summary

- This PR is a cleanup
- It uses `C++17`'s  `std::variant` instead of a collection of `unique_ptr`s for the various sampler types (only one of which could be valid at a time).
- This will help us as we add more Samplers (e.g. 2D primitives, winding numbers, different exec policies, ...) in the future

### Gotchas
- One hiccup is that it doesn't work with out gcc@8.3.1 standard library on blueos. There seems to be a bug in that version's implementation of variant:
```
/usr/tce/packages/gcc/gcc-8.3.1/include/c++/8/variant:461:14: error: dependent using declaration resolved to type without 'typename'
using _Base::_Variant_storage;
             ^
```
  i.e. the using is missing a `typename`. I'm not sure if we can work around that directly.
- I will try to see if passing in the `GNU_ROOT` to a newer gcc fixes this, but I think it would also require our users to do the same


src/axom/mir/examples/tutorial_simple/mir_tutorial_simple.cpp
```diff
@@ -137,11 +124,15 @@ int main(int argc, char **argv)
...
-  if(retval != 0)
+  try
+  {
+    params.parse(argc, argv);
+  }
+  catch(const axom::CLI::ParseError &e)
   {
-    return retval;
+    return params.m_app.exit(e);
   }
```
kennyweiss, line 134: @BradWhitlock -- I noticed that calling `./mir_tutorial_simple -h` did not exit the application after printing out the help message. This change fixes it.

src/axom/primal/geometry/NURBSPatch.hpp
kennyweiss, line 1: These changes were due to an error message from the updated blueos compilers. 

Seems like the other compilers were able to deduce the type of the 1D `axom::ArrayView`s  without the explicit template parameter.

src/axom/mir/examples/tutorial_simple/mir_tutorial_simple.cpp
BradWhitlock: @kennyweiss - thanks for adding the exception handling. I thought CUDA kernels *only* permitted accessing device memory. Why would this only be showing up now? And why would earlier compilers not generate code that failed if your suspicion is true?

Arlie-Capps (approved):
This is a definite improvement.  Thanks, Kenny.

src/axom/mir/examples/tutorial_simple/mir_tutorial_simple.cpp
kennyweiss: Those are all good questions. Presumably something changed from cuda@11.2 to cuda@11.8 that affected only these tests (?)  I'll dig deeper and let you know what I discover.

host-configs/rzansel-blueos_3_ppc64le_ib_p9-clang@16.0.6.1.cmake
gunney1, line 1: I assume these host config files will be deleted when merged, since rzansel is gone.

kennyweiss: Done.

Issue #1615, opened 2025-07-15
Windows 64-bit CI build errors


bmhan12:
I was testing the Windows build while trying out C++17 in Axom, and noticed `develop` was currently failing on both the 32-bit and 64-bit Windows images: https://github.com/LLNL/axom/actions/runs/16276700045

One of the 64-bit errors is related to overload resolution with atomics. I've attached it below, related to `FlatTable.hpp` from the 64-bit image:

<details>
<summary>Dropdown of the atomic error</summary>

```
2025-07-14T21:43:32.9013385Z D:\a\axom\axom\uberenv_libs\vcpkg\installed\x64-windows\include\RAJA\policy\atomic_builtin.hpp(614,16): error C2665: 'RAJA::detail::builtin_atomicCAS': no overloaded function could convert all the argument types [D:\a\axom\axom\build_axom\axom\core\tests\core_serial_tests.vcxproj]
2025-07-14T21:43:32.9015603Z   (compiling source file '../../../../src/axom/core/tests/core_serial_main.cpp')
2025-07-14T21:43:32.9016577Z       D:\a\axom\axom\uberenv_libs\vcpkg\installed\x64-windows\include\RAJA\policy\atomic_builtin.hpp(199,23):
2025-07-14T21:43:32.9017544Z       could be '__int64 RAJA::detail::builtin_atomicCAS(__int64 *,__int64,__int64)'
2025-07-14T21:43:32.9018493Z           D:\a\axom\axom\uberenv_libs\vcpkg\installed\x64-windows\include\RAJA\policy\atomic_builtin.hpp(614,16):
2025-07-14T21:43:32.9019696Z           '__int64 RAJA::detail::builtin_atomicCAS(__int64 *,__int64,__int64)': cannot convert argument 1 from 'T *' to '__int64 *'
2025-07-14T21:43:32.9020474Z           with
2025-07-14T21:43:32.9020733Z           [
2025-07-14T21:43:32.9020974Z               T=uint8_t
2025-07-14T21:43:32.9021258Z           ]
2025-07-14T21:43:32.9021874Z               D:\a\axom\axom\uberenv_libs\vcpkg\installed\x64-windows\include\RAJA\policy\atomic_builtin.hpp(614,34):
2025-07-14T21:43:32.9023152Z               Types pointed to are unrelated; conversion requires reinterpret_cast, C-style cast or parenthesized function-style cast
[...]

Issue #1580, opened 2025-05-21
Fix `x86-windows` TPL build


kennyweiss:
The build is currently failing due to our use of RAJA's `atomicMin`/`atomicMax` w/ `doubles`, which relies on an intrinsic that is not available in 32-bit configurations on Windows.

See: https://github.com/LLNL/RAJA/issues/1848

BradWhitlock:
The idea was to add workarounds for double into the new wrappers for ``axom::atomicMin`` and ``axom::atomicMax`` (https://github.com/LLNL/axom/blob/3666e7888c7542c50ab4177044089da539b1c933/src/axom/core/execution/atomics.hpp#L51)

kennyweiss:
Discussing this w/ @bmhan12 -- it seems likely that disabling RAJA in our x86 Windows configuration would fix the build.

PR #1635, opened 2025-08-08
64-bit Windows RAJA unsigned atomics test

bmhan12:
This draft PR:
- Tests RAJA branch LLNL/RAJA#1870 to see if it fixes issue #1615
  - Confirms that the branch does fix the 64-bit Window builds. 
- Windows GHA build for this PR can be found at : https://github.com/LLNL/axom/actions/runs/16816563826
  - Note: The 32-bit x86 Window builds are failing due to #1580, which is a separate issue from the RAJA fix.
- Some Axom code changes were necessary to address MPI guards/OpenMP usage in `quest_shape_in_memory_ex` tests.

### DIFF
--- a/src/axom/quest/examples/CMakeLists.txt
+++ b/src/axom/quest/examples/CMakeLists.txt
@@ -326,8 +326,12 @@ if((CONDUIT_FOUND OR
         )
 
     if(AXOM_ENABLE_TESTS AND AXOM_DATA_DIR)
-        # 3D shaping tests.  No 2D shaping in this feature, at least for now.
-        set(_nranks 1)
+
+        if(AXOM_ENABLE_MPI)     # this test requires MPI when available
+            set(_nranks 1)      # (but only 1 rank)
+        else()                  # but can otherwise be run without MPI
+            unset(_nranks)
+        endif()
 
         # Run the in-memory shaping example on with each enabled policy
         set(_policies "seq")
--- a/src/axom/quest/examples/CMakeLists.txt
+++ b/src/axom/quest/examples/CMakeLists.txt
@@ -336,7 +336,10 @@ if((CONDUIT_FOUND OR
         # Run the in-memory shaping example on with each enabled policy
         set(_policies "seq")
         if(RAJA_FOUND)
-            blt_list_append(TO _policies ELEMENTS "omp"  IF AXOM_ENABLE_OPENMP)
+            # Hanging issue on Windows
+            if(NOT WIN32)
+                blt_list_append(TO _policies ELEMENTS "omp"  IF AXOM_ENABLE_OPENMP)
+            endif()
             blt_list_append(TO _policies ELEMENTS "cuda" IF AXOM_ENABLE_CUDA)
             blt_list_append(TO _policies ELEMENTS "hip"  IF AXOM_ENABLE_HIP)
         endif()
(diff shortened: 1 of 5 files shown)


src/axom/quest/examples/CMakeLists.txt
```diff
@@ -326,13 +326,20 @@ if((CONDUIT_FOUND OR
...
+        endif()
 
         # Run the in-memory shaping example on with each enabled policy
         set(_policies "seq")
         if(RAJA_FOUND)
-            blt_list_append(TO _policies ELEMENTS "omp"  IF AXOM_ENABLE_OPENMP)
+            # Hanging issue on Windows
+            if(NOT WIN32)
+                blt_list_append(TO _policies ELEMENTS "omp"  IF AXOM_ENABLE_OPENMP)
+            endif()
```
kennyweiss, line 366: Do you have more information about the hang?
Could/should this be fixed?
Should we create an issue for it?

bmhan12: Unfortunately not much else outside of the Windows GHA build log.
This fix/workaround is necessary to get the Windows 64-bit GHA jobs passing, and I've updated the original issue #1615 about these test failures.

This draft PR is mainly to highlight that the RAJA atomics branch does fix our issue. The vcpkg changes here are pretty hasty/hacky, so I do not think they should get merged.

~~The non-vcpkg config, `quest_shape_in_memory_ex` changes I will make into a separate PR.~~
EDIT: Don't think this is necessary. Can be done in this PR

kennyweiss: Thanks -- do you think we should update vcpkg to RAJA@develop after the RAJA PR merges? Or should we add a patch to our RAJA?

Since we maintain our own vcpkg RAJA package, I think either option should be ok, and it'll be good to fix our Windows TPL CI.

bmhan12: > Thanks -- do you think we should update vcpkg to RAJA@develop after the RAJA PR merges? 

I think the first/former option is better. 
At first, I wasn't certain if it was possible for vcpkg to track a commit instead of a branch or tag (wasn't reading the documentation correctly :satisfied:)

src/axom/sina/tests/sina_Document.cpp
```diff
@@ -807,7 +807,10 @@ TEST(Document, test_appendOrderedCurvesToJson)
   doAppendOrderedCurveTest("json", appendDocumentToJson);
 }
 
-#ifdef AXOM_USE_HDF5
+// Windows error -
+// unknown file: error: C++ exception with description "Could not save to 'test.hdf5':
+// iostream stream error" thrown in the test body.
+#if defined(AXOM_USE_HDF5) && !defined(WIN32)
 TEST(Document, test_appendOrderedCurvesToHDF5)
```
bmhan12: I encountered a file stream related error with one of the Sina tests.
It is possibly due to a difference in behavior between Windows and Linux for file systems, I disabled the test on Windows for now.

BradWhitlock: I think the problem is [here in sina::Document.cpp](https://github.com/LLNL/axom/blob/e4fdc79c4a7fe0dd11ce45c9a87cabddef949905/src/axom/sina/core/Document.cpp#L362)  where a file is being renamed. Since HDF5 was not present, the file being renamed was not written in the first place.

bmhan12: Test fixed/renabled by using a different file name to pass to `saveDocument`(test.hdf5 --> test_curve.hdf5)

PR #1638, opened 2025-08-12
FlatMap: Iterator behavior fixes

### LINKED ISSUE #1618: FlatMap iterators should be DefaultConstructible
When using `axom::FlatMap` in the InOutOctree, I found it to be nearly a drop-in replacement for `std::unordered_map` and `sparsehash`. The only missing feature was that I couldn't use a default constructible FlatMap iterator.

See: #1614

publixsubfan:
# Summary

- Enables default-constructibility of `FlatMap` and `FlatMapView` iterators
- This change also fixes a potential bug where iterators referring to two different flat map instances might compare equivalent

Resolves #1618.


src/axom/core/FlatMapView.hpp
```diff
@@ -156,8 +158,20 @@ class FlatMapView : detail::flat_map::SequentialLookupPolicy<typename Hash::resu
...
 private:
-  IndexType m_numGroups2;
-  IndexType m_size;
+  /*!
+   * \brief Returns whether another FlatMapView points to the same base FlatMap
+   *  instance as this FlatMapView.
+   *
+   *  This is intended as a helper function for iterator comparisons.
+   */
+  AXOM_HOST_DEVICE bool viewToSameMap(const FlatMapView& other) const
```
adayton1: I would rename this to `isViewOfSameMap` or `viewHasSameMap`.

src/axom/core/FlatMap.hpp
```diff
@@ -689,6 +689,8 @@ class FlatMap<KeyType, ValueType, Hash>::IteratorImpl
   using reference = DataType&;
 
 public:
+  IteratorImpl() : m_map(nullptr), m_internalIdx(0) { }
+
   IteratorImpl(MapConstType* map, IndexType internalIdx) : m_map(map), m_internalIdx(internalIdx)
   {
     assert(m_internalIdx >= 0 && m_internalIdx <= m_map->bucket_count());
```
kennyweiss, line 697: This should probably also have a `m_map != nullptr`

src/axom/core/FlatMap.hpp
```diff
@@ -689,6 +689,8 @@ class FlatMap<KeyType, ValueType, Hash>::IteratorImpl
   using reference = DataType&;
 
 public:
+  IteratorImpl() : m_map(nullptr), m_internalIdx(0) { }
```
kennyweiss: Suggestion: 
If we change the member variables to 
```cpp
private:
  MapConstType* m_map {nullptr};
  IndexType m_internalIdx {0};
```
then this could be:
```cpp
   IteratorImpl() = default;
```

src/axom/core/FlatMapView.hpp
```diff
@@ -201,12 +217,12 @@ class FlatMapView<KeyType, ValueType, IsConst, Hash>::IteratorImpl
...
   AXOM_HOST_DEVICE friend bool operator==(const IteratorImpl& lhs, const IteratorImpl& rhs)
   {
-    return lhs.m_internalIdx == rhs.m_internalIdx;
+    return (lhs.m_map.viewToSameMap(rhs.m_map) && lhs.m_internalIdx == rhs.m_internalIdx);
   }
 
   AXOM_HOST_DEVICE friend bool operator!=(const IteratorImpl& lhs, const IteratorImpl& rhs)
   {
-    return lhs.m_internalIdx != rhs.m_internalIdx;
+    return (!lhs.m_map.viewToSameMap(rhs.m_map) || lhs.m_internalIdx != rhs.m_internalIdx);
```
kennyweiss: To ensure consistency, should this just be 
```suggestion
    return lhs != rhs;
```

kennyweiss (approved):

PR #1642, opened 2025-08-20
Enhance SamplingShaper so it can work with contours stored in MFEM meshes

BradWhitlock:
This PR adds support for shaping using MFEM files that contain 1D curves in 2D space, much like we do for C2C.

* Curve linearization was moved from C2C reader into `LinearizeCurves` class and used for C2C and MFEM data.
* `MFEMReader` class was added to read MFEM curve as array of `NURBSCurve` or `CurvedPolygon`.
* Some functions for MFEM curve to primal type were made more accessible
* `SamplingShaper` was enhanced so it can read MFEM files as either `NURBSCurve` or `CurvedPolygon`
* New `WindingNumberSampler` class uses `CurvedPolygons` to determine in/out for quad points from `SamplingShaper`
* `SamplingShaper` can use winding number sampling (on CurvedPolygons) or InOut sampling on MFEM discretized curves (same as how C2C works)
* Shaping driver has new `--sampling` argument to select between "inout" and "windingnumber" sampling. The latter only works for MFEM files currently.
* data directory contains new "heroic_roses" contours as MFEM
* Added a script for plotting the shaping output in VisIt and getting colors right for heroic_roses image


src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -199,17 +199,18 @@ class NURBSPatch
     const int knot_deg_v = m_knotvec_v.getDegree();
     SLIC_ASSERT(knot_deg_u >= -1 && knot_deg_v >= -1);
 
-    if(const bool is_empty = (knot_deg_u == -1 && knot_deg_u == -1); is_empty)
+    if(const bool is_empty = (knot_deg_u == -1 && knot_deg_v == -1); is_empty)
```
kennyweiss, line 202: 👍  Thanks!

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -199,17 +199,18 @@ class NURBSPatch
...
     {
       SLIC_ASSERT(knot_deg_u >= 0 && knot_deg_v >= 0);
+#if defined(AXOM_DEBUG)
       const int deg_u = utilities::max(0, knot_deg_u);
       const int deg_v = utilities::max(0, knot_deg_v);
-
+#endif
       const int npts_u = knotVector_u.getNumControlPoints();
       const int npts_v = knotVector_v.getNumControlPoints();
       SLIC_ASSERT(npts_u > deg_u && npts_v > deg_v);
```
kennyweiss: Thanks for this one too!

Side question -- should we start using  `[[maybe_unused]]` for this sort of variable instead of guarding with `AXOM_DEBUG`

See: https://en.cppreference.com/w/cpp/language/attributes/maybe_unused.html

src/axom/quest/io/MFEMReader.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+namespace quest
+{
+namespace internal
+{
+/*!
+ * \brief Read the MFEM file and build the desired type of geometry from it using a supplied function.
+ *
+ * \param fileName The name of the file to read.
+ * \param func The function/lambda that generates the geometry using the map of zones to curves.
```
kennyweiss: Please describe the expected signature of the lambda

kennyweiss: And also that this function only currently supports (curved) segment meshes in 2D

src/axom/quest/io/MFEMReader.cpp
```diff
@@ -0,0 +1,128 @@
...
+      }
+
+      // Use the map to build the geometry.
+      func(mesh, contourZones);
+
+      retval = READ_SUCCESS;
+    }
+    else
+    {
+      SLIC_INFO("Mesh must have dimension 1 and spatial dimension 2.");
```
kennyweiss: Suggestion:
Please make this a warning and print out the mesh and space dimensions of the provided mesh.

```suggestion
      SLIC_WARNING("Mesh must have dimension 1 and spatial dimension 2.");
```

src/axom/quest/io/MFEMReader.cpp
kennyweiss, line 1: This is neat!

src/axom/quest/io/C2CReader.hpp
```diff
@@ -63,48 +64,19 @@ class C2CReader
   virtual void log();
 
   /*!
-   * \brief Projects high-order NURBS contours onto a linear mesh using \a segmentsPerPiece
```
kennyweiss, line 66: Nice consolidation! Should we mention API changes in the RELEASE_NOTES?

src/axom/quest/interface/internal/QuestHelpers.cpp
```diff
@@ -362,124 +368,228 @@ int read_c2c_mesh_uniform(const std::string& file,
...
 
   return rc;
 }
+#endif
 
 /// Mesh Helper Methods
 
+#if defined(AXOM_USE_MFEM)
+template <typename CurveType, typename CreateFunc1, typename CreateFunc2>
+CurveType segment_to_curve_impl(const mfem::Mesh* mesh, int elem_id, CreateFunc1 &&create1, CreateFunc2 &&create2)
```
kennyweiss: Suggestion: 
`PolynomialConstructor` and `RationalConstructor` (or something similar) might be clearer than `CreateFunc1` and `CreateFunc2`.

src/axom/quest/examples/CMakeLists.txt
```diff
@@ -263,6 +263,30 @@ if(AXOM_ENABLE_MPI AND MFEM_FOUND AND MFEM_USE_MPI
...
+        set(_testname quest_shaping_driver_ex_heroic_roses_mfem_cp)
+        axom_add_test(
+            NAME ${_testname}
+            COMMAND  quest_shaping_driver_ex 
+                     -i ${shaping_data_dir}/heroic_roses_mfem_cp.yaml
+                     --method sampling --sampling windingnumber
+                     inline_mesh --min 0 0 --max 300 400 --res 300 400 -d 2
+            NUM_MPI_TASKS ${_nranks})
+        set_tests_properties(${_testname} PROPERTIES
+            PASS_REGULAR_EXPRESSION  "Volume of material 'black' is 36,645.")
```
kennyweiss, line 288: Presumably the volumes are different b/c the winding numbers are on the curved geometry and the inout sampling is discretized.

src/axom/quest/detail/shaping/shaping_helpers.hpp
```diff
@@ -224,6 +224,188 @@ void computeVolumeFractionsIdentity(mfem::DataCollection* dc,
...
+    "\t Sampling inout field '{}' took {:.3Lf} seconds (@ {:L} queries per second)",
+    inoutName,
+    timer.elapsed(),
+    static_cast<int>((NE * nq) / timer.elapsed())));
+}
+
+/**
+   * \brief Samples the inout field over the indexed geometry, possibly using a
+   * callback function to project the input points (from the computational mesh)
+   * to query points on the spatial index
```
kennyweiss: Minor: This one doesn't have a projector function.

Perhaps it should (?), but I didn't add it.

src/axom/quest/detail/shaping/WindingNumberSampler.hpp
```diff
@@ -0,0 +1,326 @@
...
+  bool inside = false;
+  double wn {};
+  for(int c = 0; c < shape.numEdges(); c++)
+  {
+    wn += axom::primal::winding_number(pt, shape[c]);
+  }
+  // A point inside the polygon should have non-zero winding number.
+  inside |= (std::round(wn) != 0);
+
+  return inside;
```
kennyweiss: Since the in/out is only determined once at the end, would this make sense?


```suggestion
  double wn {};
  for(int c = 0; c < shape.numEdges(); c++)
  {
    wn += axom::primal::winding_number(pt, shape[c]);
  }
  // A point inside the polygon should have non-zero winding number.
  return (std::round(wn) != 0);
```

src/axom/quest/detail/shaping/WindingNumberSampler.hpp
```diff
@@ -0,0 +1,326 @@
...
+{
+
+using QFunctionCollection = mfem::NamedFieldsMap<mfem::QuadratureFunction>;
+using DenseTensorCollection = mfem::NamedFieldsMap<mfem::DenseTensor>;
+
+namespace detail
+{
+
+/*!
+ * \brief Check whether point \a pt is inside \a shape.
```
kennyweiss, line 36: Please mention that this is using the non-zero rule to determine containment

src/axom/quest/detail/shaping/WindingNumberSampler.hpp
```diff
@@ -0,0 +1,326 @@
...
+
+    // Look up all of the query points. This will allocate the candidates array.
+    AXOM_ANNOTATE_BEGIN("findPoints");
+    axom::Array<axom::IndexType> offsets(numQueryPoints, numQueryPoints, allocatorID);
+    axom::Array<axom::IndexType> sizes(numQueryPoints, numQueryPoints, allocatorID);
+    axom::Array<axom::IndexType> candidates;
+    auto offsetsView = offsets.view();
+    auto sizesView = sizes.view();
+    m_bvh.findPoints(offsetsView, sizesView, candidates, numQueryPoints, queryPointsView);
+    AXOM_ANNOTATE_END("findPoints");
```
kennyweiss, line 217: Does this mean that we're only checking points against curves whose bounding box contains the curve?
If so, this will not in general compute the correct result.

As a simple example, consider the boundary of a square made up of two horizontal lines and two vertical lines. A query point will almost never be in a bounding box of any of the lines and will almost always return 0 if we're only considering the bounding box candidates.

Can you please add a test case for this simple square mesh and ensure we get the correct results?

src/axom/quest/LinearizeCurves.hpp
kennyweiss, line 1: This is a nice improvement!

Would it make sense to add some standalone unit tests for the linearization functions?
Perhaps we already have them within the c2c/mfem reader tests or the shaping tests?

kennyweiss (changes requested):
Thanks @BradWhitlock -- this is really nice and will be very useful!

My only concern is about the BVH tree in the `WindingNumberSampler`. If I'm understanding it correctly, it will skip open curves when the point is not in their bounding boxes and return an incorrect value.

Please see my suggestion below for a simple test case.

src/axom/quest/detail/shaping/WindingNumberSampler.hpp
BradWhitlock: Done.

src/axom/quest/detail/shaping/WindingNumberSampler.hpp
BradWhitlock: Fixed it. I had copied the code from an earlier version.

src/axom/quest/detail/shaping/shaping_helpers.hpp
BradWhitlock: I added the projector plumbing for `computeVolumeFractionsBaseline` into `SamplingShaper` and the samplers.

src/axom/quest/examples/CMakeLists.txt
BradWhitlock: That was my thought too.

src/axom/quest/interface/internal/QuestHelpers.cpp
BradWhitlock: I changed the template args to the names you suggested.

src/axom/quest/io/C2CReader.hpp
BradWhitlock: There is a bullet in the release notes. Is that clear enough?
```
Moves curve linearization from the quest::C2CReader into quest::LinearizeCurves so the logic can be used with other curve data.
```

src/axom/quest/io/MFEMReader.cpp
BradWhitlock: I made it a warning and included the dimensions.

src/axom/quest/io/MFEMReader.cpp
BradWhitlock: I improved the comments as suggested.

src/axom/primal/geometry/NURBSPatch.hpp
BradWhitlock: I made an `AXOM_MAYBE_UNUSED` macro and used that here instead of conditional compilation. The macro should be defined when the compiler supports c++17 or later.

I could have instead used `AXOM_UNUSED_VAR(deg_u)`, or even just put the `utilities::max()` call into the `SLIC_ASSERT()` macro as alternate ways of doing this.

src/axom/quest/detail/shaping/WindingNumberSampler.hpp
BradWhitlock: The `WindingNumberSampler` is operating on `CurvedPolygons` read from the MFEM file, not a list of separate edge contours or linearized edges. The `CurvedPolygon` bboxes are used to initialize the BVH so when we ask the BVH for candidate hits for all of the quad points in the inout function, we're getting a list of `CurvedPolygon` ids to check more closely with the winding number rules. If the point is inside any candidate, we count it as inside. Otherwise the point is outside.

Since the `CurvedPolygon` uses `BezierCurve` edges, I would expect its bounding box to be reliable, and I am assuming this. Each edge curve in the polygon should live in its control polygon, and making an axis-aligned bbox from those points ought to contain it. Then the overall polygon bbox should just be the outer extents of each edge's bbox. If the `CurvedPolygon` was using `NURBSCurves`, the bboxes would have to be computed more carefully since the curve could stray outside of a bbox computed from just the control points.

* Please correct me if I'm wrong
* I'll make a square test

kennyweiss: If it's always a bounding box for a closed `CurvedPolygon`  that is oriented such that points inside the polygon are inside then it should be ok.

It it's a bounding box over non-closed BezierCurve or NURBSCurve instances, then I don't think it would work.
Similarly if the closed curve is oriented the other way (e.g. like the inner loop of an annulus), then we'd get the wrong answer.

fwiw, I was asking about the square test for four separate edges.

BradWhitlock: CurvedPolygons with edges that are reversed can be used to subtract out regions from the polygon.
image

BradWhitlock: CurvedPolygons where the edges are not closed also seem to be tolerated.
image

src/axom/quest/LinearizeCurves.hpp
BradWhitlock: I added basic tests for `LinearizeCurves`. It also gets used in the c2c reader test.

src/axom/quest/detail/shaping/WindingNumberSampler.hpp
jcs15c: Just to give my two cents here, I think that the approach of checking bounding boxes of each "closed" loop, and returning a zero for any point outside it is a reasonable one. Numerically, it should be same as if you computed a fractional winding number field for each shape, rounded it for each shape, then added those together. That being said, that's very different from calculating the entire winding number field and then rounding that, just like how `round(a) + round(b) + round(c)` can be, and usually is, different from `round(a + b + c)`. 

For comparison, here's what the winding number field looks like for all of those shapes looked at simultaneously, and the rounded result:
image

This collection of shapes is actually a super interesting example of why this "round first" approach might be preferred. Because the winding number is a global property of the set of curves (i.e. every point looks at every curve in some way), the more open/degenerate a single loop is, the greater impact it can have on the implicit boundaries of other loops. In the lower left the containment decisions are still fairly reasonable, since the geometric errors are too, but the inverse is true in the top right: worse geometric errors near one another accumulate to a worse winding number. Even more interesting, since every missing edge is pointed in the same direction, those errors really accumulate in a particularly bad way, such as how the nearest integer to the winding number field in the top right is actually -1.

In any case, I think the approach you have here is probably the right way to do it, but *only* if you're comfortable with the assumption that each `CurvedPolygon` is supposed to represent a single closed loop. It's worth keeping in mind it definitely it's a viable approach, but not the most immediate one for winding numbers.

jcs15c: As an unnecessary aside, here's what the winding number field for just those missing straight line pieces looks like:
image
Clearly as the lines get longer, they contribute more to their _own_ winding number field, so subtracting the whole thing away has a bigger impact. Neat!

src/axom/quest/examples/make_shapes.py
kennyweiss, line 1: 👍 -- thanks for adding this!

kennyweiss (approved):
Thanks for all the changes @BradWhitlock -- really looking forward to using this!

PR #1645, opened 2025-08-21
Change MIRAlgorithm to make it clearer that it operates on a single domain.

BradWhitlock:
Early in MIR development, it was coded to iterate over multiple domains and perform MIR on each domain. The code was then pared back to operate on a single domain, with data provided using various topology, coordset, and matset views. This was done to limit the Blueprint types on which the algorithm needs to operate in order to generate smaller code.

This PR removes a vestige of domain iteration from `MIRAlgorithm` and changes comments to indicate that data are passed one domain at a time.

* Added multidomain testing for MIR
* Fixed an issue arising from domain prefix in Conduit object paths

### DIFF
--- a/src/axom/mir/MIRAlgorithm.cpp
+++ b/src/axom/mir/MIRAlgorithm.cpp
@@ -137,7 +137,7 @@ std::string MIRAlgorithm::localPath(const conduit::Node &obj) const
   std::string path(obj.path());
   const auto dpos = path.find("domain");
   const auto spos = path.find("/");
-  if(dpos == 0 && spos > dpos && obj.parent() != nullptr)
+  if(dpos == 0 && spos != std::string::npos && spos > dpos && obj.parent() != nullptr)
   {
     path = path.substr(spos + 1, path.size() - spos - 1);
   }
--- a/src/axom/mir/MIRAlgorithm.cpp
+++ b/src/axom/mir/MIRAlgorithm.cpp
@@ -135,7 +135,7 @@ void MIRAlgorithm::saveMesh(const conduit::Node &n_mesh, const std::string &file
 std::string MIRAlgorithm::localPath(const conduit::Node &obj) const
 {
   std::string path(obj.path());
-  const auto dpos = path.find("dom");
+  const auto dpos = path.find("domain");
   const auto spos = path.find("/");
   if(dpos == 0 && spos > dpos && obj.parent() != nullptr)
   {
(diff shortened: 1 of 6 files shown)


src/axom/mir/MIRAlgorithm.cpp
```diff
@@ -24,14 +24,7 @@ void MIRAlgorithm::execute(const conduit::Node &n_input,
...
   {
-    // Handle multiple domains
-    for(const auto &dom_ptr : domains)
-    {
-      const conduit::Node &n_domain = *dom_ptr;
-      conduit::Node &n_newDomain = n_output.append();
-
-      executeSetup(n_domain, n_options, n_newDomain);
-    }
+    SLIC_ERROR("The input node contains multiple domains. Pass a single domain at a time instead.");
```
kennyweiss, line 27: Do we have a multidomain example that loops over its domains and calls MIR on each one?

kennyweiss (approved):
Thanks @BradWhitlock

src/axom/mir/MIRAlgorithm.cpp
BradWhitlock: I'm changing one of the examples to do that now. It might spark some other changes.

BradWhitlock: I changed some 2D tests for EquiZ and Elvira so they make multidomain data and do MIR on it. This did expose something that Jacob ran into. I corrected the problem.

src/axom/mir/MIRAlgorithm.cpp
```diff
@@ -132,5 +132,17 @@ void MIRAlgorithm::saveMesh(const conduit::Node &n_mesh, const std::string &file
...
+std::string MIRAlgorithm::localPath(const conduit::Node &obj) const
+{
+  std::string path(obj.path());
+  const auto dpos = path.find("dom");
+  const auto spos = path.find("/");
+  if(dpos == 0 && spos > dpos && obj.parent() != nullptr)
+  {
+    path = path.substr(spos + 1, path.size() - spos - 1);
+  }
+  return path;
```
kennyweiss: Could this be replaced by 
```cpp
return axom::utilities::string::removePrefix("dom", obj.path());
```
https://github.com/LLNL/axom/blob/0398fe6f3191ff6e71aa7baf8764e34a8efc10e3/src/axom/core/utilities/StringUtilities.hpp#L136-L150

kennyweiss (approved):
Thanks @BradWhitlock !

src/axom/mir/MIRAlgorithm.cpp
BradWhitlock: I looked at that but I technically do not know what the prefix is. It could be "domain_00005", "domain_0000008", etc.

kennyweiss: Thanks -- that makes sense.

kennyweiss: Maybe update the docs for `localPath` that you're expecting the prefix to be of the form `domain_000N`?

PR #1649, opened 2025-08-29
Bugfix/bergel1/logging perf

gberg617:
# Summary

This PR adds the ability to test whether pending messages exist by performing an "all reduce" on a single MPI communicator.  This eliminates redundant flushing of streams when there are no messages to be logged, and improves performance on large problems where unnecessary flushing can be an issue.


src/axom/slic/core/LogStreamStatusMonitorMPI.cpp
```diff
@@ -0,0 +1,76 @@
...
+  if (ls->isUsingMPI() == true)
+  {
+    m_useMPI = true;
+    if (m_mpiComm == MPI_COMM_NULL && 
+      ls->comm() != MPI_COMM_NULL)
+    {
+      m_mpiComm = ls->comm();
+    }
+    else if (m_mpiComm != MPI_COMM_NULL && m_mpiComm != ls->comm()) {
+      std::cerr << "ERROR: multiple MPI communicators passed to LogStreamStatusMonitor" << std::endl;
```
gberg617: I couldn't come up with a good way of adding different MPI communicators in a consistent manner, so I'm leaving this constraint in for now.  Any thoughts on this?

src/axom/slic/tests/slic_perf.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "axom/slic/streams/SynchronizedStream.hpp"
+#include "gtest/gtest.h"
+
+#include <mpi.h>
+
+// namespace alias
+namespace slic = axom::slic;
+
+//------------------------------------------------------------------------------
+TEST(SlicPerfTest, test_many_flushes)
```
gberg617, line 18: This performance test shows a clear improvement with the changes in this PR.  The run time decreased from about 3.8s to 40ms on 32 ranks (approx. 100x improvement).

src/axom/slic/core/LogStreamStatusMonitor.cpp
```diff
@@ -0,0 +1,52 @@
...
+  int has_pending_messages = 0;
+  for (auto& stream : m_streamVec)
+  {
+    has_pending_messages += static_cast<int>(stream->hasPendingMessages());
+  }
+  return has_pending_messages > 0;
+}
+
+}
+}
```
white238:
```suggestion
}

```

src/axom/slic/core/LogStreamStatusMonitorMPI.cpp
```diff
@@ -0,0 +1,76 @@
...
+    int local_has_pending_messages = has_pending_messages;
+
+    MPI_Allreduce(&local_has_pending_messages, &has_pending_messages, 1, MPI_INT, MPI_MAX, m_mpiComm);
+  }
+
+  return has_pending_messages > 0;
+}
+
+}
+}
```
white238:
```suggestion
}

```

src/axom/slic/core/LogStreamStatusMonitorMPI.hpp
```diff
@@ -0,0 +1,59 @@
...
+{
+namespace slic
+{
+/*!
+ * \class LogStreamStatusMonitorMPI
+ *
+ * \brief Monitor log streams to see if there are any pending messages using MPI
+ *
+ */
+class LogStreamStatusMonitorMPI : public LogStreamStatusMonitor
```
white238: Is there a reason to not combine these two classes and just toggle the behavior inside?

src/axom/slic/tests/CMakeLists.txt
```diff
@@ -51,6 +52,27 @@ if (AXOM_ENABLE_MPI AND AXOM_ENABLE_LUMBERJACK)
...
+    foreach(test ${perf_slic_tests})
+        get_filename_component( test_name ${test} NAME_WE )
+        axom_add_executable(NAME       ${test_name}_test
+                            SOURCES    ${test}
+                            OUTPUT_DIR ${TEST_OUTPUT_DIRECTORY}
+                            DEPENDS_ON ${slic_tests_depends}
+                            FOLDER     axom/slic/tests )
+        axom_add_test(NAME          ${test_name}
+                      COMMAND       ${test_name}_test 
+                      NUM_MPI_TASKS 32)
```
white238: I am not entirely sure this is okay to have such a high MPI task job... the current highest to my knowledge is 8.  We have CI/scripts that run 16 tests at at time and that will overload the CPU's we have.

https://github.com/LLNL/axom/blob/c062852fd9306d9838f28b8c05b9f829fbdd002e/scripts/llnl_scripts/llnl_lc_build_tools.py#L255-L256
 
 Then whatever this math comes out to:
 
 https://github.com/LLNL/axom/blob/c062852fd9306d9838f28b8c05b9f829fbdd002e/scripts/github-actions/linux-build_and_test.sh#L33

src/axom/slic/tests/slic_log_stream_status_monitor.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+  MPI_Init(&argc, &argv);
+
+  // finalized when exiting main scope
+  result = RUN_ALL_TESTS();
+
+  MPI_Finalize();
+
+  return result;
+}
```
white238:
```suggestion
}

```

src/axom/slic/core/LogStreamStatusMonitorMPI.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+//------------------------------------------------------------------------------
+bool LogStreamStatusMonitorMPI::hasPendingMessages() const
+{
+  int has_pending_messages = LogStreamStatusMonitor::hasPendingMessages();
+
+  if (m_useMPI)
+  {
+    int rank, size;
```
BradWhitlock: These look unused.

src/axom/slic/tests/CMakeLists.txt
kennyweiss: I agree -- for our CI, we'd probably want to limit this to 4 tasks (?), but could go higher on our LC CI, if it's necessary for the test.

src/axom/slic/core/LogStreamStatusMonitorMPI.hpp
gberg617: Nope! I went ahead and combined them.

src/axom/slic/tests/CMakeLists.txt
gberg617: OK sounds good.  I changed the test to run on 4 ranks.

src/axom/slic/core/LogStreamStatusMonitorMPI.cpp
gberg617: Fixed.

src/axom/slic/core/LogStreamStatusMonitorMPI.cpp
gberg617: @white238 @bmhan12  This constraint implies that SLIC can only accept MPI-based streams that use the same MPI communicator.  This might be somewhat restrictive.  We should talk about this if you have any concerns.

src/axom/slic/core/LogStreamStatusMonitor.hpp
```diff
@@ -0,0 +1,64 @@
...
+{
+public:
+  LogStreamStatusMonitor();
+
+  /*!
+   * \brief Add LogStream pointer to vector stored in LogStreamStatusMonitor.
+   * \param [in] ls pointer to the user-supplied LogStream object.
+   *
+   */
+  void addStream(LogStream* ls);
```
kennyweiss, line 44: Would we ever need to remove a logstream, or clear the list of monitored steams?

Presumably, these will come later, as needed.

src/axom/slic/core/LogStreamStatusMonitor.cpp
```diff
@@ -0,0 +1,85 @@
...
+  if (ls->isUsingMPI() == true)
+  {
+    m_useMPI = true;
+    if (m_mpiComm == MPI_COMM_NULL && 
+      ls->comm() != MPI_COMM_NULL)
+    {
+      m_mpiComm = ls->comm();
+    }
+    else if (m_mpiComm != MPI_COMM_NULL && m_mpiComm != ls->comm()) {
+      std::cerr << "ERROR: multiple MPI communicators passed to LogStreamStatusMonitor" << std::endl;
```
kennyweiss: Would something like the following be clearer?

```suggestion
      std::cerr << "ERROR: attempting to register a logstream with an incompatible MPI communicators to LogStreamStatusMonitor's existing communicator" << std::endl;
```

src/axom/slic/core/LogStreamStatusMonitor.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+  /*!
+   * \brief Checks to see if any pending messages exist on the current MPI communicator
+   */
+  bool hasPendingMessages() const;
```
kennyweiss, line 51: Please indicate in the doxygen that `hasPendingMessages` is a collective call in MPI configurations

src/axom/slic/tests/slic_log_stream_status_monitor.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+}
+
+int main(int argc, char* argv[])
+{
+  int result = 0;
+
+  ::testing::InitGoogleTest(&argc, argv);
+
+  MPI_Init(&argc, &argv);
```
kennyweiss, line 156: In case you're not familiar with it, you can use the MPIWrapper here:
https://github.com/LLNL/axom/blob/981a7e7e191bbc9013a1fc64a32056e9c5d7cccd/src/axom/core/utilities/RAII.hpp

(You don't need to change it, just thought I'd let you know about it).

src/axom/slic/streams/LumberjackStream.cpp
```diff
@@ -269,6 +269,20 @@ void LumberjackStream::write(bool local)
...
+  return m_lj->getMessages().size() > 0;
+}
+
+//------------------------------------------------------------------------------
+bool LumberjackStream::isUsingMPI()
+{
+  return true;
+}
+
+int LumberjackStream::comm() { return m_lj->getCommunicator()->comm(); }
```
kennyweiss: Is it safe to return an `int` for the communicator?

src/axom/slic/streams/LumberjackStream.hpp
```diff
@@ -232,6 +232,27 @@ class LumberjackStream : public LogStream
...
+   * \return Returns true if this class relies on MPI
+   */
+  virtual bool isUsingMPI();
+
+  /*!
+   * \brief Get the communicator
+   *
+   * \return Returns the communicator ID if it exists, or -1 otherwise
+   */
+  virtual int comm();
```
kennyweiss: Should these be marked `override`?  (here and elsewhere)

src/axom/slic/core/Logger.cpp
```diff
@@ -32,6 +33,18 @@ Logger*& getLogger()
...
+//------------------------------------------------------------------------------
+// This is a singleton, scope-limited to this file.
+LogStreamStatusMonitor*& getLogStreamStatusMonitor()
+{
+  static LogStreamStatusMonitor* s_logStreamStatusMonitor = nullptr;
+  if (!s_logStreamStatusMonitor)
+  {
+    s_logStreamStatusMonitor = new LogStreamStatusMonitor();
+  }
+  return s_logStreamStatusMonitor;
```
kennyweiss: Is there a reason we couldn't do
```suggestion
  static LogStreamStatusMonitor s_logStreamStatusMonitor;
  return &s_logStreamStatusMonitor;
```

And then in `Logger::finalize()`, we could call `getLogSttreamStatusMonitor()->clear()`

src/axom/slic/tests/slic_log_stream_status_monitor.cpp
```diff
@@ -0,0 +1,160 @@
...
+  EXPECT_EQ(logStreamStatusMonitor.hasPendingMessages(), false);
+
+  generic_stream.append(axom::slic::message::Debug, "test message", "test tag", "test file name", 1, false, false);
+
+  /* hasPendingMessages will not probe for unflushed messages in 
+     GenericOutputStream because it relies on ofstream for flushing rather than MPI 
+   */
+  EXPECT_EQ(logStreamStatusMonitor.hasPendingMessages(), false);
+
+  generic_stream.flush();
```
kennyweiss: I'm confused about this test and comment.

Is the status monitor mainly for file based streams?
Please document which streams will be expected to have pending messages and which will flush for themselves.

src/axom/slic/tests/slic_log_stream_status_monitor.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include "axom/slic/streams/LumberjackStream.hpp"
+#include "axom/slic/streams/SynchronizedStream.hpp"
+#include "gtest/gtest.h"
+
+#include <mpi.h>
+
+// namespace alias
+namespace slic = axom::slic;
+
+TEST(SlicLogStreamMonitorTest, test_has_pending_messages)
```
kennyweiss: Please split this into separate tests for each output stream type

src/axom/slic/tests/slic_log_stream_status_monitor.cpp
```diff
@@ -0,0 +1,160 @@
...
+  logStreamStatusMonitor.addStream(&lj_stream);
+
+  EXPECT_EQ(logStreamStatusMonitor.hasPendingMessages(), false);
+
+  lj_stream.append(axom::slic::message::Debug, "test message", "test tag", "test file name", 1, false, false);
+
+  EXPECT_EQ(logStreamStatusMonitor.hasPendingMessages(), true);
+
+  lj_stream.flush();
+
```
kennyweiss: Please add another test after each flush that shows `hasPendingMessages() == false`

kennyweiss (commented):
Thanks @gberg617 

I have a few questions, but my main one relates to the comment about ostream-based streams flushing themselves.

Looking forward to the docs update.

src/axom/slic/core/LogStreamStatusMonitor.cpp
```diff
@@ -0,0 +1,85 @@
...
+ ******************************************************************************
+ *
+ * \file LogStreamStatusMonitor.cpp
+ *
+ * \brief Implementation of the LogStreamStatusMonitor class.
+ *
+ ******************************************************************************
+ */
+
+#include <mpi.h>
```
kennyweiss: This include also needs to be guarded by `AXOM_USE_MPI`  (caught by failing CI in the non-mpi jobs).

src/axom/slic/core/LogStreamStatusMonitor.hpp
gberg617: I added a `finalize()` method that clears/resets the state of logStreamStatusMonitor.  I am assuming that the streams stored in this class are the same as those stored in Logger.  Since there isn't currently a way to remove individual log streams from the Logger without clearing the state, I prefer to not include such a method in LogStreamStatusMonitor.

src/axom/slic/core/LogStreamStatusMonitor.cpp
gberg617: Fixed!

src/axom/slic/tests/slic_log_stream_status_monitor.cpp
gberg617: OK, I'll look into this soon.

src/axom/slic/streams/LumberjackStream.cpp
gberg617: I think so but I'll check on this.

src/axom/slic/streams/LumberjackStream.hpp
gberg617: Yup. Fixed!

src/axom/slic/core/Logger.cpp
gberg617: Fixed for both points.

gberg617: We could also return the ref without it being a pointer.  I prefer this way if it's fine for you.

src/axom/slic/tests/slic_log_stream_status_monitor.cpp
gberg617: Apologies for the confusion.  The main point here is that it isn't generally possible to probe for messages that are not flushed in an ostream object because that information isn't easily accessible (I removed the above comment since it is somewhat misleading).  I added a new function to the `LogStream` class and each of its inherited classes called `canTestHasPendingMessages()`.  This function will return true if we can check for unflushed messages (which is possible for SynchronizedStream and LumberjackStream, and not possible for GenericOutputStream).  I added some logic in the Logger class to flush messages for a given stream only if either `canTestHasPendingMessages()` and `hasPendingMessages()` are both true, or just `canTestHasPendingMessages()` is false to handle all the use cases.  Hopefully these changes will clarify which stream types can check for pending messages.

src/axom/slic/tests/slic_log_stream_status_monitor.cpp
gberg617: Fixed!

src/axom/slic/tests/slic_log_stream_status_monitor.cpp
gberg617: Fixed.

src/axom/slic/core/LogStreamStatusMonitor.cpp
gberg617: Fixed!

src/axom/slic/core/LogStreamStatusMonitor.hpp
gberg617: Fixed.

src/axom/slic/streams/LumberjackStream.hpp
```diff
@@ -230,7 +230,37 @@ class LumberjackStream : public LogStream
...
+  void write(bool local = false);
+
+  /*!
+   * \brief Tests whether there are any pending messages that need to be flushed.
+   * This method should only be overriden for LogStream inherited classes that can
+   * reliably test whether pending messages exist.
+   *
+   * \return Returns true if there are pending messages that need to be flushed
+   */
+  virtual bool canTestHasPendingMessages() override;
```
kennyweiss: Is this about whether we can test if there are pending messages, or whether it has pending messages?

Would  `canHavePendingMessages()` work for the function's name?

src/axom/slic/streams/LumberjackStream.hpp
```diff
@@ -230,7 +230,37 @@ class LumberjackStream : public LogStream
...
    */
-  virtual void write(bool local = false);
+  void write(bool local = false);
+
+  /*!
+   * \brief Tests whether there are any pending messages that need to be flushed.
+   * This method should only be overriden for LogStream inherited classes that can
+   * reliably test whether pending messages exist.
+   *
+   * \return Returns true if there are pending messages that need to be flushed
```
kennyweiss: I think this `return` needs to be updated.

src/axom/slic/core/LogStreamStatusMonitor.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+//------------------------------------------------------------------------------
+bool LogStreamStatusMonitor::hasPendingMessages() const
+{
+  int has_pending_messages = 0;
+
+  for (auto& stream : m_streamVec)
+  {
+    has_pending_messages += static_cast<int>(stream->hasPendingMessages());
+  }
```
kennyweiss, line 99: Based on the name -- `canTestHasPendingMessages()` -- I was expecting this loop to be something like:
```cpp
  for (auto& stream : m_streamVec)
  {
     if(stream->canTestHasPendingMessages())
     {
        has_pending_messages += static_cast<int>(stream->hasPendingMessages());
     }
  }
```

src/axom/slic/tests/slic_log_stream_status_monitor.cpp
```diff
@@ -0,0 +1,176 @@
...
+  stream->append(axom::slic::message::Debug, "test message", "test tag", "test file name", 1, false, false);
+
+  if(stream->canTestHasPendingMessages() == true)
+  {
+    EXPECT_EQ(logStreamStatusMonitor.hasPendingMessages(), true);
+  }
+  else
+  {
+    EXPECT_EQ(logStreamStatusMonitor.hasPendingMessages(), false);
+  }
```
kennyweiss: Thanks for updating the tests -- this is a lot cleaner!

One more suggestion:
```suggestion
  if(stream->canTestHasPendingMessages() == true)
  {
    EXPECT_TRUE(logStreamStatusMonitor.hasPendingMessages());
  }
  else
  {
    EXPECT_FALSE(logStreamStatusMonitor.hasPendingMessages());
  }
```

src/axom/slic/core/Logger.cpp
```diff
@@ -32,6 +33,14 @@ Logger*& getLogger()
   return s_Logger;
 }
 
+//------------------------------------------------------------------------------
+// This is a singleton, scope-limited to this file.
+LogStreamStatusMonitor& getLogStreamStatusMonitor()
+{
+  static LogStreamStatusMonitor s_logStreamStatusMonitor;
+  return s_logStreamStatusMonitor;
+}
```
kennyweiss, line 42: 👍

kennyweiss (approved):
Thanks for the updates @gberg617 

One pending question about API names (and where it should be called) and it should be good to go.

src/axom/slic/streams/LumberjackStream.cpp
gberg617: Good call.  I changed the return type to `MPI_Comm` and added an `#if defined (AXOM_USE_MPI)`.

src/axom/slic/streams/LumberjackStream.hpp
gberg617: It's whether we can have pending messages.  I went ahead and changed the name.

src/axom/slic/core/LogStreamStatusMonitor.cpp
gberg617: Fixed.

src/axom/slic/tests/slic_log_stream_status_monitor.cpp
gberg617: Fixed.

src/axom/slic/core/Logger.cpp
gberg617: Fixed.

src/axom/slic/streams/LumberjackStream.hpp
gberg617: Fixed.

src/axom/slic/core/LogStreamStatusMonitor.hpp
```diff
@@ -0,0 +1,71 @@
...
+  /*!
+   * \brief Add LogStream pointer to vector stored in LogStreamStatusMonitor.
+   * \param [in] ls pointer to the user-supplied LogStream object.
+   *
+   */
+  void addStream(LogStream* ls);
+
+  /*!
+   * \brief Checks to see if any pending messages exist on the current MPI communicator.
+   * This call is collective.  All ranks in m_mpiComm must call this function.
```
bmhan12:
```suggestion
   * \note This call is collective.  All ranks in m_mpiComm must call this function.
```
Just to help make this stick out more.

src/axom/slic/core/Logger.hpp
```diff
@@ -26,6 +27,8 @@ namespace slic
 // Forward declarations
 class LogStream;
 
+class LogStreamStatusMonitorMPI;
```
bmhan12: I think this forward declaration may not be necessary anymore?

bmhan12 (approved):
Thanks @gberg617 for debugging this! Be sure to update `RELEASE-NOTES.md` with details from this PR!

src/axom/slic/core/LogStreamStatusMonitorMPI.cpp
gberg617: This will need to be addressed in this PR with a follow-up commit.

gberg617: I pushed a fix for this.  @white238 @bmhan12 @kennyweiss Let me know if you have any thoughts on the new commits.

src/axom/slic/core/LogStreamStatusMonitor.hpp
gberg617: Fixed.

src/axom/slic/core/Logger.hpp
gberg617: Fixed.

src/axom/slic/core/LogStream.hpp
```diff
@@ -144,7 +148,40 @@ class LogStream
    */
   virtual void push() {};
 
+  /*!
+   * \brief Determines whether it is possible to reliably check if pending messages exist
+   *
+   * \return Returns True when it is possible to reliably check whether pending messages exist
+   */
+  virtual bool canHavePendingMessages() { return false; };
```
white238: In my opinion this is kind of awkward. I think its better to just have the `hasPendingMesages` and return `false` for any that will not hold onto messages by their design.

gberg617: I removed `canHavePendingMessage()` and replaced the calls with `isUsingMPI()`. I believe this can work for now because any stream that can be checked for pending messages currently uses MPI.  I also modified the `hasPendingMessages()` function to reflect this change.  We can always extend this in the future or revert back to the earlier version if needed.

For background, the function `canHavePendingMessages()` was initially put in to handle the case where we cannot check for pending messages (such as an ostream object stored in a `GenericOutputStream`).  In these cases, we have to always flush the stream's messages when calling `flushStreams()`.

src/axom/slic/core/Logger.cpp
```diff
@@ -367,13 +379,21 @@ void Logger::outputLocalMessages()
 //------------------------------------------------------------------------------
 void Logger::flushStreams()
 {
+
```
kennyweiss, line 382: Please add an implementation note about why this is here and what it's doing

src/axom/slic/core/Logger.cpp
```diff
@@ -367,13 +379,21 @@ void Logger::outputLocalMessages()
...
   //Flush for all message levels
   for(int level = message::Error; level < message::Num_Levels; ++level)
   {
     unsigned nstreams = static_cast<unsigned>(m_logStreams[level].size());
     for(unsigned istream = 0; istream < nstreams; ++istream)
     {
-      m_logStreams[level][istream]->flush();
+      const bool streamUsesMPI = m_logStreams[level][istream]->isUsingMPI();
+      if ((streamUsesMPI == true && pendingMessages) || 
+           streamUsesMPI == false)
```
kennyweiss: Is this equivalent to
```
if( !streamUsesMPI || pendingMessages)
{

}
```
If so, please use the latter form.

kennyweiss: Since this logic is repeated, it should probably be moved to a class function

src/axom/slic/core/LogStreamStatusMonitor.cpp
```diff
@@ -0,0 +1,119 @@
...
+//------------------------------------------------------------------------------
+void LogStreamStatusMonitor::addStream(LogStream* ls)
+{
+  m_streamVec.push_back(ls);
+#if defined(AXOM_USE_MPI)
+  if (ls->isUsingMPI() == true)
+  {
+    m_useMPI = true;
+
+    auto commCompareFunc =
```
kennyweiss, line 53: Please add a comment here about what this is doing and why

kennyweiss:
We might want to add a counter to `GenericOutputStream` to indicate if any messages have been logged since the last `flush()` and only call `flush()` when there are messages.  Flushing w/ no messages might be expensive since it might need to open and append (nothing) to a file.

See:
https://github.com/LLNL/axom/blob/25744e566c650e8ce53c35f466672623eab1e917/src/axom/slic/streams/GenericOutputStream.cpp#L127-L138

gberg617:
> We might want to add a counter to `GenericOutputStream` to indicate if any messages have been logged since the last `flush()` and only call `flush()` when there are messages. Flushing w/ no messages might be expensive since it might need to open and append (nothing) to a file.
> 
> See:
> 
> https://github.com/LLNL/axom/blob/25744e566c650e8ce53c35f466672623eab1e917/src/axom/slic/streams/GenericOutputStream.cpp#L127-L138

@kennyweiss It's my understanding that files are only opened once during the first write operation, and stay open until the ofstream is destructed.  In addition, the flush operation on an ofstream that has an empty buffer is essentially a no-op, so it doesn't incur a major performance penalty for this case that we were initially concerned about (other than the function call overhead).  With that said, do you still think it's worth adding a counter and checking for pending messages?

gberg617:
> > do you still think it's worth adding a counter and checking for pending messages?
> 
> Thanks @gberg617 -- It might be worthwhile in the future, but let's not make it part of this PR. Could you please add an issue for this so we'll have a record if/when it comes up again?

Added.  See https://github.com/LLNL/axom/issues/1658

src/axom/slic/core/LogStreamStatusMonitor.cpp
gberg617: I added some comments that hopefully clarify what is being done here.

src/axom/slic/core/Logger.cpp
gberg617: Fixed.

src/axom/slic/core/Logger.cpp
gberg617: I added some comments here.

PR #1651, opened 2025-09-03
FlatMap: add device-aware rehash

publixsubfan:
# Summary

Add GPU-aware parallel rehash method to `FlatMap`. We use a simple atomic-based gather step to get filled bucket indices, then construct an iterator over the indices to pass into the batched insertion method.

For now, this method is only used when the `FlatMap` is allocated in a device-accessible memory space (`Device` or `Unified`), and the underlying key-value pair is trivially-copyable.


src/axom/core/FlatMapUtil.hpp
```diff
@@ -145,6 +187,46 @@ auto FlatMap<KeyType, ValueType, Hash>::create(ArrayView<KeyType> keys,
...
+        // Bucket contains an element.
+        IndexType dest = axom::atomicAdd<ExecSpace>(counter, 1);
+        filled_bucket_idxs[dest] = bucket_idx;
+      }
+    });
+
+  FlatMapOffsetIterator bucket_begin {m_buckets.data(), filled_bucket_idx_vec.data()};
+
+  FlatMap new_map(count, m_allocator);
+  new_map.template insert<ExecSpace>(bucket_begin, bucket_begin + m_size);
```
Arlie-Capps, line 254: I had to stop at this line and read about the [template and typename keywords](https://stackoverflow.com/questions/610245/).

src/axom/core/FlatMap.hpp
```diff
@@ -615,28 +615,22 @@ class FlatMap : detail::flat_map::SequentialLookupPolicy<typename Hash::result_t
...
+  #if defined(AXOM_USE_CUDA)
+          using ExecSpace = axom::CUDA_EXEC<256>;
+  #elif defined(AXOM_USE_HIP)
+          using ExecSpace = axom::HIP_EXEC<256>;
+  #endif
+          this->parallelRehash<ExecSpace>(count);
           return;
         }
-      }
 #endif
```
kennyweiss: Should there be an `#else` here?

E.g. what if you call `rehash` from the cpu? or on a type thatt's not trivially copyable?
At the very least, we probably would want a warning/error?

publixsubfan: It looks like it was hidden by github, but the fallback logic is below the `#endif` statement. For CPU-side memory we just construct a new FlatMap on the CPU with the `begin()` and `end()` iterators and swap it with the current instance. This also covers the case where the types are not trivially-copyable.

PR #1653, opened 2025-09-09
Accelerates 2D/3D GWN queries with memoization

jcs15c:
This PR is ready to go (assuming cooperation with the Gitlab CI), but I'm waiting to merge it until I'm back in office on 9/30, since it isn't critical for the upcoming release.

# Summary

The current 2D and 3D algorithms currently compute redundant information between evaluation at different query points, namely evaluations of the curve/surface. This PR is an algorithm refactoring/enhancement that addresses this by utilizing memoization for performance improvements. A `NURBSCurveGWNCache` or `NURBSPatchGWNCache` object stores the original NURBS object along with map(s) that stores these intermediate values as they are used in the GWN algorithm:

- In 2D, the input NURBS curve is decomposed into Bezier curves, and necessary levels of subdivisions are stored for each Bezier curve. This implementation is considerably simpler than before, e.g. no longer constructs and checks polygons at each refinement, but avoids extra memory allocations.
- In 3D, quadrature nodes and tangent vectors to the surface are stored for each trimming curve in a patch. Furthermore, a copy of the original patch altered to improve the robustness of the GWN algorithm, e.g. linearly extending the parameter space of the patch so that the untrimmed geometry is larger than the visible region.

These methods are compatible with user supplied cache objects, but are also constructed automatically in cases where the GWN algorithm is called on arrays of query points to simplify use. 

Note that these acceleration techniques are currently not reflected in the winding number sampling shaper code in quest, namely because these methods are built on `CurvedPolygon` objects which contain Bezier, not NURBS curves. This will be addressed in a future PR.

Some lingering questions that can hopefully be thought about during review:
- Is this framework of dedicated "cache objects" the implementation that is most conducive to future changes, e.g., GPU implmentation? 
- Is `GWNCache` the correct label for these objects? Conceivably, there could be other kinds of memoization strategies that involve curves/surfaces, but other geometric operations, such as closest point.

## Performance Improvements

To evaluate the performance improvement of the 3D memoization scheme, I've executed this block of code on this branch and develop on my local laptop. On a release build, this code executed in 13.8 minutes on develop and 10.3 seconds on this branch.

```cpp
TEST(primal_integral, demo_test)
{
  using Point3D = primal::Point<double, 3>;
  using Vector3D = primal::Vector<double, 3>;

  // Set up an array of points to test against
  axom::Array<Point3D> points;

  for(int i = 0; i < 1e5; ++i)
  {
    auto x = axom::utilities::random_real( -2.0, 2.0 );
    auto y = axom::utilities::random_real( -2.0, 2.0 );
    auto z = axom::utilities::random_real( -2.0, 2.0 );
    points.push_back( Point3D { x, y, z } );
  }
  
  // Get the array of points
  auto sphere_faces = make_sphere_biquintic();

  // Evaluate the winding number at the points
  auto gwn = winding_number(points, sphere_faces);
}
```

For 2D, we run this block of code, which takes 108ms on the current develop branch and 18ms on the current branch.

```cpp
TEST(primal_integral, demo_test)
{
  primal::Point<double, 2> data[7] = {primal::Point<double, 2> {1.0, 0.0},
                                      primal::Point<double, 2> {1.0, 2.0},
                                      primal::Point<double, 2> {-1.0, 2.0},
                                      primal::Point<double, 2> {-1.0, 0.0},
                                      primal::Point<double, 2> {-1.0, -2.0},
                                      primal::Point<double, 2> {1.0, -2.0},
                                      primal::Point<double, 2> {1.0, 0.0}};
  double weights[7] = {1.0, 1. / 3., 1. / 3., 1.0, 1. / 3., 1. / 3., 1.0};
  double knots[11] = {0.0, 0.0, 0.0, 0.0, 0.5, 0.5, 0.5, 1.0, 1.0, 1.0, 1.0};

  primal::NURBSCurve<double, 2> circle(data, weights, 7, knots, 11);
  //primal::detail::NURBSCurveGWNCache circle_data(circle);

  // Set up an array of points to test against
  axom::Array<primal::Point<double, 2>> points;

  for(int i = 0; i < 1e5; ++i)
  {
    auto x = axom::utilities::random_real(-2.0, 2.0);
    auto y = axom::utilities::random_real(-2.0, 2.0);
    points.push_back(primal::Point<double, 2> {x, y});
  }

  // Evaluate the winding number at the points
  for(auto & pt : points)
    winding_number(pt, circle_data); // `circle` on develop
}
```

### Additional changes and bugfixes
- Updates the `quest_winding_number.cpp` example to utilize cached NURBS curves.
- An additional `NURBSPatch::expandParameterSpace` method that allows the patch to the expanded differently along each axis.
- Changes several internal intersection tolerances for consistency.
- Adds `isHalfOpen` option for ray-bezier intersection tests.
- Adds a method to evaluate average surface normals for trimmed surfaces.


src/axom/primal/operators/detail/intersect_bezier_impl.hpp
```diff
@@ -326,20 +327,20 @@ bool intersect_ray_bezier(const Ray<T, 2> &r,
                           double EPS,
                           int order,
                           double c_offset,
-                          double c_scale)
+                          double c_scale,
+                          bool isHalfOpen)
```
kennyweiss, line 332: Please add docs for this function, including the new `isHalfOpen` parameter

src/axom/primal/operators/winding_number.hpp
```diff
@@ -540,18 +654,17 @@ double winding_number(const Point<T, 3>& query,
  */
 template <typename T>
 double winding_number(const Point<T, 3>& query,
-                      const NURBSPatch<T, 3>& nPatch,
+                      const NURBSPatch<T, 3>& nurbs,
                       const double edge_tol = 1e-8,
                       const double ls_tol = 1e-8,
                       const double quad_tol = 1e-8,
+                      const double disk_size = 0.01,
                       const double EPS = 1e-8)
```
kennyweiss: Please update the docs, including the new parameter and renaming the old one

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -1568,6 +1599,40 @@ class NURBSPatch
...
+   */
+  void expandParameterSpace(double expansionAmount_u,
+                            double expansionAmount_v,
+                            bool removeTrimmingCurves = false)
+  {
+    SLIC_ASSERT(expansionAmount_u > 0.0 && expansionAmount_v > 0.0);
+    SLIC_WARNING_IF(expansionAmount_u > 1.15 * (getMaxKnot_u() - getMinKnot_u()) ||
+                      expansionAmount_v > 1.15 * (getMaxKnot_v() - getMinKnot_v()),
+                    "Expanding patch parameter space is numerically unstable "
+                    "for large values of expansionAmount.");
```
kennyweiss: Please clarify in this functions documentation if this is a value in parameter space or a percentage.
I'm thinking it's the latter since the knot span might not be normalized to `[0,1]`

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -2980,7 +3048,73 @@ class NURBSPatch
...
+      split_patches(i, 0).split_u(knot_vals_u[i + 1], split_patches(i, 0), split_patches(i + 1, 0));
+    }
+
+    for(int i = 0; i < num_knot_span_u; ++i)
+    {
+      for(int j = 0; j < num_knot_span_v - 1; ++j)
+      {
+        split_patches(i, j).split_v(knot_vals_v[j + 1], split_patches(i, j), split_patches(i, j + 1));
+      }
+    }
```
kennyweiss: Should this snippet be a function?  I.e. to decompose the NURBSPatch and trimming curves into a 2D array of single span patches?

src/axom/primal/operators/winding_number.hpp
```diff
@@ -254,6 +256,196 @@ double winding_number(const Point<T, 2>& q,
...
+ */
+template <typename T>
+double winding_number(const Point<T, 2>& q,
+                      const detail::NURBSCurveGWNCache<T>& nurbs_cache,
+                      bool& isOnCurve,
+                      double edge_tol = 1e-8,
+                      double EPS = 1e-8)
+{
+  // Early return is possible for most points + curves
+  if(!nurbs_cache.boundingBox().expand(edge_tol).contains(q))
```
kennyweiss: Do/Should we defer the bounding box expansion to the caller? 
It is likely (somewhat) more expensive to create and expand a bounding box than to just check for containment on the precomputed bounding box.

src/axom/primal/operators/winding_number.hpp
```diff
@@ -254,6 +256,196 @@ double winding_number(const Point<T, 2>& q,
...
+ * \param [in] q_arr The array of query point to test
+ * \param [in] c_arr The array of curve objects
+ * \param [in] edge_tol The physical distance level at which objects are considered indistinguishable
+ * \param [in] EPS Miscellaneous numerical tolerance level for nonphysical distances
+ * 
+ * \return The array of GWN values.
+ */
+template <typename T>
+axom::Array<double> winding_number(const axom::Array<Point<T, 2>>& q_arr,
+                                   const axom::Array<detail::NURBSCurveGWNCache<T>>& c_arr,
```
kennyweiss: I'm sure we can change this later, if necessary, but I wonder if the input arrays should be `axom::ArrayView` instead of `axom::ArrayView`.  That would give users more flexibility to call this w/ their own array types (other than `axom::Array`) without an explicit copy.

src/axom/primal/operators/winding_number.hpp
```diff
@@ -254,6 +256,196 @@ double winding_number(const Point<T, 2>& q,
...
+ * \brief Computes the GWN for an array of 2D points wrt to a 2D curved polygon
+ *
+ * \param [in] q_arr The query point to test
+ * \param [in] cpoly The CurvedPolygon object
+ * \param [in] edge_tol The physical distance level at which objects are considered indistinguishable
+ * \param [in] EPS Miscellaneous numerical tolerance level for nonphysical distances
+ *
+ * Computes the GWN for the curved polygon by summing the GWN for each curved edge
+ * 
+ * \warning Because the cache isKdiscarded immediately after computation,
```
kennyweiss: Typo?
```suggestion
 * \warning Because the cache is discarded immediately after computation,
```

src/axom/primal/operators/winding_number.hpp
kennyweiss, line 1: Really nice cleanup w/ lots of new options!

src/axom/primal/operators/detail/winding_number_2d_impl.hpp
```diff
@@ -284,6 +286,75 @@ double convex_endpoint_winding_number(const Point<T, 2>& q,
...
 }
 
+/*!
+ * \brief Computes the GWN for a 2D point wrt cached data for a 2D NURBS curve
+ *
+ * \param [in] q The query point to test
+ * \param [in] nurbs_cache The NURBS curve cached data object 
+ * \param [in] isOnEdge An returned flag if the point is on the curve
+ * \param [in] edge_tol The physical distance level at which objects are considered indistinguishable
+ * \param [in] EPS Miscellaneous numerical tolerance level for nonphysical distances
```
kennyweiss, line 300: Please update the params in the doxy comment

src/axom/primal/operators/detail/winding_number_2d_impl.hpp
```diff
@@ -284,6 +286,75 @@ double convex_endpoint_winding_number(const Point<T, 2>& q,
...
+                                                edge_tol,
+                                                EPS) +
+    detail::bezier_winding_number_memoized(q,
+                                           nurbs_cache,
+                                           bezier_idx,
+                                           refinement_level + 1,
+                                           2 * refinement_index + 1,
+                                           isOnCurve,
+                                           edge_tol,
+                                           EPS);
```
kennyweiss: Aside: Weird formatting on the `+`

src/axom/primal/operators/detail/winding_number_2d_memoization.hpp
```diff
@@ -0,0 +1,172 @@
...
+  SubdivisionData(const BezierCurve<T, 2>& a_curve, bool knownConvex) : curve(a_curve)
+  {
+    isConvexControlPolygon = knownConvex ? true : is_convex(Polygon<T, 2>(curve.getControlPoints()));
+    boundingBox = curve.boundingBox();
+  }
+
+  BezierCurve<T, 2> curve;
+  bool isConvexControlPolygon;
+  BoundingBox<T, 2> boundingBox;
+};
```
kennyweiss, line 66: Naming: 
* Should the name include `Bezier`? 
* This struct doesn't have much to do w/ subdivision

src/axom/primal/operators/detail/winding_number_2d_memoization.hpp
```diff
@@ -0,0 +1,172 @@
...
+{
+  SubdivisionData() = default;
+
+  SubdivisionData(const BezierCurve<T, 2>& a_curve, bool knownConvex) : curve(a_curve)
+  {
+    isConvexControlPolygon = knownConvex ? true : is_convex(Polygon<T, 2>(curve.getControlPoints()));
+    boundingBox = curve.boundingBox();
+  }
+
+  BezierCurve<T, 2> curve;
```
kennyweiss: Should there be a const accessor to a private curve to ensure people don't modify the curve?

src/axom/primal/operators/detail/winding_number_2d_memoization.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+private:
+  BoundingBox<T, 2> m_boundingBox;
+  int m_numControlPoints;
+  int m_degree;
+  int m_numSpans;
+
+  Point<T, 2> m_initPoint, m_endPoint;
+
+  mutable axom::Array<std::map<std::pair<int, int>, SubdivisionData<T>>> m_bezierSubdivisionMaps;
```
kennyweiss: Possible optimization for the future -- this might be more efficient w/ a hashmap (perhaps `axom::FlatMap` ?) than w/ a `std::map`.

(no need to make this change or even test it out in this PR)

src/axom/primal/operators/detail/winding_number_3d_memoization.hpp
```diff
@@ -0,0 +1,296 @@
...
+{
+  TrimmingCurveQuadratureData() = default;
+
+  /*!
+   * \brief Constructor for quadrature data from a single trimming curve on a patch
+   * 
+   * \param [in] quad_npts The number of Gaussian nodes
+   * \param [in] a_patch The 3D NURBS surface
+   * \param [in] a_curve_index The index of the trimming curve
+   * \param [in] a_refinementLevel How mnay subdivisions for the curve
```
kennyweiss: Typo:
```suggestion
   * \param [in] a_refinementLevel How many subdivisions for the curve
```

src/axom/primal/operators/detail/winding_number_3d_memoization.hpp
```diff
@@ -0,0 +1,296 @@
...
+    for(int i = 0; i < num_knot_span_u; ++i)
+    {
+      for(int j = 0; j < num_knot_span_v - 1; ++j)
+      {
+        split_patches(i, j).split_v(knot_vals_v[j + 1], split_patches(i, j), split_patches(i, j + 1));
+      }
+    }
+
+    // Bounding boxes should be defined according to the *pre-expanded* surface,
+    //  since the expanded portions are never visibile
```
kennyweiss: Typo:
```suggestion
    //  since the expanded portions are never visible
```

src/axom/primal/operators/detail/winding_number_3d_memoization.hpp
```diff
@@ -0,0 +1,296 @@
...
+                                                                                   quadNPts,
+                                                                                   refinementLevel,
+                                                                                   refinementIndex);
+    }
+
+    return m_curveQuadratureMaps[curveIndex][hash_key];
+  }
+
+private:
+  // The patch is private to present dirtying the cachce by changing the patch,
```
kennyweiss:
```suggestion
  // The patch is private to present dirtying the cache by changing the patch,
```

src/axom/primal/operators/detail/winding_number_3d_memoization.hpp
```diff
@@ -0,0 +1,296 @@
...
+  auto getMinKnot_v() const { return m_alteredPatch.getMinKnot_v(); }
+  auto getMaxKnot_v() const { return m_alteredPatch.getMaxKnot_v(); }
+  auto getTrimmingCurves() const { return m_alteredPatch.getTrimmingCurves(); };
+  auto getNumTrimmingCurves() const { return m_alteredPatch.getNumTrimmingCurves(); }
+  auto getParameterSpaceDiagonal() const { return m_pboxDiag; }
+
+  // Access precomputed data
+  auto getAverageNormal() const { return m_averageNormal; }
+  auto boundingBox() const { return m_bBox; }
+  auto orientedBoundingBox() const { return m_oBox; }
```
kennyweiss: Do we want to return copies? or const references?

kennyweiss: I suppose this should change if/when the corresponding functions in `NURBSPatch` change...

src/axom/primal/operators/detail/winding_number_3d_impl.hpp
```diff
@@ -127,16 +132,339 @@ void degenerate_surface_processing(const NURBSPatch<T, 3>& patch,
...
+{
+  // Can't rotate the patch as pre-processing if working with cached data
+  double quad = 0;
+  for(int n = 0; n < nurbs.getNumTrimmingCurves(); ++n)
+  {
+    // Get the quadrature points for the curve on the patch without any refinement
+    auto trimming_curve_data = nurbs.getTrimmingCurveQuadratureData(n, quad_npts, 0, 0);
+    double quad_coarse = stokes_gwn_component(query, ax, rotator, trimming_curve_data);
+
+    quad += 0.25 * M_1_PI *
```
kennyweiss: The compiler might do this for us, but it might be more efficient to pull the scaling term outside the loop as a `constexpr`,  e.g. 
```cpp
constexpr double sc = 0.25 * M_1_PI;
for(....)
{
   ...
   quad += sc * stokes_gwn_adaptive(...);
}
```

src/axom/primal/operators/detail/intersect_bezier_impl.hpp
jcs15c: There's documentation for this method in its declaration at the top of the file, but it was missing the new parameter. I'm happy to copy and paste it for the implementation too, but this is what's consistent with the rest of the file (to the extent that that's important).

src/axom/primal/operators/detail/winding_number_2d_impl.hpp
jcs15c: I wonder if it would be more clear with assigning each half to a variable and then adding those? I assume there wouldn't be a meaningful performance hit, and it feels like a cleaner solution than simply wrapping everything in a clang-format off.

src/axom/primal/operators/detail/winding_number_2d_memoization.hpp
jcs15c: Good point! It probably should just be called `BezierData` or `BezierCurveData`.

src/axom/primal/operators/detail/winding_number_2d_memoization.hpp
jcs15c: Good point. I was going a bit back and forth on whether or not it should be unchangeable in the first place, since it's really the cache object `NURBSCurveGWNCache` that shouldn't be able to have it's data changed arbitrarily, but at the very least you don't want the bounding box to become unmoored from the curve.

src/axom/primal/operators/detail/winding_number_2d_memoization.hpp
jcs15c: Right, I remember we talked about how `std::map` uses trees, right? It feels like it would be a natural fit for a subdivision structure like this, but I suppose there's no guarantee that map[3, 2], for example, would put the data at the 2nd index in the 3rd level (or, for that matter, if that'd even be the most efficient way to do it).

src/axom/primal/operators/detail/winding_number_3d_impl.hpp
jcs15c: Might as well!

src/axom/primal/operators/detail/winding_number_3d_memoization.hpp
jcs15c: I'm guessing that most of these methods should be const references, regardless of what the corresponding call to `NURBSPatch` is doing. There are a few though where that probably is too restrictive, and the method should just return a copy. I think `getAverageNormal()` is a good example, where a reasonable follow up operation is to call `Vector::unitVector()` on the result. Realistically, I should use `auto` a little less liberally here to make sure everything works as expected.

src/axom/primal/operators/detail/winding_number_2d_impl.hpp
BradWhitlock: Yes, please assign to temporaries and return the sum.

src/axom/primal/operators/detail/winding_number_3d_memoization.hpp
jcs15c: I also mispelled "prevent"!

BradWhitlock (approved):
Nice speedups.

src/axom/primal/operators/winding_number.hpp
jcs15c: I think that makes sense, but it means that `edge_tol` needs to be passed around in all the constructors for `NURBSCurveGWNCache` objects to ensure the expansion amount is consistent. I've changed it here and elsewhere to do this, do you think that works?

src/axom/primal/operators/winding_number.hpp
jcs15c: Presumably that should just be a drop-in replacement from `axom::Array` to `axom::ArrayView`, right?

src/axom/primal/operators/winding_number.hpp
jcs15c: Typo, and also outdated! This method is, in fact, accelerated by memoization.

src/axom/primal/geometry/NURBSPatch.hpp
jcs15c: Good idea, I forgot I have the same code in two places. I think in the future it's worth reconsidering the specific implementation, e.g. using Bezier extraction directly, but I doubt that's the bottleneck in any practical use.

src/axom/primal/operators/winding_number.hpp
kennyweiss: I think there's an implicit conversion from `Array` to `ArrayView`, but we might need to modify the calls from `winding_number(pt_arr, c_arr)` to `winding_number(pt_arr.view(), c_arr.view()` if we make this change.

src/axom/primal/operators/detail/winding_number_3d_memoization.hpp
jcs15c: Actually, I've found some really odd results swapping these to const references, so it's probably not as simple as I thought. For example, making the result of `getControlPoints()` a `const axom::Array<Point<double, 3>, 2>&` causes the method to return the wrong numerical result, and doing the equivalent for `getKnots_u()` results in a seg fault due to an invalid allocator ID. I'm going to have to think about this more...

suppressions.valgrind
kennyweiss, line 1: Presumably, this deletion was accidental and we still want the valgrind suppressions file. Please add it back.

jcs15c: Mortifying. I'll add it back promptly.

src/axom/primal/operators/detail/intersect_bezier_impl.hpp
kennyweiss: Thanks -- I think we can limit documenting the params to one place, e.g. the top of the file, but it should be consistent (e.g. document all parameters or none)

src/axom/primal/operators/detail/winding_number_3d_memoization.hpp
kennyweiss: Sounds like something to look into in the future

src/axom/primal/operators/winding_number.hpp
```diff
@@ -254,6 +256,205 @@ double winding_number(const Point<T, 2>& q,
   return ret_val;
 }
 
+/*!
+ * \brief Computes the GWN for a 2D point wrt memoized data for a 2D NURBS curve 
+ *
+ * \param [in] query The query point to test
+ * \param [in] nurbs_cache The NURBS curve cache data object containing memoized values
+ * \param [in] isOnCurve Set to true is the query point is on the curve
```
Arlie-Capps: `isOnCurve` should be [out] or [inout]

Arlie-Capps: I suspect all the Doxygen should be reviewed for this.

jcs15c: Very true. There's a couple of places where the [in]/[out]/[inout] labels are missing altogether. I'll go through and double-check everything.

PR #1654, opened 2025-09-10
Refactor clipping code

gunney1:
# Summary

- This PR is a refactoring of the clipping code in `IntersectionShaper`
- `IntersectionShaper` has 4 significant jobs that make it big and a bit unwieldy.
  1. Clip user-specified geometries against the cells of a mesh to compute overlap volumes.    Clipping accounts for more than 95% of the time spent in shaping and is the motivation for the current effort.  The intention is to provide fast clipping functionality for `IntersectionShaper` to use.
  2. Combine overlap volumes using replacement rules.
  3. Support multiple mesh formats (Blueprint and MFEM) stored in multiple hierarchy classes (Sidre and Conduit).
  4. Support multiple execution spaces to be selected by the user at run time.

- This PR
  - Rewrites the clipping part to support shape-specific approaches improve clipping performance.  Clipping involves some shape-specific operations and some common operations.  Both are re-factored to support trying faster algorithms.
  - Begin implementing shape-specific clipping approaches.  This part is work in progress.  The full functionality is complete but different approaches should be tried to get better performance.
  - Implement computation common to all shapes.  Multiple execution spaces are supported by factoring out those computations into a class templated on the execution space.
  - Provides a mesh wrapper that presents standard and derived mesh data as `ArrayView` objects, regardless of whether the mesh is Blueprint or MFEM, and stored in Sidre or Conduit.  This part is a work-in-progress in that MFEM support has not started.

# Completed work
- The `GeometryClipper` class provides the access point for the clipping functionality.
  - `GeometryClipper::clip` is the point of access for clipping functionality.
  - The `GeometryClipper::clip` method uses fast geometry-specific operations where they are available and falls back on more brute-force methods where they aren't.  (The brute-force method is currently the only choice.)
- Geometry-specific operations are implemented through the abstract base class `GeometryClipperStrategy`.
  - This is where geometry-specific operations are implemented (if available), such as fast screening operations and specialized clipping code.
    - Screening is the process of determining whether a mesh cell is completely outside or completely inside a geometry.  Cells that can be labeled this way are excluded from the much more expensive clipping operation.
    - An example of specialized clipping is for the plane geometry. which takes advantage of its simplicity.
  - The following classes implement the `GeometryClipperStrategy` interface:
    - `PlaneClipper` for a plane geometry
    - `TetClipper` for a tetrahedron
    - `HexClipper` for a hexahedron
    - `SphereClipper` for a sphere
    - `FSorClipper` for a surface of revolution where the discrete curve r(z) is a function.  Although the curve can have vertical segments, it cannot double back in z.  It can support stepped geometries but not hollow geometries.  Cones and cylinders are supported by this implementation.
    - `SorClipper` for surface of revolution where the r(z) curve can double back on itself, for example to make a bowl.
    - `TetMeshClipper` for a general geometry described by a tetrahedral mesh.
  - `ShapeeMesh` is the wrapper to that provides mesh data the clipping code.  Under the hood, does different things based on how the mesh layout.  `ShapeeMesh` also provides convenient derived data that multiple geometries may need, such as the mesh as an array of tetrahedra.


src/axom/sidre/core/View.hpp
```diff
@@ -671,6 +671,7 @@ class View
     //       a future optimization opportunity to split the
     if(m_state == EMPTY || m_state == TUPLE)
     {
+      // auto conduitAllocId = getValidConduitAllocatorID(allocID);
```
Arlie-Capps: remove if not needed

Arlie-Capps:
Brian, this is a lot of work!  I appreciate what you are trying to do here.  However, if you dump a huge PR on reviewers, it is very bad for engagement: your PR will be harder to review and take longer.

Break this up into several PRs:
- all the stuff in core
- all the stuff in primal, and klee/Geometry.hpp
- all the stuff in sidre, and the slam test

Each of these would be fairly uncontroversial and easy to review, moves you toward your goal, and helps reviewers feel interested in your progress.  This still leaves an enormous body of work within quest, which you can probably break up even further.

I figure you may not see the need to break up this PR at this stage, but please consider doing so.  Whatever the case, I'll keep on with my review.

Arlie-Capps:
Thank you for the performance comparison overview.  That's encouraging.

src/axom/quest/ShapeeMesh.hpp
Arlie-Capps, line 1: Please rename this file and class.  I suggest "ShapeMesh".

src/axom/quest/detail/Discretize_detail.hpp
```diff
@@ -165,6 +166,7 @@ int discrSeg(const Point2D &a, const Point2D &b, int levels, axom::ArrayView<Oct
   {
     return 0;
   }
+#endif
```
Arlie-Capps: Why comment out (old) lines 155 through 168?  These lines check for violated input assumptions and degenerate octs, which I still think may be valuable.  If we don't need or want these checks, why use `#if 0` ?

src/axom/quest/detail/Discretize_detail.hpp
```diff
@@ -265,7 +267,11 @@ bool discretize(const axom::ArrayView<Point2D> &polyline,
                 axom::Array<OctType> &out,
                 int &octcount)
 {
-  int allocId = axom::execution_space<ExecSpace>::allocatorID();
+  SLIC_ERROR_IF(!axom::execution_space<ExecSpace>::usesAllocId(out.getAllocatorID()),
+                axom::fmt::format("Execution space {} cannot access allocator ed {}",
```
Arlie-Capps: What is allocator ed?

src/axom/quest/detail/Discretize_detail.hpp
```diff
@@ -276,7 +282,7 @@ bool discretize(const axom::ArrayView<Point2D> &polyline,
     // invalid if a.x > b.x
     if(a[0] > b[0])
     {
-      stillValid = false;
+      // stillValid = false;
```
Arlie-Capps: Not sure we want to remove sanity checks, unless the algorithm deals properly with this input.  In that case, let's delete the entire if-statement and the comment on (old) line 282.

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,380 @@
...
+
+   * The base class owns this hierachy, but subclasses use it
+   * for their specific data.
+
+   * Use cases:
+   * - Construct object from Klee input in the form of hierarchy
+   *   data.
+   * - Print the object.
+
+   * Most if not all data should be in host memory.
```
Arlie-Capps: Please make this comment more specific.

src/axom/quest/MeshClipperStrategy.cpp
```diff
@@ -0,0 +1,140 @@
...
+namespace quest
+{
+namespace experimental
+{
+namespace internal
+{
+/*!
+ * \brief Implementation of a GeometryOperatorVisitor for processing klee shape operators
+ *
+ * This class extracts the matrix form of supported operators and marks the operator as unvalid otherwise
```
Arlie-Capps: What does "and marks the operator as unvalid otherwise" mean?

gunney1:
Depends on https://github.com/LLNL/axom/pull/1661

gunney1:
Depends on https://github.com/LLNL/axom/pull/1662

src/axom/quest/detail/Discretize_detail.hpp
gunney1: The reason is that we need to support the cases being asserted against.  Our SOR clipper needs to allow curves that have vertical segments and even double back in z.  I wasn't sure it the current implementation would work, so I kept it and just commented out the checks.  It does work.  I'll remove the commented-out code.

src/axom/quest/ShapeeMesh.hpp
gunney1: Done.

src/axom/sidre/core/View.hpp
gunney1: Done.

PR #1661, opened 2025-09-21
Fix crash querying whether an exec space uses an allocator id.

gunney1:
We can query whether an execution space uses a certain axom allocator id.  If we pass the invalid allocator id, it crashes.  This fixes the code to return false instead of crashing.

### DIFF
--- a/src/axom/core/memory_management.hpp
+++ b/src/axom/core/memory_management.hpp
@@ -29,7 +29,7 @@
 namespace axom
 {
 // To co-exist with Umpire allocator ids, use negative values here.
-constexpr int INVALID_ALLOCATOR_ID = -1;  //!< Place holder for no allocator
+constexpr int INVALID_ALLOCATOR_ID = -1;  //!< Place holder for no/unknown allocator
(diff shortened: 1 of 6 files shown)


src/axom/core/memory_management.hpp
```diff
@@ -455,6 +455,14 @@ inline int getAllocatorID<MemorySpace::Malloc>()
   return axom::MALLOC_ALLOCATOR_ID;
 }
 
+/**
+ * @brief Return the Axom MemorySpace for the given Axom allocator id.
+ *
+ *  For Umpire allocator ids, the MemorySpace is the Umpire memory
+ *  space.  For MALLOC_ALLOCATOR_ID, the MemorySpace is
+ *  MemorySpace::Malloc.  Other values have no corresponding MemorySpace
+ *  and will cause an unrecoverable exception.
```
kennyweiss: Clarify: It's calling `processAbort` -- is it fair to call that an unrecoverable exception? Please consider rephrasing.

gunney1: Will do.  Thanks.

BradWhitlock (approved):
Looks good to me.

Issue #1628, opened 2025-07-23
Improve tranformation support for Klee's Geometry operators


kennyweiss:
Klee shapes have `operators` that apply transformations (scale, translate, rotate, ...) to their geometry.

However, we don't make it easy to determine what the combined transformation matrix for a set of operators will be.
This should be easier now that we have several transformation functions in core -- https://github.com/LLNL/axom/blob/develop/src/axom/core/numerics/transforms.hpp

We should also add an optional `center` field to the `scale` operator to make it easier to specify the desired transformation.

PR #1667, opened 2025-09-24
Initial clipping code re-factor

gunney1:
# Summary

- This PR is an initial refactoring of the clipping code in `IntersectionShaper`.

- `IntersectionShaper` has 4 significant jobs that make it big and a bit unwieldy.
  1. Clip user-specified geometries against the cells of a mesh to compute overlap volumes.    Clipping accounts for more than 95% of the time spent in shaping and is the motivation for the current effort.  The intention is to provide fast clipping functionality for `IntersectionShaper` to use.
  2. Combine overlap volumes using replacement rules.
  3. Support multiple mesh formats (Blueprint and MFEM) stored in multiple hierarchy classes (Sidre and Conduit).
  4. Support multiple execution spaces to be selected by the user at run time.

- This PR
  - Implements the clipping code outside of `IntersectionShaper`.  The idea is to eventually remove the `IntersectionShaper`'s clipping code and use the new clipping code instead.
  - Introduces well-defined interfaces to support shape-specific approaches to improve clipping performance.  Clipping involves some shape-specific operations and some common operations.  Both have been re-factored to support a variety of potentially faster algorithms.
  - Begin implementing shape-specific clipping approaches.  The tetrahedral shape is included as an example.  More will follow in https://github.com/LLNL/axom/pull/1654.
  - Implement computation common to all shapes.  Multiple execution spaces are supported by factoring out those computations into a class templated on the execution space.
  - Provides a mesh wrapper that presents basic and derived mesh data as `ArrayView` objects, regardless of whether the mesh is Blueprint or MFEM, and stored in Sidre or Conduit.  This part is a work-in-progress in that MFEM support has not started.

PR https://github.com/LLNL/axom/pull/1654 depends on this one.

### DIFF
--- a/src/axom/quest/MeshClipper.cpp
+++ b/src/axom/quest/MeshClipper.cpp
@@ -1,0 +1,250 @@
+// Copyright (c) 2017-2025, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level LICENSE file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include "axom/config.hpp"
+
+#include "axom/quest/MeshClipper.hpp"
+#include "axom/quest/detail/clipping/MeshClipperImpl.hpp"
+#include "axom/core/execution/execution_space.hpp"
+#include "axom/core/execution/runtime_policy.hpp"
+#include "axom/slic/interface/slic_macros.hpp"
+#include "axom/fmt.hpp"
+
+namespace axom
+{
+namespace quest
+{
+namespace experimental
+{
+
+MeshClipper::MeshClipper(quest::experimental::ShapeMesh& shapeMesh,
+                         const std::shared_ptr<quest::experimental::MeshClipperStrategy>& strategy)
+  : m_shapeMesh(shapeMesh)
+  , m_strategy(strategy)
+  , m_impl(newImpl())
+  , m_verbose(false)
+{ }
+
+void MeshClipper::clip(axom::Array<double>& ovlap)
+{
+  const int allocId = m_shapeMesh.getAllocatorID();
+  const axom::IndexType cellCount = m_shapeMesh.getCellCount();
+
+  // Resize output array and use appropriate allocator.
+  if(ovlap.size() < cellCount || ovlap.getAllocatorID() != allocId)
+  {
+    AXOM_ANNOTATE_SCOPE("MeshClipper::clip_alloc");
+    ovlap = axom::Array<double>(ArrayOptions::Uninitialized(), cellCount, cellCount, allocId);
+  }
+  clip(ovlap.view());
+}
+
+/*
+ * Orchestrates the geometry clipping by using the capabilities of the
+ * MeshClipperStrategy implementation.
+ *
+ * If the strategy can label cells as inside/on/outside geometry
+ * boundary, use that to reduce reliance on expensive clipping methods.
+ *
+ * Regardless of labling, try to use specialized clipping first.
+ * If specialized methods aren't implemented, resort to discretizing
+ * geomety into tets or octs for clipping against mesh cells.
+ */
+void MeshClipper::clip(axom::ArrayView<double> ovlap)
+{
+  const int allocId = m_shapeMesh.getAllocatorID();
+  const axom::IndexType cellCount = m_shapeMesh.getCellCount();
+  SLIC_ASSERT(ovlap.size() == cellCount);
+
+  // Try to label cells as inside, outside or on shape boundary
+  axom::Array<char> cellLabels;
+  bool withCellInOut = m_strategy->labelCellsInOut(m_shapeMesh, cellLabels);
+
+  bool done = false;
+
+  if(withCellInOut)
+  {
+    SLIC_ERROR_IF(
+      cellLabels.size() != m_shapeMesh.getCellCount(),
+      axom::fmt::format("MeshClipperStrategy '{}' did not return the correct array size of {}",
+                        m_strategy->name(),
+                        m_shapeMesh.getCellCount()));
+    SLIC_ERROR_IF(cellLabels.getAllocatorID() != allocId,
+                  axom::fmt::format("MeshClipperStrategy '{}' failed to provide cellLabels data "
+                                    "with the required allocator id {}",
+                                    m_strategy->name(),
+                                    allocId));
+
+    if(m_verbose)
+    {
+      logLabelStats(cellLabels, "cells");
+    }
+
+    AXOM_ANNOTATE_BEGIN("MeshClipper::processInOut");
+
+    m_impl->initVolumeOverlaps(cellLabels.view(), ovlap);
+
+    axom::Array<axom::IndexType> cellsOnBdry;
+    m_impl->collectOnIndices(cellLabels.view(), cellsOnBdry);
+
+    axom::Array<char> tetLabels;
+    bool withTetInOut = m_strategy->labelTetsInOut(m_shapeMesh, cellsOnBdry.view(), tetLabels);
+
+    axom::Array<axom::IndexType> tetsOnBdry;
+
+    if(withTetInOut)
+    {
+      if(m_verbose)
+      {
+        logLabelStats(tetLabels, "tets");
+      }
+      m_impl->collectOnIndices(tetLabels.view(), tetsOnBdry);
+      m_impl->remapTetIndices(tetsOnBdry, cellsOnBdry);
+
+      SLIC_ASSERT(tetsOnBdry.getAllocatorID() == m_shapeMesh.getAllocatorID());
+      SLIC_ASSERT(tetsOnBdry.size() <= cellsOnBdry.size() * TETS_PER_HEXAHEDRON);
+
+      m_impl->addVolumesOfInteriorTets(cellsOnBdry.view(), tetLabels.view(), ovlap);
+    }
+
+    AXOM_ANNOTATE_END("MeshClipper::processInOut");
+
+    //
+    // If implementation has a specialized clip, use it.
+    //
+    if(withTetInOut)
+    {
+      done = m_strategy->specializedClipTets(m_shapeMesh, ovlap, tetsOnBdry);
+    }
+    else
+    {
+      done = m_strategy->specializedClipCells(m_shapeMesh, ovlap, cellsOnBdry);
+    }
+
+    if(!done)
+    {
+      AXOM_ANNOTATE_SCOPE("MeshClipper::clip3D_limited");
+      if(withTetInOut)
+      {
+        m_impl->computeClipVolumes3DTets(tetsOnBdry.view(), ovlap);
+      }
+      else
+      {
+        m_impl->computeClipVolumes3D(cellsOnBdry.view(), ovlap);
+      }
+    }
+
+    m_localCellInCount = cellsOnBdry.size();
+  }
+  else  // !withCellInOut
+  {
+    std::string msg =
+      axom::fmt::format("MeshClipper strategy '{}' did not provide in/out cell labels.\n",
+                        m_strategy->name());
+    SLIC_INFO(msg);
+    m_impl->initVolumeOverlaps(ovlap);
+    done = m_strategy->specializedClipCells(m_shapeMesh, ovlap);
+
+    if(!done)
+    {
+      AXOM_ANNOTATE_SCOPE("MeshClipper::clip3D");
+      m_impl->computeClipVolumes3D(ovlap);
+    }
+
+    m_localCellInCount = cellCount;
+  }
+}
+
+/*!
+ * @brief Allocate a Impl for the execution-space computations
+ * of this clipper.
+ */
+std::unique_ptr<MeshClipper::Impl> MeshClipper::newImpl()
+{
+  using RuntimePolicy = axom::runtime_policy::Policy;
+
+  auto runtimePolicy = m_shapeMesh.getRuntimePolicy();
+
+  std::unique_ptr<Impl> impl;
+  if(runtimePolicy == RuntimePolicy::seq)
+  {
+    impl.reset(new detail::MeshClipperImpl<axom::SEQ_EXEC>(*this));
+  }
+#ifdef AXOM_RUNTIME_POLICY_USE_OPENMP
+  else if(runtimePolicy == RuntimePolicy::omp)
+  {
+    impl.reset(new detail::MeshClipperImpl<axom::OMP_EXEC>(*this));
+  }
+#endif
+#ifdef AXOM_RUNTIME_POLICY_USE_CUDA
+  else if(runtimePolicy == RuntimePolicy::cuda)
+  {
+    impl.reset(new detail::MeshClipperImpl<axom::CUDA_EXEC<256>>(*this));
+  }
+#endif
+#ifdef AXOM_RUNTIME_POLICY_USE_HIP
+  else if(runtimePolicy == RuntimePolicy::hip)
+  {
+    impl.reset(new detail::MeshClipperImpl<axom::HIP_EXEC<256>>(*this));
+  }
+#endif
+  else
+  {
+    SLIC_ERROR(axom::fmt::format("MeshClipper has no impl for runtime policy {}", runtimePolicy));
+  }
+  return impl;
+}
+
+void MeshClipper::logLabelStats(axom::ArrayView<const LabelType> labels, const std::string& labelType)
+{
+  axom::IndexType countsa[4];
+  axom::IndexType countsb[4];
+  getLabelCounts(labels, countsa[0], countsa[1], countsa[2]);
+  countsa[3] = m_shapeMesh.getCellCount();
+#ifdef AXOM_USE_MPI
+  MPI_Reduce(countsa, countsb, 4, axom::mpi_traits<axom::IndexType>::type, MPI_SUM, 0, MPI_COMM_WORLD);
+#endif
+  std::string msg = axom::fmt::format(
+    "MeshClipper strategy '{}' globally labeled {} {} inside, {} on and {} outside, for mesh with "
+    "{} cells ({} tets)\n",
+    m_strategy->name(),
+    labelType,
+    countsb[0],
+    countsb[1],
+    countsb[2],
+    countsb[3],
+    countsb[3] * TETS_PER_HEXAHEDRON);
+  SLIC_INFO(msg);
+}
+
+void MeshClipper::getClippingStats(axom::IndexType& localCellInCount,
+                                   axom::IndexType& globalCellInCount,
+                                   axom::IndexType& maxLocalCellInCount) const
+{
+  localCellInCount = m_localCellInCount;
+#ifdef AXOM_USE_MPI
+  MPI_Reduce(&localCellInCount,
+             &globalCellInCount,
+             1,
+             axom::mpi_traits<axom::IndexType>::type,
+             MPI_SUM,
+             0,
+             MPI_COMM_WORLD);
+  MPI_Reduce(&localCellInCount,
+             &maxLocalCellInCount,
+             1,
+             axom::mpi_traits<axom::IndexType>::type,
+             MPI_MAX,
+             0,
+             MPI_COMM_WORLD);
+#else
+  maxLocalCellInCount = localCellInCount;
+  globalCellInCount = localCellInCount;
+#endif
+}
+
+}  // namespace experimental
+}  // end namespace quest
+}  // end namespace axom
--- a/src/axom/quest/MeshClipper.hpp
+++ b/src/axom/quest/MeshClipper.hpp
@@ -1,0 +1,206 @@
+// Copyright (c) 2017-2025, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#ifndef AXOM_QUEST_MESHCLIPPER_HPP
+#define AXOM_QUEST_MESHCLIPPER_HPP
+
+#include "axom/config.hpp"
+
+#include "axom/klee/Geometry.hpp"
+#include "axom/quest/MeshClipperStrategy.hpp"
+#include "axom/quest/ShapeMesh.hpp"
+
+namespace axom
+{
+namespace quest
+{
+namespace experimental
+{
+
+/*!
+ * @brief Abstract base class encapsulating the clipping task in
+ * intersection shaping, to be specialized for geometry-specific
+ * implementations.
+ *
+ * Each object operates is associated with a single klee::Geometry object.*
+ *
+ * This class defines the interfaces for specializing computations
+ * that can run fast on specific geometries.  This allows for
+ * geometry-specific optimizations.  If no such specialization
+ * is provided, default methods are used.
+ */
+class MeshClipper
+{
+public:
+  //!@brief Whether an element is in, out or on shape boundary.
+  using LabelType = MeshClipperStrategy::LabelType;
+
+  static constexpr axom::IndexType TETS_PER_HEXAHEDRON = MeshClipperStrategy::TETS_PER_HEXAHEDRON;
+
+  /*!
+   * @brief Construct a shape clipper
+   *
+   * @param [in/out] bpMesh Single-domain Blueprint mesh
+   *   to shape into.
+   * @param [in] strategy Strategy where external
+   *   shape-dependent operations are implemented.
+   *
+   * @c bpMesh must be an unstructured hex mesh.
+   * That is the only type currently supported.
+   */
+  MeshClipper(quest::experimental::ShapeMesh& shapeMesh, const std::shared_ptr<MeshClipperStrategy>& strategy);
+
+  //!@brief The mesh.
+  ShapeMesh& getShapeMesh() { return m_shapeMesh; }
+
+  //!@brief Allocator id to be used for all array data.
+  int getAllocatorID() const { return m_shapeMesh.getAllocatorID(); }
+
+  void setVerbose(bool verbose) { m_verbose = verbose; }
+
+  /*!
+   * @brief Clip
+   *
+   * @param ovlap [out] Shape overlap volume of each cell
+   *   in the shapee mesh.
+   */
+  void clip(axom::Array<double>& ovlap);
+
+  /*!
+   * @brief Clip
+   *
+   * @param ovlap [out] Shape overlap volume of each cell
+   *   in the shapee mesh.
+   */
+  void clip(axom::ArrayView<double> ovlap);
+
+  //!@brief Dimension of the shape (2 or 3)
+  int dimension() const { return m_shapeMesh.dimension(); }
+
+  /*!
+   * @brief Single interface for methods implemented with
+   * execution space templates.
+   *
+   * These methods require messy instantiation of
+   * execution spaces and their runtime selection.
+   *
+   * The implementations are in class detail::MeshClipperImpl,
+   * which is templated on execution space.
+   */
+  struct Impl
+  {
+    Impl(MeshClipper& impl) : m_myClipper(impl) { }
+    virtual ~Impl() = default;
+
+    static constexpr axom::IndexType TETS_PER_HEXAHEDRON = MeshClipperStrategy::TETS_PER_HEXAHEDRON;
+
+    /*!
+     * @brief Initialize overlap volumes to full for cells completely
+     * inside the shape and zero for cells outside or on shape boundary.
+     */
+    virtual void initVolumeOverlaps(const axom::ArrayView<MeshClipperStrategy::LabelType>& labels,
+                                    axom::ArrayView<double> ovlap) = 0;
+
+    //! @brief Initialize overlap volumes to zero.
+    virtual void initVolumeOverlaps(axom::ArrayView<double> ovlap) = 0;
+
+    //!@brief Collect unlabeled LABEL_ON indices into an index list.
+    virtual void collectOnIndices(const axom::ArrayView<LabelType>& labels,
+                                  axom::Array<axom::IndexType>& onIndices) = 0;
+
+    //!@brief Change tet indices from sparse-set values to full-set values.
+    virtual void remapTetIndices(axom::ArrayView<axom::IndexType> tetsOnBdry,
+                                 axom::ArrayView<const axom::IndexType> cellsOnBdry) = 0;
+
+    //!@brief Add volumes of tets inside the geometry to the volume data.
+    virtual void addVolumesOfInteriorTets(axom::ArrayView<const axom::IndexType> cellsOnBdry,
+                                          axom::ArrayView<const LabelType> tetLabels,
+                                          axom::ArrayView<double> ovlap) = 0;
+
+    //!@brief Compute clip volumes for every cell.
+    virtual void computeClipVolumes3D(axom::ArrayView<double> ovlap) = 0;
+
+    //!@brief Compute clip volumes for cell in an index list.
+    virtual void computeClipVolumes3D(const axom::ArrayView<axom::IndexType>& cellIndices,
+                                      axom::ArrayView<double> ovlap) = 0;
+
+    /*!
+     * @brief Compute clip volumes for cell tets in an index list.
+     *
+     * The tets are the results from decomposing each cell hex into
+     * TETS_PER_HEXAHEDRON tets and stored consecutively.
+     */
+    virtual void computeClipVolumes3DTets(const axom::ArrayView<axom::IndexType>& tetIndices,
+                                          axom::ArrayView<double> ovlap) = 0;
+
+    //!@brief Count the number of labels of each type.
+    virtual void getLabelCounts(axom::ArrayView<const LabelType> labels,
+                                axom::IndexType& inCount,
+                                axom::IndexType& onCount,
+                                axom::IndexType& outCount) = 0;
+
+    ShapeMesh& getShapeMesh() { return m_myClipper.m_shapeMesh; }
+
+    MeshClipperStrategy& getStrategy() { return *m_myClipper.m_strategy; }
+
+  private:
+    //!@brief The MeshClipper that owns this Impl.
+    MeshClipper& m_myClipper;
+  };
+
+  //! @brief For assessments, not general use.
+  void getClippingStats(axom::IndexType& localCellInCount,
+                        axom::IndexType& globalCellInCount,
+                        axom::IndexType& maxLocalCellInCount) const;
+
+private:
+  friend Impl;
+
+  quest::experimental::ShapeMesh& m_shapeMesh;
+
+  //! @brief Shape-specific operations in clipping.
+  std::shared_ptr<quest::experimental::MeshClipperStrategy> m_strategy;
+
+  //! @brief Object where execution space code is instantiated.
+  std::unique_ptr<Impl> m_impl;
+
+  /* NOTE: MeshClipperStrategy is for shape-specific functions,
+   * implemented externally.  Impl implements internal algorithms
+   * for multiple execution spaces.
+   */
+
+  ///@{
+  //! @name Statistics
+  axom::IndexType m_localCellInCount {0};
+  ///@}
+
+  bool m_verbose;
+
+#if defined(__CUDACC__)
+public:
+#endif
+  //!@brief Allocate a delegate for m_shapeMesh's runtime policy.
+  std::unique_ptr<Impl> newImpl();
+
+  //@{
+  //!@name Convenience methods
+  //!@brief Count the number of labels of each type.
+  void getLabelCounts(const axom::Array<LabelType>& labels,
+                      axom::IndexType& inCount,
+                      axom::IndexType& onCount,
+                      axom::IndexType& outCount)
+  {
+    m_impl->getLabelCounts(labels, inCount, onCount, outCount);
+  }
+
+  void logLabelStats(axom::ArrayView<const LabelType> labels, const std::string& labelType);
+  //@}
+};
+
+}  // namespace experimental
+}  // namespace quest
+}  // namespace axom
+
+#endif  // AXOM_QUEST_MESHCLIPPER_HPP
--- a/src/axom/quest/MeshClipperStrategy.cpp
+++ b/src/axom/quest/MeshClipperStrategy.cpp
@@ -1,0 +1,141 @@
+// Copyright (c) 2017-2025, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level LICENSE file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include "axom/config.hpp"
+
+#include "axom/quest/MeshClipperStrategy.hpp"
+#include "axom/klee/GeometryOperators.hpp"
+
+namespace axom
+{
+namespace quest
+{
+namespace experimental
+{
+namespace internal
+{
+/*!
+ * \brief Implementation of a GeometryOperatorVisitor for processing klee shape operators
+ *
+ * This class extracts the matrix form of supported operators.
+ * For other operators, it sets the valid flag to false.
+ * To use, check the \a isValid() function after visiting and then call the \a getMatrix() function.
+ */
+class AffineMatrixVisitor : public klee::GeometryOperatorVisitor
+{
+public:
+  AffineMatrixVisitor() : m_matrix(4, 4) { }
+
+  void visit(const klee::Translation& translation) override
+  {
+    m_matrix = translation.toMatrix();
+    m_isValid = true;
+  }
+  void visit(const klee::Rotation& rotation) override
+  {
+    m_matrix = rotation.toMatrix();
+    m_isValid = true;
+  }
+  void visit(const klee::Scale& scale) override
+  {
+    m_matrix = scale.toMatrix();
+    m_isValid = true;
+  }
+  void visit(const klee::UnitConverter& converter) override
+  {
+    m_matrix = converter.toMatrix();
+    m_isValid = true;
+  }
+
+  void visit(const klee::CompositeOperator&) override
+  {
+    SLIC_WARNING("CompositeOperator not supported for Shaper query");
+    m_isValid = false;
+  }
+  void visit(const klee::SliceOperator&) override
+  {
+    SLIC_WARNING("SliceOperator not yet supported for Shaper query");
+    m_isValid = false;
+  }
+
+  const numerics::Matrix<double>& getMatrix() const { return m_matrix; }
+  bool isValid() const { return m_isValid; }
+
+private:
+  bool m_isValid {false};
+  numerics::Matrix<double> m_matrix;
+};
+
+}  // end namespace internal
+
+MeshClipperStrategy::MeshClipperStrategy(const klee::Geometry& kGeom)
+  : m_info(kGeom.asHierarchy())
+  , m_extTrans(computeTransformationMatrix(kGeom.getGeometryOperator()))
+{ }
+
+const std::string& MeshClipperStrategy::name() const
+{
+  static const std::string n = "UNNAMED";
+  return n;
+}
+
+const axom::primal::BoundingBox<double, 2>& MeshClipperStrategy::getBoundingBox2D() const
+{
+  static const axom::primal::BoundingBox<double, 2> invalidBb2d;
+  return invalidBb2d;
+}
+
+const axom::primal::BoundingBox<double, 3>& MeshClipperStrategy::getBoundingBox3D() const
+{
+  static const axom::primal::BoundingBox<double, 3> invalidBb3d;
+  return invalidBb3d;
+}
+
+numerics::Matrix<double> MeshClipperStrategy::computeTransformationMatrix(
+  const std::shared_ptr<const axom::klee::GeometryOperator>& op) const
+{
+  const auto identity4x4 = numerics::Matrix<double>::identity(4);
+  numerics::Matrix<double> transformation(identity4x4);
+  if(op)
+  {
+    auto composite = std::dynamic_pointer_cast<const klee::CompositeOperator>(op);
+    if(composite)
+    {
+      // Concatenate the transformations
+
+      // Why don't we multiply the matrices in CompositeOperator::addOperator()?
+      // Why keep the matrices factored and multiply them here repeatedly?
+      // Combining them would also avoid this if-else logic.  BTNG
+      for(auto op : composite->getOperators())
+      {
+        // Use visitor pattern to extract the affine matrix from supported operators
+        internal::AffineMatrixVisitor visitor;
+        op->accept(visitor);
+        if(!visitor.isValid())
+        {
+          continue;
+        }
+        const auto& matrix = visitor.getMatrix();
+        numerics::Matrix<double> res(identity4x4);
+        numerics::matrix_multiply(matrix, transformation, res);
+        transformation = res;
+      }
+    }
+    else
+    {
+      internal::AffineMatrixVisitor visitor;
+      op->accept(visitor);
+      if(visitor.isValid())
+      {
+        transformation = visitor.getMatrix();
+      }
+    }
+  }
+  return transformation;
+}
+
+}  // namespace experimental
+}  // end namespace quest
+}  // end namespace axom
--- a/src/axom/quest/MeshClipperStrategy.hpp
+++ b/src/axom/quest/MeshClipperStrategy.hpp
@@ -1,0 +1,377 @@
+// Copyright (c) 2017-2025, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#ifndef AXOM_QUEST_GEOMETRYCLIPPERSTRATEGY_HPP
+#define AXOM_QUEST_GEOMETRYCLIPPERSTRATEGY_HPP
+
+#include "axom/config.hpp"
+
+#include "axom/core/Array.hpp"
+#include "axom/klee/Geometry.hpp"
+#include "axom/quest/ShapeMesh.hpp"
+#include "axom/primal.hpp"
+
+// Requires Conduit for storing hierarchy-form data.
+#include "conduit_blueprint.hpp"
+
+namespace axom
+{
+namespace quest
+{
+namespace experimental
+{
+
+/*!
+ * @brief Strategy base class for clipping operations for specific
+ * geometry instances.
+
+ * Key methods to implement:  (Some combination of these is required.)
+
+ * -# @c getBoundingBox2D or @c getBoundingBox3D: Axis-alligned
+ *    bounding box for the geometry.
+
+ * -# @c labelCellsInOut: Label whether the cells in a mesh is inside,
+ *    outside or on the shape boundary.  If a cell cannot be
+ *    determined, you can conservatively label it as on the boundary.
+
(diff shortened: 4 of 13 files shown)


src/axom/quest/ShapeMesh.hpp
```diff
@@ -0,0 +1,325 @@
...
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#ifndef AXOM_QUEST_SHAPEEMESH_HPP
+#define AXOM_QUEST_SHAPEEMESH_HPP
+
+#ifndef AXOM_USE_CONDUIT
+  #error "ShapeMesh requires Conduit"
+// TODO: Support MFEM as well.
+#endif
```
kennyweiss, line 20: You need to 
```
#include "axom/config.hpp"
```
before checking for compile defines like `AXOM_USE_CONDUIT`

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+namespace experimental
+{
+
+/*!
+ * @brief Strategy base class for clipping operations for specific
+ * geometry instances.
+
+ * Key methods to implement:  (Some combination of these is required.)
+
+ * -# @c getBoundingBox2D or @c getBoundingBox3D: Axis-alligned
```
Arlie-Capps: please correct to "axis-aligned"

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+ *    bounding box for the geometry.
+
+ * -# @c labelCellsInOut: Label whether the cells in a mesh is inside,
```
Arlie-Capps: Please rephrase to "Label each cell in a mesh as inside,"

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+ *    available) to clip the cells in a mesh.  Implementation should
+ *    use special knowledge of the geometry.  One version of this
+ *    method clips all cells in the mesh and the other clips only
+ *    cells in a provided index list.  The latter works in
+ *    conjunction with @c labelCellsInOut.
+
+ * Every method should return true if it fulfilled the request, or
+ * false if it was a no-op.
+
+ * Implementations of this strategy must provide either
```
Arlie-Capps: Suggestion: change to "Subclasses of MeshClipperStrategy must implement either"

This changes the meaning of the sentence, because now you're discussing the C++ class rather than the idea (the strategy) embodied in the class.  I think it makes the Doxygen more specific and therefore clearer.

src/axom/quest/MeshClipperStrategy.hpp
Arlie-Capps: (throughout files)

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+  virtual const axom::primal::BoundingBox<double, 2>& getBoundingBox2D() const;
+
+  /*!
+   * @brief Get the 3D axis-alligned bounding box for the geometry,
+   * if it's applicable and available.
+  */
+  virtual const axom::primal::BoundingBox<double, 3>& getBoundingBox3D() const;
+
+  /*!
+   * @brief Label the cells in the mesh as inside, outside or
```
Arlie-Capps: Suggest: "Label each cell in the mesh as..."

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+   * The cell labels should be set to
+   * - @c labelIn if the cell is completely inside the shape,
+   * - @c labelOut if the cell is completely outside, and
+   * - @c labelOn if the cell is both inside and outside (or
+   *   cannot be easily determined).
+   *
+   * The output labels are used in optimizing the clipping algorithm.
+   * Subclasses should implementation this if it's cost-effective, and
+   * skip if it's not.  It's safe to label cells as on the boundary if
+   * it can't be possitively determined as inside or outside.
```
Arlie-Capps:
```suggestion
   * it can't be positively determined as inside or outside.
```

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+   *
+   * @return Whether the operation was done.  (A false means
+   * not done.)
+   *
+   * If implemenation returns true, it should ensure these
```
Arlie-Capps:
```suggestion
   * If implementation returns true, it should ensure these
```

Arlie-Capps: Please check for this typo throughout the files, as well as "possitively".

src/axom/quest/MeshClipperStrategy.hpp
Arlie-Capps: I think "each cell" or "each tet" is clearer and more specific.  But I realize that you have this pattern throughout the doxygen of this file (and possibly outside).  I think it should be consistently one way or the other, and I realize this adds to the work.  So take my suggestion for what it is (just a style suggestion).

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+  /*!
+   * @brief Label the tetrahedra in certain cells, if possible.
+   *
+   * @param [in] shapeMesh Blueprint mesh to shape into.
+   * @param [in] cellIds Indices of cells whose constituent
+   *   tets should be labeled.
+   * @param [out] tetLabels Output
+   *
+   * Only the cells labeled as ON the boundary are subjected to
+   * this labeling.
```
Arlie-Capps: I'm not sure I understand or agree with this sentence.  The argument `cellIds` specifies which cells are subjected to this labeling.  It should be perfectly fine for the calling code to just ask for random cells throughout the mesh to have tets labeled.  Please remove this sentence, unless I'm completely missing something.

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+  */
+  virtual bool labelCellsInOut(quest::experimental::ShapeMesh& shapeMesh, axom::Array<LabelType>& cellLabels)
+  {
+    AXOM_UNUSED_VAR(shapeMesh);
+    AXOM_UNUSED_VAR(cellLabels);
+    return false;
+  }
+
+  /*!
+   * @brief Label the tetrahedra in certain cells, if possible.
```
Arlie-Capps: We need documentation on the ordering of tets inside cells.  I'm sure that there is a well-known convention in code that expects to call this routine, but that needs to be documented.  If it's here already and I didn't see it, please add a reference from this routine's doxygen and anywhere else that depends on the ordering so users can understand it.

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+    return false;
+  }
+
+  /*!
+   * @brief Clip with a fast geometry-specialized method if
+   * possible.
+   *
+   * @param [in] shapeMesh Blueprint mesh to shape into.
+   * @param ovlap [out] Shape overlap volume of each cell
+   *   in the shapee mesh.  It's initialized to zeros.
```
Arlie-Capps:
```suggestion
   *   in \a shapeMesh.  It's initialized to zeros.
```

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+    AXOM_UNUSED_VAR(tetLabels);
+    return false;
+  }
+
+  /*!
+   * @brief Clip with a fast geometry-specialized method if
+   * possible.
+   *
+   * @param [in] shapeMesh Blueprint mesh to shape into.
+   * @param ovlap [out] Shape overlap volume of each cell
```
Arlie-Capps: Does this need to be changed?  I'm not sure.
```suggestion
   * @param [out] ovlap Shape overlap volume of each cell
```

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+                                   const axom::ArrayView<IndexType>& tetIds)
+  {
+    AXOM_UNUSED_VAR(shapeMesh);
+    AXOM_UNUSED_VAR(ovlap);
+    AXOM_UNUSED_VAR(tetIds);
+    return false;
+  }
+
+  /*!
+   * @brief Get the geometry as discrete tetrahedra, or return false.
```
Arlie-Capps, line 311: As noted earlier, we need to document the canonical ordering of tets and octs within a hexahedron, or if it's already documented, provide a pointer from here to the information.

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+    AXOM_UNUSED_VAR(shapeMesh);
+    AXOM_UNUSED_VAR(octs);
+    return false;
+  }
+
+  //@}
+
+protected:
+  /*!
+   * @brief Free-form representation of the concrete object.
```
Arlie-Capps: "Free-form?"  Do you mean Conduit Blueprint?  Please explain.

src/axom/quest/MeshClipperStrategy.hpp
gunney1: Good suggestion!

src/axom/quest/MeshClipperStrategy.hpp
gunney1: You're right.

src/axom/quest/MeshClipperStrategy.hpp
gunney1: How about this:  Indices [i*NUM_TETS_PER_HEX, (i+1)*NUM_TETS_PER_HEX) in \c tetLabels correspond to parent cell index \c c = \c cellIds[i].  The \c NUM_TETS_PER_HEX tets in cell \c c have indices [c*NUM_TETS_PER_HEX, (c+1)*NUM_TETS_PER_HEX) in \c shapeMesh.getCellsAsTets().

It's a bit complex because of the extra level of indirection in tetLabels.

Note that the order of tets inside cells is not up to this class.  It's up to `ShapeMesh`, hence the refernce to `shapeMesh.getCellsAsTets()`.

src/axom/quest/MeshClipperStrategy.hpp
gunney1: I think it does.

src/axom/quest/MeshClipperStrategy.hpp
gunney1: The ordering is up to `ShapeMesh`, through the method `getCellsAsTets`, and only indirectly.  It's actually in `primal::Hexahedron::trianglate`.  I'll add references to `getCellsAsTets` in the doxygen comments.  I'm not sure it's a good idea to spell out exactly what that ordering is in `quest` code.  `Hexahedron` may change.  And, even if it doesn't, my plans include trying other triangulations of a hex in order to improve clipping performance.

src/axom/quest/MeshClipperStrategy.hpp
gunney1: Free-form means `MeshClipperStrategy` doesn't specify the format.  It merely holds the information so the subclasses can access it.  It's the subclasses that actually uses the information and therefore defines the content and structure of that information.  I think the detailed documentation for `m_info` should be in the subclasses and will add it for the `TetClipper` class introduced in this PR.

src/axom/quest/MeshClipperStrategy.hpp
Arlie-Capps: Thank you.  I think you are right: we need a pointer here (and everywhere in this class that depends on the user knowing tets-order) to `ShapeMesh::getCellsAsTets`.  I'm putting more comments on that method in this PR.

src/axom/quest/MeshClipperStrategy.hpp
Arlie-Capps: More discussion at line 297 in this PR.

src/axom/quest/ShapeMesh.hpp
```diff
@@ -0,0 +1,325 @@
...
+   * This data is dynamically generated as needed, and cached for use
+   * by multiple geometry clippers.  The idea is to eliminate redundant
+   * code and computations.
+   */
+  /*!
+   * @brief Tetrahedral version of mesh cells with cell i having tet ids
+   * [i*NUM_TETS_PER_HEX, (i+1)*NUM_TETS_PER_HEX).
+  */
+  axom::ArrayView<const TetrahedronType> getCellsAsTets();
+  axom::ArrayView<const HexahedronType> getCellsAsHexes();
```
Arlie-Capps, line 175: In the Doxygen here, please write that the order of tets in each cell (returned by `getCellsAsTets`) depends on `primal::Hexahedron::triangulate`.

In the Doxygen for `primal::Hexahedron::triangulate`, please describe the order of tets returned.  It really would be best to get that description into this PR, but if it doesn't make it into this PR, it absolutely has to get in to a future PR (perhaps the one moving this code out of `experimental`).  Without documenting the tet order, I am afraid this functionality will be much less useful, since the user would have to go read the source.  Thankfully Axom is open source and that is possible, but let's make it easy on the users.

I also think we need a Sphinx page explaining all this.  Not necessary for this PR.  I'll be happy to help with writing that.

src/axom/quest/MeshClipperStrategy.hpp
Arlie-Capps: I see your point.  Then, please rephrase by removing "Free-form", since it isn't free-form---it's determined by `Geometry::populateGeomInfo`.  To me, "Free-form" means that the Node has no rules.

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+   *
+   * The constructor initializes this as a deep copy of the source
+   * klee::Geometry hierarchy data.  Subclasses may use and change this
+   * data as needed.
```
Arlie-Capps: Please refer readers to `Geometry::asHierarchy`, since that's the public interface that determines what goes into `m_info`.

Also, please flesh out the Doxygen for `Geometry::asHierarchy`, hopefully in this PR, but if not, most definitely needs to be in a future PR.  And I'm willing to help write Sphinx documentation discussing this convention.  (See conversation at `ShapeMesh::getCellsAsHexs`, ShapeMesh.hpp line 158.)

src/axom/quest/ShapeMesh.hpp
gunney1: I'm not sure it's essential to document the tet order.  All the info needed to describe the tet is in the `Tetrahedron` class.

When I loop through the 24 tets, I get a `Tetrahedron` object  each loop with all the info I need.  It doesn't matter which 4 out of the 15 points define the tet, (it's 15, because we add 7 points when triangulating) or what order the tets come in.

You may be thinking how can two independent parts of the code be consistent if one doesn't know the order that the other uses.  They are consistent because they both use the same tet array.  I guess this is another reason to have `ShapeMesh` compute the tet array.  BTW, in the future, we may use a different triangulation to improve performance.

Let me know if this isn't clear.

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  /*!
+   * @brief Get the geometry as discrete tetrahedra, or return false.
+   *
+   * @param [in] shapeMesh Blueprint mesh to shape into.
+   * @param [out] tets Array of tetrahedra filling the space of the shape,
+   * fully transformed.
+   *
+   * All vertex coordinates close to zero should be snapped to zero.
```
Arlie-Capps: Please rephrase to "Subclasses implementing this routine should snap to zero any output vertex coordinate that is close to zero."  This makes it clear that determining "close to zero" and doing "snap to zero" are the responsibilities of the implementing subclass.

src/axom/quest/MeshClipperStrategy.hpp
gunney1: Definitely!  While `klee::Geometry` is transitioning to using hierarchy information exclusively, `m_info` is populated by data passed to the constructor.  I believe that the appropriate design pattern is called "abstract data holder".  The best practice for it (according to LivChat) is have the subclasses validate the hierarchy data, fail fast if it's not valid and document requirements.  `TetClipper` already does the first 2 and I'll add the last one.

Let me know what you think.

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,377 @@
...
+   *
+   * @return Whether the shape can be represented as octahedra.
+   *
+   * If implemenation returns true, it should ensure these
+   * post-conditions hold:
+   * @post octs.getAllocatorID() == shapeMesh.getAllocatorId()
+   */
```
Arlie-Capps, line 350: Please update this routine's doxygen to match any changes in `getGeometryAsTets`.

src/axom/quest/MeshClipperStrategy.hpp
gunney1: OK

src/axom/quest/ShapeMesh.hpp
Arlie-Capps: Yes, I think your point is clear.

For the use case you are proposing, yes, we don't need to document the order of tets coming out of `Hexahedron::triangulate`.  So I guess the no-added-docs route might be considered part of a minimum viable product.  But adding that documentation would increase the utility of `triangulate` and make it much more accessible to users.

If we don't document the tet order, the output from `triangulate` (a public API) is only useful in the sense of "iterate and accumulate over all tets."  You don't know which one you're getting.  This is reminds me of functions and objects in the CGAL computer geometry library, where you would get a "circulator" object and use that to iterate in some unknown order over facets or vertices or whatever.  This was particularly frustrating when I was trying to use that library.  As a user, I want more info, not less.

I understand we may use a different triangulation in the future, as needed.  And if we document the order, and we change it, yes, we will need to update the documentation.

src/axom/quest/MeshClipperStrategy.hpp
Arlie-Capps: I think that will work.

src/axom/quest/MeshClipperStrategy.hpp
gunney1: Done!

src/axom/quest/MeshClipperStrategy.hpp
gunney1: Got it.  Thanks.

src/axom/quest/ShapeMesh.cpp
```diff
@@ -0,0 +1,918 @@
...
+  const axom::ArrayView<const double>& vY = vertexCoords[1];
+  const axom::ArrayView<const double>& vZ = vertexCoords[2];
+
+  axom::ArrayView<const IndexType, 2> connView = getCellNodeConnectivity();
+
+  m_cellsAsHexes =
+    axom::Array<HexahedronType>(ArrayOptions::Uninitialized(), m_cellCount, m_cellCount, m_allocId);
+  axom::ArrayView<HexahedronType> cellsAsHexesView = m_cellsAsHexes.view();
+
+  constexpr double ZERO_THRESHOLD = 1.e-10;
```
Arlie-Capps: `ZERO_THRESHOLD` controls which points are snapped to zero.  I'm not comfortable defining a parameter that could affect the stability of a user's algorithm as a local variable with no way for the user to adjust it.

src/axom/quest/MeshClipper.hpp
```diff
@@ -0,0 +1,206 @@
...
+  //! @brief Shape-specific operations in clipping.
+  std::shared_ptr<quest::experimental::MeshClipperStrategy> m_strategy;
+
+  //! @brief Object where execution space code is instantiated.
+  std::unique_ptr<Impl> m_impl;
+
+  /* NOTE: MeshClipperStrategy is for shape-specific functions,
+   * implemented externally.  Impl implements internal algorithms
+   * for multiple execution spaces.
+   */
```
Arlie-Capps, line 187: Thanks for this comment.

src/axom/quest/ShapeMesh.cpp
gunney1: I'll add a `setZeroThreshold(double threshold)` method.

src/axom/quest/ShapeMesh.hpp
```diff
@@ -0,0 +1,325 @@
+// Copyright (c) 2017-2025, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#ifndef AXOM_QUEST_SHAPEEMESH_HPP
+#define AXOM_QUEST_SHAPEEMESH_HPP
```
kennyweiss:
```suggestion
#ifndef AXOM_QUEST_SHAPEMESH_HPP
#define AXOM_QUEST_SHAPEMESH_HPP
```

src/axom/quest/ShapeMesh.hpp
```diff
@@ -0,0 +1,325 @@
...
+  conduit::Node& getMeshConduitPath(conduit::Node& node,
+                                    const std::string& path,
+                                    const conduit::DataType& dtype);
+};
+
+}  // namespace experimental
+}  // namespace quest
+}  // namespace axom
+
+#endif  // AXOM_QUEST_SHAPEEMESH_HPP
```
kennyweiss:
```suggestion
#endif  // AXOM_QUEST_SHAPEMESH_HPP
```

src/axom/quest/detail/clipping/TetClipper.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+namespace axom
+{
+namespace quest
+{
+namespace experimental
+{
+
+/*!
+ * @brief Geometry clipping operations for sphere geometries.
```
kennyweiss: Typo:

```suggestion
 * @brief Geometry clipping operations for tetrahedral geometries.
```

kennyweiss: Please clarify in the class doxygen if this is for a single tetrahedron.

src/axom/quest/detail/clipping/TetClipper.hpp
```diff
@@ -0,0 +1,68 @@
...
+   * - v0, v1, v2, v3: each contains a 3D coordinates of the
+   *   tetrahedron vertices, in the order used by primal::Tetrahedron.
+   */
+  TetClipper(const klee::Geometry& kGeom, const std::string& name = "");
+
+  virtual ~TetClipper() = default;
+
+  const std::string& name() const override { return m_name; }
+
+  bool getGeometryAsTets(quest::experimental::ShapeMesh& shappeMesh, axom::Array<TetrahedronType>& tets) override;
```
kennyweiss, line 49: Please add a doxy comment that `tets` will contain the one tetrahedron from this class

src/axom/quest/detail/clipping/TetClipper.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+bool TetClipper::getGeometryAsTets(quest::experimental::ShapeMesh& shapeMesh, axom::Array<TetrahedronType>& tets)
+{
+  AXOM_ANNOTATE_SCOPE("TetClipper::getGeometryAsTets");
+  int allocId = shapeMesh.getAllocatorID();
+  if(tets.getAllocatorID() != allocId || tets.size() != 1)
+  {
+    tets = axom::Array<TetrahedronType>(1, 1, allocId);
+  }
+  axom::copy(tets.data(), &m_tet, sizeof(TetrahedronType));
```
kennyweiss, line 44: Is this using `axom::copy` instead of assignment (e.g. `tets[0] = m_tet`) b/c of memory space/allocations? 
Or just b/c that's how you chose to do it?
If it's the former, please add a comment.

src/axom/quest/detail/clipping/TetClipper.cpp
```diff
@@ -0,0 +1,64 @@
...
+  const auto v2 = m_info.fetch_existing("v2").as_double_array();
+  const auto v3 = m_info.fetch_existing("v3").as_double_array();
+  for(int d = 0; d < 3; ++d)
+  {
+    m_tetBeforeTrans[0][d] = v0[d];
+    m_tetBeforeTrans[1][d] = v1[d];
+    m_tetBeforeTrans[2][d] = v2[d];
+    m_tetBeforeTrans[3][d] = v3[d];
+  }
+}
```
kennyweiss, line 65: Does this need validation that the data in `m_info` conforms to our expectations?

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -0,0 +1,906 @@
...
+namespace axom
+{
+namespace quest
+{
+namespace experimental
+{
+namespace detail
+{
+
+template <typename ExecSpace>
```
kennyweiss, line 37: Please add a short class documentation about what this class does, and that we expect to have specific derived classes, e.g. `TetClipper`

Note: This class hierarchy is a bit unexpected and might warrant a comment:
```cpp
class MeshClipperImpl : public MeshClipper::Impl
```

src/axom/quest/MeshClipper.hpp
```diff
@@ -0,0 +1,206 @@
...
+{
+namespace experimental
+{
+
+/*!
+ * @brief Abstract base class encapsulating the clipping task in
+ * intersection shaping, to be specialized for geometry-specific
+ * implementations.
+ *
+ * Each object operates is associated with a single klee::Geometry object.*
```
kennyweiss: Slight clarification:
```suggestion
 * Each MeshClipper object operates is associated with a single klee::Geometry object.
```

src/axom/quest/ShapeMesh.cpp
```diff
@@ -0,0 +1,917 @@
...
+  return rval;
+}
+
+void ShapeMesh::setMatsetFromVolume(const std::string& materialName,
+                                    const axom::ArrayView<double>& volumes,
+                                    bool isFraction)
+{
+  SLIC_ERROR_IF(m_matsetName.empty(),
+                "Cannot use material set in ShapeMesh: Matset name was not provided, and no "
+                "default matset was found.");
```
Arlie-Capps, line 302: I think `m_matsetName` is the default matset, correct?  Then shouldn't the `SLIC_ERROR_IF` predicate also look at argument `materialName`?

src/axom/quest/ShapeMesh.cpp
```diff
@@ -0,0 +1,917 @@
...
+   * If this changes in the future, this code must adapt.
+   */
+  conduit::Node info;
+  rval = conduit::blueprint::mesh::verify(bpMesh, info);
+
+  if(rval)
+  {
+    std::string topoType = bpMesh.fetch("topologies")[m_topoName]["type"].as_string();
+    rval = topoType == "unstructured";
+    info[0].set_string("Topology is not unstructured.");
```
Arlie-Capps: I think that this line (and line 281 below) should be guarded by `if (!rval)` .

src/axom/quest/ShapeMesh.cpp
Arlie-Capps: The `SLIC_ERROR_IF` at line 367 looks similar to this one.

Arlie-Capps: Maybe I don't understand what is going on here.

src/axom/quest/MeshClipper.cpp
```diff
@@ -0,0 +1,250 @@
...
+/*
+ * Orchestrates the geometry clipping by using the capabilities of the
+ * MeshClipperStrategy implementation.
+ *
+ * If the strategy can label cells as inside/on/outside geometry
+ * boundary, use that to reduce reliance on expensive clipping methods.
+ *
+ * Regardless of labling, try to use specialized clipping first.
+ * If specialized methods aren't implemented, resort to discretizing
+ * geomety into tets or octs for clipping against mesh cells.
```
Arlie-Capps: Typos: please correct to "labeling" and "geometry"

src/axom/quest/MeshClipper.cpp
```diff
@@ -0,0 +1,250 @@
...
+      AXOM_ANNOTATE_SCOPE("MeshClipper::clip3D");
+      m_impl->computeClipVolumes3D(ovlap);
+    }
+
+    m_localCellInCount = cellCount;
+  }
+}
+
+/*!
+ * @brief Allocate a Impl for the execution-space computations
```
Arlie-Capps: Suggest "an Impl"

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -0,0 +1,906 @@
...
+      axom::copy(&onIndices[0], &zero, sizeof(zero));
+    }
+
+    axom::for_all<ExecSpace>(
+      1,
+      labelCount,
+      AXOM_LAMBDA(axom::IndexType i) {
+        if(tmpLabelsView[i] != tmpLabelsView[i - 1])
+        {
+          onIndicesView[tmpLabelsView[i - 1]] = i;
```
Arlie-Capps, line 164: Maybe I'm missing something, but I fear a write conflict here if multiple elements of `tmpLabelsView` hold the same value.

src/axom/quest/MeshClipper.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+    /*!
+     * @brief Initialize overlap volumes to full for cells completely
+     * inside the shape and zero for cells outside or on shape boundary.
+     */
+    virtual void initVolumeOverlaps(const axom::ArrayView<MeshClipperStrategy::LabelType>& labels,
+                                    axom::ArrayView<double> ovlap) = 0;
+
+    //! @brief Initialize overlap volumes to zero.
+    virtual void initVolumeOverlaps(axom::ArrayView<double> ovlap) = 0;
```
Arlie-Capps: Please rename this to zeroVolumeOverlaps.  Sharing a name with initVolumeOverlaps(labels, ovlap) is confusing.

src/axom/quest/MeshClipper.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+    //!@brief Collect unlabeled LABEL_ON indices into an index list.
+    virtual void collectOnIndices(const axom::ArrayView<LabelType>& labels,
+                                  axom::Array<axom::IndexType>& onIndices) = 0;
+
+    //!@brief Change tet indices from sparse-set values to full-set values.
+    virtual void remapTetIndices(axom::ArrayView<axom::IndexType> tetsOnBdry,
+                                 axom::ArrayView<const axom::IndexType> cellsOnBdry) = 0;
+
+    //!@brief Add volumes of tets inside the geometry to the volume data.
```
Arlie-Capps, line 132: Please expand to explain that tetLabels should hold labels for the tets indexed by cellsOnBdry, and that the volume of each tet labeled "inside" will be added to its cell within ovlap.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -0,0 +1,906 @@
...
+          {
+            const axom::IndexType tetId = hexId * NUM_TETS_PER_HEX + it;
+            const auto& tet = meshTets[tetId];
+            ovlap[hexId] += tet.volume();
+          }
+        }
+      });
+  }
+
+  //! @brief Make an list of indices where labels have value LABEL_ON.
```
Arlie-Capps: "Make a list"

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -0,0 +1,906 @@
...
+        auto cellId = cellsOnBdry[cellIdId];
+        auto tetIdInCell = tetIdId % NUM_TETS_PER_HEX;
+        auto tetId = cellId * NUM_TETS_PER_HEX + tetIdInCell;
+        tetsOnBdry[i] = tetId;
+      });
+  }
+
+  /*
+   * Clip tets of from the mesh with tets or octs from the clipping
+   * geomnetry.  This implemenation was lifted from IntersectionShaper
```
Arlie-Capps: Rephrase: use "Clip tets from the mesh"
Typos: correct to "geometry", "implementation"

Arlie-Capps: There are several comments below that should be checked for these typos.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -0,0 +1,906 @@
...
+  }  // end of computeClipVolumes3D() function
+
+  /*
+   * Clip tets of from the mesh with tets or octs from the clipping
+   * geomnetry.  This implemenation is like the above except that it
+   * limits clipping to a subset of mesh cells labeled as potentially
+   * on the boundary.
+   */
+  void computeClipVolumes3D(const axom::ArrayView<axom::IndexType>& cellIndices,
+                            axom::ArrayView<double> ovlap) override
```
Arlie-Capps, line 431: All the computeClipVolumes3D routines have some extensive logic.  The best would be to refactor these three complex routines down into one.  The interface might stay the same (three routines), but two would call the most general of the three.  This lessens code maintenance burden and reduces the chance of bugs.

src/axom/quest/ShapeMesh.cpp
gunney1: Matset name and material name are 2 separate parts of the path.  Both should be non-empty strings.  I'll add the check for `materialName.empty()` so we can abort with a helpful message, wherever it's used.

src/axom/quest/ShapeMesh.cpp
gunney1: Right.  Thank you.

src/axom/quest/MeshClipper.cpp
gunney1: Done.  Thanks.

src/axom/quest/MeshClipper.cpp
gunney1: OK

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: It looks that way, but the algorithm is designed to not have write conflicts.  `tmpLabelsView` is a monotonically non-decreasing discrete integer function, and where its value doesn't change, nothing is written.  (That's what the if-check is for.)

src/axom/quest/MeshClipper.hpp
gunney1: Done.

src/axom/quest/MeshClipper.hpp
gunney1: Done.  However, the insided-ness of the cell listed is not relevant to this method per se.  So I renamed the input arrays to not imply that.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: Right.  Thanks.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: Done.  Thanks.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: Agreed.  I plan to do this after the re-factoring and performance refinements, which should give indications of which parts can be re-factored.

Issue https://github.com/LLNL/axom/issues/1704

src/axom/quest/ShapeMesh.cpp
gunney1: Correct.  Thank you.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
Arlie-Capps: Thanks for the Webex discussion.  I can see that there will be no write-conflicts because `tmpLabelsView` holds non-decreasing values, and we're only writing to `onIndicesView` where `tmpLabelsView` is actually increasing.  So we're always going to be writing to distinct offsets.  I appreciate you said you would add a clarifying comment (also crediting Jason Burmark for pointing out this algorithm).

src/axom/quest/ShapeMesh.hpp
```diff
@@ -0,0 +1,343 @@
...
+ *
+ * This wrapper class:
+ * - encapsulate mesh-dependent data
+ * - provide a common interface to mesh data regardless of the mesh format
+ * - avoid redundant work
+ * - provides some convenience tools typically used in shaping
+ *
+ * The wrapped mesh must have an unstructured 3D hex topology.  That
+ * is the only topology currently supported.  It can be extended to
+ * support 2D.
```
kennyweiss, line 50: Is this class intended to support more general topologies? Or is it only intended to support 3D hex meshes? If it's the latter, perhaps it should be renamed.

More generally, (and longer term) will this class persist or will this capability be migrated to `bump`?

src/axom/quest/ShapeMesh.hpp
```diff
@@ -0,0 +1,343 @@
...
+   * !@brief Return computational mesh as a sidre::Group if it has
+   * that form, or nullptr otherwise.
+  */
+  sidre::Group* getMeshAsSidre() { return m_bpGrpExt; }
+
+  /*!
+   * @brief Return computational mesh as a conduit::Node if it has
+   * that form, or nullptr otherwise.
+  */
+  conduit::Node* getMeshAsConduit() { return m_bpNodeExt; }
```
kennyweiss, line 141: Is there a way for the user to determine if their mesh is sidre or conduit?

src/axom/quest/ShapeMesh.hpp
```diff
@@ -0,0 +1,343 @@
...
+  /*!
+   * @brief Allocator id set in constructor.
+   */
+  int getAllocatorID() const { return m_allocId; }
+
+  /*
+   * !@brief Return computational mesh as a sidre::Group if it has
+   * that form, or nullptr otherwise.
+  */
+  sidre::Group* getMeshAsSidre() { return m_bpGrpExt; }
```
kennyweiss, line 135: If we're guarding sidre, this needs to be guarded as well.
Please check that you can successfully build/run the code w/ `AXOM_ENABLE_SIDRE=OFF`

src/axom/quest/ShapeMesh.hpp
```diff
@@ -0,0 +1,343 @@
...
+   * that may be used in clipping.
+   *
+   * Mesh-dependent data are computed as needed and cached, but this
+   * computes them all at once.
+  */
+  void precomputeMeshData();
+  //@}
+
+  /*!
+   * @brief Check whether mesh meets requirements for shaping.
```
kennyweiss, line 198: Please document the requirements (at a high level)

src/axom/quest/ShapeMesh.cpp
```diff
@@ -0,0 +1,925 @@
...
+                "Topology name was not provided, and no default topology was found.");
+
+  const int hostAllocId = axom::execution_space<axom::SEQ_EXEC>::allocatorID();
+
+  // We want unstructured topo but can accomodate structured.
+  const auto& typeNode =
+    m_bpNodeExt->fetch_existing("topologies").fetch_existing(m_topoName).fetch_existing("type");
+  const std::string topoType = typeNode.as_string();
+  SLIC_ERROR_IF(topoType != "unstructured",
+                "ShapeMesh currently only works with unstructured mesh, not " + topoType + ".");
```
kennyweiss: This comment (about allowing structured) doesn't match the `SLIC_ERROR`

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,400 @@
+// Copyright (c) 2017-2025, Lawrence Livermore National Security, LLC and
+// other Axom Project Developers. See the top-level COPYRIGHT file for details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#ifndef AXOM_QUEST_GEOMETRYCLIPPERSTRATEGY_HPP
+#define AXOM_QUEST_GEOMETRYCLIPPERSTRATEGY_HPP
+
```
kennyweiss: `AXOM_QUEST_GEOMETRYCLIPPERSTRATEGY_HPP` -> `AXOM_QUEST_MESHCLIPPERSTRATEGY_HPP`

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,400 @@
...
+   * @brief A type to denote whether something is inside,
+   * on or outside the boundary of a geometry.
+  */
+  using LabelType = char;
+  //!@brief Denotes something inside a shape boundary.
+  static constexpr LabelType LABEL_IN = 0;
+  //!@brief Denotes something on a shape boundary.
+  static constexpr LabelType LABEL_ON = 1;
+  //!@brief Denotes something outside a shape boundary.
+  static constexpr LabelType LABEL_OUT = 2;
```
kennyweiss: Should these be an enum?, 

e.g. `primal::OrientationResult` -- https://github.com/LLNL/axom/blob/develop/src/axom/primal/geometry/OrientationResult.hpp

src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -0,0 +1,400 @@
...
+   * @brief Get the 2D axis-aligned bounding box for the geometry,
+   * if it's applicable and available.
+  */
+  virtual const axom::primal::BoundingBox<double, 2>& getBoundingBox2D() const;
+
+  /*!
+   * @brief Get the 3D axis-aligned bounding box for the geometry,
+   * if it's applicable and available.
+  */
+  virtual const axom::primal::BoundingBox<double, 3>& getBoundingBox3D() const;
```
kennyweiss, line 147: If these are bounding boxes for the entire mesh, should it just be returned by value? That would avoid the need to have a static invalid `BoundingBox` in the `MeshClipperStrategy.cpp`

src/axom/quest/MeshClipperStrategy.cpp
```diff
@@ -0,0 +1,141 @@
...
+namespace internal
+{
+/*!
+ * \brief Implementation of a GeometryOperatorVisitor for processing klee shape operators
+ *
+ * This class extracts the matrix form of supported operators.
+ * For other operators, it sets the valid flag to false.
+ * To use, check the \a isValid() function after visiting and then call the \a getMatrix() function.
+ */
+class AffineMatrixVisitor : public klee::GeometryOperatorVisitor
```
kennyweiss, line 26: Note (likely for a future PR) -- I think these should be available directly in Klee rather than in Quest, 
e.g. here as well: 

https://github.com/LLNL/axom/blob/adc75e2af3abce9ef134172410b0e55dc615009f/src/axom/quest/DiscreteShape.cpp#L21-L74

See also: https://github.com/LLNL/axom/issues/1628

src/axom/quest/MeshClipperStrategy.cpp
```diff
@@ -0,0 +1,141 @@
...
+{
+  static const std::string n = "UNNAMED";
+  return n;
+}
+
+const axom::primal::BoundingBox<double, 2>& MeshClipperStrategy::getBoundingBox2D() const
+{
+  static const axom::primal::BoundingBox<double, 2> invalidBb2d;
+  return invalidBb2d;
+}
```
kennyweiss, line 88: I think this should return by value to avoid the static instance (see comment in header)

src/axom/quest/tests/CMakeLists.txt
```diff
@@ -231,6 +231,54 @@ if(ENABLE_MPI AND MFEM_FOUND AND MFEM_USE_MPI
...
+    # Run the geometry clipping test on with each enabled policy
+    set(_policies "seq")
+    if(RAJA_FOUND)
+      blt_list_append(TO _policies ELEMENTS "omp"  IF AXOM_ENABLE_OPENMP)
+      blt_list_append(TO _policies ELEMENTS "cuda" IF AXOM_ENABLE_CUDA)
+      blt_list_append(TO _policies ELEMENTS "hip"  IF AXOM_ENABLE_HIP)
+    endif()
+
+    set(_testgeoms "tet" )
+    set(_meshTypes "bpSidre" "bpConduit")
```
kennyweiss, line 258: Should there be a check for sidre's availability?

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -0,0 +1,898 @@
...
+    }
+    auto onIndicesView = onIndices.view();
+
+    LabelType firstLabel = '\0';
+    axom::copy(&firstLabel, &labels[0], sizeof(firstLabel));
+    if(firstLabel == 1)
+    {
+      axom::IndexType zero = 0;
+      axom::copy(&onIndices[0], &zero, sizeof(zero));
+    }
```
kennyweiss: This would likely be clearer via named values (e.g. enum values)

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -0,0 +1,898 @@
...
+    if(useTets == useOcts)
+    {
+      SLIC_ERROR(
+        axom::fmt::format("Problem with MeshClipperStrategy implementation '{}'."
+                          "  Implementations that don't provide a specializedClip function"
+                          " must provide exactly one getGeometryAsOcts() or getGeometryAsTets()."
+                          "  This implementation provides {}.",
+                          strategy.name(),
+                          int(useOcts) + int(useTets)));
+    }
```
kennyweiss, line 230: Suggestion (non-binding): Instead of implementing both algorithms (tets and octs) within the same function, would it make sense to to figure out which case we're in and pass it to an implementation function?

kennyweiss (approved):
Thanks @gunney1 -- lots of good work in here.

I have some concerns related to the generality of ShapeMesh and the overlap in functionality between this and bump, but I think we'll work these out in the future when we pull this out of the experimental namespace.

In the immediate term, since you're guarding sidre in some cases, I want to make sure that this compiles properly when `AXOM_ENABLE_SIDRE` is off.

src/axom/quest/detail/clipping/TetClipper.hpp
gunney1: Done.

src/axom/quest/detail/clipping/TetClipper.cpp
gunney1: It's the former.  Comment added.

src/axom/quest/detail/clipping/TetClipper.cpp
gunney1: Probably a good idea.  Adding checks on the array lengths.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: Done.  But note that `TetClipper` is orthogonal to this class.  See comments at https://github.com/LLNL/axom/blob/88d9466edb940c3a6c872252758ba5d88381517b/src/axom/quest/MeshClipper.hpp#L183C1-L185C36

src/axom/quest/MeshClipper.hpp
gunney1: Oh dear.  That whole doxygen block is from before I pulled out the geometry-specific operations into a strategy class.  Totally rewritten.  Sorry.

src/axom/quest/ShapeMesh.hpp
gunney1: It's 3D for now but should be extended to 2D also.  Regarding the second point, I'm not sure. @BradWhitlock and I need to figure that out.

src/axom/quest/ShapeMesh.hpp
gunney1: The user could check on the null-ness of the returned pointers.  But they should have this information already, as they constructed the `ShapeMesh`.

src/axom/quest/ShapeMesh.hpp
gunney1: It turns out I had to guard the whole class against non-sidre builds.  One important requirement is to allocate all data in the same memory pool, by using the same memory allocator through out.  The mechanism for this is `sidre::ConduitMemory`.

src/axom/quest/ShapeMesh.hpp
gunney1: Good idea.  Done.

src/axom/quest/MeshClipperStrategy.hpp
gunney1: Thanks.  Done.

src/axom/quest/ShapeMesh.cpp
gunney1: Not sure I follow.  It looks right to me.

kennyweiss: The comment a few lines up says
```cpp
    // We want unstructured topo but can accomodate structured.
```

but this line errors out if it's not "unstructured"
```cpp
   SLIC_ERROR_IF(topoType != "unstructured", "ShapeMesh currently only works with unstructured mesh, not " + topoType + ".");
```

src/axom/quest/MeshClipperStrategy.hpp
gunney1: OK.  Done.

src/axom/quest/tests/CMakeLists.txt
gunney1: Yes.  Done.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: Agreed.  Done.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: Certainly reasonable to do that.  The counter argument is that both functions would have a ton of codes in common.  It was the same thing in the `IntersectionShaper` code, which implemented these methods by templates.  In any case, we need to look at how to factor out the common parts, as @Arlie-Capps recommended in his comments.

src/axom/quest/MeshClipperStrategy.hpp
gunney1: I don't think it makes a difference.  The default implementation will never be used because there's no associated geometry to bound.  But I didn't want to make the method abstract because it would require a pointless implementation for the re-implementation of the original "no geometry" case in `klee::Geometry`.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
kennyweiss: >  In any case, we need to look at how to factor out the common parts

That's what I had in mind.  It can certainly be deferred to a follow-up PR.

gunney1: For future reference, the refactoring is in issue https://github.com/LLNL/axom/issues/1704

src/axom/quest/ShapeMesh.cpp
gunney1: I see.  Thanks.

PR #1669, opened 2025-09-25
made members in LumberjackStream protected instead of private

gberg617:
Convert private members to protected in LumberjackStream to allow other codes to more easily inherit from it.

### DIFF
--- a/src/axom/slic/streams/LumberjackStream.hpp
+++ b/src/axom/slic/streams/LumberjackStream.hpp
@@ -253,11 +253,11 @@ public:
    */
   virtual MPI_Comm comm() override;
 
-private:
+protected:
   void initializeLumberjack(MPI_Comm comm, int ranksLimit);
   void finalizeLumberjack();
 
-  /// \name Private Members
+  /// \name Protected Members
   /// @{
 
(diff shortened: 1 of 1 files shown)


bmhan12 (approved):

src/axom/slic/streams/LumberjackStream.hpp
```diff
@@ -253,11 +253,11 @@ class LumberjackStream : public LogStream
    */
   virtual MPI_Comm comm() override;
 
-private:
+protected:
   void initializeLumberjack(MPI_Comm comm, int ranksLimit);
   void finalizeLumberjack();
 
-  /// \name Private Members
+  /// \name Protected Members
```
white238, line 260: Do you need these member variables protected? This should be done sparingly

rhornung67 (approved):

src/axom/slic/streams/LumberjackStream.hpp
gberg617: I think I only need the Lumberjack instance `m_lj`, but I made the others protected as well.  I can change this if needed!

PR #1672, opened 2025-09-29
Enhance CurvedPolygon so WN memoization works with sampling shaper

jcs15c:
This PR will be merged after #1683.

# Summary

In order that the SamplingShaper in quest is compatible with recent GWN accelerations via memoization, the `primal::CurvedPolygon<T, NDIMS>` class is enhanced so that it is templated directly on curve type, i.e. `primal::CurvedPolygon<CurveType>`. Currently supported curve types are `BezierCurve`, `NURBSCurve`, and the new `NURBSCurveGWNCache`. Because `CurvedPolygon` is used in many of primal's operators, this requires many updates:
- All 2D GWN algorithms (and by association methods in `in_curved_polygon` have been made compatible with all curve types.
- All 2D integral evaluation methods are compatible with all curve types.
- Methods in `compute_moments.hpp` are currently _only_ compatible with `CurvedPolygon<BezierCurve<T, 2>>`, since the methods within expect Bezier curves, and would require a non-trivial extension to work on `NURBSCurve` objects via Bezier extraction.

The tests for these methods are updated in their respective suites to ensure compatibility with the new template.

On the quest side of things, this PR makes the following changes related to SamplingShaper
- Updates `MFEMReader` so that the default input for an MFEM mesh is a `CurvedPolygon<NURBS>` instead of Bezier.
- Updates `WindingNumberSamper.hpp` so that the sampler stores an array of "cached contours," i.e. `CurvedPolygon<NURBSCurveGWNCache>` instead of a view of `CurvedPolygon<BezierCurve>`s.

### Performance Improvements
These changes reduce the time to complete the test `quest_shaping_driver_ex_heroic_roses_mfem_cp` from 85.93 seconds on a debug build of `develop` to 56.68 seconds (or from 12.67s to 6.91s on a release build) on the current branch. Accounting for the time to process the curves and construct/access the BVH (~31.31s debug/~4.47s release) this is about a 2x improvement.

### DIFF
--- a/src/axom/primal/geometry/CurvedPolygon.hpp
+++ b/src/axom/primal/geometry/CurvedPolygon.hpp
@@ -124,7 +124,6 @@ public:
     m_edges.resize(other_poly.numEdges());
     for(int e = 0; e < other_poly.numEdges(); ++e)
     {
-      //this->addEdge(CurveType(other_poly[e]));
       m_edges[e] = CurveType(other_poly[e]);
     }
   }


src/axom/primal/geometry/BezierCurve.hpp
```diff
@@ -256,6 +259,9 @@ class BezierCurve
   /// Retrieves the control point at index \a idx
   const PointType& operator[](int idx) const { return m_controlPoints[idx]; }
 
+  PointType getInitPoint() const { return m_controlPoints[0]; }
+  PointType getEndPoint() const { return m_controlPoints[m_controlPoints.size() - 1]; }
```
kennyweiss: Should these return by reference instead of by value?

src/axom/primal/geometry/CurvedPolygon.hpp
```diff
@@ -74,7 +114,7 @@ class CurvedPolygon
   }
 
   /// Constructor from an array of \a nEdges curves
-  CurvedPolygon(BezierCurveType* curves, int nEdges)
+  CurvedPolygon(CurveType* curves, int nEdges)
```
kennyweiss: Minor: Should we convert this to an `ArrayView<CurveType>`?
(Perhaps in a follow-up PR)

src/axom/primal/operators/detail/winding_number_2d_memoization.hpp
```diff
@@ -59,26 +59,50 @@ struct BezierCurveData
   auto isConvexControlPolygon() const { return m_isConvexControlPolygon; }
   auto getBoundingBox() const { return m_boundingBox; }
 
+  friend bool operator==(const BezierCurveData<T>& lhs, const BezierCurveData<T>& rhs)
+  {
+    // isConvexControlPolygon will be equal if the curves are
+    return (lhs.m_curve == rhs.m_curve) && (lhs.m_boundingBox == rhs.m_boundingBox) &&
```
kennyweiss, line 65: Is it safe to check for (exact) equality of bounding boxes?

kennyweiss (commented):
Looks reasonable so far.

Lingering questions for this PR include:

* Many operations on a CurvedPolygon<NURBSCurveGWNCache> are forbidden at compile time (e.g. split and reverseOrientation) to ensure that the GWN cache is not made invalid. Is this the right approach? Or should the methods be allowed after the cache is cleared, and a warning message displayed to inform the user that this has happened?

    * Seems reasonable to start by disallowing these operations (as long as this is documented) since invalidating the cache can have severe performance costs. If we need to allow this in the future, we can add support for it.


* Some template boilerplate is required in CurvedPolygon.hpp to deduce the numeric type from the curve type. Is this implemented in the right way? Should those template methods be in a detail file?

    * I'd leave it as is since it's reasonably self contained. Perhaps move it inside an anonymous namespace or a `detail` namespace (within the same file)?

src/axom/primal/geometry/CurvedPolygon.hpp
jcs15c: I think you're right that it's probably best to make the change now, have it work like the reorganization of the curve classes themselves

src/axom/primal/operators/detail/winding_number_2d_memoization.hpp
jcs15c: I guess it depends on what your expectations are. The way I'm used to using `operator==` equality is from less of a math perspective, and more of a data perspective, i.e. if you call a copy constructor, then the result should be `operator==` to the input. If two objects were initialized with the same shape, but different values for `bbExpansionAmount` (or an erroneous value for `knownConvex` for that matter) then I don't think they should be considered equal.

Is that what you mean?

src/axom/primal/geometry/BezierCurve.hpp
jcs15c: Yep! Fixed.

src/axom/primal/geometry/CurvedPolygon.hpp
```diff
@@ -105,36 +158,43 @@ class CurvedPolygon
...
+  /// Appends a curve to the list of edges
+  void addEdge(const CurveType& c1) { m_edges.push_back(c1); }
+
+  /// Consumes then appends a curve to the list of edges
+  void addEdge(CurveType&& c1) { m_edges.push_back(std::move(c1)); }
 
   /// Splits an edge "in place"
   void splitEdge(int idx, T t)
   {
     SLIC_ASSERT(idx < static_cast<int>(m_edges.size()));
```
kennyweiss: Might as well also check that `idx` is non-negative.

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -57,13 +59,69 @@ inline double evaluate_scalar_line_integral_component(const primal::BezierCurve<
...
+ * Must accept a 2D point as input and return a double
+ * \param [in] npts The number of quadrature points in the rule
+ * \return the value of the integral
+ */
+template <class Lambda, typename T, int NDIMS>
+inline double evaluate_scalar_line_integral_component(const primal::NURBSCurve<T, NDIMS>& n,
+                                                      Lambda&& scalar_integrand,
+                                                      const int npts)
+{
+  auto beziers = n.extractBezier();
```
kennyweiss: Minor:

```suggestion
  const auto beziers = n.extractBezier();
```

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -90,6 +148,60 @@ inline double evaluate_vector_line_integral_component(const primal::BezierCurve<
...
+ * Must accept a 2D point as input and return a 2D axom vector object
+ * \param [in] npts The number of quadrature points in the rule
+ * \return the value of the integral
+ */
+template <class Lambda, typename T, int NDIMS>
+inline double evaluate_vector_line_integral_component(const primal::NURBSCurve<T, NDIMS>& n,
+                                                      Lambda&& vec_field,
+                                                      const int npts)
+{
+  auto beziers = n.extractBezier();
```
kennyweiss: Likewise:

```suggestion
  const auto beziers = n.extractBezier();
```

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -232,15 +172,16 @@ double evaluate_vector_line_integral(const primal::NURBSCurve<T, NDIMS>& n,
  *
  * \return the value of the integral
  */
-template <typename Lambda, typename T, int NDIMS>
-double evaluate_vector_line_integral(const axom::Array<primal::NURBSCurve<T, NDIMS>>& narray,
+template <typename Lambda, typename CurveType>
+double evaluate_vector_line_integral(const axom::Array<CurveType>& carray,
```
kennyweiss, line 186: Possible generalization for the future -- perhaps the exposed API should use `axom::ArrayView` instead of `axom::Array` ?

src/axom/quest/detail/shaping/WindingNumberSampler.hpp
```diff
@@ -81,10 +81,13 @@ class WindingNumberSampler
    * \param geomView A view that contains the shapes being queried.
    *
    */
-  WindingNumberSampler(const std::string& shapeName, GeometryView geomView)
-    : m_shapeName(shapeName)
-    , m_geometryView(geomView)
-  { }
+  WindingNumberSampler(const std::string& shapeName, GeometryView geomView) : m_shapeName(shapeName)
+  {
+    for(auto & contour : geomView)
```
kennyweiss: Minor:

```suggestion
    for(const auto & contour : geomView)
```

Assuming that caching doesn't change the contour, and mostly to check if all the required functions are marked `const`

src/axom/primal/tests/primal_integral.cpp
```diff
@@ -448,12 +449,74 @@ TEST(primal_integral, evaluate_nurbs_surface_normal)
...
+  using NCache = primal::NURBSCurve<double, 2>;
+  using CPolygon = primal::CurvedPolygon<NCache>;
+  double abs_tol = 1e-10;
+
+  // Quadrature nodes. Should be sufficiently high to pass tests
+  int npts = 20;
+
+  // Test integrals with same integrand and curves as `evaluate_integral_rational`,
+  //  but with curves added to detail::NURBSCurveGWNCache objects to ensure template compatibility,
+  //  even if there isn't compelling reason to use GWN caches for this purpose
```
kennyweiss: minor:

```suggestion
  //  even if there isn't a compelling reason to use GWN caches for this purpose
```

kennyweiss (approved):
Thanks @jcs15c -- really nice improvement!

kennyweiss:
> Performance Improvements
>
> These changes reduce the time to complete the test quest_shaping_driver_ex_heroic_roses_mfem_cp from 85.93 seconds on a __debug build__ of develop to 56.68 seconds on the current branch. Accounting for the time to process the curves and construct/access the BVH (~31.31s) this is about a 2x improvement.

Thanks for adding the performance improvement.  Could you please also adding the performance change for a Release config?

Arlie-Capps (approved):
Thanks, Jacob!

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -57,13 +59,69 @@ inline double evaluate_scalar_line_integral_component(const primal::BezierCurve<
...
+inline double evaluate_scalar_line_integral_component(const primal::NURBSCurve<T, NDIMS>& n,
+                                                      Lambda&& scalar_integrand,
+                                                      const int npts)
+{
+  const auto beziers = n.extractBezier();
+  double total_integral = 0.0;
+  for(int i = 0; i < beziers.size(); ++i)
+  {
+    total_integral +=
+      detail::evaluate_scalar_line_integral_component(beziers[i], scalar_integrand, npts);
```
BradWhitlock: It may be good to add `std::forward<Lambda>(scalar_integrand)` here and in other places like this.

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -57,13 +59,69 @@ inline double evaluate_scalar_line_integral_component(const primal::BezierCurve<
...
+ * The cache object has already decomposed the NURBS curve into Bezier segments, 
+ *  which are used to evaluate the integral over each
+ *
+ * \param [in] n The NURBS curve object
+ * \param [in] integrand the lambda function representing the integrand. 
+ * Must accept a 2D point as input and return a double
+ * \param [in] npts The number of quadrature points in the rule
+ * \return the value of the integral
+ */
+template <class Lambda, typename T, int NDIMS>
```
BradWhitlock: Is `class` better here than `typename`? Also, please add `\tparam` doc strings to the methods where the function takes template arguments.

src/axom/primal/operators/detail/winding_number_2d_memoization.hpp
```diff
@@ -184,6 +238,13 @@ class NURBSCurveGWNCache
   mutable axom::Array<std::map<std::pair<int, int>, BezierCurveData<T>>> m_bezierSubdivisionMaps;
```
BradWhitlock, line 238: Is some other object hanging onto a const reference or pointer to this cache, making it necessary to mark `getSubdivisionData` `const`, leading to this `mutable`?

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -258,8 +204,8 @@ double evaluate_vector_line_integral(const axom::Array<primal::NURBSCurve<T, NDI
  * \param [in] npts_P the number of quadrature points to evaluate the antiderivative
  * \return the value of the integral
  */
-template <class Lambda, typename T>
-double evaluate_area_integral(const primal::CurvedPolygon<T, 2> cpoly,
+template <class Lambda, typename CurveType>
+double evaluate_area_integral(const primal::CurvedPolygon<CurveType> cpoly,
```
BradWhitlock: Do you want to pass `cpoly` by value here? Or, is it just missing a reference?

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -291,25 +237,22 @@ double evaluate_area_integral(const primal::CurvedPolygon<T, 2> cpoly,
...
- *   is computed using npts_Q * npts_P on each segment
  * 
  * \note The numerical result is only meaningful if the curves enclose a region
  * 
  * \return the value of the integral
  */
-template <class Lambda, typename T>
-double evaluate_area_integral(const axom::Array<primal::NURBSCurve<T, 2>> narray,
+template <class Lambda, typename CurveType>
+double evaluate_area_integral(const axom::Array<CurveType> carray,
```
BradWhitlock: Passing by value.

src/axom/primal/operators/winding_number.hpp
```diff
@@ -418,7 +417,8 @@ axom::Array<double> winding_number(const axom::Array<Point<T, 2>>& query_arr,
 /*!
  * \brief Computes the GWN for an array of 2D points wrt to a 2D curved polygon
  *
- * \param [in] query_arr The query point to test
+ * \tparam CurveType The BezierCurve, NURBSCurve, or NURBSCurveGWNCache which represents the curve
+ * \param [in] q_arr The query point to test
```
BradWhitlock: It looks like an array of points so "query points" would be more apt.

src/axom/primal/tests/primal_curved_polygon.cpp
```diff
@@ -219,23 +245,55 @@ TEST(primal_curvedpolygon, isClosed_BiGon)
...
+  EXPECT_FALSE(NURBSPolygon(bPoly2).isClosed());
+  EXPECT_FALSE(NURBSCachePolygon(bPoly2).isClosed());
+}
+
+template <typename CurveType>
+void test_split_edge(primal::CurvedPolygon<CurveType>& curved_polygon)
+{
+  using PointType = primal::Point<double, 2>;
+  
+  // Needs to be std::vector to use .assign
```
BradWhitlock: Is this comment needed?

BradWhitlock (approved):
I had a few small suggestions. Overall, looks good. I like how CurvedPolygon is more general now.

src/axom/primal/operators/evaluate_integral.hpp
jcs15c: Just missing the reference, I'll fix it here and elsewhere. Sorry!

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
jcs15c: I think it should probably be `typename`, same as the other methods.

src/axom/primal/tests/primal_curved_polygon.cpp
jcs15c: Nope, old copy-paste error.

src/axom/primal/operators/evaluate_integral.hpp
jcs15c: I think this is a good idea, as we discussed earlier, but I'm also running into a case where it's useful to template the return type of the lambda function so that you can perform integrals on vector-type functions. Since `primal::Vector<T, NDIMS>` already support addition and scalar multiplication, it's a pretty straightforward addition, and would be really helpful for computing things like the centroid of geometric objects, since it reduces the overhead of evaluating the function for a single input double. I think I'm going to put both ideas in a new feature branch.

src/axom/primal/operators/detail/winding_number_2d_memoization.hpp
jcs15c: I think the idea is that this allows you to pass cache objects by const reference between functions, and use the same cache for each whiles still dynamically updating it. In a practical sense, calling `getSubdivisionData` is the primary purpose of these cache objects, so if it wasn't const, you couldn't realistically use const objects, but I don't know if that's especially satisfying. For the most part, this class was set up by pattern matching the design of the older `MemoizedSectorAreaWeight` class, so I'm open to the possibility that what worked there might not be best here.

Issue #1674, opened 2025-09-29
Group::deepCopyToConduit drops View-less groups


gunney1:
`deepCopyToConduit` was modeled after `createNativeLayout` which, after copying, removed detstination `Groups` that have no `View` under them.  We discussed why `createNativeLayout` does this at a project meeting but couldn't recall exactly why view-less nodes are dropped.  This behavior in `deepCopyToConduit` is breaking other codes that require empty Groups to be kept.

`deepCopyToConduit` is a new function that has not been exposed to users, so we should remove the behavior of removing empty `Group`s.

Please chime in if you have thoughts or knowledge of this behavior in `createNativeLayout`.

PR #1675, opened 2025-09-29
Keep View-less Groups in deepCopyToConduit

gunney1:
# Summary

- This PR is a bug fix
- It deletes a few lines of code that removed `Groups` post-copy if they have no `View`s under them.

The deletion is unexpected and breaking some mint code that requires empty Groups in the destination hierarchy.

Fixes https://github.com/LLNL/axom/issues/1674


Arlie-Capps (approved):

kennyweiss (approved):
Thanks @gunney1

PR #1679, opened 2025-10-01
Feature/gunney/extend primal geometries (second try)

gunney1:
# Summary

This is a redo of https://github.com/LLNL/axom/pull/1666, which was approved by @kennyweiss, @Arlie-Capps and @bmhan12.  Please re-approve.  I screwed up the merge step somehow and it didn't merge into `develop`.  The description below is exactly the same.

- This PR is a (work-in-progress) feature.
- It does the following:
  - Introduces a `conduit::Node` for internally storing geomety data and allow for new geometry formats and information without further changes to the code.  Currently, this information is stored in a variety of objects:
    1. primal geometries (`Tetrahedron`, `Hexahedron`, `Sphere`, etc.)
    2. aggregate data for geometries not supported by primal (for example, 2D function and refinement level for surface of revolution)
    3. file names for geometries stored in files (for example, ProE mesh)
    4. `sidre::Group` and topology name (for geometries stored in blueprint mesh convention)
   - Provides a way to populate, store and access geometry data without requiring numerous accessors and constructors.

Problems with using primitives in `Geometry`
  1. They don't have all the parameters needed for shaping (which is why `Geometry` has to keep supplemental data).
  2. We have to add new primitives and change `Geometry` to support new shapes.
For these reasons, primitives will be removed in favor of the hierarchy representation.  `Geometry` is supporting both during the transition.

This is a WIP.  It is being used in the on-going work to re-factor and improve clipping performance and may still change to support that goal.
# Dependencies
The mesh clipping refactor includes multiple PRs.  Please review these before this one.
- https://github.com/LLNL/axom/pull/1661
- https://github.com/LLNL/axom/pull/1662


kennyweiss (approved):

Arlie-Capps (approved):
Thanks, Brian.

Issue #1529, opened 2025-04-01
Add functionality to compute/access Gaussian quadrature nodes and weights


kennyweiss:
We currently get Gaussian quadrature nodes and weights from MFEM, and use these to integrate along curves. As a consequence, we cannot rely on this functionality in configurations without MFEM support.

It should be relatively straightforward for axom to directly provide this functionality.

adayton1:
We would definitely appreciate this!

PR #1683, opened 2025-10-03
Add Gauss-Legendre quadrature detached from MFEM include

jcs15c:
# Summary

This PR is a feature/enhancement that adds to `core/numerics` a method of directly computing Gauss-Legendre quadrature nodes and weights without relying on the analogous methods in MFEM. These methods are used in `winding_number.hpp` and `evaluate_integral.hpp`. This significantly reduces the number of `#ifdef` guards that need to be used to check for an MFEM dependency, particularly relevant for PR #1672.

Quadrature nodes are calculated in `calculate_gauss_legendre(npts)` through a Newton's method search which finds the roots of the Legendre polynomial of degree `npts`. Although the Newton iterations converge rapidly from the initial guess (which correctly assumes the nodes are nearly equally distributed in $$\arccos(x_i)$$), an additional interface is provided through `get_gauss_legendre(npts)` which stores the resulting quadrature rules in a static map so that repeated calls do not have to recompute the rule each time. General quadrature rules are stored in a `numerics::QuadratureRule` object which stores one `axom::Array<double>` of nodes and one of weights. This framework is readily extensible to definitions of other types of more specialized quadrature rules.

The accuracy and stability of the quadrature nodes is tested against the MFEM rules for even large values of `npts`, up to around ~10,000, but a rigorous implementation would use asymptotic formulas for evaluation of high-degree rules instead of Newton's method, as they could be evaluated significantly faster for such large orders. Because the new implementation is essentially identical to the original MFEM implementation, the two methods take about the same time to evaluate.


src/axom/primal/CMakeLists.txt
kennyweiss, line 1: 👍

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -26,32 +29,65 @@ namespace primal
...
+#else
+public:
+  QuadratureRuleView(const axom::numerics::QuadratureRule& rule) : m_quad_rule(rule) { }
+
+  double node(int idx) const { return m_quad_rule.node(idx); }
+  double weight(int idx) const { return m_quad_rule.weight(idx); }
+
+private:
+  const axom::numerics::QuadratureRule& m_quad_rule;
+#endif
```
kennyweiss: This is really nice!

Since it's only a few lines of code, would it be cleaner/clearer to define the entire class within the ifdef, e.g.

```cpp
#ifdef AXOM_USE_MFEM
class QuadratureRuleView
{
public:
  QuadratureRuleView(const mfem::IntegrationRule& rule) : m_quad_rule(rule) { }

  double node(int idx) const { return m_quad_rule.IntPoint(idx).x; }
  double weight(int idx) const { return m_quad_rule.IntPoint(idx).weight; }

private:
  const mfem::IntegrationRule& m_quad_rule;
};
#else
class QuadratureRuleView
{
public:
  QuadratureRuleView(const axom::numerics::QuadratureRule& rule) : m_quad_rule(rule) { }

  double node(int idx) const { return m_quad_rule.node(idx); }
  double weight(int idx) const { return m_quad_rule.weight(idx); }

private:
  const axom::numerics::QuadratureRule& m_quad_rule;
};
#endif
```

I'm not sure which is better...

kennyweiss: It would probably also be useful to have a function to get the order and/or num nodes/weights

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -60,30 +96,39 @@ inline double evaluate_scalar_line_integral_component(const primal::BezierCurve<
...
 inline double evaluate_vector_line_integral_component(const primal::BezierCurve<T, NDIMS>& c,
                                                       Lambda&& vec_field,
-                                                      const mfem::IntegrationRule& quad)
+                                                      const int npts)
 {
+#ifdef AXOM_USE_MFEM
+  static mfem::IntegrationRules my_IntRules(0, mfem::Quadrature1D::GaussLegendre);
+  QuadratureRuleView quad(my_IntRules.Get(mfem::Geometry::SEGMENT, 2 * npts - 1));
+#else
+  SLIC_WARNING_IF(npts > 100, "Axom's Gauss-Legendre nodes may be less precise for n > 100");
```
kennyweiss: Consider augmenting this warning with a note that configuring Axom w/ MFEM should provide more precise quadrature weights.

kennyweiss: I wonder if it's worthwhile to modify this warning to only print it out once. E.g if we call this on a loop w/ lots of BezierCurves, we'll get too many warnings about this.

(This might be better as a builtin SLIC macro -- `SLIC_*_ONCE` -- e.g. `SLIC_WARNING_IF_ONCE` )

src/axom/core/tests/numerics_quadrature.hpp
kennyweiss, line 1: This is a nice set of tests!

Please also add some tests to check that the axom Gaussian quadrature nodes/weights match those of MFEM.  (Obviously, only when Axom is configured with MFEM).

Note: We can have this test depend on MFEM, when Axom is configured w/ MFEM, even though `core` doesn't (and shouldn't) depend on MFEM. Alternatively, we can add those tests somewhere in the primal tests.

src/axom/core/numerics/quadrature.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+public:
+  QuadratureRule() = default;
+
+  //! \brief Accessor for quadrature nodes
+  double node(size_t idx) const { return m_nodes[idx]; };
+
+  //! \brief Accessor for quadrature weights
+  double weight(size_t idx) const { return m_weights[idx]; };
+
```
kennyweiss, line 41: I think we'll want to also expose the number of quadrature weights/nodes as a class member.

Would it make sense to also expose the order? Or does that only make sense for some quadrature rules. 
(It might be good to discuss how general we'd like this design to be beyond 1D Gauss-Legendre quadrature).

src/axom/core/numerics/quadrature.hpp
```diff
@@ -0,0 +1,84 @@
...
+class QuadratureRule
+{
+  // Define friend functions so rules can only be created via compute_rule() methods
+  friend QuadratureRule compute_gauss_legendre(int npts);
+
+public:
+  QuadratureRule() = default;
+
+  //! \brief Accessor for quadrature nodes
+  double node(size_t idx) const { return m_nodes[idx]; };
```
kennyweiss, line 36: Design question -- do we always want this to be `double`? Or would we ever want other floating point types?

I think starting w/ double makes sense, but wanted to raise the question.

src/axom/core/numerics/quadrature.cpp
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * \note If this method has already been called for a given order, it will reuse the same quadrature points
+ *  without needing to recompute them
+ *
+ * \return The `QuadratureRule` object which contains axom::Array<double>'s of nodes and weights
+ */
+const QuadratureRule& get_gauss_legendre(int npts)
+{
+  // Define a static map that stores the GL quadrature rule for a given order
+  static std::map<int, QuadratureRule> rule_library;
```
kennyweiss: Minor: Should this be a `std::map` or a hashmap (e.g. `std::unordered_map` or `axom::FlatMap`)?

I'm not sure if it matters too much for this PR, but we might want to revisit this in the future.

src/axom/core/numerics/quadrature.cpp
```diff
@@ -0,0 +1,138 @@
...
+  if(rule_library.find(npts) == rule_library.end())
+  {
+    rule_library[npts] = compute_gauss_legendre(npts);
+  }
```
kennyweiss: We might have thread safety concerns on this line in the future. 

(Should be fine for this PR, but maybe add a comment?)

kennyweiss (approved):
Thanks @jcs15c -- this is a really nice contribution!

It should resolve https://github.com/LLNL/axom/issues/1529

Please also update the RELEASE-NOTES.

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
jcs15c: I think you're right that putting the whole thing in the ifdef block is more clear, and probably translates better to future quadrature methods.

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
jcs15c: I totally agree, and the only reason it's not like that now is because I'm not entirely sure the full extent of SLIC's capabilities, or how to go about implementing it myself. Some kind of static boolean?

src/axom/core/tests/numerics_quadrature.hpp
jcs15c: Sounds reasonable enough to put them in primal_integral

src/axom/core/numerics/quadrature.cpp
jcs15c: No reason it can't be an `axom::FlatMap`, defaulting to `std::map` is a bad habit of mine. I've changed it now.

src/axom/core/numerics/quadrature.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+      if(done)
+      {
+        break;
+      }
+
+      // Compute the Newton method step size
+      dz = P_n / Pp_n;
+
+      if(std::fabs(dz) < 1e-16)
```
Arlie-Capps: This is the termination condition.  I feel hesitant to hard-wire the error tolerance to 1e-16.  Thoughts?

src/axom/core/numerics/quadrature.cpp
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  const int n = npts;
+  const int m = (npts + 1) / 2;
+
+  // Nodes are mirrored across x = 0.5, so only need to evaluate half
+  for(int i = 1; i <= m; ++i)
+  {
+    // Each node is the root of a Legendre polynomial,
+    //  which are approximately uniformly distirbuted in arccos(xi).
```
BradWhitlock: typo "distirbuted"

src/axom/core/numerics/quadrature.hpp
```diff
@@ -0,0 +1,88 @@
...
+  double weight(size_t idx) const { return m_weights[idx]; };
+
+  //! \brief Accessofr for the size of the quadrature rule
+  int getNumPoints() const { return m_npts; }
+
+private:
+  //! \brief Private constructor for use in compute_<rule>() methods. Only allocates memory
+  QuadratureRule(int npts) : m_nodes(npts), m_weights(npts), m_npts(npts) { };
+
+  axom::Array<double> m_nodes;
```
BradWhitlock: I would prefer to see `QuadratureRule` use `axom::ArrayView<double>` instead. The array ownership could be still done in your friend function. That function should probably also take an optional allocatorID. Then I'd like to see `QuadratureRule` decorated with `AXOM_HOST_DEVICE` and tested in some GPU kernels.

Or maybe instead of doing that, provide a method for returning a view of the QuadratureRule.
```
struct QuadratureRule
{
  struct View
  { AXOM_HOST_DEVICE double node(axom::IndexType index) const { return m_nodes[index]; }
    AXOM_HOST_DEVICE double weight(axom::IndexType index) const { return m_weights[index]; }
    AXOM_HOST_DEVICE axom::IndexType size() const { return m_nodes.size(); }
    axom::ArrayView<double> m_nodes, m_weights;
  };

  View view() const
  {
    return View{m_nodes.view(), m_weights.view()};
  }
};
```

BradWhitlock: Another thing - since the `axom::Array` contains a size, why also have a member to keep track of the number of points?

src/axom/core/numerics/quadrature.hpp
jcs15c: In my opinion, the order of a quadrature rule is important, but not nearly as descriptive as the number of points. As an example, a 10th order Gauss-Legendre rule has 6 points, but a 10th order Gauss-Lobatto rule has 8 points. Since we don't currently tie the `QuadratureRule` object to a specific method, I think it's better to leave it out for now.

src/axom/core/numerics/quadrature.cpp
jcs15c: Do you think replacing it with a `axom::numerics::floating_point_limits<double>::epsilon()` is better? I personally don't think it needs to be/should be an exposed tolerance, since its ambiguous (to me at least) how using approximate quadrature nodes impacts the accuracy of the downstream numerical integration. 

As it is, the algorithm already converges very quickly, and is very insensitive to the numerical stopping criteria. If you replace it with 10^-10, you get essentially the same number of iterations, since Newton's method converges quadratically.

src/axom/core/numerics/quadrature.hpp
BradWhitlock: I think you answered my npts question above. n/m

jcs15c: No, you're totally right, `QuadratureRule::size()` (or `.getNumPoints()`) should just call `m_nodes.size()` (which is always equal to `m_weigths.size()`, regardless of the method) instead of storing the extra integer. If it was an actual `.getOrder()` method, then it would have to be separate like this.

jcs15c: @BradWhitlock To your earlier point about views, just to make sure I understand correctly, that would that require that the static unordered_map in `get_gauss_legendre` instead have values of a pair of `axom::Array`s (or two maps of one array each), so that the `QuadratureRule` object can contain only views of those arrays, right? That would work, but I think I'd have to make `calculate_gauss_legendre` return the arrays themselves instead of a `QuadratureRule` object, or populate a const reference to the map, or just move its entire functionality into `get_gauss_legendre`. 

My original idea was to let `calculate_gauss_legendre` bypass the caching altogether and just compute the nodes on the fly, but I'm not sure what the utility of that would really be. Let me implement this view version and see if it's closer to what you have in mind.

src/axom/quest/LinearizeCurves.cpp
kennyweiss, line 1: 👍

src/axom/primal/tests/primal_integral.cpp
kennyweiss, line 1: Thanks for adding this!

src/axom/core/numerics/quadrature.cpp
```diff
@@ -0,0 +1,150 @@
...
+ * \note If this method has already been called for a given order, it will reuse the same quadrature points
+ *  without needing to recompute them
+ *
+ * \return The `QuadratureRule` object which contains axom::Array<double>'s of nodes and weights
+ */
+const QuadratureRule& get_gauss_legendre(int npts)
+{
+  // Define a static map that stores the GL quadrature rule for a given order
+  static axom::FlatMap<int, QuadratureRule> rule_library;
+  if(rule_library.find(npts) == rule_library.end())
```
kennyweiss: Consider adding a SLIC_WARNING if `npts < 1`

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -2489,7 +2488,8 @@ class NURBSPatch
    *  then computes the 2D area of that projection to get the corresponding
    *  component of the average surface normal.
    *  
-   * \note This requires the MFEM third-party library
+   * Evaluates line integrals with MFEM integration rules if available, 
+   *  or axom/numerics/quadrature otherwise, which are less robust for large npts (> ~100)
```
Arlie-Capps: Consider updating the comment here and (new) line 3103, since you did find good stability for many more npts.

src/axom/core/numerics/quadrature.cpp
jcs15c: I think it should actually be a stricter slic assert, since there's no meaning to a "0 point" quadrature rule.

src/axom/core/numerics/quadrature.hpp
BradWhitlock: Hi @jcs15c , I think you get what I mean but in case this helps clear it up...
```
// template it in case we wanted to call from a kernel and pass axom::StackArray or axom::StaticArray.
template <typename Container>
AXOM_HOST_DEVICE
void calculate_gauss_legendre(int npts, Container &nodes, Container &weights)
{
   // existing logic using these output arrays.
}

QuadratureRule get_gauss_legendre(int npts, int allocatorID = <some default value>)
{
   // Assumes std::map<std::pair<int,int>, std::pair<axom::Array<double>,axom::Array<double>>>;
   const auto key = std::make_pair<int,int>(npts, allocatorID);
   auto it = cache.find(key);
   if(it == cache.end())
   {
     it = cache[key];

     auto nodes = axom::Array<double>(npts);
     auto weights = axom::Array<double>(npts);
     calculate_gauss_legendre(npts, nodes, weights);

     // copy nodes, weights to arrays in target memory space
     it->second.first = axom::Array<double>(npts, npts, allocatorID);
     it->second.second = axom::Array<double>(npts, npts, allocatorID);
     axom::copy(it->second.first.data(), nodes.data(), npts * sizeof(double));
     axom::copy(it->second.second.data(), weights.data(), npts * sizeof(double));
   }
   return QuadratureRule{it->second.first.view(), it->second.second.view()};
}

Then later in other code, we could do
template <typename ExecSpace>
void func()
{
   // Get the rule for the right memory space, creating it if needed.
   // I was assuming that the function would be called prior to the axom::for_all. If we need to
   // call it for different numbers of points from inside the lambda then that is a problem with
   // what I proposed.
   const auto rule = get_gauss_legendre(5, axom::execution_space<ExecSpace>::allocatorID());

   // Use the rule in a lambda
   axom::for_all<ExecSpace>(n, AXOM_LAMBDA(axom::IndexType i)
   {
     double sum = 0.;
     for(int j = 0; j < rule.numPoints(); j++)
     {
        // These QuadratureRule methods would be AXOM_HOST_DEVICE to use them here
        sum += rule.weight(j) * f(rule.node(j));

        // Maybe we can do this now too...
        axom::StackArray<double, MAXNODES> nodes, weights;
        int nnodes = // something dynamic
        calculate_gauss_legendre(nnodes, nodes, weights);
        QuadratureRule newQR{axom::ArrayView<double>(nodes.data(), nnodes),
                                         axom::ArrayView<double>(weights.data(), nnodes)};
        // use newQR
     }
     someArray[i] = sum;
   });
}
```

jcs15c: Okay great, I think I got it pretty close then, aside from a handful of redundant cache lookups. Thank you!!

jcs15c: @BradWhitlock Would you mind taking another look at my implementation of the ArrayView version, and the test I'm using in `numerics_quadrature.hpp` (reproduced below) just to make sure this looks practical for future applications? In particular, I don't think it's necessary to be able generate rules from within the lambda, since the generation procedure isn't especially expensive, and by definition only needs to be done once for a given rule. 

On the other hand, the node computation procedure _within_ a certain rule is pretty well parallelizeable, and could hypothetically benefit from being ported to the GPU, but I don't think there's many contexts in which anyone could use a rule with an order high enough to make it worth it. Even generating ~100,000 points only takes about half a minute, and that was on a debug build. 

```cpp
template <typename ExecSpace>
struct test_device_quadrature
{
  static void test()
  {
    const int npts = 15;

    // Create the rule with the proper allocator
    const auto rule =
      axom::numerics::get_gauss_legendre(npts, axom::execution_space<ExecSpace>::allocatorID());

    // Use the rule in a lambda to integrate the volume under std::sin(pi * x) on [0, 1]
    axom::ReduceSum<ExecSpace, double> quadrature_sum(0.0);
    axom::for_all<ExecSpace>(
      npts,
      AXOM_LAMBDA(axom::IndexType i) { quadrature_sum += rule.weight(i) * sin(rule.node(i)); });

    EXPECT_NEAR(quadrature_sum.get(), 0.459697694132, 1e-6);
  }
};

TEST(numerics_quadrature, get_nodes_seq) { test_device_quadrature<axom::SEQ_EXEC>::test(); }

#if defined(AXOM_RUNTIME_POLICY_USE_OPENMP)
TEST(numerics_quadrature, get_nodes_omp) { test_device_quadrature<axom::OMP_EXEC>::test(); }
#endif

#if defined(AXOM_RUNTIME_POLICY_USE_CUDA)
TEST(numerics_quadrature, get_nodes_cuda) { test_device_quadrature<axom::CUDA_EXEC<256>>::test(); }
#endif

#if defined(AXOM_RUNTIME_POLICY_USE_HIP)
TEST(numerics_quadrature, get_nodes_hip) { test_device_quadrature<axom::HIP_EXEC<256>>::test(); }
#endif
```

src/axom/core/numerics/quadrature.cpp
jcs15c: Actually, the simplest reason it shouldn't be a hashmap is because there's no built-in way to hash a pair of ints. I've accidentally avoided the same issue in the GWN memoization code, which also has maps keyed by `pair<RefinementLevel, RefinementIndex>` by (in retrospect incorrectly) using std::map there too. 

It's easy enough to make and use a simple custom hash for a pair of ints, but ideally that implementation would be shared by this and the memoization code, and be more general to arbitrary pairs. For now, I think it's simplest to use `std::map` here, and figure out later where that shared implementation should go. Maybe in `FlatTable.hpp`?

kennyweiss: Thanks @jcs15c -- can you please create an issue for this?

Tag: @publixsubfan

src/axom/core/numerics/quadrature.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+      // Take the Newton step, repeat the process
+      z -= dz;
+    }
+
+    nodes[i - 1] = xi;
+    nodes[n - i] = 1.0 - xi;
+
+    // Evaluate the weight as w_i = 2 / (1 - z^2) / P'_n(z)^2, with z \in [-1, 1]
+    weights[i - 1] = weights[n - i] = 1.0 / (4.0 * xi * (1.0 - xi) * Pp_n * Pp_n);
```
kennyweiss: The comment doesn't (appear to) match the code -- presumably it's in how $z^2$ and $xi$ are defined

kennyweiss (approved):
I took another pass through the latest changes -- looks great to me.

src/axom/core/numerics/quadrature.cpp
jcs15c: Fair point. The lines of code above say that `xi = (1 - z) / 2`, but it was a bit confusing where the extra factor of 1/2 came from. I've updated the comment to be more clear.

PR #1696, opened 2025-10-22
Enable larger size for statically-sized polygons used in MIR and BUMP.

BradWhitlock:
MIR and BUMP use a class called `PrimalAdaptor` to wrap Blueprint shapes as primal geometry so we can do operations such as clipping. Clipping is used in `bump::TopologyMapper` and `mir::ElviraAlgorithm` to figure out shape overlaps using clip operations.

I was using `TopologyMapper` and ran into a case where it was clipping a 4-gon vs a 5-gon. This caused clip to fail in a `SLIC_ASSERT` because for statically-sized polygons, the default number of max vertices is 8 and there is a check in clip that wants to make sure that polygon max vertices is greater than or equal to the sum of the 2 input polygons. This failed since (4+5 > 8).

This PR adds a `MAX_VERTS=12` default template argument to some classes to make static polygons a little bigger by default and so the size is user-settable so clients that need larger sizes than that can bump it up when instantiating these classes. This affects:
* `axom::bump::PrimalAdaptor`
* `axom::bump::TopologyMapper`
* `axom::mir::ElviraAlgorithm`

`TopologyMapper` and `ElviraAlgorithm` pass their `MAX_VERTS` value to their internal `PrimalAdaptors` too so clipping can operate on the larger polygon sizes.

Other issues fixed:
* Copying an empty conduit::Node resulted in a problem due to some allocator checks
* MIR no-op case was not renaming objects when object names were supplied in the options.
* MIR on clean matset with selected zones was not working.

I added some new tests:
* New TopologyMapper test that does oct-oct clipping and sets MAX_VERTS to 16
* New "mir_coupling3d" test for MIR with the TopologyMapper on 3D strided-structured meshes. I found/fixed a bug in one of the functions to make 3D strided-structured topology views that was causing MIR to mess up.


src/axom/bump/TopologyMapper.hpp
```diff
@@ -405,6 +405,8 @@ AXOM_HOST_DEVICE double shapeOverlap(const VariableShape<T, 3> &shape1,
  * \tparam SrcMatsetView The view type for the source matset.
  * \tparam TargetTopologyView The view type for the target topology.
  * \tparam TargetCoordsetView The view type for the target coordset.
+ * \tparam MAX_VERTS The maximum number of vertices allowed in a polygon. This
+ *                   value is used only in 2D.
```
kennyweiss: Would it make sense to name it something like `MAX_VERTS_2D`?

BradWhitlock: I made that change.

src/axom/mir/tests/mir_coupled3d.cpp
```diff
@@ -0,0 +1,534 @@
...
+#include <cmath>
+#include <cstdlib>
+
+namespace utils = axom::bump::utilities;
+namespace views = axom::bump::views;
+
+std::string baselineDirectory()
+{
+  return pjoin(dataDirectory(), "mir", "regression", "mir_coupled3d");
+}
```
Arlie-Capps, line 25: I don't see where `baselineDirectory()` gets used in this file.  If it is used, consider returning "mir_coupling3d" to match the name of the file.  Otherwise, if there's no use for this, please remove.

BradWhitlock: baselineDirectory() is used by the TestApp object so it knows where to look for baseline files. I renamed the "mir_coupling" tests to "mir_coupled" so the names match the baselines.

PR #1711, opened 2025-11-07
Improves array accessors in Bezier and NURBS classes

### LINKED ISSUE #1528: Improvements to `BezierCurve`, `BezierSurface`, `NURBSCurve` and `NURBSPatch` interfaces
While using the `NURBSCurve` class, I noticed several usability enhancements:
- [ ] Simplify the constructors -- there are numerous constructors for these classes, taking arrays of control points, weights and knot spans. Each has overloads for axom::Array and c-style arrays (+extra size parameters). We should simplify this to take array-like template parameters for classes that have a `size()` function and a subscript operator. This should then work for `std::vector`, axom::Array` and `axom::ArrayView`. It will be straightforward for users to wrap their c-style arrays in an axom::Array.
- [ ] Provide references to control points, weights and knot spans. Without these, it can be awkward to copy/manipulate the curves and surfaces

### LINKED ISSUE #1709: sina docs formatting issues
Looks like some of the bullets here have issues:
https://axom.readthedocs.io/en/develop/axom/sina/docs/sphinx/documents.html#appending-documents-to-existing-sina-files

for example, the continuation of a sentence that starts in a bullet in a new paragraph:

Image

kennyweiss:
# Summary

- This PR improves the accessors to the underlying arrays in primal's `BezierCurve`, `BezierPatch`, `NURBSCurve`, `NURBSPatch`, `CurvesPolygon` and `KnotVector` classes.
- The API previously returned a copy of the arrays by value. It now returns by reference. I also added const reference accessors
- These changes allowed me to improve the way we broadcast NURBS curve data in our `PC2CReader` class
- Misc: I also made some minor optimizations to the `linear_winding_number` function and fixed some other minor errors/warnings along the way, e.g. for configs without Lua, and in the generated sphinx docs
- Resolves #1528 
- Resolves #1709

### DIFF
--- a/src/axom/primal/geometry/NURBSCurve.hpp
+++ b/src/axom/primal/geometry/NURBSCurve.hpp
@@ -660,10 +660,10 @@ public:
     m_weights[idx] = weight;
   }
 
-  /// \brief Returns a reference to the NURBS curve's control points
+  /// \brief Returns a reference to the NURBS curve's weights
   WeightsVec& getWeights() { return m_weights; }
 
-  /// \brief Returns a const reference to the NURBS curve's control points
+  /// \brief Returns a const reference to the NURBS curve's weights
   const WeightsVec& getWeights() const { return m_weights; }
 
   /// \brief Returns an axis-aligned bounding box containing the NURBS curve
@@ -745,10 +745,10 @@ public:
   /// \brief Return a const reference to the knot vector
   const KnotVectorType& getKnots() const { return m_knotvec; }
 
-  /// \brief Get minimum knot
+  /// \brief Get minimum knot value
   T getMinKnot() const { return m_knotvec.getMinKnot(); }
 
-  /// \brief Get maximum knot
+  /// \brief Get maximum knot value
   T getMaxKnot() const { return m_knotvec.getMaxKnot(); }
 
   /// \brief Reverses the order of the NURBS curve's control points and weights
--- a/src/axom/sina/docs/sphinx/documents.rst
+++ b/src/axom/sina/docs/sphinx/documents.rst
@@ -161,7 +161,7 @@ first document to the filesystem, and when you reach a point where you would lik
 use an append function matching the filetype (JSON is recommended for small files, namely anything
 not involving timeseries, and HDF5 for anything beefier). There's some nuance to how appending works:
 
-- If your new document contains records not present in the file on disk, it will add them in.
+- If your new document contains records not present in the file on disk, it will add them.
   This is generally how you'll want to do snapshots; it determines whether records match by checking the ID,
   so as long as each snapshot has a unique ID, they'll accumulate in the target file.
 
@@ -170,19 +170,19 @@ not involving timeseries, and HDF5 for anything beefier). There's some nuance to
   to accumulate timeseries values. Simply ensure that the record containing new values has the same
   ID as the one containing what you have so far.
 
-- For data, files, user defined, and anything else that isn't curve sets or libraries: Sina will
-  go through field-by-field. If there's a field not already present (ex: runtime), then it will be added in.
+- For data, files, user defined, and anything else that isn't curve sets or libraries, it will
+  go through field-by-field. If there's a field not already present, it will be added.
   If it IS already present, an optional argument allows you to define the behavior. By default, newest wins,
   but you can also have oldest win or refuse the write.
 
-- For curve sets: Sina will append new values to the existing dependents/independents. There is some
+- For curve sets, it will append new values to the existing dependents/independents. There is some
   checking to ensure all timeseries WITHIN A CURVE SET end up the same length, so while it is possible
   to add new dependents (or independents) to a set of curves while a simulation is running, ensure that
   you're backfilling the required number of values (ex: you dump every 10 cycles, you've dumped 20 times, each
   curve has 200 values. If on your next dump you also add brand_new_curve, it MUST have 210 values.) You'll
-  almost always want to define curves up front and simply add in new values.
+  almost always want to define curves up front and add in new values.
 
-- For library_data, append will recurse, following the above rules
+- For library_data, append will recurse, following the above rules.
 
 In general, appending is very powerful, but a bit complicated; if Sina encounters any issues, it will
 write them to the returned conduit node for troubleshooting.


src/axom/sina/docs/sphinx/documents.rst
```diff
@@ -154,35 +154,35 @@ properly queried:
...
+  but you can also have oldest win or refuse the write.
+
+- For curve sets: Sina will append new values to the existing dependents/independents. There is some
+  checking to ensure all timeseries WITHIN A CURVE SET end up the same length, so while it is possible
+  to add new dependents (or independents) to a set of curves while a simulation is running, ensure that
+  you're backfilling the required number of values (ex: you dump every 10 cycles, you've dumped 20 times, each
+  curve has 200 values. If on your next dump you also add brand_new_curve, it MUST have 210 values.) You'll
+  almost always want to define curves up front and simply add in new values.
+
+- For library_data, append will recurse, following the above rules
```
kennyweiss: @doutriaux1 -- This fixes a sphinx warning in the sina docs (https://github.com/LLNL/axom/issues/1709). 
Here's what it looks like w/o the fixes:

Image

See: https://axom.readthedocs.io/en/v0.12.0/axom/sina/docs/sphinx/documents.html#appending-documents-to-existing-sina-files

src/axom/primal/operators/detail/winding_number_2d_impl.hpp
kennyweiss, line 1: I profiled a release config of the 2D  `quest_winding_number` application w/ memoization, and this function accounted for about 80% of the runtime! (mostly in the `acos` call at the end)

src/axom/sina/docs/sphinx/documents.rst
doutriaux1: thanks @kennyweiss

src/axom/primal/operators/detail/winding_number_2d_impl.hpp
jcs15c: I think that's good news, right? This method and the `acos` is really the fundamental "mathy" step of the whole GWN process, so it's reassuring that we're not spending too much of the runtime just processing the geometry. To my understanding, the `acos` itself is pretty unavoidable, except for maybe swapping it with some sort of `atan` or maybe even a `log`.

kennyweiss: I agree -- I was mostly noting why I was looking at this function in this PR. I wanted to see which of the Bezier/NURBS functions showed up after my optimizations and this function was dominating the runtime (at least in the examples that I looked at).

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -458,24 +466,25 @@ class KnotVector
...
+      axom::utilities::clampVal(target_multiplicity - multiplicity, 0, m_deg - multiplicity);
 
     insertKnotBySpan(span, t, r);
   }
 
   /// \brief Checks if given parameter is in knot span (to a tolerance)
   bool isValidParameter(T t, T EPS = 1e-5) const
   {
-    return t >= m_knots[0] - EPS && t <= m_knots[m_knots.size() - 1] + EPS;
+    return (t >= m_knots[0] - EPS) && (t <= m_knots[m_knots.size() - 1] + EPS);
```
adayton1: Should `getMinKnot()` and `getMaxKnot()` be used here?

adayton1: And in the function below?

src/axom/primal/geometry/NURBSCurve.hpp
```diff
@@ -654,8 +657,11 @@ class NURBSCurve
     m_weights[idx] = weight;
   }
 
-  /// \brief Returns a copy of the NURBS curve's control points
-  WeightsVec getWeights() const { return m_weights; }
+  /// \brief Returns a reference to the NURBS curve's control points
```
adayton1: I think the comment needs to be updated to mention weights instead of control points.

src/axom/quest/io/PC2CReader.cpp
```diff
@@ -41,37 +41,34 @@ int PC2CReader::read()
...
+  // handle rank 0
   case 0:
     rc = C2CReader::read();
+
     if(rc == READER_SUCCESS)
     {
       bcast_int(m_nurbsData.size());
 
-      for(const auto& curve : m_nurbsData)
+      for(auto& curve : m_nurbsData)
```
adayton1, line 52: Could this be `const auto& curve`?

src/axom/quest/io/PC2CReader.cpp
```diff
@@ -41,37 +41,34 @@ int PC2CReader::read()
...
 
-        AXOM_UNUSED_VAR(numKnots);
-
-        // broadcast control points as an array of doubles
-        auto pt_v = numCP > 0
-          ? axom::ArrayView<double>(const_cast<double*>(curve[0].data()), numCP * DIM)
-          : axom::ArrayView<double>(nullptr, 0);
+        // broadcast control points
+        // since they're contiguous, we can broadcast them as an array of doubles
+        auto pt_v = numCP > 0 ? axom::ArrayView<double>(curve[0].data(), numCP * DIM)
```
adayton1, line 61: Is there a `curve.getControlPoints()` method you could use?

kennyweiss: This line is treating an array of `numCP` `primal::Point<double,DIM>` as an array of `numCP * DIM` doubles, so it's grabbing a pointer to the first coordinate of the first point and creating an `axom::ArrayView` from that pointer.

If I do `curve.getControlPoint()`, the type is PointType, and I would still need to cast it.

I tried to comment about this, but perhaps it wasn't explicit enough.

src/axom/quest/io/PC2CReader.cpp
kennyweiss: Thanks -- that's what I had originally, but the type(s) didn't work out -- if I did that, I'd need to `const_cast` the pointers (at the least the control point pointer), and that seemed messier than just using a non-const iterator here.

src/axom/primal/geometry/NURBSCurve.hpp
rhornung67: And in the method below...

src/axom/sina/docs/sphinx/documents.rst
```diff
@@ -154,35 +154,35 @@ properly queried:
...
-    you're backfilling the required number of values (ex: you dump every 10 cycles, you've dumped 20 times, each
-    curve has 200 values. If on your next dump you also add brand_new_curve, it MUST have 210 values.) You'll
-    almost always want to define curves up front and simply add in new values.
-
-    - For library_data, append will recurse, fulfilling a and b above.
+  the data they contain. This is most useful when you have a longer-running simulation where you want
+  to accumulate timeseries values. Simply ensure that the record containing new values has the same
+  ID as the one containing what you have so far.
+
+- For data, files, user defined, and anything else that isn't curve sets or libraries: Sina will
```
rhornung67:
```suggestion
- For data, files, user defined, and anything else that isn't curve sets or libraries, Sina will
```

src/axom/sina/docs/sphinx/documents.rst
```diff
@@ -154,35 +154,35 @@ properly queried:
...
+  go through field-by-field. If there's a field not already present (ex: runtime), then it will be added in.
```
rhornung67: I don't understand the parenthetical statement here. Do you mean something is added at runtime?

Arlie-Capps: Or is it that the field "runtime" will be added if not already present?  You could either write 

"field not already present, such as \"runtime\", then it will"

or even leave it out entirely:

"field not already present, then it will"

PR #1715, opened 2025-11-08
Enhance material views with iterators to traverse zone id/vf data.

BradWhitlock:
Material views in Axom take a MAXMATS template argument that sets their IDList and VFList types' sizes. These are used to get ids and volume fractions for a zone. In real applications the MAXMATS argument might need to be large O(100+). This could make kernels that call a material view's `zoneMaterials()` method require a fair amount of stack space or GPU registers.

This PR:
* Enhanced material views with a const_iterator class that can be used to traverse data for a particular zone. This is an alternate way of getting the ids/vfs with smaller stack space costs.
* Enhanced material views with a `zoneMaterials()` overload that takes `ArrayView`.
* Changed algorithms that used `zoneMaterials()` so they use iterators instead.
* Changed `ElviraAlgorithm` so it uses const_iterators or `zoneMaterials(ArrayView)` where possible
* Added heavily_mixed MIR example that makes a matset with many materials per zone and made sure MIR can operate on it
* Reduced clip tolerances in `ElviraAlgorithm` to work around a clipping problem

```
// Use matset view iterators
const auto end = matsetView.endZone(zoneIndex);
for(auto zoneMat = matsetView.beginZone(zoneIndex); zoneMat != end; zoneMat++)
{
  std::cout << "id=" << zoneMat.material_id() <<  ", vf=" << zoneMat.volume_fraction() << std::endl;
}
```

Heavily mixed
image


RELEASE-NOTES.md
kennyweiss, line 1: 👍

src/axom/mir/examples/concentric_circles/runMIR.hpp
```diff
@@ -10,6 +10,50 @@
...
+        {
+          sum += it.volume_fraction();
+        }
+        vfSum += sum;
+      });
+    vf2 = vfSum.get();
+  }
+
+  std::cout << "test_matset_traversal: vf1=" << vf1 << ", vf2=" << vf2
+            << ", nzones=" << matsetView.numberOfZones() << std::endl;
```
kennyweiss: It looks like `vf1` and `vf2` are supposed to be (nearly) the same -- should this function compare them? or at least indicate that they are supposed to be (nearly) the same

src/axom/bump/views/MaterialView.hpp
```diff
@@ -163,6 +183,104 @@ class UnibufferMaterialView
     return false;
   }
 
+  /*!
+   * \brief An iterator class for iterating over the mat/vf data in a zone.
+   */
```
kennyweiss, line 190: Please document what can be done with these iterators.

Are these more like const iterators, in the sense that they can only be used to access the values of the volume fractions/indexes, but cannot modify the materials in each zone or the volume fractions of the materials in the zone?

kennyweiss (approved):
Thanks @BradWhitlock -- the only question I had was about the documentation (and perhaps naming) of the iterator since it appears to offer read-only access to the data

src/axom/mir/examples/concentric_circles/runMIR.hpp
BradWhitlock: I added a comparison and changed to `SLIC_INFO`.

src/axom/bump/views/MaterialView.hpp
BradWhitlock: I renamed these classes to `const_iterator` and tried to improve the documentation.

PR #1717, opened 2025-11-09
Update to shroud-0.14.0

ltaylor16:
# Summary

- This PR is a devtool update, issue #1716
- The PR updates the Shroud wrappers to use version 0.14.0
- It uses the public install of shroud on LC instead of updating TPL.
- Changes the host-configs files to use this version. Update spack files to use correct path after next TPL update.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -21,6 +21,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 ### Added
 - Adds the `AXOM_ENABLE_TUTORIALS` configuration variable (`ON` by default)
 - Adds a tutorial on shaping in Axom and associated infrastructure in `quest`, `klee`, `inlet` and `sidre`.
+- Added iterators and Attribute class to the Python interface for sidre.
 
 ###  Changed
 - Treatment of materials on strided-structured Blueprint meshes has changed in `axom::mir`.
--- a/scripts/spack/packages/axom/package.py
+++ b/scripts/spack/packages/axom/package.py
@@ -378,23 +378,20 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
             )
             hip_link_flags += "-lmpi_gtl_hsa "
 
+            if spec.satisfies("^hip@6.0.0:"):
+                hip_link_flags += "-L{0}/lib/llvm/lib -Wl,-rpath,{0}/lib/llvm/lib ".format(rocm_root)
+            else:
+                hip_link_flags += "-L{0}/llvm/lib -Wl,-rpath,{0}/llvm/lib ".format(rocm_root)
+            # Only amdclang requires this path; cray compiler fails if this is included
+            if spec.satisfies("%llvm-amdgpu"):
+                hip_link_flags += "-L{0}/lib -Wl,-rpath,{0}/lib ".format(rocm_root)
+            hip_link_flags += "-lpgmath -lompstub "
+
             # Fixes for mpi for rocm until wrapper paths are fixed
             # These flags are already part of the wrapped compilers on TOSS4 systems
             if spec.satisfies("+fortran") and self.is_fortran_compiler("amdflang"):
-
                 hip_link_flags += "-Wl,--disable-new-dtags "
-
-                if spec.satisfies("^hip@6.0.0:"):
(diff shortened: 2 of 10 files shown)


src/axom/sidre/interface/c_fortran/utilSidre.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Release library allocated memory.
+void SIDRE_SHROUD_memory_destructor(SIDRE_SHROUD_capsule_data *cap)
+{
+  void *ptr = cap->addr;
+  switch(cap->idtor)
```
kennyweiss, line 21: Where are these `idtor` values mapped from integers to types?
e.g. 
* 1 -> `string`
* 2->`DataStore`

src/axom/sidre/interface/c_fortran/wrapDataStore.cpp
```diff
@@ -44,31 +42,59 @@ SIDRE_DataStore *SIDRE_DataStore_new(SIDRE_DataStore *SHC_rv)
   // splicer begin class.DataStore.method.new
   axom::sidre::DataStore *SHCXX_rv = new axom::sidre::DataStore();
   SHC_rv->addr = static_cast<void *>(SHCXX_rv);
-  SHC_rv->idtor = 1;
+  SHC_rv->idtor = 2;
```
kennyweiss, line 45: I see the `2` here as the `idtor` for `DataStore` -- is this tracked internally by shroud?

src/axom/slic/interface/c_fortran/utilSLIC.cpp
```diff
@@ -0,0 +1,134 @@
...
+    std::string *cxx_ptr = reinterpret_cast<std::string *>(ptr);
+    delete cxx_ptr;
+    break;
+  }
+  case 3:  // new_string
+  {
+    std::string *cxx_ptr = reinterpret_cast<std::string *>(ptr);
+    delete cxx_ptr;
+    break;
+  }
```
kennyweiss, line 58: Similar question here -- how/where is the mapping defined? (This one has two `std::string` mappings)

src/axom/sidre/interface/c_fortran/wrapDataStore.cpp
ltaylor16: The mapping is tracked internally by Shroud so that the wrapper knows how to delete the memory.
I have thought about creating some #define or an enum instead of using the integers directly. When a new destructor is required it can often renumber the `idtor` values creating a lot of meaningless differences. The user never needs to know the `idtor` value.
The other option is to create a function for each destructor. This is similar to what the Python wrapper does for the capsule argument. While Fortran can do function pointers, the integer index was much simpler.

src/axom/slic/interface/c_fortran/utilSLIC.cpp
ltaylor16: It uses a text string to decide when it needs a destructor. `std::string` vs `new_string`. They should be reconciled into one use case. Currently they are created via two different paths. One is generated for the typemap, `std::string`, and one from a `destructor_name` in `fc-statements.json`.

src/axom/sidre/interface/c_fortran/typesSidre.h
```diff
@@ -21,45 +33,84 @@ extern "C" {
...
+struct s_SIDRE_SHROUD_capsule_data
 {
-  void *addr; /* address of C++ memory */
-  int idtor;  /* index of destructor */
+  void *addr;    /* address of C++ memory */
+  int idtor;     /* index of destructor */
+  int cmemflags; /* memory flags */
+};
+typedef struct s_SIDRE_SHROUD_capsule_data SIDRE_SHROUD_capsule_data;
+#if 0
```
Arlie-Capps, line 44: Buffer and DataStore are now not visible?

ltaylor16: Currently only the `s_SIDRE_SHROUD_capsule_data` type is used for all classes which has a `void *addr` field. So every time it is used in a wrapper it needs to be cast to the correct type. This code is a path towards having a capsule struct for each class so that all of the casting can be avoided.

Arlie-Capps (approved):

PR #1720, opened 2025-11-11
Bugfix/probinso/no openmp build fixes

robinson96:
# Summary

- This PR is a bugfix
  - Fixes build when ENABLE_OPENMP=0 and OPENMP_FOUND=1

### DIFF
--- a/src/cmake/axom-config.cmake.in
+++ b/src/cmake/axom-config.cmake.in
@@ -33,7 +33,7 @@ if(NOT AXOM_FOUND)
   set(AXOM_USE_CUDA           "@AXOM_USE_CUDA@")
   set(AXOM_USE_HIP            "@AXOM_USE_HIP@")
   set(AXOM_USE_MPI            "@AXOM_USE_MPI@")
-  set(AXOM_USE_OPENMP         "@AXOM_ENABLE_OPENMP@")
+  set(AXOM_USE_OPENMP         "@AXOM_USE_OPENMP@")
 
   # Axom components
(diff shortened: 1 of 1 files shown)


rhornung67 (approved):
Thank you @robinson96

cyrush:
It would be good to know what cases you see `ZZZ_FOUND`, when `ENABLE_ZZZ` is not used.

Not sure how that could happen in a product using BLT unless you are using a fused cmake build system.

robinson96:
> It would be good to know what cases you see `ZZZ_FOUND`, when `ENABLE_ZZZ` is not used.
> 
> Not sure how that could happen in a product using BLT unless you are using a fused cmake build system.

We in fact, are using a fused cmake build system, so that's a possibility. I encountered this while doing a openmp disabled build of our code base, where we globally set ENABLE_OPENMP to 0.

I'm not sure where OPENMP_FOUND is being set, exactly.  All I know is that ENABLE_* is consistently the pattern for user requests, and OPENMP_FOUND is an internal variable somewhere. It's also possible it's being exported by a dependency of axom which was configured . 

Unfortunately, not sure how easy it will be to recreate the situation, and I'm sorry I missed the conversation on this until now.

robinson96:
Hey all, I took some time today to try to reproduce the original issue that led me to this change. So far, I've been unable to do so. I'm going to revert my codebase ,close this PR, and I'll reopen if I bump into it again with better documentation as to what led to it.

PR #1721, opened 2025-11-12
Use std::invoke_result_t instead of deprecated/removed std::result_of

adayton1:
# Summary

- This PR is a bugfix
- It does the following:
  - std::result_of was deprecated in C++17 and removed in C++20
  - Uses std::invoke_result_t instead
  - Handles the first case in issue #1718.


kennyweiss (approved):
Thanks @adayton1 !

rhornung67 (approved):
@adayton1 did you say earlier that there are similar changes needed to some other files?

PR #1728, opened 2025-12-03
Array optimizations for insertions

publixsubfan:
# Summary

- Implements the following optimizations which apply to `Array::push_back` and `Array::emplace_back`:
  - Make `ArrayOpsBase` a member variable, and use the member variable for operations. This avoids extraneous queries of the Umpire memory space for the current allocator ID.
  - Add a new method `Array::reserveForPushBack` which simplifies the more general logic from `Array::reserveForInsert` for a single insertion at the end of the elements.
  - Mark test for resize in `Array::reserveForPushBack` with an unlikely macro (see next bullet point), to optimize for the more common case of insertions that do not require a reallocation.
- Adds new macros `AXOM_LIKELY` and `AXOM_UNLIKELY` to be applied to `if`-statements. These serve as compiler hints to indicate that a branch path is likely/unlikely to be taken, and are implemented as follows:
  - With C++20 support: using the new `[[likely]]`/`[[unlikely]]` attributes
  - Without C++20, but on GCC/Clang: using `__builtin_expect`

# Performance data

**TL;DR**: Our slowdown in the worst case of `std::vector<int>` to `axom::Array<int>` went from ~100x to 2-3x.

I believe the remaining slowdown is attributable to two issues: the use of virtual functions for `Array::updateNumElements`, and an if-statement on the `ArrayOps::emplace()` path to handle a push-back on an Array allocated in device-only memory. These will be bigger code changes, and I'll push them out in a follow-on PR.

Using the `core_benchmark_array` test @kennyweiss added, here's the current speedup:

### `push_back` without reserving memory

  | `int` | `pair<int, int>` | `std::string`
-- | -- | -- | --
std::vector push_back | 40,799 | 332,789 | 3,474,727
Array push_back | 4,585,356 | 4,851,556 | 7,863,931
Make ArrayOps a member var | 206,508 | 521,558 | 3,501,949
Specialize reserveForInsert | 138,097 | 442,305 | 3,503,908
Mark resize as unlikely | 101,703 | 389,399 | 3,480,322
**Slowdown before** | 112.38x | 14.58x | 2.26x
**Slowdown after** | 2.49x | 1.17x | 1.00x

### `push_back` with pre-reserved axom::Array/std::vector

  | int | pair | string
-- | -- | -- | --
std::vector push_back | 35,579 | 35,804 | 3,382,912
Array push back | 4,588,348 | 4,567,329 | 7,706,043
Make ArrayOps a member var | 195,521 | 195,462 | 3,311,500
Specialize reserveForInsert | 101,944 | 126,184 | 3,268,466
Mark resize as unlikely | 96,485 | 108,503 | 3,316,208
**Slowdown before** | 128.96x | 127.56x | 2.28x
**Slowdown after** | 2.71x | 3.03x | 0.98x

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -27,6 +27,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   the append function that matches your protocol.
 - Adds Sina fortran functions `sina_set_curves_order` and `sina_set_record_curves_order`
 - Sidre: Added iterators and Attribute class to the Python interface.
+- Adds new optimization hint macros `AXOM_LIKELY` and `AXOM_UNLIKELY` to mark likely/unlikely
+  paths in if-statements.
 
 ###  Changed
 - Treatment of materials on strided-structured Blueprint meshes has changed in `axom::mir`.
@@ -49,6 +51,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 
 ###  Fixed
 - Sina's Fortran tests are now running (instead of silently failing)
+- Optimized `Array::push_back()` and `Array::emplace_back()` operations.
 
 ###  Deprecated
 


src/axom/core/Array.hpp
```diff
@@ -1667,26 +1670,29 @@ inline T* Array<T, DIM, SPACE>::reserveForInsert(IndexType n, IndexType pos)
...
+  }
+  IndexType end = m_num_elements;
+  updateNumElements(m_num_elements + 1);
+  return end;
 #else
   // Device path: supports insertion while m_num_elements < m_capacity
   // Does not support insertions which require reallocating the underlying
   // buffer.
-  IndexType new_pos = axom::atomicAdd<axom::auto_atomic>(&m_num_elements, n);
+  IndexType new_pos = axom::atomicAdd<axom::auto_atomic>(&m_num_elements, 1);
```
kennyweiss: I think this needs to be cast to an `axom::IndexType`, i.e.

```suggestion
  IndexType new_pos = axom::atomicAdd<axom::auto_atomic>(&m_num_elements, axom::IndexType{1});
```

I ran into this here -- https://github.com/LLNL/axom/pull/1553/files#diff-3b666631d68cdbdedc67abd9461d88f819f40c8d329b965b4ffb19f76c54074e

src/axom/core/Array.hpp
```diff
@@ -956,12 +954,10 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM, SPACE>>
...
    * \brief Make space for a subsequent insertion into the array.
    *
-   * \param [in] n the number of elements to insert.
-   *
    * \note This version supports concurrent GPU insertions.
-   * \note Reallocation is not supported.
+   * \note Reallocation is not supported on the GPU.
    */
-  AXOM_DEVICE IndexType reserveForDeviceInsert(IndexType n);
+  AXOM_HOST_DEVICE IndexType reserveForPushBack();
```
kennyweiss, line 960: Is this public? If so, please note the change in the RELEASE_NOTES.

kennyweiss (approved):
Thanks @publixsubfan -- awesome performance improvements!

Please don't forget to update the RELEASE_NOTES.

src/axom/core/Array.hpp
kennyweiss: Looks like you caught this one already!

src/axom/core/Array.hpp
publixsubfan: This is a private method, I don't anticipate any changes to the external interface for users in this changeset.

PR #1731, opened 2025-12-04
Devirtualize axom::Array

publixsubfan:
# Summary

- Add a `StoragePolicy` template argument to `axom::Array` to implement custom handling for Sidre-based arrays, in lieu of using virtual functions for `Array::updateNumElements()` and `Array::dynamicRealloc()`.
  Each `StoragePolicy` consists of the following interface:
  - `onShapeUpdate()`: called to allow for passing updated dimensions of the valid data elements back to Sidre. This is a no-op for the default `axom::Array`.
  - `reallocate()`: called from `Array::dynamicRealloc()` to get a buffer with a new size.
  - `deallocate()`: called when the buffer is no longer needed. No-op for view-like types like `sidre::Array`.

  Three storage policies have been added to Axom:
  -  `DefaultStoragePolicy`: the standard storage policy for Axom. Uses `axom::allocate/deallocate` to handle memory.
  - `SidreStoragePolicy`: storage policy for `sidre::Array`. Holds a `sidre::View` pointer and manages the array memory through this view.
  - `ExternalStoragePolicy`: storage policy for `mint::ExternalArray`. Treats the memory as externally-managed, by leaving `reallocate()` and `deallocate()` as no-ops.

- Add a new helper class to Mint, `mint::detail::ArrayWrapper<T>`, which serves to replace the polymorphic usage of `axom::Array` in `mint::ConnectivityArray`. This class is implemented as a simple wrapper of a `std::variant` holding one of the array types.

# Performance

From the `core_benchmark_array` test:

### `push_back` without reserving memory

  | `int` | `pair<int, int>` | `std::string`
-- | -- | -- | --
std::vector push_back | 40799 | 332789 | 3474727
Array push back (after PR #1728) | 101703 | 389399 | 3480322
De-virtualize Array | 59025 | 336242 | 3478506
**Speedup this PR** | 1.72x | 1.16x | 1.00x
**Slowdown before** | 2.49x | 1.17x | 1.00x
**Slowdown after** | 1.45x | 1.01x | 1.00x

### `push_back` with pre-reserved memory

  | `int` | `pair<int, int>` | `std::string`
-- | -- | -- | --
std::vector push_back | 35579 | 35804 | 3382912
Array push back (after PR #1728) | 96485 | 108503 | 3316208
De-virtualize Array | 47269 | 50642 | 3269664
**Speedup this PR** | 2.04x | 2.14x | 1.01x
**Slowdown before** | 2.71x | 3.03x | 0.98x
**Slowdown after** | 1.33x | 1.41x | 0.97x

This PR speeds up push-back performance by up to 2x, and the current slowdown in push-back relative to `std::vector` is now down to ~1.5x.

### DIFF
--- a/src/axom/mint/utils/ExternalArray.hpp
+++ b/src/axom/mint/utils/ExternalArray.hpp
@@ -42,8 +42,16 @@ struct ExternalStoragePolicy
    * \brief Reallocates a buffer. No-op for ExternalArray.
    */
   template <typename Func>
-  T* reallocate(T*, int, int, int, Func&&)
+  T* reallocate(T* AXOM_UNUSED_PARAM(old_data),
+                int old_capacity,
+                int AXOM_UNUSED_PARAM(allocator_id),
+                int new_capacity,
+                Func&& AXOM_UNUSED_PARAM(nontrivial_move))
   {
+    if(old_capacity != new_capacity)
+    {
+      SLIC_ERROR("Cannot increase capacity of an ExternalArray.");
+    }
     return nullptr;
   }
 
--- a/src/axom/core/Array.hpp
+++ b/src/axom/core/Array.hpp
@@ -65,7 +65,7 @@ struct DefaultStoragePolicy
    * \brief Reallocates a buffer.
    *
    * \param [in] old_data pointer to the old buffer
-   * \param [in] size the number of elements stored in the array
+   * \param [in] old_capacity the capacity of the currently allocated buffer
    * \param [in] allocator_id the allocator ID to use
    * \param [in] new_capacity the capacity to allocate
    * \param [in] nontrivial_move a callback to move elements that aren't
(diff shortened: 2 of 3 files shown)


src/axom/mint/mesh/ConnectivityArray.hpp
```diff
@@ -709,11 +706,7 @@ class ConnectivityArray
   CellType m_cell_type;
   IndexType m_stride;
   StorageMode m_storageMode;
-  // We keep a unique_ptr to an axom::Array to polymorphically hold:
-  //  * axom::Array if we own the memory
-  //  * sidre::Array if the memory is stored in Sidre
-  //  * mint::utilities::ExternalArray if the memory is externally-owned
-  std::unique_ptr<axom::Array<IndexType, 2>> m_values;
+  detail::ArrayWrapper<IndexType, 2> m_values;
```
kennyweiss, line 709: Nice solution!

Arlie-Capps (approved):
Thank you for the careful work.

src/axom/mint/utils/ExternalArray.hpp
```diff
@@ -17,6 +17,44 @@ namespace axom
...
+namespace detail
+{
+/*!
+ * \class ExternalStoragePolicy
+ *
+ * \brief Externally-managed storage policy.
+ *  Reallocation is not supported.
+ */
+template <typename T>
+struct ExternalStoragePolicy
```
BradWhitlock, line 29: It looks like this could cause some unused parameter warnings. Maybe add `AXOM_UNUSED_PARAM()` in these methods or remove the parameter names.

BradWhitlock (approved):
Good stuff!

src/axom/mint/utils/ExternalArray.hpp
publixsubfan: Done.

PR #1732, opened 2025-12-04
Matrix host-configs and CI

bmhan12:
This PR:

- Adds host-configs for matrix
  - Requires updating caliper version to a commit be compatible with newer CUDA.
- Updates uberenv commit
- (Tentatively/Temporarily) Adds workarounds for tests failing on matrix. Either switching from device to unified memory, or disablement when running with CUDA.

TODO:
- [x] Decide on cadence for matrix job (nightly? weekends? etc.)
- [x] Resolve or work around unit test failures

For now, jobs run every PR commit, but are allowed to fail.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -37,6 +37,12 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   results in polygons for 3D inputs and lines for 2D inputs.
 - Adds a reader for STEP files to `quest/io`. The geometry can either be returned
   as an array of `NURBSPatch` with trimming curves or it can be triangulated into triangle mesh.
+- Adds a `components` variant to Axom's spack package. Users can now provide an explicit list 
+  of desired components, e.g. `components=sidre` and Spack adds any missing dependencies, 
+  e.g. the `slic` component and the `conduit` dependency. By default (when `components` is not provided, 
+  or when `components="all"`), all components are enabled. All components can be disabled via `components=none`.
+- Adds a `conduit` spack variant. Conduit was previously a required dependency in our spack package, and is now enabled by default.
+- Adds spack variants for `adiak` and `caliper`. These replace the previous `profiling` variant which enabled both at the same time.
 
 ###  Changed
 - Treatment of materials on strided-structured Blueprint meshes has changed in `axom::mir`.
--- a/scripts/docker/dockerfile_clang-19
+++ b/scripts/docker/dockerfile_clang-19
@@ -29,7 +29,7 @@ RUN git clone --recursive --branch $branch --single-branch --depth 1 https://git
 # Build/install TPLs via spack and then remove the temporary build directory on success
 RUN cd axom_repo && python3 ./scripts/uberenv/uberenv.py --spack-env-file=./scripts/spack/configs/docker/ubuntu24/spack.yaml \
                                                          --project-json=.uberenv_config.json \
-                                                         --spec="+mfem+raja+umpire+profiling %clang_19" --prefix=/home/axom/axom_tpls -k \
+                                                         --spec="+mfem+raja+umpire+adiak+caliper %clang_19" --prefix=/home/axom/axom_tpls -k \
                  && rm -rf /home/axom/axom_tpls/build_stage /home/axom/axom_tpls/spack
 
 RUN mkdir -p /home/axom/export_hostconfig
--- a/scripts/docker/dockerfile_gcc-12_cuda-12
+++ b/scripts/docker/dockerfile_gcc-12_cuda-12
@@ -36,7 +36,7 @@ RUN git clone --recursive --branch $branch https://github.com/LLNL/axom.git axom
 # Build/install TPLs via spack and then remove the temporary build directory on success
 RUN cd ${HOME}/axom_repo && python3 ./scripts/uberenv/uberenv.py --spack-env-file=./scripts/spack/configs/docker/ubuntu22_cuda/spack.yaml \
                                                                  --project-json=.uberenv_config.json \
-                                                                 --spec="%gcc@12.3.0+mfem+profiling+cuda cuda_arch=70 ^raja+openmp+cuda ^umpire+openmp+cuda" \
+                                                                 --spec="%gcc@12.3.0+mfem+adiak+caliper+cuda cuda_arch=70 ^raja+openmp+cuda ^umpire+openmp+cuda" \
                                                                  --prefix=${HOME}/axom_tpls -k \
                          && rm -rf ${HOME}/axom_tpls/build_stage ${HOME}/axom_tpls/spack
 
--- a/scripts/docker/dockerfile_gcc-13
+++ b/scripts/docker/dockerfile_gcc-13
@@ -29,7 +29,7 @@ RUN git clone --recursive --branch $branch --single-branch --depth 1 https://git
 # Build/install TPLs via spack and then remove the temporary build directory on success
 RUN cd axom_repo && python3 ./scripts/uberenv/uberenv.py --spack-env-file=./scripts/spack/configs/docker/ubuntu24/spack.yaml \
                                                          --project-json=.uberenv_config.json \
-                                                         --spec="+mfem+raja+umpire+profiling %gcc_13" --prefix=/home/axom/axom_tpls -k \
+                                                         --spec="+mfem+raja+umpire+adiak+caliper %gcc_13" --prefix=/home/axom/axom_tpls -k \
                  && rm -rf /home/axom/axom_tpls/build_stage /home/axom/axom_tpls/spack
 
 RUN mkdir -p /home/axom/export_hostconfig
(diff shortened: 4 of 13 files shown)


bmhan12:
For now, I have the matrix job set to run every PR commit, but allowed to fail. 
I will open up an issue tracking the current unit test failures and workarounds.

I've noticed even with the `wbronze` bank set as default for our runner, getting an allocation can still be difficult because the CI partition is overlapped with pbatch: https://lc.llnl.gov/confluence/spaces/GITLAB/pages/785285769/Special+partitions+for+CI+batch+jobs
(e.g. users can request pbatch nodes for the whole day, so no CI jobs will run)

As a byproduct of allowing the matrix job to fail, the matrix job timeout won't prevent PR's from merging.

src/axom/bump/utilities/conduit_memory.hpp
```diff
@@ -97,7 +97,16 @@ class ConduitAllocateThroughAxom
    */
   static void *internal_allocate(size_t items, size_t item_size)
   {
-    const auto axomAllocatorID = axom::execution_space<ExecSpace>::allocatorID();
+    int axomAllocatorID;
+#if defined(AXOM_USE_UMPIRE) && defined(AXOM_USE_GPU)
+    constexpr bool on_device = axom::execution_space<ExecSpace>::onDevice();
+
+    axomAllocatorID = on_device ? axom::getUmpireResourceAllocatorID(umpire::resource::Unified)
```
BradWhitlock, line 104: Is this compensating for code elsewhere that mistakenly attempted to use the wrong type of memory? If so, were lassen/rzansel masking the problem automatically? Or, is this specific fix something unique to matrix? Thanks.

bmhan12: > Is this compensating for code elsewhere that mistakenly attempted to use the wrong type of memory?

Yes, where I have added this I think the code is trying to access device memory on the host. The code segfaults otherwise without these changes.

> If so, were lassen/rzansel masking the problem automatically? Or, is this specific fix something unique to matrix? 

Yes, by default, blueos had the Address Translation Service (ATS) provided by IBM that was automatically handling cases with host/device mismatch: https://hpc.llnl.gov/documentation/tutorials/using-lc-s-sierra-systems

Matrix as far as I know does not have ATS.

Someone else can chime in, but I think something similar to ATS must exist with the AMD hardware, otherwise we should have been observing these problems on tioga/tuolumne.

scripts/spack/packages/mfem/package.py
kennyweiss: @bmhan12 -- can we remove the override of `get_make_config_options` from our copy of mfem's `package.py`?  Here's the relevant change for `^hipblas` in the spack-packages repo

* In our SHA of the spack-packages repo
https://github.com/spack/spack-packages/blob/a75a7f75182ffc7a51c6ca7f0fec4bf9b2705be8/repos/spack_repo/builtin/packages/mfem/package.py#L1038-L1042
* In spack-packages@develop
https://github.com/spack/spack-packages/blob/c7a51222ec6bdb3247fa8a53496bc5874618fbd6/repos/spack_repo/builtin/packages/mfem/package.py#L1048-L1052

scripts/spack/packages/mfem/package.py
```diff
@@ -12,6 +12,11 @@ class Mfem(BuiltinMfem):
         extension="tar.gz",
     )
```
kennyweiss: @bmhan12 -- can we remove this version override from our local `mfem/package.py`?

It's available in the SHA of spack-packages that we're pointing to  (`a75a7f75`):
https://github.com/spack/spack-packages/blob/a75a7f75182ffc7a51c6ca7f0fec4bf9b2705be8/repos/spack_repo/builtin/packages/mfem/package.py#L56-L61

scripts/spack/configs/versions.yaml
```diff
@@ -6,9 +6,10 @@ packages:
   adiak:
     require: 
       - spec: "@0.4.0"
+  # 7/23/25 - Commit after 2.13.1 release, has fix for CUDA 12.9.1
   caliper:
     require:
-      - spec: "@2.12.1"
+      - spec: "@git.527e1ae3a16487e425707e928e05c43d372e144f=develop"
```
kennyweiss: Thanks for the comment.  Do you think we need to regenerate the other host-configs for this change, or is it sufficient to regenerate the cuda-based host-configs for now?

kennyweiss: Looking at the copy of `caliper/package.py` that we're pointing to from spack-packages (and also spack-packages@develop), it doesn't look like there's a `develop` version. 

So, I think we either need to tell spack that this hash is equivalent to the `master` version of caliper, or add a caliper package to axom's local spack packages w/ a 2.13.1 version

If we're going w/ the former, I think we need
```suggestion
  # 7/23/25 - Commit after 2.13.1 release, has fix for CUDA 12.9.1
caliper:
require:
      - spec: "@git.527e1ae3a16487e425707e928e05c43d372e144f=master"
```

See: https://github.com/spack/spack-packages/blob/a75a7f75182ffc7a51c6ca7f0fec4bf9b2705be8/repos/spack_repo/builtin/packages/caliper/package.py#L38

host-configs/matrix-toss_4_x86_64_ib-gcc@13.3.1_cuda.cmake
```diff
@@ -0,0 +1,150 @@
...
+set(CMAKE_BUILD_TYPE "Release" CACHE STRING "")
+
+#------------------------------------------------------------------------------
+# Compilers
+#------------------------------------------------------------------------------
+# Compiler Spec: gcc@13.3.1/hpiynx6mdjb57v77us3h6yyzxymdbr6d
+#------------------------------------------------------------------------------
+if(DEFINED ENV{SPACK_CC})
+
+  set(CMAKE_C_COMPILER "/usr/WS1/axom/libs/toss_4_x86_64_ib/2025_11_25_10_51_58/none-none/compiler-wrapper-1.0-hxzxhboumtniaqs7qivgliluv4h2lbxt/libexec/spack/gcc/gcc" CACHE PATH "")
```
kennyweiss, line 24: Do you know why this is pointing to a `none-none` subdirectory? 
Likely a variable that's not defined...

kennyweiss (approved):
Thanks @bmhan12 

If you'd prefer to merge sooner, please create an issue to track the follow-up fixes we need to do for our cuda build, including the patches you added for umpire allocators and the failing tests.

host-configs/matrix-toss_4_x86_64_ib-gcc@13.3.1_cuda.cmake
bmhan12: > Do you know why this is pointing to a none-none subdirectory?

There's probably a more official spack documentation page on this, but this is the discussion thread I learned about what this is: https://spackpm.slack.com/archives/C01UZDL35C4/p1745272891140599?thread_ts=1743630583.342759&cid=C01UZDL35C4

`none-none` contains packages that don't depend on a compiler (hence `none-none` for "compiler"-"version" as opposed to gcc-13.3.1 for example.)

scripts/spack/configs/versions.yaml
bmhan12: I think just the CUDA host-configs are sufficient for now, I didn't notice any difference building & testing the non-cuda host-configs locally with newer caliper.

scripts/spack/packages/mfem/package.py
bmhan12: mfem recipe updated! It now contains just the line for the hypre change.

bmhan12:
> If you'd prefer to merge sooner, please create an issue to track the follow-up fixes we need to do for our cuda build, including the patches you added for umpire allocators and the failing tests.

I added an issue #1743 to track these unit test failures, and added my thoughts on the errors and current workarounds if they were added.

publixsubfan:
Sorry, that looks like a typo on my end. I think the check should be `space == MemorySpace::Device`.

scripts/spack/packages/mfem/package.py
kennyweiss: Thanks!

src/axom/bump/extraction/TableBasedExtractor.hpp
```diff
@@ -1170,6 +1170,8 @@ class TableBasedExtractor
     axom::for_all<ExecSpace>(
       selectedZonesView.size(),
       AXOM_LAMBDA(axom::IndexType szIndex) {
+        // Avoid first-capture in constexpr-if context error
+        (void)selection;
```
bmhan12, line 1174: Capture these variables outside of "if constexpr".

Documentation about this:
https://docs.nvidia.com/cuda/cuda-programming-guide/05-appendices/cpp-language-support.html#extended-lambda-restrictions
 
The exact rule line in question: "A variable cannot be implicitly captured inside an if-constexpr block that is lexically nested inside an extended lambda unless the variable has been implicitly captured outside the if-constexpr block or appears in the extended lambda’s explicit capture list."

PR #1735, opened 2025-12-06
Polygon clipping support in ClipField and EquiZ MIR.

BradWhitlock:
I tried `EquiZAlgorithm` on a mesh with octagons and found that it could not handle polygons (just tri/quad).

This PR:

* Added a `TopologyMapper` test that intersects 2 meshes with octagons and runs EquiZ on the resulting matset.
* Enhanced `ClipField` and `EquiZAlgorithm` to handle polygons with 5-8 sides.
* Added polygonal clipping tables for polygons with 4-8 sides. The clipped shapes are kept in larger polygons rather than splitting to quad/tri. This makes fewer zone fragments (also looks better and could be a little faster).
* Fixes a glitch in the "originalElements" field when we remove degenerate zone fragments.
* The line count looks large here but it is mostly look up tables.

Here we see VisIt and Axom. VisIt looks wrong for quad shapes here and successive refinement into quad/tri causes more irregular edges _(also not showing all mesh edges it makes)_. Axom keeps zone fragments together as polygons through clipping so fewer fragments and edges are needed.
image

Here is another example of how clipping polygons results in fewer zone fragments:
image


src/axom/bump/clipping/split/convert_clip_cases.py
kennyweiss, line 1: It's not immediately clear to me which files are generated and which ones are input for a generation process.

Could you please update the README file (one directory up) about how to generate/update the files (and anything else that might be relevant)?

src/axom/bump/clipping/split/convert_clip_cases.py
```diff
@@ -0,0 +1,286 @@
...
+    f.write("} // namespace tables\n")
+    f.write("} // namespace clipping\n")
+    f.write("} // namespace bump\n")
+    f.write("} // namespace axom\n")
+    f.close()
+
+def main():
+   #example()
+   outSizes, outOffsets, outShapes = convert_clip_cases("ClipCasesQua.cpp", "Qua", 4)
+   write_new_tables("polygonalClipCasesQua.cpp", "Qua", outSizes, outOffsets, outShapes)
```
kennyweiss, line 271: These file names don't appear to match the checked in ones. Did you rename them after generating?

BradWhitlock: I did rename them.

src/axom/bump/clipping/split/convert_clip_cases.py
BradWhitlock: It might make more sense to do this in the follow-up PR (#1738) since I reorganized the directory and changed the conversion script.

BradWhitlock: I added more information in #1738 that addresses these points.

src/axom/bump/clipping/split/convert_clip_cases.py
BradWhitlock: I changed the script in my other PR so when it writes the files, it uses the target filename instead of a filename that needs to be changed.

Arlie-Capps (approved):
Thanks, @BradWhitlock .  That's a goodly chunk of work.

PR #1737, opened 2025-12-09
ArrayOps C++17 cleanup

publixsubfan:
# Summary

Consolidates code in `ArrayOps` implementation by using C++17 `if constexpr`. This gets rid of the following:

- Custom type traits and tag types used to implement a switchyard for `ArrayOpsBase::init()`, to avoid instantiating a device kernel when types aren't constructible on the device
- Removes the `ArrayOps`-to-`ArrayOpsBase` indirection layer. Both host and device implementation code is now contained solely within the `ArrayOps::[op]()` methods.
- Removes the `MemSpaceTraits` custom type trait.
- Removes the `OperationSpace` enum.

No impact to user code is expected.

### DIFF
--- a/src/axom/core/ArrayBase.hpp
+++ b/src/axom/core/ArrayBase.hpp
@@ -881,6 +881,9 @@ struct DeviceStagingBuffer
         axom::copy(m_staging_buf, m_data + begin, sizeof(T) * nelems);
       }
     }
+#else
+    AXOM_UNUSED_VAR(space);
+    AXOM_UNUSED_VAR(read_from_data);
 #endif
   }
 
@@ -939,6 +942,9 @@ struct ArrayOps
   #else
   using ExecSpace = axom::HIP_EXEC<256>;
   #endif
+#else
+  // CPU-only placeholder for memory space.
+  constexpr static MemorySpace space {MemorySpace::Dynamic};
 #endif
   using StagingBuffer = DeviceStagingBuffer<T>;
 


src/axom/core/ArrayBase.hpp
```diff
@@ -840,316 +840,93 @@ struct all_types_are_integral
...
+                      bool read_from_data = false)
     : m_data(data)
     , m_begin(begin)
     , m_num_elems(nelems)
+#if defined(AXOM_USE_CUDA) && defined(AXOM_USE_UMPIRE)
+    // CUDA device memory is inaccessible from the host. This is the only case
+    // where mirroring data is required.
+    , m_deviceStage(space == MemorySpace::Device)
+#else
+    // In all other contexts, mirroring data is not required. HIP device memory
```
rhornung67, line 869: Is a sync required to make sure data tables for host and device are consistent?

src/axom/core/ArrayBase.hpp
```diff
@@ -840,316 +840,93 @@ struct all_types_are_integral
...
+    , m_deviceStage(false)
+#endif
   {
-    int allocator_id = 0;
-  #ifdef AXOM_USE_UMPIRE
-    allocator_id = axom::detail::getAllocatorID<axom::MemorySpace::Host>();
-  #endif
-    m_staging_buf = axom::allocate<T>(nelems, allocator_id);
-    if(read_from_data)
+#if defined(AXOM_USE_CUDA) && defined(AXOM_USE_UMPIRE)
```
rhornung67, line 874: @LLNL/axom-contrib should we consider reducing the number of macro variables we have in the code. For example, if CUDA (or HIP) is used, Umpire is required, correct? Is there a reason we need one for use CUDA and one for use Umpire?

rhornung67 (approved):
@publixsubfan very nice code cleanup!

src/axom/core/ArrayBase.hpp
publixsubfan: That's a good question: I believe that for host-to-device consistency, memory fencing is used internally before launching a device kernel to ensure that any prior host writes are visible.

The issue that bit us in another project is that it's not always true for data flowing from device-to-host. I think there's two cases to be concerned about here:
- **Asynchronous device kernels**: I think it's fair to push the responsibility of syncing before Array operations onto the caller, if they're using async kernels. We don't use any async kernels in the `ArrayOps` function.
- **Device-to-device `hipMemcpy` treated as async with the host**: This is a more salient concern, as this bit us in another project. But we use Umpire via `axom::copy`, and Umpire was going to push a fix/workaround at some point IIRC?

rhornung67: I agree with your thinking for the first case. I was unaware of the Umpire thing.

src/axom/core/ArrayBase.hpp
rhornung67: Please note that this comment is not meant to be asking to do this for this PR. Simply, pinging everyone to see if such macro variable consolidation makes send and/or would help reduce code complexity.

kennyweiss: Umpire is always required in our GPU configs, so technically, the extra `AXOM_USE_UMPIRE` test is superfluous here. 
On the other hand, we can/do use umpire in non-GPU configs, so we need separate knobs/defines

Our build system has a config variable for each enabled feature (typically of the form `[AXOM_]ENABLE_<FOO>` for features or `<FOO>_FOUND` for dependencies), and each of these corresponds to a separate compiler define in `axom/config.hpp` of the form `AXOM_USE_<FOO>`



See https://github.com/LLNL/axom/blob/df7fef005ffb2c40284ef22d0be789304ab51935/src/axom/config.hpp.in#L51-L77

src/axom/core/ArrayBase.hpp
```diff
@@ -1361,294 +1178,29 @@ struct ArrayOpsBase
...
-      HostOp::realloc_move(array, nelems, values);
-    }
-    else
+#ifdef AXOM_USE_CUDA
+    // CUDA-only: we require non-trivial types to be trivially-relocatable.
+    // This enables us to do simple memcpys for move operations.
+    bool presume_trivially_relocatable = (SPACE == OperationSpace::Device);
+#else
+    bool presume_trivially_relocatable = false;
+#endif
```
kennyweiss, line 1188: Would/Could it help to mark these as `constexpr`?

publixsubfan: I think we're prevented from taking full advantage of `constexpr` here because in the CUDA case, we should be testing against the runtime `space` variable (that above code has a bug that I just noticed). So that `presume_trivially_relocatable` has to be non-constexpr, which means we can't make the if-statement below`constexpr`. At best we can hope that the compiler is smart enough to do the right thing in non-CUDA cases.

kennyweiss: Thanks -- It's hard to tell from github -- I assumed `SPACE` was a template parameter

src/axom/core/ArrayBase.hpp
rhornung67: Yeah, that's what I was thinking, but I was looking through the code and trying to see if there was a way to simplify.... :-(

src/axom/core/ArrayBase.hpp
publixsubfan: I ended up removing the template parameter entirely from `ArrayOps`, so we should only be using the runtime variable `space` now (which is what we want). I did make `presume_trivially_relocatable` a `constexpr` variable in the non-CUDA case to try and help out the compiler a little.

kennyweiss: Thanks -- that makes sense

PR #1739, opened 2025-12-11
Axom package update

adrienbernede:
# Summary

The goal of this PR is to address discussions from #1726 and in particular, step 2 of https://github.com/spack/spack-packages/pull/2276#issuecomment-3613675495.

The changes are divided in three types:
- Sync with spack-packages (as of https://github.com/spack/spack-packages/pull/2558), including:
  - Style changes
  - Updated constraints on dependencies versions
- Simplified handling of rpath leveraging external compilers extra_rpaths


scripts/spack/configs/toss_4_x86_64_ib_cray/spack.yaml
```diff
@@ -41,6 +43,7 @@ spack:
     - spec: '%cray-mpich@8.1.29.rocm_6_3_1'
       when: '%mpi'
     cce_20:
+    - spec: fflags=-ef
```
white238, line 48: What is the benefit of having these in the toolchain vs the compiler spec?

scripts/spack/packages/axom/package.py
```diff
@@ -369,25 +376,20 @@ def initconfig_hardware_entries(self):
 
             hip_link_flags = ""
 
-            rocm_root = os.path.dirname(spec["llvm-amdgpu"].prefix)
+            rocm_root = spec["llvm-amdgpu"].prefix
```
white238: I thought this required a removal of the `llvm` directory in the prefix of `llvm-amdgpu`...

scripts/spack/configs/toss_4_x86_64_ib_cray/spack.yaml
adrienbernede: The rationale is not entirely clear to me, but I trust @bmhan12 with it:
https://github.com/LLNL/axom/discussions/1726#discussioncomment-15134170

Now, personally, I see those configs as layers, from generic to specialized: 
The compiler spec should be as generic as possible.
If we need to tune a compiler, the toolchain allows us do keep the compilers generic.
If a flag is only needed by a package, not the full stack, then it should be set in the package.

In this case, if `Mfreeform` and `ef` are Axom specific, I would set them in Axom package.
On the other hand, the extra-rpath appeared needed by any package built with those compilers, so we can define them in the compiler without loosing generality.

scripts/spack/packages/axom/package.py
adrienbernede: Yes... that's an oversight. It still works because the CI is using a version of Spack where the CachedCMakePackage still defines ROCM_PATH as `os.path.dirname(spec["llvm-amdgpu"].prefix)`.

Let me do a little experiment to demonstrate that:

1. In the next commit, I’ll update axom reference to spack-packages so that it has the change in CachedCMakePackage (since https://github.com/spack/spack-packages/commit/a229f5467023b3408001243fa5b0401643f6f148)
That will cause the CI to fail on Tioga and Tuolumne.

2. Then I’ll update Axom package to not set ROCM_ROOT_DIR, and update the llvm-amdgpu prefix to the correct path to fix the failure.

adrienbernede: hmm... this won't work because Axom's CI is using the predefined host-configs.

adrienbernede: So I’ll just push the fix in the spack configs along with the removal of ROCM_ROOT_DIR.
How should I test those changes? I'm thinking you might want to trigger host-config generation when there are changes in the local spack configs and packages.

white238: we generate and test our spack recipes manually then check in the generated host-config.

bmhan12:
@adrienbernede 
I'll close this PR, the dependency changes have been added to axom recipe and upstream `spack-packages`. 
The `extra_rpaths` change unfortunately did not work in my testing.

PR #1744, opened 2025-12-15
Generalizes `primal::evaluate_integral` methods for componentwise integration

jcs15c:
# Summary

This PR is an enhancement, adding an additional `primal::evaluate_line_integral` method to `evaluate_integral.hpp` which permits integral evaluation over any lambda function which has an "integrable" return type, i.e. one for which addition and scalar multiplication are defined. Most prominently, this permits evaluation over integrands which return `primal::Vector` objects. The same functionality is added to `primal::evaluate_area_integral`. These methods are supported by additional tests in `primal_integral.cpp`.

More generally, this addition is meant to simplify the process of evaluating physical quantities for geometric objects. For example, each component of the average surface normal of a 3D NURBS surface is computed by integrating each component of the surface normal vector over the surface. Evaluating this as three separate integrals results in a significant amount of redundant computation which can now be avoided.

### A note on names
Implementing this feature necessarily introduces some amount of naming ambiguity between methods in `evaluate_integral.hpp`, as now there are two types of "vector field integrals" with distinct meanings. Currently defined in `develop` are methods `evaluate_vector_line_integral`, which evaluates integrals of the form 

$$\int_C \vec{F}(\vec{x}) \cdot d\vec{r},$$

which assume that $\vec{F}$ : $\mathbb{R}^n \to \mathbb{R}^n$. Typically $n$ is equal to 2 or 3, but technically other values are possible mathematically. Either way, this method always returns a scalar value, i.e. a double, as would be suggested by the dot product in the formulation.

The proposed changes introduces methods of evaluating integrals of the form 

$$\int_C \vec{F}(\vec{x}) dr,$$

which assumes that $\vec{F}$ : $\mathbb{R}^n \to \mathbb{R}^m$, with an output in $\mathbb{R}^m$. While $n$ will most often be 2 or 3, $m$ can be significantly larger. Specifically, I'm targeting a future use with $m = 39$, corresponding to a function with a return type `primal::Vector<double, 39>`. These could also be characterized as "vector field integrals," even though they behave more like "component-wise scalar field integrals." This ambiguity is essentially unavoidable, because even the input for "scalar field integrals" ($\mathbb{R}^n$) are themselves a "vector space over a scalar field" in a mathematical sense. 

To make sure any ambiguity is inherently resolved by templates, the functionality of `evaluate_scalar_line_integral` methods in `evaluate_integral.hpp` is merged into an `evaluate_line_integral` method which performs this more general component-wise integration. Conveniently, this mirrors the `evaluate_area_integral` method.


src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -3121,16 +3121,13 @@ class NURBSPatch
...
+      };
 
-        // Find the area of the resulting projection
-        ret_vec[N] +=
-          evaluate_area_integral(nPatch.getTrimmingCurves(), avg_surface_normal_integrand, npts);
-      }
+      // Find the area of the resulting projection
+      ret_vec +=
+        evaluate_area_integral(nPatch.getTrimmingCurves(), avg_surface_normal_integrand, npts);
     }
```
kennyweiss: Nice cleanup!

Would something like this work?

```cpp
VectorType ret_vec;
for(const auto& nPatch : extractTrimmedBezier())
{
    ret_vec += evaluate_area_integral(
                nPatch.getTrimmingCurves(),
                [&nPatch](Point2D x) -> Vector<T,3>{ return nPatch.normal(x[0], x[1]); },
                npts);
}
return ret_vec;
```

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -18,101 +18,137 @@
...
 namespace axom
 {
 namespace primal
 {
 namespace detail
 {
+namespace internal
+{
+///@{
+/// \name Boilerplate to support integrals of functions with general return types,
```
kennyweiss:
```suggestion
/// \name Type traits to support integrals of functions with general return types,
```

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -18,101 +18,137 @@
...
+  RetType total_integral = RetType {};
   for(int i = 0; i < beziers.size(); ++i)
   {
     total_integral +=
-      detail::evaluate_scalar_line_integral_component(beziers[i],
-                                                      std::forward<Lambda>(scalar_integrand),
-                                                      npts);
+      detail::evaluate_line_integral_component(beziers[i], std::forward<Lambda>(integrand), npts);
   }
   return total_integral;
```
kennyweiss: Similar suggestion to another above about using range-for instead of explicit loop

Would the following work?
```cpp
  RetType total_integral = RetType {};
  for(const auto& bez : n.extractBeziers())
  {
    total_integral += detail::evaluate_line_integral_component(bez, std::forward<Lambda>(integrand), npts);
  }
  return total_integral;
```

Feel free to ignore, but it might be more efficient since the Bezier curves are shorter lived.
Then again, it might be a wash.

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -18,101 +18,137 @@
...
 }
 
 /*!
- * \brief Evaluate the scalar integral on a single NURBS curve with cached data for GWN evaluation.
+ * \brief Evaluate a integral on a single NURBS curve with cached data for GWN evaluation.
```
kennyweiss:
```suggestion
 * \brief Evaluate an integral on a single NURBS curve with cached data for GWN evaluation.
```

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
kennyweiss, line 1: @jcs15c -- since there are several options w/ somewhat confusing names/distinctions, here and in the parent directory (`primal/operators/evaluate_integral.hpp`) it would likely be helpful to have a file-level doxygen comment (or a group doxygen comment) that briefly discusses the various options

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -199,6 +207,107 @@ double evaluate_vector_line_integral(const axom::Array<CurveType>& carray,
   return total_integral;
 }
 
+/*!
+ * \brief Evaluate a line integral along the boundary of a CurvedPolygon object 
+ *        for a function with an arbitrary return type.
+ *
+ * The line integral is evaluated on each curve in the CurvedPolygon, and added
+ * together to represent the total integral. The Polygon need not be connected.
```
kennyweiss: Suggestion:  The capitalization of "Polygon" threw me off since we have a Polygon class. I'd suggest using lower case (i.e. "Polygon" -> "polygon") and possibly adding the word curved (i.e. "The curved polygon").

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -209,27 +318,37 @@ double evaluate_vector_line_integral(const axom::Array<CurveType>& carray,
  * For algorithm details, see "Spectral Mesh-Free Quadrature for Planar 
  * Regions Bounded by Rational Parametric Curves" by David Gunderman et al.
  * 
- * \tparam Lambda A callable type taking an axom::Point type returning a numeric type
+ * \tparam Lambda A callable type taking an CurveType's PointType and returning an integrable type
```
kennyweiss:
```suggestion
 * \tparam Lambda A callable type taking a CurveType's PointType and returning an integrable type
```

src/axom/primal/tests/primal_integral.cpp
```diff
@@ -91,6 +91,71 @@ TEST(primal_integral, evaluate_area_integral)
...
+  using ReturnType = primal::Vector<double, 3>;
+  double abs_tol = 1e-10;
+
+  // Quadrature nodes. Should be sufficiently high to pass tests
+  int npts = 20;
+
+  // Define anonymous function for testing
+  auto aggregate_integrand = [](Point2D x) -> ReturnType {
+    return ReturnType {1.0, x[0] * x[1] * x[1], std::sin(x[0] * x[1])};
+  };
```
kennyweiss: Please add a brief comment about this lambda. Is this essentially defining three integrands (one per component in the return type) ? --  constant, $x \cdot y^2$ and $sin(x\cdot y)$

kennyweiss (approved):
Thanks @jcs15c -- nice contribution!

src/axom/primal/geometry/NURBSPatch.hpp
jcs15c: It absolutely would! Reminds me I should probably do the same change in the new step reader stuff too. 

EDIT: I realize now the main thrust of the suggestion was the range-based for-loop, not the fully anonymous lambda function. Fixed both here, and the latter in the step reader.

src/axom/primal/operators/evaluate_integral.hpp
jcs15c: Fixed here and in 10 other places. Whoops!

src/axom/primal/tests/primal_integral.cpp
jcs15c: Added more detail here and in the other aggregate example

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
jcs15c: I added file-level comments, and some groups that hopefully make things more clear. It's definitely worth clarifying the difference between a "scalar-field integral with a function of vector return type" and a "vector-field integral, which always has vector return type"

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
jcs15c: Makes perfect sense!

src/axom/primal/operators/evaluate_integral.hpp
```diff
@@ -295,24 +445,22 @@ double evaluate_area_integral(const axom::Array<CurveType>& carray,
...
-    for(int j = 0; j < beziers.size(); j++)
+    for(const auto& bez : carray[i].extractBezier())
     {
-      total_integral +=
-        detail::evaluate_area_integral_component(beziers[j],
-                                                 std::forward<Lambda>(scalar_integrand),
-                                                 int_lb,
-                                                 npts_Q,
-                                                 npts_P);
+      total_integral += detail::evaluate_area_integral_component(beziers[j],
```
kennyweiss: Looks like this needs to be updated:

```suggestion
      total_integral += detail::evaluate_area_integral_component(bez,
```

kennyweiss: Thanks for the other changes!

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -3111,26 +3111,16 @@ class NURBSPatch
...
-          evaluate_area_integral(nPatch.getTrimmingCurves(), avg_surface_normal_integrand, npts);
-      }
-    }
+    // Split the patch along the unique knot values to improve convergence
+    for(const auto& nPatch : extractTrimmedBezier())
+    {
+      // Integrate the surface normal over the patches
+      ret_vec += evaluate_area_integral(
+        nPatch.getTrimmingCurves(),
+        [&nPatch](Point2D x) -> Vector<T, 3> { return nPatch.normal(x[0], x[1]); } npts);
```
BradWhitlock: Where did the comma in front of "npts" go? Isn't it still needed?

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -3111,26 +3111,16 @@ class NURBSPatch
...
+    };
```
BradWhitlock: Extra semicolon. Delete it.

jcs15c: This whole function call is a mess, apparently. Sorry about that, I'll fix it and double check the rest.

src/axom/primal/tests/primal_integral.cpp
```diff
@@ -309,12 +431,8 @@ TEST(primal_integral, evaluate_integral_rational)
   EXPECT_NEAR(evaluate_area_integral(quarter_ellipse, transc_integrand, npts), 0.472951736306, abs_tol);
 
   // Test line integral on scalar domain againt values computed with external software
-  EXPECT_NEAR(evaluate_scalar_line_integral(ellipse_arc, const_integrand, npts),
-              2.42211205514,
-              abs_tol);
-  EXPECT_NEAR(evaluate_scalar_line_integral(ellipse_arc, transc_integrand, npts),
-              1.38837959326,
-              abs_tol);
+  EXPECT_NEAR(evaluate_line_integral(ellipse_arc, const_integrand, npts), 2.42211205514, abs_tol);
```
kennyweiss, line 434: Please update the RELEASE-NOTES w/ the changed function name

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -18,126 +36,171 @@
...
  * \return the value of the integral
  */
-template <typename Lambda, typename T, int NDIMS>
-inline double evaluate_scalar_line_integral_component(const primal::BezierCurve<T, NDIMS>& c,
-                                                      Lambda&& scalar_integrand,
-                                                      const int npts)
+template <typename Lambda,
+          typename T,
+          int NDIMS,
+          typename LambdaRetTyp = std::invoke_result_t<Lambda, typename BezierCurve<T, NDIMS>::PointType>>
```
Arlie-Capps: Throughout this file, sometimes I read `LambdaRetType` (as in (new) line 87 in the Doxygen), sometimes `LambdaRetTyp` (as here), and sometimes `LambdaRetTYpe` (as at (new) line 300).  Please make them all consistent.  If they have to be disambiguated, please make it a larger string or conceptual distance.

jcs15c: Entirely typos! As neat as it is that they were consistent enough for things to still compile, thank you for catching these. They're all fixed now.

PR #1736, opened 2025-12-08
Partial change from larger clipping performance work

gunney1:
# Summary
- This PR is a feature and a refactoring.  It's part of the planned follow-ups to the recently merged PR https://github.com/LLNL/axom/pull/1667
- It does the following:
  - Implements screening in `TetClipper` to reduce use of expensive clipping method.
  - Adds `PlaneClipper`, for clipping the 3D plane geometry.  This is now the second geometry supported in the new clipping code.
  - Factors out repetitive parts of `MeshClipperImpl`, addressing issue https://github.com/LLNL/axom/issues/1704
  - Switches to decomposing a hex into 18 tets instead of 24.  Switching back and forth is currently supported by a compile-time parameter in `ShapeMesh`, for evaluating performance as we test against diverse sets of configuratios.
  - Adds a flexible way to collect and log statistics, for evaluating performance.  We can now add statistics without changing the core code.
  - Add a developer parameter to control how much screening to use.  In general, more screening is better, but there are overhead costs to screening.

Not included in this PR: shapes for the sphere, SOR (including cone and cylinder), hex, tet mesh.  Further optimizations.

# Performance
These are the number of clips used and the clipping time for the 2 shapes in this PR (tet and plane).  They came from the same domain, with 3 mesh resolutions: 50^3, 100^3 and 200^3, respectively M50, M100 and M200.  There are three levels of screening: L0 is no screening.  L1 is screening on the hex cells.  L2 is screening on the 18 individual tets in each hex.  Values in the L1 and L2 columns are normalized by the L0 column for the same mesh, so they are ratios.  Smaller mean more effective screening.

<div class="outline-text-3" id="text-1-8">

quantity | M50_L0 | M50_L1 | M50_L2 | M100_L0 | M100_L1 | M100_L2 | M200_L0 | M200_L1 | M200_L2
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
tetTime | 0.15326 | 0.227 | 0.221 | 1.09257 | 0.137 | 0.14 | 7.986933 | 0.085 | 0.086
planeTime | 0.379523 | 0.102 | 0.052 | 2.933372 | 0.054 | 0.03 | 22.81227 | 0.03 | 0.017
tetClips | 975240.0 | 0.074 | 0.023 | 7711200.0 | 0.037 | 0.011 | 59687712.0 | 0.019 | 0.006
[...]

PR #1746, opened 2025-12-18
Allow overriding default allocator in deepCopyToConduit

gunney1:
# Summary

- This PR is a feature
- It does the following:
  - Allows user to override the default allocator id when deep-copying a `sidre::View` to a `conduit::Node`.
  - Add a test for copying with allocator override.

This PR takes a bit out of https://github.com/llnl/axom/pull/1736 to make that one smaller, as requested by the code review.


src/axom/sidre/tests/sidre_view.cpp
```diff
@@ -2018,6 +2008,58 @@ TEST(sidre_view, deep_copy_to_conduit)
...
+
+    // dst2 sets and uses allocator allocIds[si]
+    conduit::Node dst2;
+    dst2.set_allocator(ConduitMemory::axomAllocIdToConduit(allocIds[si]));
+    srcView->deepCopyToConduit(dst2);
+    EXPECT_EQ(ConduitMemory::conduitAllocIdToAxom(dst2.allocator()), allocIds[si]);
+
+    // dst3 sets allocator allocIds[si] but overrides it in the copy
+    conduit::Node dst3;
+    dst3.set_allocator(ConduitMemory::axomAllocIdToConduit(allocIds[si]));
```
BradWhitlock, line 2053: Do you need this line? It looks like passing the allocator ID to `deepCopyToConduit` will cause the allocator to get set in that method.

BradWhitlock (approved):
Looks fine to me. Thanks.

src/axom/sidre/tests/sidre_view.cpp
gunney1: Yes, we want that line to set the allocator so we can check that it's overridden by the copy.  In practice, we would skip the explicit call to `set_allocator`, but for testing, we want to see it change.

BradWhitlock (approved):

src/axom/sidre/tests/sidre_view.cpp
```diff
@@ -2018,6 +2008,58 @@ TEST(sidre_view, deep_copy_to_conduit)
...
+    srcView->deepCopyToConduit(dst3, allocIds[0]);
+    EXPECT_EQ(ConduitMemory::conduitAllocIdToAxom(dst3.allocator()), allocIds[0]);
+  }
+
+  ds.getRoot()->destroyGroup("src");
+}
```
kennyweiss, line 2059: This new test is checking that we can successfully call `deepCopyToConduit` in several ways, but it does not verify that the data was actually copied. Is there a way we can check the data?

One idea might be to copy using the data using the specified allocator and then to copy it back to a new buffer using the host allocator and then to check that the data is the same.
Another might be to to do some sort of checksum on the data, and then check that the checksums are the same after copying.

gunney1: We have other tests that verify the copied data.

PR #1748, opened 2025-12-20
Changes supporting GPU shaping branch

gunney1:
# Summary

- This PR contains misc small changes separated from the clipping performance work
- It does the following:
  - Adds a sign-of utility function
  - Add sphere containment functionality to `Sphere`
  - Add a `CoordinateTransformer` constructor to build a transformer using terminus points.
  - Fix an error in setting `klee::Geometry` from a `primal::Cone`


src/axom/primal/tests/primal_sphere.cpp
```diff
@@ -150,6 +150,34 @@ void check_sphere_intersection()
...
+  PointType center {3.0, 4.0, 0.0};
+  double radius = 5.0;
+
+  SphereType S0(center, radius);
+  SphereType S1(PointType {1.5, 2.0, 0.0}, radius / 2);
+  SphereType S2(S1.getCenter(), radius);
+
+  double tol = 1e-12;
+
+  // STEP 0: test fully containing
```
BradWhitlock, line 169: Can we see S0 contain S0 with tol=0. Looking at `contains()`, I guess a sphere does not contain itself. That is counterintuitive.

src/axom/primal/geometry/CoordinateTransformer.hpp
```diff
@@ -66,6 +66,16 @@ class CoordinateTransformer
    */
   CoordinateTransformer(const numerics::Matrix<T>& matrix) { setMatrix(matrix); }
 
+  /*!
+   * @brief Construct transformer that moves 4 starting points to 4
+   * destination points.
+   */
+  AXOM_HOST_DEVICE CoordinateTransformer(const primal::Point<T, 3>* startPts,
```
BradWhitlock, line 74: A little ASCII art diagram could be useful. I assume it's something like this:

```
      *[2]
      |  
      |
      *[0]-----*[1]
    /
  /
*[3]

```

src/axom/core/utilities/Utilities.hpp
```diff
@@ -348,6 +348,16 @@ inline AXOM_HOST_DEVICE bool isNearlyEqualRelative(RealType a,
   // return abs(a-b) <= max(absThresh, relThresh * maxFabs );
 }
 
+/*!
+ * \brief Sign of a value of any type that supports comparison and
```
rhornung67, line 352: Should there be a check for eps < 0? For example, `sign_of(-0.1, -0.01)` returns -1, but `sign_of(-0.01, -0.1)` returns 1.

Maybe only a comment is needed?

src/axom/primal/geometry/Sphere.hpp
```diff
@@ -178,6 +178,18 @@ class Sphere
   AXOM_HOST_DEVICE
   inline bool intersectsWith(const Sphere<T, NDIMS>& sphere, double TOL = 1.e-9) const;
 
+  /*!
+   * \brief Tests if this sphere completely contains another sphere.
+   *
+   * \param [in] other The sphere object to check for containment
+   * \param [in] margin Amount that this sphere must contain the other sphere by.
+   *   Positive means that the other sphere is "more inside".
```
rhornung67:
```suggestion
   *   Positive means that the other sphere is "more inside".
```

Not sure what you mean by this. It would help to explain, maybe with a simple example.

src/axom/core/utilities/Utilities.hpp
gunney1: Yes.  Good idea!

src/axom/primal/geometry/CoordinateTransformer.hpp
gunney1: Sorry, I'm not sure I understand.  The 4 starting and ending points can be anywhere.  This constructor simply calls `setByTerminus`, which explains a bit about the input requirements.  I'll add that info to the constructor docs.  Let me know if you think it needs more.

BradWhitlock: I guess I don't understand what the 4 points represent then. That's what I would want to know.

src/axom/primal/geometry/Sphere.hpp
gunney1: I was going for an intuitive explanation but obviously that didn't work.  Maybe a mathematical explanation would be better?  Like: "Margin is added to other's radius before comparing."

src/axom/primal/tests/primal_sphere.cpp
gunney1: Good point!  I'll replace `>` with `>=`.

src/axom/core/tests/core_utilities.hpp
```diff
@@ -207,3 +207,24 @@ TEST(core_utilities, sort_multiple_values)
     }
   }
 }
+
+TEST(core_utilties, sign_of)
```
Arlie-Capps: typo for "utilities"

src/axom/primal/geometry/Sphere.hpp
rhornung67: I like that statement better as it is more clear.

src/axom/primal/geometry/CoordinateTransformer.hpp
Arlie-Capps: @gunney1 , correct me if I'm wrong, but I think you use this ctor to transform an arbitrary tet to a unit tet like the one @BradWhitlock diagrammed.  Perhaps you could make ASCII art showing the tet before transformation and use Brad's as the tet after transformation.

gunney1: It can transform an arbitrary tet to a unit tet but only if you specify a unit tet as the second 4 points.  We can transform any tet to any other tet.

gunney1: @BradWhitlock you can think of the 4 points as providing the required information to define the transformation.  There are 12 unknowns in the matrix so you need 12 constraints, which are the 4 3D points.  I guess if it helps to think of the points as tet vertices, I could write that...

src/axom/primal/geometry/Sphere.hpp
gunney1: OK.  Thanks and sorry for the confusion.

src/axom/core/tests/core_utilities.hpp
gunney1: Thanks.

src/axom/primal/geometry/CoordinateTransformer.hpp
gunney1: Maybe something like "If you think of each 4 points as the vertices of a tetrahedron, the transformation moves the starting tet to the destination tet."

src/axom/primal/geometry/Cone.hpp
```diff
@@ -143,7 +143,8 @@ class Cone
   }
 
   /*!
-   * \brief Returns the algebraic volume of the cone
+   * \brief Returns the algebraic volume of the cone,
```
kennyweiss: (I might have seen a similar comment elsewhere) 
What does `algebraic volume` imply vs. `signed volume`?
I'd suggest `signed volume` instead of `algebraic`.

kennyweiss (approved):

src/axom/primal/geometry/Cone.hpp
gunney1: Done.  thanks.

PR #1749, opened 2025-12-22
Improves test throughput, especially for threaded tests

kennyweiss:
# Summary

This PR focuses on our test throughput, especially in configurations with OpenMP enabled.
- The threaded tests were not all using the `OMP_NUM_THREADS` environment variable, and were blocking all other tests leading to excessive test times. ~in excess of 300 seconds on a beefy node. After this PR, it's down to about 80 seconds~
    - **Update:** It seems that I overstate the benefits of adding the PROCESSORS tag. It might even slow our tests down.
       Nevertheless, this branch does improve our test times by about 20%
    - It seems that I inadvertently disabled `C2C` and `MFEM` for my initial testing, which removed a bunch of the more expensive tests.
- It also improves the consistency of how we're setting up the number of MPI ranks and OpenMP threads, via the `NUM_MPI_TASKS` and `NUM_OMP_THREADS` parameters to `axom_add_test`, which forwards them to `blt_add_test`
- It relies on a concurrent PR to blt which properly marks unit test with the number of requested processors using the CMake `PROCESSORS` test property.  https://github.com/llnl/blt/pull/749
   - **Update:** Since the PROCESSORS property did not pan out (at least in its current state), I removed it from this PR.
- Introduces `AXOM_TEST_NUM_OMP_THREADS` config variable to provide a single, configurable default thread count for OpenMP tests.
- Converts multiple `blt_add_test()` call sites to `axom_add_test()` and removes ad-hoc per-test `PROCESSORS`/`OMP_NUM_THREADS` handling in favor of BLT’s centralized behavior.
- Reduces the resolution of the quest_shaping_in_memory tests from `30^3` to `10^3` since the former were taking over 100 seconds each in debug configs when using the ctest `PROCESSORS` property.

## Test throughput improvements
Comparing test runtimes on an exclusively allocated toss4 node on LC using `ctest -j100`

| time        | branch        | uses PROCESSORS property | uses RESOURCE_LOCK |
|-------------|---------------|--------------------------|---------------------|
| 349 seconds | axom@develop  | ✗                        | ✗                   |
| 364 seconds | axom@develop  | ✗                        | ✓                   |
| 356 seconds | axom@develop  | ✓                        | ✗                   |
| 364 seconds | axom@develop  | ✓                        | ✓                   |
| 296 seconds | this branch   | ✗                        | ✗                   |
| 295 seconds | this branch   | ✗                        | ✓                   |
| 305 seconds | this branch   | ✓                        | ✗                   |
| 314 seconds | this branch   | ✓                        | ✓                   |
|---|-----|----|----|
| **250 seconds** | **this branch (w/ lower heroic_roses res)**   | ✗                        | ✗                   |

**Notes:** 
* After running some experiments, it looks like adding the `PROCESSORS` test property is actually slowing us down by about 3%, and we should consider other approaches to increasing our test throughput, e.g. CMake's resource spec and resource groups -- https://cmake.org/cmake/help/latest/prop_test/RESOURCE_GROUPS.html
* `RESOURCE_LOCK` serializes invocation of the MPI tests to better see how long each test takes.
```cmake
  # Optionally, add RESOURCE_LOCK property to mpi tests for timing purposes
   if(arg_NUM_MPI_TASKS AND BLT_ENABLE_MPI)
       set_property(TEST ${arg_NAME}
                    APPEND PROPERTY RESOURCE_LOCK lock_name)
   endif()

```

### DIFF
--- a/src/axom/quest/examples/CMakeLists.txt
+++ b/src/axom/quest/examples/CMakeLists.txt
@@ -282,10 +282,10 @@ if(AXOM_ENABLE_MPI AND MFEM_FOUND AND MFEM_USE_MPI
             COMMAND  quest_shaping_driver_ex 
                      -i ${shaping_data_dir}/heroic_roses_mfem.yaml
                      --method sampling
-                     inline_mesh --min 0 0 --max 300 400 --res 300 400 -d 2
+                     inline_mesh --min 0 0 --max 300 400 --res 150 200 -d 2
             NUM_MPI_TASKS ${_nranks})
         set_tests_properties(${_testname} PROPERTIES
-            PASS_REGULAR_EXPRESSION  "Volume of material 'black' is 36,?637.")
+            PASS_REGULAR_EXPRESSION  "Volume of material 'black' is 36,?595.")
 
         # Test 2D MFEM mesh (winding number)
         set(_testname quest_shaping_driver_ex_heroic_roses_mfem_cp)
@@ -294,10 +294,10 @@ if(AXOM_ENABLE_MPI AND MFEM_FOUND AND MFEM_USE_MPI
             COMMAND  quest_shaping_driver_ex 
                      -i ${shaping_data_dir}/heroic_roses_mfem_cp.yaml
                      --method sampling --sampling windingnumber
-                     inline_mesh --min 0 0 --max 300 400 --res 300 400 -d 2
+                     inline_mesh --min 0 0 --max 300 400 --res 150 200 -d 2
             NUM_MPI_TASKS ${_nranks})
         set_tests_properties(${_testname} PROPERTIES
-            PASS_REGULAR_EXPRESSION  "Volume of material 'black' is 36,?645.")
+            PASS_REGULAR_EXPRESSION  "Volume of material 'black' is 36,?633.")
 
         # 3D shaping tests
         set(_nranks 3)


src/axom/core/examples/CMakeLists.txt
```diff
@@ -73,26 +73,36 @@ if(AXOM_ENABLE_TESTS)
...
+        if(_pol STREQUAL "omp")
+            set(_num_threads ${AXOM_TEST_NUM_OMP_THREADS})
+        endif()
+
         axom_add_test(
-            NAME    "core_array_perf_1d_${_pol}"
-            COMMAND core_array_perf_ex --policy ${_pol} --shape 1200000 -r 10)
+            NAME            core_array_perf_1d_${_pol}
+            COMMAND         core_array_perf_ex --policy ${_pol} --shape 1200000 -r 10
+            NUM_OMP_THREADS ${_num_threads})
```
kennyweiss, line 90: Note:  This PR relies on allowing empty arguments to `NUM_OMP_THREADS`, in which case `if(arg_NUM_OMP_THREADS)` will evaluate to `false`

src/axom/core/tests/CMakeLists.txt
```diff
@@ -183,14 +183,16 @@ axom_add_executable(NAME       utils_annotations_test
...
+  set(_num_ranks 1)
   if(AXOM_ENABLE_MPI)
-    axom_add_test(NAME    ${_test_name}
-                  COMMAND utils_annotations_test -m ${_mode}
-                  NUM_MPI_TASKS 2)
-  else()
-    axom_add_test(NAME    ${_test_name}
-                  COMMAND utils_annotations_test -m ${_mode})
+    set(_num_ranks 2)
   endif()
```
kennyweiss, line 193: Note: A recent BLT PR allows for `NUM_MPI_TASKS` to be `1` in non-mpi configs. 
This allows us to simplify some test setup logic.

src/axom/quest/examples/CMakeLists.txt
```diff
@@ -381,8 +396,9 @@ if((CONDUIT_FOUND OR
                            --scale 0.75 0.75 0.75
                            --dir 0.2 0.4 0.8
                            --meshType bpSidre
-                           inline_mesh --min -2 -2 -2 --max 2 2 2 --res 30 30 30
-                NUM_MPI_TASKS ${_nranks})
+                           inline_mesh --min -2 -2 -2 --max 2 2 2 --res 10 10 10
```
kennyweiss, line 399: @gunney1 -- in a debug config, these tests were each taking ~100 seconds at a resolution of $30^3$. Changing it to $10^3$ reduces the runtime significantly to around 1-10 seconds per test.

src/cmake/AxomOptions.cmake
```diff
@@ -47,6 +47,16 @@ option(AXOM_ENABLE_TOOLS "Enables Axom Tools" ON)
...
 
+#------------------------------------------------------------------------------
+# Test execution controls
+#------------------------------------------------------------------------------
+if(AXOM_ENABLE_OPENMP)
+    set(AXOM_TEST_NUM_OMP_THREADS 4 CACHE STRING "Default number of OpenMP threads for tests")
+else()
+    set(AXOM_TEST_NUM_OMP_THREADS 0 CACHE STRING "Default number of OpenMP threads for tests")
+endif()
+mark_as_advanced(AXOM_TEST_NUM_OMP_THREADS)
```
kennyweiss, line 58: Note: I added a new config variable to control the default number of threads per test. Users/Devs can mostly ignore it.

src/axom/quest/tests/CMakeLists.txt
```diff
@@ -191,38 +185,48 @@ if(ENABLE_MPI AND MFEM_FOUND AND MFEM_USE_MPI
     endforeach()
 
     # Invoke the intersection shaper test program as many tests that run a subset of the tests.
-    set(_numOMPThreads 8) # limit omp parallelism
-    set(_policies seq)
+    set(_policies "seq")
     if(RAJA_FOUND)
```
BradWhitlock, line 189: Longer term, what about putting this conditional logic in a function that returns the supported policies? It must occur in multiple places.

kennyweiss: Good idea -- I was thinking about something like that. 
Maybe in a follow up PR so it's easier to distinguish from this set of changes.

kennyweiss: Update: https://github.com/llnl/axom/issues/1750

RELEASE-NOTES.md
```diff
@@ -43,6 +43,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   or when `components="all"`), all components are enabled. All components can be disabled via `components=none`.
 - Adds a `conduit` spack variant. Conduit was previously a required dependency in our spack package, and is now enabled by default.
 - Adds spack variants for `adiak` and `caliper`. These replace the previous `profiling` variant which enabled both at the same time.
+- Adds the `AXOM_TEST_NUM_OMP_THREADS` configuration variable to control the default OpenMP thread count for tests.
```
rhornung67, line 46:
```suggestion
- Adds the `AXOM_TEST_OMP_NUM_THREADS` configuration variable to control the default OpenMP thread count for tests.
```

Minor nit. I think it's better to be consistent with the `OMP_NUM_THREADS` env variable or OpenMP library function with a similar name. Here and throughout.

Also, the terms `ranks` and `tasks` are used interchangeably as they relate to MPI. Should that be made consistent also?

kennyweiss: +1 for consistency, but it might be tricky here.

The keywords/parameters for `blt_add_test` are `NUM_MPI_TASKS` and `NUM_OMP_THREADS`, while the ENV variable for OpenMP is `OMP_NUM_THREADS`.  It's not clear that renaming the blt parameters would be appropriate or an improvement, and would affect all blt users.

What do you suggest?

Arlie-Capps: Aargh--- naming is important and consistency is a desirable quality.  I would say, let's match `blt_add_test`, and in any discussion let's point out the discrepancy to OpenMP environment variables.

src/cmake/AxomMacros.cmake
```diff
@@ -248,16 +248,26 @@ macro(axom_add_test)
                  NUM_OMP_THREADS ${arg_NUM_OMP_THREADS}
```
Arlie-Capps, line 248: I recommend that we don't change the names of the args to `axom_add_test`, and that in the documentation comment for this function, that you please add something like "The OpenMP environment variable is OMP_NUM_THREADS, but BLT uses NUM_MPI_TASKS and NUM_OMP_THREADS.  We are electing to match BLT."

RELEASE-NOTES.md
rhornung67: Probably best to discuss with Chris and others. I would defer to OpenMP which is a common, well-known standard, whereas BLT it not.

kennyweiss: I think it would be confusing to switch it. This is an existing macro that's been in the code for a long time and is entirely Axom developer focused. 

We're calling `axom_add_test()` which directly calls `blt_add_test()` and all other keywords/parameters are the same.

white238: My vote is to match BLT since it is influencing a BLT macro. The reason it is named that way is because it follows the pattern of the BLT MPI variable which was created long before the OpenMP parameter which was before I knew that `OMP_NUM_THREADS` environment variable existed.

src/cmake/AxomMacros.cmake
kennyweiss: Thanks. Since this is already tested, I'll update the docs separately.

**Update:**  https://github.com/llnl/axom/pull/1751/changes#diff-1d2f627492d8794d6cd7e721307fa5cf4f527762eaee41820a132532fb3ecd52

PR #1751, opened 2025-12-24
Updates mfem dependency to mfem@4.9

kennyweiss:
# Summary

- This PR update our mfem dependency to mfem@4.9
- It also fixes some unsafe calls within the sample-based shaping infrastructure to access/update quadrature data. 
    - We now use `mfem::Reshape()` instead of `QuadratureFunction::GetValues()` (which was failing w/ mfem@4.9). 
    - The update should work with mfem@4.8 and mfem@4.9

### TODO:
- [x] Update rz host-configs
- [x] Update cz host-configs
- [x] Update docker images
- [x] Update vcpkg
- [x] Update RELEASE-NOTES

### DIFF
--- a/src/cmake/AxomMacros.cmake
+++ b/src/cmake/AxomMacros.cmake
@@ -228,8 +228,13 @@ endmacro(axom_add_library)
 ##               NUM_OMP_THREADS [n]
 ##               CONFIGURATIONS  [config1 [config2...]])
 ##
-## Wrapper around blt_add_test() that handles functionality that Axom applies to all
-## tests.
+## Wrapper around blt_add_test() that handles functionality 
+## that Axom applies to all tests.
+##
+## Note that NUM_OMP_THREADS delegates to the corresponding argument 
+## in blt_add_test() and sets the OpenMP environment variable OMP_NUM_THREADS.
+## When AXOM_ENABLE_OPENMP is set and NUM_OMP_THREADS is not provided, 
+## this macros also sets the environment variable OMP_NUM_THREADS=1.
 ##------------------------------------------------------------------------------
 macro(axom_add_test)
 


src/axom/core/numerics/quadrature.hpp
```diff
@@ -41,7 +41,7 @@ class QuadratureRule
 
   //! \brief Accessor for the size of the quadrature rule
   AXOM_HOST_DEVICE
-  int getNumPoints() const { return m_nodes.size(); }
+  int getNumPoints() const { return static_cast<int>(m_nodes.size()); }
```
kennyweiss, line 44: I fixed some type conversion warnings from msvc along the way.  
(There are many more)

src/axom/quest/detail/shaping/shaping_helpers.hpp
kennyweiss, line 1: Note: Some mfem-based shaping tests were failing after the upgrade to mfem@4.9. 
Some users were also experiencing memory-related issues w/ shaping against newer versions of mfem. 

I think these fixes should resolve the user problem and work against mfem@4.8 as well.

src/cmake/AxomMacros.cmake
```diff
@@ -228,8 +228,13 @@ endmacro(axom_add_library)
...
 ##
-## Wrapper around blt_add_test() that handles functionality that Axom applies to all
-## tests.
+## Wrapper around blt_add_test() that handles functionality 
+## that Axom applies to all tests.
+##
+## Note that NUM_OMP_THREADS delegates to the corresponding argument 
+## in blt_add_test() and sets the OpenMP environment variable OMP_NUM_THREADS.
+## When AXOM_ENABLE_OPENMP is set and NUM_OMP_THREADS is not provided, 
+## this macros also sets the environment variable OMP_NUM_THREADS=1.
```
kennyweiss, line 237: This update is for a request in https://github.com/llnl/axom/pull/1749

Tag: @Arlie-Capps  @rhornung67

src/axom/quest/detail/shaping/shaping_helpers.hpp
```diff
@@ -18,6 +18,7 @@
 
 #if defined(AXOM_USE_MFEM)
   #include "mfem.hpp"
+  #include "mfem/linalg/dtensor.hpp"
```
white238, line 21: Is this not included in the above header? Should this be reported to mfem?

white238 (approved):

src/axom/quest/detail/shaping/shaping_helpers.hpp
kennyweiss: It's not:
* https://github.com/mfem/mfem/blob/8e3a7b90fc1934c33b66baafee10fabd1eca0dea/mfem.hpp#L46
* https://github.com/mfem/mfem/blob/master/linalg/linalg.hpp
* https://github.com/mfem/mfem/blob/master/linalg/dtensor.hpp

@v-dobrev, @tzanio -- is `dtensor.hpp` intentionally left out of `mfem.hpp`  / `linalg.hpp` ?

src/cmake/AxomMacros.cmake
Arlie-Capps: Thanks, @kennyweiss .  I'd suggest to even add words here like 
```
## 
## Watch out for confusion between the names of macro argument
## OMP_NUM_THREADS and environment variable NUM_OMP_THREADS.
## These names are close, and related, but not the same.
```
Maybe that's beating a dead horse, but I can just see future @Arlie-Capps writing a bug because of hubris and hasty assumptions.  Documentation gives future me a fighting chance.

Arlie-Capps (approved):

src/cmake/AxomMacros.cmake
kennyweiss: Thanks @Arlie-Capps -- since `axom_add_test()` and `blt_add_test()` both have explicit logic to set the `OMP_NUM_THREADS` environment variable, and this has not been an issue in the past, it seems reasonable (to me) to stick with the current docs update and perhaps make it more explicit in the future if this trips anyone up.

PR #1752, opened 2025-12-30
Enable Windows CI w/ TPLs

kennyweiss:
# Summary

- This PR reworks our CI on Windows to use TPL builds from the "manual Windows TPL" CI, similar to our docker-based workflow.
- We can now generate a TPL set via the "manual Windows TPL" github actions, which will cache the vcpkg builds
- In our ci-tests.yaml github action, we pull in the cached TPLs and use then in our testing
    - For this initial setup, we're only testing the `x64-windows` Debug config in our on-commit CI-tests
    - We can easily add the other Windows configs (`{x64, x86}`, `{Debug, Release}`) as desired
- Resolves https://github.com/llnl/axom/issues/546

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -14,7 +14,7 @@ concurrency:
 env:
   CLANG_DOCKER_IMAGE: axom/tpls:clang-19_12-24-25_01h-51m
   GCC_DOCKER_IMAGE: axom/tpls:gcc-13_12-24-25_01h-51m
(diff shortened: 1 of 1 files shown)


kennyweiss:
Reverted to draft...
It seemed like this was working, but the `ci-test` plan is not finding the cache from the "manual Windows build" plan.

Might be a mismatch in my expectations for caches. From the [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache):
> Workflow runs cannot restore caches created for child branches or sibling branches.

PR #1753, opened 2026-01-02
Decompose hex to 18 tets for clipping

gunney1:
# Summary

- This PR is a feature
- It does the following:
  - Implement 18-tet decomposition for a hex for clipping.

The 18-tet decomposition should reduce work compared to the default 24-tet decomposition, because it generates fewer tets.  If a hex face has 4 coplanar vertices, there will be degenerate tets, so we check for them.  For on-going performance evaluation, we are supporting both decompositions for now.

This change is part of the bigger change for improving mesh clipping performance.

### DIFF
--- a/src/axom/quest/MeshClipper.cpp
+++ b/src/axom/quest/MeshClipper.cpp
@@ -101,7 +101,7 @@ void MeshClipper::clip(axom::ArrayView<double> ovlap)
         logLabelStats(tetLabels, "tets");
       }
       m_impl->collectOnIndices(tetLabels.view(), tetsOnBdry);
-      m_impl->remapTetIndices(tetsOnBdry, cellsOnBdry);
+      m_impl->remapTetIndices(cellsOnBdry, tetsOnBdry);
 
       SLIC_ASSERT(tetsOnBdry.getAllocatorID() == m_shapeMesh.getAllocatorID());
       SLIC_ASSERT(tetsOnBdry.size() <= cellsOnBdry.size() * NUM_TETS_PER_HEX);


src/axom/quest/ShapeMesh.hpp
```diff
@@ -153,6 +166,21 @@ class ShapeMesh
   // zero to zero.  Default threshold is 1e-10.
   void setZeroThreshold(double threshold) { m_zeroThreshold = threshold; }
 
+  /*!
+   * @brief Decompose a hexahedron into NUM_TETS_PER_HEX tetrahedra.
+   * @param hex [in] The hexahedron
+   * @param tets [out] Pointer to space for NUM_TETS_PER_HEX tetrahedra.
```
Arlie-Capps, line 172: Please add (redundant) information: something like "Caller must ensure tets points to sufficient space for the tetrahedra generated by this method.  This method neither checks the pointer nor reallocates the space."

You can add this either here or in a warning.  This amplifies the caution that readers should feel from what you already wrote.

gunney1: Done.

src/axom/quest/ShapeMesh.hpp
```diff
@@ -345,6 +384,125 @@ class ShapeMesh
                                     const conduit::DataType& dtype);
 };
 
+AXOM_HOST_DEVICE inline void ShapeMesh::hexToTets(const HexahedronType& hex, TetrahedronType* tets)
```
bmhan12, line 387: For future cleanup, we'll want to have `hexToTets` be part of the `Hexahedron` class, whether that be replacing or co-existing with the 24-tet triangulation.

gunney1: I'll make a note of that in the code.

PR #1754, opened 2026-01-02
Generalize statistics collection code for MeshClipper

gunney1:
# Summary

- This PR is a refactoring
- It does the following:
  - Refactor and generalize the mesh clipper's statistics collection, reduction and logging code.

Statistics are stored in the key-value pairs of a `conduit::Node`.  New quantities can be added without modifying the mesh clipper implementation.  The mesh clipper can just add, globally reduce and log statistics without having to know what quantities they are.

The screen level parameter is also added.  It's a run-time parameter for studying how effective screening is.  It's integrated with statistics so it's included on this branch.

This change is part of the clipping performance work and has been separated to make smaller change sets.  Some of the statistics code introduced here may seem unused, but they will be used in up-coming PRs.

### DIFF
--- a/src/axom/quest/ShapeMesh.hpp
+++ b/src/axom/quest/ShapeMesh.hpp
@@ -62,7 +62,7 @@ public:
   using BoundingBox3DType = primal::BoundingBox<double, 3>;
 
   /*!
-   * @brief Number of tetrahedra per hexahedron decomposes into
+   * @brief Number of tetrahedra that a hexahedron decomposes into
    * @see hexToTets()
    *
    * @internal Only values of 24 and 18 are valid.  18 is likely more
@@ -176,8 +176,10 @@ public:
    * point at the average of the face vertices and decompose the face
    * into 4 triangles.
    *
-   * It is expected that this method will be used in long inner
-   * loops, so it is bare-bones for best performance.
+   * It is expected that this method will be used in long inner loops,
+   * so it is bare-bones for best performance.  Caller must ensure
+   * tets points to at least NUM_TETS_PER_HEX objects. This method
+   * neither checks the pointer nor reallocates the space.
    */
   AXOM_HOST_DEVICE inline static void hexToTets(const HexahedronType& hex, TetrahedronType* tets);
 
--- a/src/axom/quest/detail/clipping/MeshClipperImpl.hpp
+++ b/src/axom/quest/detail/clipping/MeshClipperImpl.hpp
@@ -111,7 +111,7 @@ public:
       return;
     };
 
-    AXOM_ANNOTATE_SCOPE("MeshClipper::collect_unlabeleds");
+    AXOM_ANNOTATE_SCOPE("MeshClipper:collect_indices");
     /*!
      * 1. Generate tmpLabels, having a value of 1 where labels is LABEL_ON and zero elsewhere.
      * 2. Inclusive scan on tmpLabels to generate values that step up at LABEL_ON cells.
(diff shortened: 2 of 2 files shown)


src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -231,16 +232,21 @@ class MeshClipperStrategy
...
   virtual bool specializedClipCells(quest::experimental::ShapeMesh& shapeMesh,
-                                    axom::ArrayView<double> ovlap)
+                                    axom::ArrayView<double> ovlap,
+                                    conduit::Node& statistics)
   {
     AXOM_UNUSED_VAR(shapeMesh);
     AXOM_UNUSED_VAR(ovlap);
+    AXOM_UNUSED_VAR(statistics);
     return false;
   }
```
kennyweiss, line 251: > * @param [out] statistics Optional statistics to record
      consisting of child nodes with integer values.

Since `statistics` is given by reference, rather than a pointer which could be `nullptr` (or perhaps a `std::optional`), it looks like it is required, rather than optional.

Is there a performance cost to passing in the statistics? E.g. if the statistics are written to the conduit node within an inner loop, and then ignored.

src/axom/quest/MeshClipper.hpp
```diff
@@ -81,6 +82,50 @@ class MeshClipper
...
 
+  //@{
+
+  /*!
+   * @brief Log clipping statistics.
+   * Intended for developer use.
+   *
+   * This is a collective method if MPI-parallel.
+   */
+  void logClippingStats(bool local = false, bool sum = true, bool max = false) const;
```
kennyweiss, line 97: Please document these three parameters

src/axom/quest/MeshClipper.hpp
```diff
@@ -81,6 +82,50 @@ class MeshClipper
...
+   *
+   * This is a collective method if MPI-parallel.
+   */
+  conduit::Node getGlobalClippingStats() const;
+
+  /*!
+   * @brief Set the level of screening,
+   * intended for developer use.
+   */
+  void setScreenLevel(int screenLevel) { m_screenLevel = screenLevel; }
```
kennyweiss, line 127: Please document the valid levels and their meanings

src/axom/quest/MeshClipper.hpp
```diff
@@ -81,6 +82,50 @@ class MeshClipper
...
+   */
+  const conduit::Node& getClippingStats() const { return m_counterStats; }
+
+  /*!
+   * @brief Get global assorted clipping statistics,
+   * intended for developer use.
+   *
+   * This is a collective method if MPI-parallel.
+   */
+  conduit::Node getGlobalClippingStats() const;
```
kennyweiss, line 111: Gut-check: Is this supposed to return by value?

src/axom/quest/MeshClipper.cpp
```diff
@@ -90,16 +113,24 @@ void MeshClipper::clip(axom::ArrayView<double> ovlap)
...
     {
       if(m_verbose)
       {
-        logLabelStats(tetLabels, "tets");
+        getLabelCounts(tetLabels, tetsInCount, tetsOnCount, tetsOutCount);
+        logClippingStats();
       }
+
       m_impl->collectOnIndices(tetLabels.view(), tetsOnBdry);
       m_impl->remapTetIndices(cellsOnBdry, tetsOnBdry);
```
kennyweiss, line 135: Gut check: Was this change intentional? Or perhaps a bug fix? I don't see an API change in this PR for `Impl::remapTetIndices()`

kennyweiss (approved):
Looks like a useful feature @gunney1 

Please check that this doesn't have a performance overhead in cases where you don't want/need to collect statistics.

src/axom/quest/MeshClipperStrategy.hpp
gunney1: I don't access the `Node` in inner loops.  That doesn't work anyway when the inner loop is a GPU kernel.  A typical use is to use a RAJA sum object in the inner loop then get the value after the loop.  I have not seen any significant cost to computing statistics this way.

src/axom/quest/MeshClipper.hpp
gunney1: Done.

src/axom/quest/MeshClipper.hpp
gunney1: Done.

src/axom/quest/MeshClipper.hpp
gunney1: Yes.   The global clipping stats are not stored internally.

src/axom/quest/MeshClipper.cpp
gunney1: It fixes an error I made when I split up the huge PR into smaller ones.  I didn't catch the error immediately because this code wasn't exercised until the PRs I submitted later.  So this is a correction to a code that wasn't being exercised, but I thought it made sense to keep the correction as close as possible to the place where the error was made.  In the original huge PR, it was correct.

PR #1755, opened 2026-01-03
Reduce device copies when collecting indices

gunney1:
# Summary

- This PR is a performance fix
- It does the following:
  - Rewrite the code that builds an index list of cells/tets on the shape boundary.  The rewrites removes 2 device copies for the cost of a RAJA reduce, which is faster.

This change is part of the clipping performance work and has been separated to make smaller change sets.

### DIFF
--- a/src/axom/quest/MeshClipper.cpp
+++ b/src/axom/quest/MeshClipper.cpp
@@ -132,7 +132,7 @@ void MeshClipper::clip(axom::ArrayView<double> ovlap)
       }
 
       m_impl->collectOnIndices(tetLabels.view(), tetsOnBdry);
-      m_impl->remapTetIndices(tetsOnBdry, cellsOnBdry);
+      m_impl->remapTetIndices(cellsOnBdry, tetsOnBdry);
 
       SLIC_ASSERT(tetsOnBdry.getAllocatorID() == m_shapeMesh.getAllocatorID());
       SLIC_ASSERT(tetsOnBdry.size() <= cellsOnBdry.size() * NUM_TETS_PER_HEX);


src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -126,42 +126,42 @@ class MeshClipperImpl : public MeshClipper::Impl
...
+    auto onIndicesView = onIndices.view();
     axom::for_all<ExecSpace>(
       1,
-      labelCount,
+      1 + labelCount,
       AXOM_LAMBDA(axom::IndexType i) {
         if(tmpLabelsView[i] != tmpLabelsView[i - 1])
         {
-          onIndicesView[tmpLabelsView[i - 1]] = i;
+          onIndicesView[tmpLabelsView[i] - 1] = i - 1;
```
Arlie-Capps, line 164: Thank you for the fixes here and in `MeshClipper::clip()`.  Do you have tests that catch this?

gunney1: Actually this was not a fix.  It's just compensating for the fact that the values in `tmpLabelViews` has shifted by one from before this change (see line 140 and the changed range at line 160).  The code would crash pretty quickly if this was wrong, so I don't have a unit test for it.

gunney1: The fix in `MeshClipper::clip` was a real fix.  The error wasn't originally caught in this PR because it wasn't exercised by the `TetClipper` class on this branch.  I found it when I was creating PRs for other shapes.  I had it correct in the original huge PR but messed it up when I split up the changes into smaller PRs.

src/axom/quest/MeshClipper.cpp
```diff
@@ -132,7 +132,7 @@ void MeshClipper::clip(axom::ArrayView<double> ovlap)
       }
 
       m_impl->collectOnIndices(tetLabels.view(), tetsOnBdry);
-      m_impl->remapTetIndices(tetsOnBdry, cellsOnBdry);
+      m_impl->remapTetIndices(cellsOnBdry, tetsOnBdry);
```
kennyweiss, line 135: I also noticed this change in: https://github.com/llnl/axom/pull/1754#discussion_r2666218495

Issue #1704, opened 2025-10-31
Re-factor MeshClipperImpl::computeClipVolumes3D


gunney1:
PR https://github.com/LLNL/axom/pull/1667 leaves several methods `quest::experimental::detail::MeshClipperImpl::computeClipVolumes3D` that have been lifted and modified from `quest::IntersectionShaper`.  We should try to re-factor these, since they have a lot of commonalities.  Note that there are very complex logic in the code.

gunney1:
Resolved by PR https://github.com/llnl/axom/pull/1778

PR #1756, opened 2026-01-03
Use 2-pass candidate search for mesh clipping

gunney1:
# Summary

- This PR is a performance improvement and re-factor
- It does the following:
  - Implement a 2-pass BVH candidate search in the `MeshClipperImpl::computeClipVolumes3DTets`.
  - Factor common parts of that function to make it more compact and for re-use in a later PR.

The 2-pass BVH candidate search is more selective than the single-pass bounding box search, because we can use a more accurate screening at the leaves.  The 2-pass is implemented only for tet-screened clips in this PR.  The non-screened version and the hex-screened version will be updated in a later PR.  (They are less likely to be used when performance is critical.)

The re-factoring is a partial fix for issue https://github.com/LLNL/axom/issues/1704.

### DIFF
--- a/src/axom/quest/MeshClipper.cpp
+++ b/src/axom/quest/MeshClipper.cpp
@@ -27,7 +27,7 @@ MeshClipper::MeshClipper(quest::experimental::ShapeMesh& shapeMesh,
   , m_verbose(false)
   , m_screenLevel(3)
 {
-  // Initialize statistics.
+  // Initialize statistics used by this class.
   m_counterStats["cellsIn"].set_int64(0);
   m_counterStats["cellsOn"].set_int64(0);
   m_counterStats["cellsOut"].set_int64(0);
--- a/src/axom/quest/MeshClipper.hpp
+++ b/src/axom/quest/MeshClipper.hpp
@@ -88,6 +88,10 @@ public:
    * @brief Log clipping statistics.
    * Intended for developer use.
    *
+   * @param local [in] Log the MPI local statistics.
+   * @param sum [in] Log the statistics summed over MPI ranks.
+   * @param max [in] Log the max statistics over MPI ranks.
+   *
    * This is a collective method if MPI-parallel.
    */
   void logClippingStats(bool local = false, bool sum = true, bool max = false) const;
@@ -109,6 +113,16 @@ public:
   /*!
    * @brief Set the level of screening,
    * intended for developer use.
+   * @param screenLevel [in] the value to control how much screening
+   *   is done to avoid using the slower clipping functions.
+   *
+   * Screenlevel values:
+   * - 0 allows no screening
+   * - 1 allows screening cells.  @see MeshClipperStrategy::labelCellsInOut.
+   * - 2 also allows screening tets.  @see MeshClipperStrategy::labelTetsInOut.
+   * - 3 also allows a check on the tet from mesh and a piece of the discretized
+   *     geometry to (conservatively) determine whether the piece is completely
+   *     inside or completely outside of the tst.
    */
   void setScreenLevel(int screenLevel) { m_screenLevel = screenLevel; }
 
--- a/src/axom/quest/ShapeMesh.hpp
+++ b/src/axom/quest/ShapeMesh.hpp
@@ -62,7 +62,7 @@ public:
   using BoundingBox3DType = primal::BoundingBox<double, 3>;
 
   /*!
-   * @brief Number of tetrahedra per hexahedron decomposes into
+   * @brief Number of tetrahedra that a hexahedron decomposes into
    * @see hexToTets()
    *
    * @internal Only values of 24 and 18 are valid.  18 is likely more
@@ -176,8 +176,10 @@ public:
    * point at the average of the face vertices and decompose the face
    * into 4 triangles.
    *
-   * It is expected that this method will be used in long inner
-   * loops, so it is bare-bones for best performance.
+   * It is expected that this method will be used in long inner loops,
+   * so it is bare-bones for best performance.  Caller must ensure
+   * tets points to at least NUM_TETS_PER_HEX objects. This method
+   * neither checks the pointer nor reallocates the space.
    */
   AXOM_HOST_DEVICE inline static void hexToTets(const HexahedronType& hex, TetrahedronType* tets);
 


src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -822,97 +783,588 @@ class MeshClipperImpl : public MeshClipper::Impl
...
+    const auto bvhTraverser = bvh.getTraverser();
+    /*
+     * Predicate for traversing the BVH.  We enter BVH nodes
+     * whose bounding boxes intersect the query bounding box.
+     */
+    auto traversePredBox = [] AXOM_HOST_DEVICE(const BoundingBoxType& queryBbox,
+                                               const BoundingBoxType& bvhBbox) -> bool {
+      return queryBbox.intersectsWith(bvhBbox);
+    };
+    AXOM_UNUSED_VAR(traversePredBox);
```
Arlie-Capps: Please remove, unless there's a very good reason for keeping.  If we do need this code, please clearly comment.  Also (new) lines 821 and 839.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -822,97 +783,588 @@ class MeshClipperImpl : public MeshClipper::Impl
...
+                                         clipStats,
+                                         screenLevel);
         });
     }
+    AXOM_ANNOTATE_END("MeshClipper:clipLoop_tetScreened");
 
     clipStats.copyTo(statistics);
     statistics["clipsCandidates"].set_int64(candidates.size());
 
+    SLIC_DEBUG(axom::fmt::format(""));
```
Arlie-Capps: Please remove, or make this more useful.

gunney1: Removed.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: Removed.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -822,97 +783,588 @@ class MeshClipperImpl : public MeshClipper::Impl
...
+                              TetrahedronType(oct[0], oct[3], oct[4], oct[5]),
+                              TetrahedronType(oct[0], oct[3], oct[5], oct[1])};
+    double octVol = 0.0;
+    for(int i = 0; i < 4; ++i)
+    {
+      double tetVol = tets[i].signedVolume();
+      SLIC_ASSERT(tetVol >= -EPS);  // Tet may be degenerate but not inverted.
+      octVol += axom::utilities::abs(tetVol);
+    }
+    return octVol;
```
kennyweiss, line 1061: Should this be implemented within the `Octahedron` class as `Octahedron::signedVolume()` ?

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -822,97 +783,588 @@ class MeshClipperImpl : public MeshClipper::Impl
...
+  AXOM_HOST_DEVICE static inline LabelType computeMeshTetGeomPieceOverlap(
+    const TetrahedronType& meshTet,
+    const TetOrOctType& geomPiece,
+    double* overlapVolume,
+    const ClippingStats& clipStats,
+    int screenLevel)
+  {
+    using ATOMIC_POL = typename axom::execution_space<ExecSpace>::atomic_policy;
+    constexpr bool tryFixOrientation = false;
+    if(screenLevel >= 3)
```
kennyweiss, line 1086: Would it make sense to define a `ScreenLevel` enum w/ meaningful names for the different levels?

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -822,97 +783,588 @@ class MeshClipperImpl : public MeshClipper::Impl
...
+    }
+    return LabelType::LABEL_ON;
+  }
+
+  /*!
+   * @brief Whether a tet and a bounding box (possibly) intersect.
+   * Answer may be a false positive but never a false negative.
+   */
+  AXOM_HOST_DEVICE static inline bool tetBoxCollision(const TetrahedronType& tet,
+                                                      const BoundingBoxType& box)
```
kennyweiss, line 1184: Is there a reason that this is implemented here as opposed to `primal::intersect(Tetrahedron, BoundingBox) -> bool` along with appropriate tests?  E.g. the functions here are not fully general, and are customized for the needs of the MeshClipper?

Similarly for the other intersection tests in this file:
* `intersect(Tet, Tet)`
* `intersect(Tet, Oct)`

kennyweiss (approved):

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: I'm not sure.  The octs from discretizing spheres and cones should convex, but general octs may be concave.  I don't know if that matters but it seems like something to think about.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: I don't think so.  It's just a way for me to do some experimentation to understand the effects of various things I was trying and to compare to similar codes.  Once that is done, I believe that screening will prove to be low cost and usually helpful and I plan to remove the parameter.

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: The main reason is that this implementation needs to be fast, and we can tolerate false positives for our needs.  I think we want `primal::intersect` to never give false results.  So yes, the algorithm here is not fully general.

kennyweiss: Makes sense -- maybe add a comment to that effect so future readers know the expectations.  (Looks like it was implied by the current docs, but could be more explicit)

Arlie-Capps (approved):

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: Done.

gunney1: OK

PR #1757, opened 2026-01-04
Sphere geometry for `MeshClipper`

gunney1:
# Summary

- This PR is a feature
- It does the following:
  - Add `SphereClipper`, the sphere geometry implemenation for `MeshClipper`
  - Add test for `SphereClipper`
The framework code for `MeshClipper` uses geometry-specific implenentations of the `MeshClipperStrategy` class.  This PR implements the sphere geometry through that strategy class.

### DIFF
--- a/src/axom/quest/detail/clipping/SphereClipper.cpp
+++ b/src/axom/quest/detail/clipping/SphereClipper.cpp
@@ -83,6 +83,9 @@ void SphereClipper::labelCellsInOutImpl(quest::experimental::ShapeMesh& shapeMes
       }
       const auto& hex = cellsAsHexes[cellId];
       cellLabel = polyhedronToLabel(hex, sphere);
+      // Note: cellLabel may be set to LABEL_ON if polyhedronToLabel
+      // cannot efficiently determine whether the hex is IN or OUT.
+      // See MeshClipperStrategy::labelCellsInOut().
     });
   return;
 }
@@ -162,6 +165,9 @@ void SphereClipper::labelTetsInOutImpl(quest::experimental::ShapeMesh& shapeMesh
         }
         const TetrahedronType& tet = cellTets[ti];
         tetLabel = polyhedronToLabel(tet, sphere);
+        // Note: cellLabel may be set to LABEL_ON if polyhedronToLabel
+        // cannot efficiently determine whether the tet is IN or OUT.
+        // See MeshClipperStrategy::labelTetsInOut().
       }
     });
   return;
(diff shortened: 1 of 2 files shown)


src/axom/quest/MeshClipper.cpp
```diff
@@ -132,7 +132,7 @@ void MeshClipper::clip(axom::ArrayView<double> ovlap)
       }
 
       m_impl->collectOnIndices(tetLabels.view(), tetsOnBdry);
-      m_impl->remapTetIndices(tetsOnBdry, cellsOnBdry);
+      m_impl->remapTetIndices(cellsOnBdry, tetsOnBdry);
```
kennyweiss, line 135: This change (likely a bugfix) is here again. 

I saw it in https://github.com/llnl/axom/pull/1754 and https://github.com/llnl/axom/pull/1755

src/axom/quest/detail/clipping/SphereClipper.cpp
```diff
@@ -0,0 +1,275 @@
...
+      TetrahedronType cellTets[NUM_TETS_PER_HEX];
+      ShapeMesh::hexToTets(hex, cellTets);
+
+      for(IndexType ti = 0; ti < NUM_TETS_PER_HEX; ++ti)
+      {
+        LabelType& tetLabel = tetLabels[ci * NUM_TETS_PER_HEX + ti];
+        const axom::IndexType tetId = cellId * NUM_TETS_PER_HEX + ti;
+        if(axom::utilities::isNearlyEqual(tetVolumes[tetId], 0.0, EPS))
+        {
+          tetLabel = LabelType::LABEL_OUT;
```
kennyweiss, line 163: Consistency: Can/should we use `primal::OrientationResult` instead of `LabelType`?

https://github.com/llnl/axom/blob/ffda074ec0d1fed5dfc86acea37abd59ab96e194/src/axom/primal/geometry/OrientationResult.hpp#L20-L28

kennyweiss (approved):

src/axom/quest/detail/clipping/SphereClipper.cpp
gunney1: It can be done, I think.  Will defer about whether we should.

src/axom/quest/detail/clipping/SphereClipper.cpp
```diff
@@ -0,0 +1,275 @@
...
+  const SphereType& sphere) const
+{
+  /*
+    If bounding box of polyhedron is more than the radius distance
+    from center, it is LABEL_OUT.  (Comparing vertices for this check
+    can miss intersections by edges and facets, so we compare bounding
+    box.)
+
+    Otherwise, polyhedron either LABEL_ON or LABEL_IN.  Sphere is
+    convex, so polyhedron is IN only if all vertices are inside.
```
Arlie-Capps: I see that the function implements the logic you wrote here.  However, here's a case that doesn't get caught.  I'm stating the 2D case for ease of communication, but it generalizes to 3D:
* Circle of radius 1 centered at origin
* Triangle with vertices at a = (1.5, 0), b = (1, 1), and c = (0, v), where v < 3.

Slide v down toward the origin. Before the line segment ac touches the circle, it should be labeled as LABEL_OUT.  But by this algorithm, its bounding box will intersect the circle, and thus get LABEL_ON.

If this is not a problem, please clearly document that this can label some polyhedra as LABEL_ON when they should be LABEL_OUT.

gunney1: This is not a problem because we are only trying to eliminate the use of the expensive clip function where we can.  LABEL_ON just means the clip function has to be used.  The clip function will give the correct results.

I'll document this.

PR #1759, opened 2026-01-05
Screening for `TetClipper`

gunney1:
# Summary

- This PR is a feature
- It does the following:
  - Add screening for the tet clipper.  `TetClipper` was introduced a while back as part of the shaping refactor, but it was a simple version.  This version implements screening to reduce the number of times we call the primitive clipping function.
  - Allow users to specify a tet with negative volume and a flag to automatically fix the tet orientation to make it positive.


kennyweiss (approved):
Thanks for splitting up the PRs @gunney1 -- made it much easier to understand the pieces and review quickly

bmhan12 (approved):

PR #1760, opened 2026-01-06
Updates MFEMReader and winding number example

kennyweiss:
# Summary

This PR adds a few improvements centered around improvements to the quest winding number example

- Improves and cleans up `MFEMReader`
  - Can now read in multispan NURBS meshes (no longer hardcodes knot vector for a single span)
  - Fixed a bug in mapping attributes to curves. We no longer assume the attributes start at 1 and are contiguous.
    Instead, we add an optional array to map the returned curves to their attributes.
- Improves quest winding number example
  - Adds flag to enable/disable memoization (on by default)
  - Adds optional caliper support
  - Updates to `svg2contours` script which convets an SVG to an mfem Mesh file
    - When finding the SVG resolution, prefer `viewBox` info, followed by `height` and `width`
    - Updates dependency to svgpathtools@1.7.2, which incorporates our patch (and removes the patch
    - Also updates environment setup to use `uv` instead of `venv`
- Improves error reporting in Klee/Shaping
  - We also check the return code when reading a file
  - We now throw an error when the requested file does not exist (or cannot be opened)
  - This was a user request since missing files in the shaping app are not a recoverable error, and the warning message was obscured by later resulting failures
- Slic: Adds `slic::getAbortFunction()` to return the function pointer associated with the slic error handler
  - Also adds an RAII utility class slic::ScopedAbortToThrow which allow temporarily setting the abort function to throw an exception. This is helpful for testing that `SLIC_ERROR`/`SLIC_ASSERT` calls happen without gtest Death tests (which cannot run in MPI-based tests).
- Misc:
  - `MPIWrapper` now only finalizes MPI if it initialized it
  - Fixes a few tests so they can run through MPI
  - Fixes some warnings about unused variables

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -44,6 +44,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Adds a `conduit` spack variant. Conduit was previously a required dependency in our spack package, and is now enabled by default.
 - Adds spack variants for `adiak` and `caliper`. These replace the previous `profiling` variant which enabled both at the same time.
 - Adds the `AXOM_TEST_NUM_OMP_THREADS` configuration variable to control the default OpenMP thread count for tests.
+- Slic: Adds `slic::getAbortFunction()` to return a function pointer to the abort handler called during logging functions
+
 
 ###  Changed
 - Evaluation methods for line integrals in `axom::primal` have been generalized, and 
@@ -79,12 +81,15 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   Refer to the new `StoragePolicy` interface for substitute functionality.
 - The `axom::bump::clipping` namespace was renamed to `axom::bump::extraction` since it now
   contains additional algorithms.
-- Updates to [mfem version 4.9][https://github.com/mfem/mfem/releases/tag/v4.9].
-  Also updates mfem's hypre dependency to [hypre version 2.27][https://github.com/hypre-space/hypre/releases/tag/v2.27.0].
+- Updates to [mfem version 4.9](https://github.com/mfem/mfem/releases/tag/v4.9).
+  Also updates mfem's hypre dependency to [hypre version 2.27](https://github.com/hypre-space/hypre/releases/tag/v2.27.0).
+- Quest: MFEMReader can now read in multispan 1D NURBS meshes
 
 ###  Fixed
 - Sina's Fortran tests are now running (instead of silently failing)
 - Optimized `Array::push_back()` and `Array::emplace_back()` operations.
+- Quest: In Shaping applications, we now check return code after attempting to load a mesh
+  and throw an error for unsuccessful loads.
 
 ###  Deprecated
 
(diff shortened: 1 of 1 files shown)


src/axom/quest/io/MFEMReader.cpp
```diff
@@ -28,110 +28,231 @@ namespace internal
...
+      }
     }
 
-    delete mesh;
+    return false;
+  };
+
+  // Examine the mesh attributes and group all of the related curves w/ same attribute
+  // Assumption is that they're part of the same contour
+  if(const bool isNURBS = dynamic_cast<const mfem::NURBSFECollection *>(fec) != nullptr; isNURBS)
```
Arlie-Capps: Style comment: I think this would be clearer if you moved the definition of `isNURBS` to the previous line, out of the test of the `if`, as you did at (new) line 154.

src/axom/quest/examples/quest_winding_number.cpp
```diff
@@ -25,67 +25,73 @@
...
+  // Utility function to get query point from query index
+  const auto num_query_points = query_mesh->GetNodalFESpace()->GetNDofs();
+  auto query_point = [&query_mesh](int idx) -> Point2D {
+    Point2D pt;
+    query_mesh->GetNode(idx, pt.data());
+    return pt;
+  };
+
+  // Query the winding numbers at each degree of freedom (DoF) of the query mesh.
+  // The loop below independently checks (and adaptively refines) every curve for each query point.
```
jcs15c:
```suggestion
  // The loop below independently checks every curve for each query point.
```
This comment is a bit outdated now that the function is templated, since the "adaptive refinement" originally referred to the memoization.

src/axom/quest/io/MFEMReader.cpp
kennyweiss: I like the new `if-with-initialization` formulation for variables that are only necessary for the if/else scope (where appropriate), but, I can move it out if you think it's clearer

src/axom/slic/interface/slic.hpp
```diff
@@ -193,6 +194,59 @@ bool isAbortOnWarningsEnabled();
...
+ */
+class ScopedAbortToThrow
+{
+public:
+  ScopedAbortToThrow()
+    : m_prev_abort_on_error(axom::slic::isAbortOnErrorsEnabled())
+    , m_prev_abort_function(axom::slic::getAbortFunction())
+  {
+    axom::slic::enableAbortOnError();
+    axom::slic::setAbortFunction(&ScopedAbortToThrow::abort_to_throw);
```
Arlie-Capps, line 228: Do these lines need to be guarded by `if (axom::slic::isInitialized())`, as in the destructor?

kennyweiss: I don't think so since `enableAbortOnError()` calls `ensureInitialized()`

https://github.com/llnl/axom/blob/ffda074ec0d1fed5dfc86acea37abd59ab96e194/src/axom/slic/interface/slic.cpp#L103-L112

Do you think a comment would help?

kennyweiss: The `if(isInitialized)` in the destructor is guarding against re-initializing just to change the error handler back to its previous state.

Arlie-Capps (approved):
Thanks, Kenny!

src/axom/quest/io/MFEMReader.cpp
Arlie-Capps: I think it's clearer if isNURBS gets put on the previous line, but I can see your point.  I meant this as just a style comment, not a request.

kennyweiss: I updated it.

src/axom/quest/examples/quest_winding_number.cpp
kennyweiss: Done.

PR #1761, opened 2026-01-06
Hexahedron shape for `MeshClipper`

gunney1:
# Summary

- This PR is a feature
- It does the following:
  - Adds the hexahedron implemenation `HexClipper` for `MeshClipper`.
  - Adds tests for `HexClipper`.


src/axom/quest/detail/clipping/HexClipper.cpp
```diff
@@ -0,0 +1,345 @@
...
+AXOM_HOST_DEVICE inline MeshClipperStrategy::LabelType HexClipper::polyhedronToLabel(
+  const Polyhedron& verts,
+  const BoundingBox3DType& vertsBb,
+  const BoundingBox3DType& hexBb,
+  const axom::ArrayView<const TetrahedronType>& hexTets,
+  const axom::StackArray<Triangle3DType, 24>& surfaceTriangles) const
+{
+  /*
+    If vertsBb and hexBb don't intersect, nothing intersects.
+    This check is not technically needed, because the thecks
```
kennyweiss: Typo:
```suggestion
    This check is not technically needed, because the checks
```

src/axom/quest/detail/clipping/HexClipper.hpp
```diff
@@ -0,0 +1,105 @@
...
+  template <typename ExecSpace>
+  void labelCellsInOutImpl(quest::experimental::ShapeMesh& shapeMesh,
+                           axom::ArrayView<LabelType> label);
+
+  template <typename ExecSpace>
+  void labelTetsInOutImpl(quest::experimental::ShapeMesh& shapeMesh,
+                          axom::ArrayView<const axom::IndexType> cellIds,
+                          axom::ArrayView<LabelType> tetLabels);
+
+  //!@brief Compute LabelType for a polyhedron (hex or tet in our case).
```
kennyweiss: Please expand this doxy comment to be more explicit about what this function does. 

It looks like it is computing the labels (in/out) for the vertices of a polyhedron with respect to the volume of the hexahedron. The input polyhedron is expected to be a tet or a hex (but presumably, the only requirement is that it is convex).

Is that right?

kennyweiss (approved):

src/axom/quest/detail/clipping/HexClipper.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+namespace axom
+{
+namespace quest
+{
+namespace experimental
+{
+
+/*!
+ * @brief Geometry clipping operations for sphere geometries.
```
bmhan12:
```suggestion
 * @brief Geometry clipping operations for hexahedron geometries.
```

src/axom/quest/detail/clipping/HexClipper.cpp
gunney1: Thanks.

src/axom/quest/detail/clipping/HexClipper.hpp
gunney1: We don't have a convexity requirement.  The hex may be concave.  This method can mark something as ON the boundary when it's inside or out.  This is a conservative error and doesn't lead to wrong computations, like all the screening that we do.  I'll add a note about this.

src/axom/quest/detail/clipping/HexClipper.hpp
gunney1: Thank you.

PR #1762, opened 2026-01-07
SOR implementation for `MeshClipper`

gunney1:
# Summary

- This PR is a feature
- It does the following:
  - Add `MonotonicZSORClipper`, for clipping a surface-of-revolution where the 2D curve doesn't double back in the axial direction.  For such a curve, the area under the curve is inside the geometry; the area above the curve is outside.
  - Add `SORClipper`, for clipping a general SOR where the curve can double back in the axial direction.  This works by splitting up the curve where it changes directions and use `MotonicZSORClipper` on each section.
  - Add tests for `MotonicZSORClipper` and `SORClipper`.
  - Modify the SOR discretizer so it allows the curve in the -z direction, and remove checks that exclude this situation.
  - Fix a couple of minor errors from recent changes to the `MeshClipper` code.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -45,7 +45,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Adds spack variants for `adiak` and `caliper`. These replace the previous `profiling` variant which enabled both at the same time.
 - Adds the `AXOM_TEST_NUM_OMP_THREADS` configuration variable to control the default OpenMP thread count for tests.
 - Slic: Adds `slic::getAbortFunction()` to return a function pointer to the abort handler called during logging functions
-
+- Quest: Adds a 3D winding number example and impoves WN performance in 2D and 3D
 
 ###  Changed
 - Version of `quest::discretize` that approximates a surface-of-revolution from a polyline
@@ -86,8 +86,9 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   contains additional algorithms.
 - Updates to [mfem version 4.9](https://github.com/mfem/mfem/releases/tag/v4.9).
   Also updates mfem's hypre dependency to [hypre version 2.27](https://github.com/hypre-space/hypre/releases/tag/v2.27.0).
-- Quest: MFEMReader can now read in multispan 1D NURBS meshes
+- Quest: `MFEMReader` can now read in multispan 1D NURBS meshes
 - fmt: Axom's built-in version of the fmt formatting library was updated to [fmt version 12.1.0](https://github.com/fmtlib/fmt/releases/tag/12.1.0).
+- Primal: Improves robustness of `linear_winding_number` by using `atan2` instead of `acos`
 
 ###  Fixed
 - Sina's Fortran tests are now running (instead of silently failing)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -145,7 +145,7 @@ if(AXOM_ENABLE_TOOLS)
 endif()
 
 # Installation examples ('Using Axom') and/or tutorials
-if (AXOM_ENABLE_EXAMPLES OR AXOM_ENABLE_TUTORIALS)
+if(AXOM_ENABLE_EXAMPLES OR AXOM_ENABLE_TUTORIALS)
     add_subdirectory(examples)
 endif()
 
--- a/src/axom/CMakeLists.txt
+++ b/src/axom/CMakeLists.txt
@@ -11,7 +11,7 @@
 
 # Core is an essential part of Axom and cannot be turned off
 axom_add_component(COMPONENT_NAME core DEFAULT_STATE ON)
-if(DEFINED AXOM_ENABLE_CORE)
+if(DEFINED AXOM_ENABLE_CORE AND NOT AXOM_ENABLE_CORE)
     message(FATAL_ERROR "Core is an essential part of Axom and cannot be turned off")
 endif()
 
--- a/src/axom/core/numerics/Determinants.hpp
+++ b/src/axom/core/numerics/Determinants.hpp
@@ -33,8 +33,7 @@ inline AXOM_HOST_DEVICE
 real determinant(const real& a00, const real& a01,
                  const real& a10, const real& a11)
 {
-  const real det = a00 * a11 - a10 * a01;
-  return det;
+  return axom::utilities::fma(a00, a11, -a01*a10);
 }
 
 /*!
@@ -56,12 +55,13 @@ real determinant(const real& a00,  const real& a01,  const real& a02,
                  const real& a10,  const real& a11,  const real& a12,
                  const real& a20,  const real& a21,  const real& a22)
 {
-  const real m01 = a00 * a11 - a10 * a01;
-  const real m02 = a00 * a21 - a20 * a01;
-  const real m12 = a10 * a21 - a20 * a11;
+  const real m01 = axom::utilities::fma(a00, a11, - a10 * a01);
+  const real m02 = axom::utilities::fma(a00, a21, - a20 * a01);
+  const real m12 = axom::utilities::fma(a10, a21, - a20 * a11);
 
-  const real det = m01 * a22 - m02 * a12 + m12 * a02;
-  return det;
+  // det = m01 * a22 - m02 * a12 + m12 * a02;
+  const real t = axom::utilities::fma(-m02, a12, m12 * a02);
+  return axom::utilities::fma(m01, a22, t);
 }
 
 /*!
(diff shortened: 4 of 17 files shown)


src/axom/quest/tests/quest_mesh_clipper.cpp
```diff
@@ -458,6 +461,155 @@ axom::klee::Geometry createGeom_Sphere(const std::string& geomName)
...
+  axom::primal::Vector<double, 3> sorDirection = params.direction.empty()
+    ? primal::Vector3D {1.0, 0.0, 0.0}
+    : primal::Vector3D {params.direction.data()};
+  // discreteFunction is discrete (z,r) pairs describing the r(z) function
+  // to be rotated around the z axis.
+  using Point2DType = axom::primal::Point<double, 2>;
+  double zLen = 0.5 * (params.length < 0 ? 2.40 : params.length);
+  double maxR = params.radius < 0 ? 1.10 : params.radius;
+  axom::Array<Point2DType> discretePts(0, 10);
+#if 1
```
gunney1: I will remove the #else part of this.

src/axom/quest/ShapeMesh.hpp
```diff
@@ -176,8 +176,10 @@ class ShapeMesh
...
    * into 4 triangles.
    *
-   * It is expected that this method will be used in long inner
-   * loops, so it is bare-bones for best performance.
+   * It is expected that this method will be used in long inner loops,
+   * so it is bare-bones for best performance.  Caller must ensure
+   * tets points to at least NUM_TETS_PER_HEX objects. This method
+   * neither checks the pointer nor reallocates the space.
    */
   AXOM_HOST_DEVICE inline static void hexToTets(const HexahedronType& hex, TetrahedronType* tets);
```
kennyweiss, line 184: Once all the PRs are updated, would it make sense to revise the API to take `axom::ArrayView`s instead of raw pointers? They can easily wrap a raw pointer, and carry their sizes, and allocators with them.

src/axom/quest/tests/quest_mesh_clipper.cpp
```diff
@@ -29,6 +29,8 @@
 #include "axom/quest.hpp"
 #include "axom/quest/detail/clipping/HexClipper.hpp"
 #include "axom/quest/detail/clipping/Plane3DClipper.hpp"
+#include "axom/quest/detail/clipping/FSorClipper.hpp"
+#include "axom/quest/detail/clipping/SorClipper.hpp"
 #include "axom/quest/detail/clipping/SphereClipper.hpp"
 #include "axom/quest/detail/clipping/TetClipper.hpp"
```
kennyweiss, line 35: If we're including `quest.hpp`, we should not need to include any specific quest headers

src/axom/quest/tests/quest_mesh_clipper.cpp
```diff
@@ -460,6 +463,145 @@ axom::klee::Geometry createGeom_Sphere(const std::string& geomName)
...
+  axom::ArrayView<const double, 2> discreteFunction((const double*)discretePts.data(),
+                                                    discretePts.size(),
+                                                    2);
+
+  auto compositeOp = std::make_shared<axom::klee::CompositeOperator>(startProp);
+  addScaleOperator(*compositeOp);
+  addTranslateOperator(*compositeOp);
+
+  axom::klee::Geometry sorGeometry =
+    createGeometry_Sor(sorBase, sorDirection, discreteFunction, compositeOp);
```
kennyweiss: Minor: It's a bit confusing/odd that `createGeom_Sor()` calls `createGeometry_Sor()` - is there a better name for this? Perhaps it would be sufficient to add brief comments for these utility functions?

(I realize that this is a utility function a unit test)

src/axom/quest/detail/Discretize_detail.hpp
```diff
@@ -293,7 +292,8 @@ bool discretize(const axom::ArrayView<Point2D> &polyline,
   // That was the octahedron count for one segment.  Multiply by the number
   // of segments we will compute.
   int totaloctcount = segoctcount * segmentcount;
-  out = axom::Array<OctType>(totaloctcount, totaloctcount, allocId);
+  out.empty();
```
kennyweiss: Is this supposed to be `out.clear()` ?

src/axom/quest/detail/clipping/MonotonicZSORClipper.cpp
```diff
@@ -0,0 +1,738 @@
...
+    Point2DType vertOnXPlane {vert[1], vert[2]};
+    Point2DType vertOnRz {
+      vert[0],
+      std::sqrt(numerics::dot_product(vertOnXPlane.data(), vertOnXPlane.data(), 2))};
+    bbInRz.addPoint(vertOnRz);
+
+    double angle = atan2(vertOnXPlane[1], vertOnXPlane[0]);
+    minAngle = angle < minAngle ? angle : minAngle;
+    maxAngle = angle > minAngle ? angle : minAngle;
+  }
```
kennyweiss, line 302: Minor: You could use a 1D bounding box for this 1D range (`primal::BoundingBox<double, 1>`)

No need to change the code unless you want to.

src/axom/quest/detail/clipping/MonotonicZSORClipper.cpp
```diff
@@ -0,0 +1,738 @@
...
+  /*
+    The geometry can be closer to the rotation axis than its
+    individual vertices are, depending on the angle (about the axis)
+    between the vertices.  Given the angle, scale the bottom of bbInRz
+    for the worst case.
+  */
+  double angleRange = maxAngle - minAngle;
```
kennyweiss: Is there an assumption that the range won't include the x-axis? E.g. from -30 degrees to 30 degrees.

src/axom/quest/ShapeMesh.hpp
gunney1: Because this method is used in long inner loops I'd be concerned about constructing and destructing an `ArrayView` every cycle.  That cost can be optimized away in theory, but does it always happen?  My reference precedence for this is the `Hexahedron::triangulate` method, in which the `Tetrahedron` container is a `TetIndexable` template parameter.  It doesn't check the container size.

Moreover, verifying the `ArrayView` size here is no more than verifying that we claimed the correct size when constructing the `ArrayView`.  I'm not sure if that's very meaningful.

`hexToTets` is a private method.  If there isn't enough space for the output, it's not a user error.  It'd be an error in the short `computeCellsAsTetsImpl` method.

src/axom/quest/tests/quest_mesh_clipper.cpp
gunney1: Should all files be included in `quest.hpp`?  I assumed that some files are omitted, because I don't see any file that I placed in the `detail` subdirectory.

src/axom/quest/tests/quest_mesh_clipper.cpp
gunney1: I'll change the utility function name to `makeSorGeometry` and add a brief comment.

src/axom/quest/detail/Discretize_detail.hpp
gunney1: Fixed.

src/axom/quest/detail/clipping/MonotonicZSORClipper.cpp
gunney1: That's cool.  I'll do it.

src/axom/quest/detail/clipping/MonotonicZSORClipper.cpp
gunney1: No.  The x-axis (that is, the axis of rotation) is included when `angleRange >= M_PI`.  We account for this when we use `angleRange` to compute `factor`.

src/axom/quest/detail/clipping/MonotonicZSORClipper.hpp
```diff
@@ -0,0 +1,221 @@
...
+{
+
+/*!
+ * @brief Geometry clipping operations for simple 3D
+ * surface-of-revolution geometries.
+ *
+ * This implementation requires the SOR curve r(z) have monotonic (but
+ * not strictly monotonic) z coordinates.  The curve may have vertical
+ * slopes but it may not double back in z.  For SOR curves that double back,
+ * use SORClipper.
```
Arlie-Capps: Slight rephrase, for clarity:

```suggestion
 * @brief Geometry clipping operations for simple 3D
 * surface-of-revolution (SOR) geometries.
 *
 * This implementation requires the SOR curve r(z) have monotonic in z.
 * The values of z that make up the polyline specifying the curve must
 * either be non-decreasing or non-increasing.  A curve may have
 * segments with vertical slope but it may not double back in z.
 * For SOR curves that double back, use \a SORClipper.
```

src/axom/quest/detail/clipping/MonotonicZSORClipper.hpp
```diff
@@ -0,0 +1,221 @@
...
+  /*!
+   * @brief Constructor.
+   *
+   * @param [in] kGeom Describes the shape to place
+   *   into the mesh.
+   * @param [in] name To override the default strategy name
+   *
+   * \c kGeom.asHierarchy() must contain the following data:
+   * - sorOrigin: 3D coordinates of the point (z,r) = (0,0)
+   * - sorDirection: orientation of the SOR axis in 3D space.
```
Arlie-Capps: I don't like "orientation."

Suggestions:
I think `sorDirection` needs to be a triple that is a unit vector.  Is this correct?  If so, please say "unit vector" instead of "orientation."  If `sorDirection` doesn't need to be a unit vector, just any vector pointing along the SOR axis, that is helpful too: please say something like "sorDirection: a vector parallel to the SOR axis."

src/axom/quest/detail/clipping/MonotonicZSORClipper.hpp
```diff
@@ -0,0 +1,221 @@
...
+   *   to approximate the sphere with octahedra.  The number
+   *   of octs grows exponentially with level
+   *   (@see quest::discretize(const axom::ArrayView<Point2D> &polyline, int pointcount, int levels, axom::Array<OctType> &out, int &octcount)).
+   *   In practice, keep this number to 11 or less.
+   */
+  MonotonicZSORClipper(const klee::Geometry& kGeom, const std::string& name = "");
+
+  /*!
+   * @brief Construct with parameters to override the specificaitons
+   * in the klee::Geometry.
```
Arlie-Capps: typo for "specifications"

src/axom/quest/detail/clipping/MonotonicZSORClipper.cpp
```diff
@@ -0,0 +1,736 @@
...
+  axom::ArrayView<const Point2DType> pts)
+{
+  const axom::IndexType segCount = pts.size() - 1;
+  SLIC_ASSERT(segCount > 0);
+
+  // boundaryIdx is where curve's axial direction changes, plus end points.
+  axom::Array<axom::IndexType> boundaryIdx(0, 2);
+  boundaryIdx.push_back(0);
+
+  constexpr double eps = 1e-14;
```
Arlie-Capps, line 659: I'd rather have this user-accessible, but perhaps it doesn't need to be.  Not for this PR.  Noted in issue #1775 .

src/axom/quest/detail/clipping/MonotonicZSORClipper.hpp
gunney1: Got it.  Thanks!

src/axom/quest/detail/clipping/MonotonicZSORClipper.hpp
gunney1: Good suggestion.  Thanks!

src/axom/quest/detail/clipping/MonotonicZSORClipper.hpp
gunney1: Thank you.

src/axom/quest/detail/clipping/MonotonicZSORClipper.cpp
gunney1: Ok.  Thanks for the new issue.

src/axom/quest/detail/clipping/MonotonicZSORClipper.hpp
```diff
@@ -0,0 +1,222 @@
...
+
+/*!
+ * @brief Geometry clipping operations for simple 3D
+ * surface-of-revolution geometries.
+ *
+ * This implementation requires the SOR curve r(z) to be monotonic in z.
+ * The values of z that make up the polyline specifying the curve must
+ * be either non-decreasing or non-increasing.  A curve may have segments
+ * with vertical slope, but it may not double back in z.
+ * For SOR curves that double back, use \a SORClilpper.
```
Arlie-Capps: Typo: should be "SORClipper".

gunney1: Thanks.  Fixed.

PR #1767, opened 2026-01-11
Bugfix for batched insertion into core's FlatMap

kennyweiss:
# Summary

- This PR is a bugfix for batched insertion into a FlatMap that has some deleted elements
   - This test was occasionally failing our CI, even on CPU machines, so I decided to hunt down the cause
   - It turned out to be related to batched insertion into a FlatMap
   - We already had a test for this, so I modified it to run several times using a set of seeds that reproduced the failures
   - I can reliably reproduce the failure with
    ```
     > OMP_NUM_THREADS=32 AXOM_FLATMAP_TEST_STRESS_ITERS=2000 ./tests/core_serial_tests --gtest_filter=core_flatmap_for_all*
    ```
- My proposed fix is to rehash the FlatMap before batched insertion if there are deleted elements
   - It seems like there \*should\* be way to account for the deleted entries during the insertion
   - @publixsubfan -- if you think that would be a better idea and have time to hunt it down that'd likely be more performant than the proposed solution

#### Misc

- I also set an explicit number of threads in the core_serial tests that used OpenMP
- I renamed the `core_flatmap_for_all`  tests (from `core_flatmap_forall`) so they can be properly filtered w/ gtest

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -90,6 +90,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Optimized `Array::push_back()` and `Array::emplace_back()` operations.
 - Quest: In Shaping applications, we now check return code after attempting to load a mesh
   and throw an error for unsuccessful loads.
+- Core: Bugfix for batched insertion into a FlatMap with deleted entries
 
 ###  Deprecated
 


src/axom/core/tests/core_flatmap_for_all.hpp
```diff
@@ -504,118 +504,126 @@ AXOM_TYPED_TEST(core_flatmap_forall, insert_multiple_batch_with_dups)
...
   }
 
-  // Add a second set of duplicates.
-  // Since these ones are at the end of the sequence, they should be the ones
-  // actually written into the map.
-  std::shuffle(shuffled_indexes.begin(),
-               shuffled_indexes.end(),
-               std::mt19937 {std::random_device {}()});
-  for(int i = 0; i < NUM_ELEMS; i++)
+  for(int trial = 0; trial < num_trials; ++trial)
```
kennyweiss, line 526: Note -- the for loop indents all lines. It might be easier to review by telling github to hide whitespace changes (in the diff view settings).

src/axom/core/FlatMapUtil.hpp
```diff
@@ -267,12 +267,18 @@ void FlatMap<KeyType, ValueType, Hash>::insert(InputIt kv_begin, InputIt kv_end)
...
+  // Batched insertion assumes probing sequences are gap-free
+  // (i.e., there are no tombstones from prior erase() operations).
+  // When tombstones exist, the parallel insertion logic can mishandle duplicates
+  // under contention (e.g. OpenMP) and produce incorrect size/value results.
+  //
+  // If tombstones exist, rehash to compact the table and restore the invariants required by this algorithm.
+  if(this->m_loadCount != static_cast<std::uint64_t>(this->m_size))
+  {
+    this->rehash(this->m_size + num_elems);
+  }
```
kennyweiss, line 279: @publixsubfan -- is this a reasonable solution? 
Or would it be better to rework the insertion algorithm to account for deleted elements?

src/axom/core/tests/core_flatmap_for_all.hpp
```diff
@@ -504,118 +504,126 @@ AXOM_TYPED_TEST(core_flatmap_forall, insert_multiple_batch_with_dups)
...
-  MapType test_map_gpu(test_map, axom::Allocator {this->getKernelAllocatorID()});
-
-  axom::Array<std::pair<int, double>> second_batch_pairs(NUM_ELEMS * 2);
-
-  // Add some duplicate key values through the batched interface.
-  std::shuffle(shuffled_indexes.begin(),
-               shuffled_indexes.end(),
-               std::mt19937 {std::random_device {}()});
-  for(int i = 0; i < NUM_ELEMS; i++)
+  // Repeat the test as necessary to catch scheduling dependent regression
```
kennyweiss, line 519: I mentioned it in my PR summary, but it's worth repeating in context. 
I was able to reliably reproduce the error before the bugfix by running this test as
```
OMP_NUM_THREADS=32 AXOM_FLATMAP_TEST_STRESS_ITERS=2000 ./tests/core_serial_tests --gtest_filter=core_flatmap_for_all*
```

I was unable to reproduce this error after the bugfix.

Arlie-Capps (approved):

src/axom/core/FlatMapUtil.hpp
publixsubfan: That's a perfectly fine solution from my perspective. I'll try to chase down the underlying issue at some point in the future.

Issue #1685, opened 2025-10-06
Add slic macros that only log the first time they're called


kennyweiss:
A discussion in https://github.com/LLNL/axom/pull/1683 prompted the question about adding slic macros that only log/warn the first time they're encountered.

I'd suggest adopted something like the following (based on code that @tomstitt wrote a few years ago):
```cpp
#define SLIC_INFO_ONCE_IF(cond, msg) \
   do                               \
   {                                \
      static bool once = true;      \
      if (cond)                     \
      {                             \
         if (once)                  \
         {                          \
            SLIC_INFO(msg);          \
            once = false;           \
         }                          \
      }                             \
   } while (false)

#define SLIC_INFO_ONCE(msg) SLIC_INFO_ONCE_IF(true, msg)

#define SLIC_INFO_ONCE_ROOT_IF(cond, msg) \
   do                                    \
   {                                     \
      static bool once = true;           \
      if (cond)                          \
      {                                  \
         if (once)                       \
         {                               \
            SLIC_INFO_ROOT(msg);          \
            once = false;                \
         }                               \
      }                                  \
   } while (false)

#define SLIC_INFO_ONCE_ROOT(msg) SLIC_INFO_ONCE_ROOT_IF(true, msg)
```

We should also add similar macros for warnings and possibly debug.

jcs15c:
When implemented, these methods will be particularly useful in

- [ ] `NURBSPatch.hpp` during parameter space evaluation methods.
- [ ] `intersect_patch_impl.hpp` if excessive intersections are recorded.

bmhan12:
Done in #1831.

PR #1768, opened 2026-01-12
Adds 3D winding number example to quest

kennyweiss:
# Summary

- This PR is a feature
- It adds a quest example for running 3D winding number queries on a STEP file
- It is modeled after the 2D winding number example, and we might eventually want to merge the two examples. In the short term, during development, it seemed easiest to keep them separate.
- I cherry-picked a few commits from @jcs15c's #1765 (all the changes in this PR to `STEPReader.cpp`), so that should be merged first
- Several unnecessarily intrusive SLIC_ASSERTS were removed/made more informative in `NURBSPatch.hpp` to accommodate the fact that many in-the-wild CAD examples have trimming curves which extend beyond the range of the patch's parameter space. In the future, these could be handled by the methods described in #1685, which would alert the user without cluttering output.
- Includes some performance and robustness improvements to 2D winding number computations
   - `linear_winding_number` is now based on `atan2` instead of `acos`. This was the bottleneck in 2D WN computations and shaves about 25% of runtimes
   - determinants are now implemented in terms of a portable `axom::utilities::fma` function (with the expectation that this will improve robustness)
- Adds some helpful features to 2D winding number example, e.g. using the curve collection bounding box when not specified by the user, and output some helpful runtime stats about query times and rates

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -145,7 +145,7 @@ if(AXOM_ENABLE_TOOLS)
 endif()
 
 # Installation examples ('Using Axom') and/or tutorials
-if (AXOM_ENABLE_EXAMPLES OR AXOM_ENABLE_TUTORIALS)
+if(AXOM_ENABLE_EXAMPLES OR AXOM_ENABLE_TUTORIALS)
     add_subdirectory(examples)
 endif()
 
--- a/src/axom/CMakeLists.txt
+++ b/src/axom/CMakeLists.txt
@@ -11,7 +11,7 @@
 
 # Core is an essential part of Axom and cannot be turned off
 axom_add_component(COMPONENT_NAME core DEFAULT_STATE ON)
-if(DEFINED AXOM_ENABLE_CORE)
+if(DEFINED AXOM_ENABLE_CORE AND NOT AXOM_ENABLE_CORE)
     message(FATAL_ERROR "Core is an essential part of Axom and cannot be turned off")
 endif()
 


src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -2163,11 +2163,23 @@ class NURBSPatch
...
+                        getMinKnot_u(),
+                        getMaxKnot_u(),
+                        1e-5));
+    SLIC_ASSERT_MSG(
+      isValidParameter_v(v, 1e-5),
+      axom::fmt::format("Requested v-parameter {} is outside valid range [{},{}] with tolerance {}",
+                        v,
+                        getMinKnot_v(),
+                        getMaxKnot_v(),
+                        1e-5));
```
kennyweiss: @jcs15c and I spoke about this -- this is triggered called when evaluating the normal for a trimmed patch, and was failing for some files due to trimming curves that are slightly outside the parametric bounding box. Since the user has no control over this, we should probably preprocess/fix this before processing the NURBS patches,

src/axom/primal/operators/detail/winding_number_3d_impl.hpp
```diff
@@ -179,14 +179,14 @@ double stokes_gwn_adaptive(const Point<T, 3>& query,
...
-                                                                    refinement_level + 1,
-                                                                    2 * refinement_index + 1);
+  const auto& trimming_curve_data_1 = nurbs.getTrimmingCurveQuadratureData(curve_index,
+                                                                           quad_npts,
+                                                                           refinement_level + 1,
+                                                                           2 * refinement_index);
+  const auto& trimming_curve_data_2 = nurbs.getTrimmingCurveQuadratureData(curve_index,
+                                                                           quad_npts,
+                                                                           refinement_level + 1,
+                                                                           2 * refinement_index + 1);
```
kennyweiss, line 190: As far as I can tell, this avoids unnecessary copies.
@jcs15c -- please correct me if I'm looking at this wrong.

src/axom/quest/io/STEPReader.cpp
kennyweiss, line 1: Note: Changes to this file were cherry-picked from @jcs15c's commits in #1765 which should probably be merged first.

src/axom/primal/geometry/NURBSPatch.hpp
jcs15c: To further keep record of this conversation, at this point it's unrealistic to clip or clamp the trimming curves to the correct parameter space for a given patch. Instead, the methods in `NURBSPatch` (which already clamp the inputs to the parameter space) should probably give a one-time warning if out-of-bounds input is recorded, then move on. Since #1765 is mostly concerned with the `STEPReader`, I'll make those changes here instead.

src/axom/primal/operators/detail/winding_number_3d_impl.hpp
jcs15c: It does, thanks for the catch!

src/axom/primal/tests/primal_winding_number.cpp
```diff
@@ -480,6 +481,28 @@ TEST(primal_winding_number, nurbs_winding_numbers)
...
+                "NURBSPatchGWNCache::getWeights() must match NURBSPatch signature");
+  static_assert(std::is_same_v<decltype(std::declval<const Cache&>().getKnots_u()),
+                               decltype(std::declval<const Patch&>().getKnots_u())>,
+                "NURBSPatchGWNCache::getKnots_u() must match NURBSPatch signature");
+  static_assert(std::is_same_v<decltype(std::declval<const Cache&>().getKnots_v()),
+                               decltype(std::declval<const Patch&>().getKnots_v())>,
+                "NURBSPatchGWNCache::getKnots_v() must match NURBSPatch signature");
+  static_assert(std::is_same_v<decltype(std::declval<const Cache&>().getTrimmingCurves()),
+                               decltype(std::declval<const Patch&>().getTrimmingCurves())>,
+                "NURBSPatchGWNCache::getTrimmingCurves() must match NURBSPatch signature");
```
jcs15c, line 503: These are great tests, thank you! It's worth keeping in mind that a future implementation of `NURBSPatchGWNCache` might have trimming curves of type `NURBSCurveGWNCache` instead of `NURBSCurve`, since that seems easier than templating the type of `NURBSPatch` on trimming curve type.

src/axom/core/utilities/Utilities.hpp
```diff
@@ -109,6 +109,22 @@ inline AXOM_HOST_DEVICE void swap(T& a, T& b)
...
 }
 
+/*!
+ * \brief Returns the fused-multiply-add of a, b, and c
+ * \accelerated
+ * \param [in] a,b,c the values
+ * \return a*b+c computed using an fma operation
+ */
+template <typename T>
+inline AXOM_HOST_DEVICE T fma(const T& a, const T& b, const T& c)
```
BradWhitlock, line 119: I can think of a bunch of places to use this!

BradWhitlock (approved):
Nice optimizations!

PR #1769, opened 2026-01-12
Spack 1.1 &  TPL v2025.12.0 updates

bmhan12:
This PR:

- Updates Spack to 1.1
- Updates the RAJA suite of packages to v2025.12.x
  - Updates Docker images

TODO:
- [x] Fix Windows x64 builds (fails to compile with RAJA)
  - ~Latest attempt: https://github.com/llnl/axom/actions/runs/20835993048~
  - Bypass error by disabling openmp. Latest run: https://github.com/llnl/axom/actions/runs/21222202624


scripts/spack/packages/axom/package.py
```diff
@@ -434,7 +431,7 @@ def initconfig_hardware_entries(self):
 
             hip_link_flags = ""
 
-            rocm_root = os.path.dirname(spec["llvm-amdgpu"].prefix)
+            rocm_root = spec["llvm-amdgpu"].prefix
```
bmhan12, line 435: Here and in `toss_4_x86_64_ib_cray/spack.yaml`, I incorporated changes from @adrienbernede 's PR #1739 to get things working with Spack 1.1 .

scripts/vcpkg_ports/raja/vcpkg_raja_openmp.patch
```diff
@@ -0,0 +1,16 @@
...
+--- a/include/RAJA/policy/openmp/thread.hpp
++++ b/include/RAJA/policy/openmp/thread.hpp
+@@ -22,6 +22,11 @@
+ 
+ #if defined(RAJA_OPENMP_ACTIVE)
+ 
++// Include OpenMP header for Windows
++#if defined(_OPENMP)
++#include <omp.h>
++#endif
```
bmhan12: This patch file is part of my attempt to get raja v25.12.0 working with Windows.

Following error without this patch:
```
D:\a\axom\axom\uberenv_libs\vcpkg\buildtrees\raja\src\v2025.12.0-beae201c76.clean\include\RAJA/policy/openmp/thread.hpp(35): error C3861: 'omp_get_max_threads': identifier not found
D:\a\axom\axom\uberenv_libs\vcpkg\buildtrees\raja\src\v2025.12.0-beae201c76.clean\include\RAJA/policy/openmp/thread.hpp(41): error C3861: 'omp_get_thread_num': identifier not found
```

bmhan12:
> @bmhan12 to be clear: it works with RAJA 2025.09.0 and breaks with RAJA 2025.09.1 and newer?

Yes.

scripts/vcpkg_ports/axom/vcpkg.json
```diff
@@ -4,11 +4,11 @@
   "homepage": "https://github.com/llnl/axom",
   "description": "Host-config generator for Axom TPLs",
   "dependencies": ["blt"],
+  "$comment":"OpenMP removed from default-features due to RAJA compilation errors, umpire removed from default-features due to fmt link errors",
   "default-features": [
     "conduit",
     "lua",
     "mfem",
-    "openmp",
```
bmhan12, line 11: After talking with @kennyweiss, I decided we can disable OpenMP on Windows for now, and can enable it again at a later date. 

It was also mentioned that Umpire is actually disabled in our Windows testing, I left a comment in this file to track these changes.

Tag: @johnbowen42

Arlie-Capps (approved):

scripts/spack/configs/versions.yaml
```diff
@@ -28,12 +29,14 @@ packages:
...
-      - spec: "@2025.09.0"
+      - spec: "@git.e827035c630e71a9358e2f21c2f3cf6fd5fb6605=develop"
   scr:
     require:
       - spec: "@3.0.1"
+  # v2025.12.0
   umpire:
     require:
-      - spec: "@2025.09.0"
+      - spec: "@git.0372fbd6e1f17d7e6dd72693f8b857f3ec7559e9=develop"
```
adrienbernede, line 42: Why use the commit here compared to the version tag ?

scripts/spack/configs/toss_4_x86_64_ib/spack.yaml
```diff
@@ -143,11 +143,11 @@ spack:
     mvapich2:
       buildable: false
       externals:
-      - spec: mvapich2@2.3.7.clang_19~alloca~cuda~debug~hwloc_graphics~hwlocv2+regcache+wrapperrpath build_system=autotools ch3_rank_bits=32 fabrics=mrail file_systems=auto process_managers=slurm threads=multiple
+      - spec: mvapich2@2.3.7.clang_19~alloca~cuda~debug~hwloc_graphics~hwlocv2+regcache+wrapperrpath ch3_rank_bits=32 fabrics=mrail file_systems=auto process_managers=slurm threads=multiple
```
adrienbernede, line 147: I'm curious how you verify those variants for external installs. Are they provided by the spack external detection tool ?

scripts/spack/configs/versions.yaml
bmhan12: 2025.12.0 isn't in upstream spack yet, so I went with the commit hash.

scripts/spack/configs/toss_4_x86_64_ib/spack.yaml
bmhan12: Yes, that is where I originally got all these variants, using spack to detect them. 
Turns out this is probably not the best idea, since the packages and detection can change between Spack versions.

white238: You actually don't need all the variants anymore. It was needed when you couldn't specify which mvapich you needed per compiler before toolchains. Basically we were forcing the concretizer to select the right one. You can do this now:

https://github.com/llnl/smith/blob/c00589ac9cb73deb2fe4d810f3f98b63aa2948e6/scripts/spack/configs/toss_4_x86_64_ib/spack.yaml#L94-L100

https://github.com/llnl/smith/blob/c00589ac9cb73deb2fe4d810f3f98b63aa2948e6/scripts/spack/configs/toss_4_x86_64_ib/spack.yaml#L28-L29

cyrush: @white238 so is just the version sufficient now for externals?

white238: yes, as long as you specify it in the toolchain

rhornung67 (commented):
Thank you for working through this stuff @bmhan12 LGTM!

PR #1771, opened 2026-01-14
Tetrahedral mesh implementation for `MeshClipper`

gunney1:
# Summary

- This PR is a feature
- It does the following:
  - Add `TetMeshClipper` implementation for `MeshClipper`.  This supports shapes defined by tetrahedral mesh, including ProE shapes.
  - Add tests for `TetMeshClipper`, with 2 configurations: a simple 3-tet mesh and the mesh in `cup.proe`.

This is the last of the supported geometries for `MeshClipper`.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -45,7 +45,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Adds spack variants for `adiak` and `caliper`. These replace the previous `profiling` variant which enabled both at the same time.
 - Adds the `AXOM_TEST_NUM_OMP_THREADS` configuration variable to control the default OpenMP thread count for tests.
 - Slic: Adds `slic::getAbortFunction()` to return a function pointer to the abort handler called during logging functions
-
+- Quest: Adds a 3D winding number example and impoves WN performance in 2D and 3D
 
 ###  Changed
 - Version of `quest::discretize` that approximates a surface-of-revolution from a polyline
@@ -86,8 +86,9 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   contains additional algorithms.
 - Updates to [mfem version 4.9](https://github.com/mfem/mfem/releases/tag/v4.9).
   Also updates mfem's hypre dependency to [hypre version 2.27](https://github.com/hypre-space/hypre/releases/tag/v2.27.0).
-- Quest: MFEMReader can now read in multispan 1D NURBS meshes
+- Quest: `MFEMReader` can now read in multispan 1D NURBS meshes
 - fmt: Axom's built-in version of the fmt formatting library was updated to [fmt version 12.1.0](https://github.com/fmtlib/fmt/releases/tag/12.1.0).
+- Primal: Improves robustness of `linear_winding_number` by using `atan2` instead of `acos`
 
 ###  Fixed
 - Sina's Fortran tests are now running (instead of silently failing)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -145,7 +145,7 @@ if(AXOM_ENABLE_TOOLS)
 endif()
 
 # Installation examples ('Using Axom') and/or tutorials
-if (AXOM_ENABLE_EXAMPLES OR AXOM_ENABLE_TUTORIALS)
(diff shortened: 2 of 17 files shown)


src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -1015,10 +1015,10 @@ class MeshClipperImpl : public MeshClipper::Impl
...
         pieceBbsView.size(),
         AXOM_LAMBDA(axom::IndexType i) {
-          pieceBbsView[i] = primal::compute_bounding_box<double, 3>(geomTetsView[i]);
-#if defined(AXOM_DEBUG)
-          SLIC_ASSERT(!geomTetsView[i].degenerate());
-#endif
+          if(!geomTetsView[i].degenerate())
+          {
+            pieceBbsView[i] = primal::compute_bounding_box<double, 3>(geomTetsView[i]);
+          }
```
kennyweiss, line 1021: Would it make sense to filter out degenerate tets before running the algorithm?

src/axom/quest/detail/clipping/TetMeshClipper.cpp
```diff
@@ -0,0 +1,981 @@
...
+ *   the center of m_tetMeshBb.
+ * - Use BVH::findBoundingBoxes and BVH::findRay to get surface
+ *   triangles near the bounding boxes and rays.
+ * - Loop through the hexes.
+ *   - If hex bb is near any surface triangle bb, label the hex ON.
+ *   - Else, the hex is either IN or OUT.  It can't possibly be ON.
+ *     Count number of surface triangles that the hex's ray intersects,
+ *     @see intersect(const Triangle<T, 3>& tri, const Ray<T, 3>& ray)
+ *     If the count is odd, hex is IN, if even, OUT.
+ */
```
kennyweiss, line 123: Can you please clarify this algorithm?

For each hex, it has a ray from the hex center and it's counting the intersections from the center to each triangle that it encounters within the hex bounding box and returning the parity?

Did I get that right?

Can you post some performance stats for this? Specifically, how expensive is this step (+ the query) vs. a simpler strategy of returning IN if the bounding box of a tet intersects any tets, and returning OUT otherwise?  Or perhaps just checking the bounding box of each hex w/ the bounding boxes of the tets?

kennyweiss (approved):
Thanks @gunney1

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: Yes, that would make sense, and I originally had a `SLIC_ERROR_IF` there to catch it.  But as a user who is given an array of tets generated elsewhere, I found it annoying to have to find and filter out the degenerate ones.  The check here makes Axom more robust at a negligible cost.

src/axom/quest/detail/clipping/TetMeshClipper.cpp
```diff
@@ -0,0 +1,981 @@
...
+
+  copy_hierarchy_with_array_allocator(m_tetMesh, oldTopoPath, newTopoPath, allocId);
+  copy_hierarchy_with_array_allocator(m_tetMesh, oldCoordsetPath, newCoordsetPath, allocId);
+
+  const std::string newCoordsetName = axom::fmt::format("{}.{}", m_coordsetName, allocId);
+  m_tetMesh[newTopoPath + "/coordset"].set(newCoordsetName);
+}
+
+// Run cellKernel through a cell loop.
+// Run faceKernel through a face loop.
```
bmhan12: placeholder comments?

src/axom/quest/detail/clipping/TetMeshClipper.cpp
gunney1: You got that right.  I will rewrite the comments along those lines.  I plan to post performance stats for all the geometries soon.  From what I've seen in my toy test configuration, this approach of counting intersections with the boundaries is not expensive.  I've also seen that not identifying as many IN tets as possible will degrade fast with mesh resolution, because there will be many that are IN.

src/axom/quest/detail/clipping/TetMeshClipper.cpp
gunney1: Yes.  Will remove.  Thanks.

PR #1773, opened 2026-01-14
Finish refactoring cell clipper methods

gunney1:
# Summary

- This PR is a refactoring
- It does the following:
  - Re-factors the two `computeClipVolumes3D` methods, taking advantage of the smaller methods created when I re-factored `computeClipVolumes3DTet` (see PR https://github.com/llnl/axom/pull/1756).

The relevant smaller methods are those that get discrete parts of the geometry into BVHs and clipping them against discrete parts of the mesh.

This PR, along with https://github.com/llnl/axom/pull/1756, resolves issue https://github.com/llnl/axom/issues/1704.

This PR does NOT additionally implement the 2-pass candidate search.  The 2-pass still is only used by `computeClipVolumes3DTet` at this point.  I plan to implement that in another PR in this series.

### DIFF
--- a/src/axom/quest/detail/clipping/MeshClipperImpl.hpp
+++ b/src/axom/quest/detail/clipping/MeshClipperImpl.hpp
@@ -607,7 +607,7 @@ public:
   }  // end of computeClipVolumes3D() function
 
   /*
-   * Clip tets of from the mesh with tets or octs from the clipping
+   * Clip tets from the mesh with tets or octs from the clipping
    * geometry.  This implementation is like the two above except that
    * it limits clipping to a subset of mesh tets labeled as
    * potentially on the boundary.


src/axom/quest/MeshClipperStrategy.hpp
```diff
@@ -343,7 +343,7 @@ class MeshClipperStrategy
    * output vertex coordinate that is close to zero.
    *
    * Degenerate tets are allowed, but they may cause a reduction in
-   * clipping performance.
+   * clipping performance, so it's best to omit them.
```
kennyweiss, line 346: Please clarify "it's best to omit them" -- does that mean the user should omit them before calling the clipping algorithm? Or that the clipping algorithm filters them?

gunney1: I will clarify with "omit them from the output".

kennyweiss: I think I see the disconnect. This comment appears to be directed at implementers of this interface rather than at users of the interface. (The previous paragraph makes this more explicit).

It might be clearer to make this and the previous paragraph a  `@note` and call out that this is for developers.

Also, something like "..., so it's best to filter them out" might be clearer.

PR #1776, opened 2026-01-15
Remove unused CMake variable AXOM_ENABLE_CORE

bmhan12:
This PR:
- Removes unused CMake variable `AXOM_ENABLE_CORE`

This variable was noticed by @johnbowen42 while testing Axom builds.
~The variable is only defined here and not used anywhere else in the CMake build.~
The variable is only mentioned here and not anywhere else in the CMake build.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -145,7 +145,7 @@ if(AXOM_ENABLE_TOOLS)
 endif()
 
 # Installation examples ('Using Axom') and/or tutorials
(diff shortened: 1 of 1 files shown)


white238 (changes requested):
Update if to include allowing users to set it to ON, though mildly pointless. Dont remove helpful error message.

rhornung67:
> Update if to include allowing users to set it to ON, though mildly pointless. Dont remove helpful error message.

Pointless, yes. But some people may find it cathartic.... 😄

PR #1777, opened 2026-01-17
Performance improvements to 3D winding number query

kennyweiss:
# Summary

- This branch contains a slew of optimizations and performance improvements to the 3D winding number query
- Speeds up 3D winding number evaluations on trimmed NURBS patches by reducing per-query overhead in the NURBS derivative and Stokes-integral hot paths.
- Adds `KnotVector::DerivativeBasisWorkspace` struct and workspace-based `derivativeBasisFunctionBySpan()` to avoid repeated allocations
- Adds an optimized specialization for first derivatives of NURBS patches in `NURBSPatch::evaluateFirstDerivatives()` -- a hotspot for WN evaluation
- Improves 3D WN memoization via improved cache lookup and avoiding repeated scaling of weights
- Enhances svg2contours.py script to report stats and bounding box

## Performance vs. develop
  - Command: quest_winding_number_3d_ex -i data/quest/step/gear.step --no-vis query_mesh --res 12 12 12 (RelWithDebInfo, -DAXOM_DEBUG_DEFINE=OFF, -march=native)
  - develop: preprocess 24.090s, query 52.372s (23.84 ms/query)
  - this branch: preprocess 3.773s, query 29.164s (13.2745 ms/query)
  - Speedup: preprocess 6.38x, query 1.80x

### DIFF
--- a/src/axom/primal/geometry/NURBSPatch.hpp
+++ b/src/axom/primal/geometry/NURBSPatch.hpp
@@ -1688,7 +1688,7 @@ public:
       {
         for(int j = 0; j < m; ++j)
         {
-          min_weight = std::min(min_weight, m_weights(i, j));
+          min_weight = axom::utilities::min(min_weight, m_weights(i, j));
         }
       }
 
--- a/host-configs/matrix-toss_4_x86_64_ib-llvm@19.1.3_cuda.cmake
+++ b/host-configs/matrix-toss_4_x86_64_ib-llvm@19.1.3_cuda.cmake
@@ -83,7 +83,7 @@ set(ENABLE_CUDA ON CACHE BOOL "")
 
 set(CMAKE_CUDA_SEPARABLE_COMPILATION ON CACHE BOOL "")
 
-set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -restrict --expt-extended-lambda " CACHE STRING "" FORCE)
+set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -restrict --expt-extended-lambda --expt-relaxed-constexpr" CACHE STRING "" FORCE)
 
 # nvcc does not like gtest's 'pthreads' flag
 
--- a/src/axom/bump/data/MeshTester.cpp
+++ b/src/axom/bump/data/MeshTester.cpp
@@ -421,21 +421,21 @@ static void addCircleMaterial(const TopoView& topoView,
   center[1] = circleCenter[1];
 
   const TopoView deviceTopologyView(topoView);
-  axom::for_all<axom::SEQ_EXEC>(
-    topoView.numberOfZones(),
-    AXOM_LAMBDA(axom::IndexType zoneIndex) {
-      const auto zone = deviceTopologyView.zone(zoneIndex);
-      // NOTE: node ordering shuffled because function takes different node ordering.
-      auto vf = calculatePercentOverlapMonteCarlo(numSamples,
-                                                  center,
-                                                  circleRadius,
-                                                  coordsetView[zone.getId(3)],
-                                                  coordsetView[zone.getId(0)],
-                                                  coordsetView[zone.getId(1)],
-                                                  coordsetView[zone.getId(2)]);
-      greenView[zoneIndex] = vf;
-      blueView[zoneIndex] = 1.0 - vf;
-    });
+  axom::for_all<axom::SEQ_EXEC>(topoView.numberOfZones(),
+                                [=] /*AXOM_LAMBDA*/ (axom::IndexType zoneIndex) {
+                                  const auto zone = deviceTopologyView.zone(zoneIndex);
+                                  // NOTE: node ordering shuffled because function takes different node ordering.
+                                  auto vf =
+                                    calculatePercentOverlapMonteCarlo(numSamples,
+                                                                      center,
+                                                                      circleRadius,
+                                                                      coordsetView[zone.getId(3)],
+                                                                      coordsetView[zone.getId(0)],
+                                                                      coordsetView[zone.getId(1)],
+                                                                      coordsetView[zone.getId(2)]);
+                                  greenView[zoneIndex] = vf;
+                                  blueView[zoneIndex] = 1.0 - vf;
+                                });
 
   // Figure out the material buffers from the volume fractions.
   std::vector<int> material_ids, sizes, offsets, indices;
--- a/src/axom/bump/tests/blueprint_testing_helpers.hpp
+++ b/src/axom/bump/tests/blueprint_testing_helpers.hpp
@@ -12,8 +12,11 @@
 #include <conduit/conduit_relay_io.hpp>
 #include <conduit/conduit_relay_io_blueprint.hpp>
 #include <algorithm>
+#include <cstdlib>
 #include <string>
 #include <vector>
+#include <chrono>
+#include <thread>
 
 //------------------------------------------------------------------------------
 // clang-format off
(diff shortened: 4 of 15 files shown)


src/axom/quest/examples/CMakeLists.txt
```diff
@@ -778,14 +778,113 @@ if(MFEM_FOUND AND OPENCASCADE_FOUND)
...
     if(AXOM_ENABLE_TESTS AND AXOM_DATA_DIR)
-        axom_add_test(
-            NAME quest_winding_number_3d_tet_slice
-            COMMAND quest_winding_number_3d_ex
-                    --input ${quest_data_dir}/step/tet.step
-                    --output-prefix wn3d_tet_slice
-                    query_mesh --min -0.5 -0.5 --max 0.5 0.5 --res 3 3 --slice-z 0.0 --order 1)
-        set_tests_properties(quest_winding_number_3d_tet_slice PROPERTIES
-            PASS_REGULAR_EXPRESSION  "4 patches .* 12 trimming curves")
+        # 3D query regression-style runs (20^3 grid) over several STEP models.
```
kennyweiss: I added a bunch of tests to track performance and correctness as I worked through these optimizations.
We should not merge this into develop as is since it'll be too slow for debug configs.

Before merging, I'll update the tests to run at a reasonable resolution for normal testing.

src/axom/primal/operators/detail/winding_number_3d_memoization.hpp
```diff
@@ -232,19 +247,20 @@ class NURBSPatchGWNCache
...
-                                                                                   curveIndex,
-                                                                                   quadNPts,
-                                                                                   refinementLevel,
-                                                                                   refinementIndex);
-    }
-
-    return m_curveQuadratureMaps[curveIndex][hash_key];
+    // Cache quadrature data per trimming curve keyed by (refinementLevel, refinementIndex).
+    // Note: `quadNPts` is fixed in the 3D winding-number implementation (currently 15).
+    const auto make_key = [](int level, int index) -> std::uint64_t {
```
BradWhitlock, line 253: These map enhancements are related to things I did in my branch. I combined the curveIndex, refinementLevel and refinementIndex into a single std::uint64_t and ditched the axom::Array too. It was worth it in my timings, and it sped up creating the caches on multiple threads.

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -757,6 +757,273 @@ class KnotVector
     return ders;
   }
 
+  struct DerivativeBasisWorkspace
+  {
+    axom::Array<T> ders;
```
BradWhitlock, line 773: There are up to 5 array allocations here. Based on how these are used, it looks like they could be replaced with views that point into a single axom::Array that contains the storage for all of the arrays. On thinking about this some more, it looks like they probably do not resize that often.

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -2309,12 +2313,111 @@ class NURBSPatch
...
-    Dv = ders[0][1];
+    const int deg_u = getDegree_u();
+    const int deg_v = getDegree_v();
+
+    const int du = axom::utilities::min(1, deg_u);
+    const int dv = axom::utilities::min(1, deg_v);
+
+    const bool isRationalPatch = isRational();
+
+    const auto span_u = m_knotvec_u.findSpan(u);
```
jcs15c, line 2362: Is this setup for `evaluateFirstDerivatives` a good enough improvement that it should be mirrored for `evaluateLinearDerivatives` too? Or at least the `du` and `dv` methods? I assume that this method gets called the most out of these.

src/axom/quest/examples/quest_winding_number_3d.cpp
```diff
@@ -123,6 +185,7 @@ int main(int argc, char** argv)
   bool memoized {true};
   bool vis {true};
   bool validate {false};
+  bool stats {true};
```
jcs15c: This is mostly just personal preference, but does `true` have to be the default for stats? I'm open to suggestions, but I don't know if any of the fields themselves have a useful meaning/interpretation, besides being a nice sanity check if there's a catastrophic error.

kennyweiss: done.

src/axom/primal/geometry/NURBSPatch.hpp
kennyweiss: I refactored it so the implementation is in `evaluatedDerivatives()` and both `evaluateFirstDerivative()` and `evaluateLinearDerivative()` get this.

kennyweiss: Actually, `evaluateLinearDerivative` needed partials, so it needed `d=2`, rather than `d=1`

jcs15c: I think I've actually come around on your original implementation here. It makes sense to have a specialization for just getting the `d == 1` derivatives, since those are (at the moment) the only ones that get used anywhere, and `evaluateFirstDerivatives` makes the most sense as to where those should go. 

Furthermore, the reason `evaluateLinearDerivatives` didn't work was _because_ it went into this specialization, which doesn't evaluate the mixed derivative. In general, you don't need to evaluate second derivatives to get `dudv`. Since I wouldn't expect the mixed derivatives to be used nearly as often as the two first-order derivatives, it'd be a waste to have `evaluateDerivatives` calculate them every time `d == 1`. If there comes a time where we actually call `du`, `dv`, or `evaluateLinearDerivatives()` outside of testing, maybe we can come up with a more unifying solution, but this works for now. 

Sorry to request the change only to undo it!

kennyweiss: No worries -- I'll revert that change

jcs15c: In looking at this, I realized that the default behavior of the NURBS Book algorithm computes derivatives of maximum _total_ order `d`, so `evaluateDerivatives(d==1)` was never designed to compute a `DuDv` derivative. Since the function returns a matrix, I think it makes sense for that matrix to be fully populated by default, i.e. computing derivatives of maximum _component_ order `d`. Technically, if you're searching for a specific mixed partial derivative, it's more efficient to do it with maximum component order instead of maximum, since you don't need to compute as many basis functions. 

I've refactored this function with an extra argument that clarifies this distinction, with the default being maximum component order.

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -2170,24 +2170,31 @@ class NURBSPatch
...
    * \pre Requires \a u, v be in the span of the knots (up to a small tolerance)
    * 
    * \note If u/v is outside the knot span up this tolerance, it is clamped to the span
    */
-  void evaluateDerivatives(T u, T v, int d, axom::Array<VectorType, 2>& ders) const
+  void evaluateDerivatives(T u,
+                           T v,
+                           int d,
+                           axom::Array<VectorType, 2>& ders,
+                           bool evalByMaxOrder = false) const
```
kennyweiss: If we're changing the signature of a public function, we should note this in the `RELEASE_NOTES`.

It's not entirely clear (to me) from the doxygen comment what this function will generate based on the `d` parameter and the new `evalByMaxOrder` parameter.

Since we're already changing the public signature, would it be clearer to replace these two parameters with a (bitwise) enum that makes this explicit? 

e.g.  
```cpp
enum DerivsToEvaluate {
   du  = 1 <<0,
   dv   = 1 <<1,
   dudv = 1 <<2,
   d2u   = 1<< 3,
    ....
}
```

This might allow us to share/deduplicate more of the implementations

jcs15c: The default argument `evalByMaxOrder = false` reflects what the function was supposed to be doing before (in my opinion) but was actually failing to do correctly, so relative to develop I think this change is more of a bugfix than anything. 

I'm not sure about an enum, since in principle this is the function that allows evaluation of derivatives of any order (in contrast to what are essentially aliases in `du`, `evaluateSecondDerivatives`, etc.). Since there's only two fundamental modes (maximum order and total order), maybe there's a solution which moves `evaluateDerivatives` to `private` and is called by two new functions:
`evaluateDerivativesByTotalOrder(T u, T v, int ord, axom::Array<VectorType, 2>& ders)` (the behavior described in the NURBS Book) which for `ord == 2` produces a 3x3 array with derivatives
```
       | eval , S_u  , S_uu |
ders = | S_v  , S_uv ,    0 |
       | S_vv ,    0 ,    0 |
```
 and `evaluateDerivativesByMaxOrder(T u, T v, int ord, axom::Array<VectorType, 2>& ders)` (what I thought the code was doing all along) which for `ord == 2` produces a 3x3 array with derivatives
```
       | eval ,  S_u  ,   S_uu |
ders = | S_v  ,  S_uv ,  S_uuv |
       | S_vv , S_vvu , S_uuvv |
```
I think we should expose some sort of distinction between them, as the former method is more efficient for evaluating, say, `S_{u^10}` and the latter is more efficient for evaluating `S_{u^5, v^5}`. I have no idea why anyone would want to evaluate such high order derivatives in the first place, but the comparative advantage is still there if you want `S_u` vs `S_uv` too.

kennyweiss: Interesting -- in that case, let's keep your parameters, and add the above ascii art to the doxygen comment to better illustrate what we're computing.

Arlie-Capps: @jcs15c , please do note this in the "Fixed" section of the `RELEASE_NOTES`.  You're fixing the function's behavior to what should have been.

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -2469,7 +2606,7 @@ class NURBSPatch
   VectorType dudv(T u, T v) const
   {
     axom::Array<VectorType, 2> ders;
-    evaluateDerivatives(u, v, 2, ders);
+    evaluateDerivatives(u, v, 1, der s);
```
kennyweiss:
```suggestion
    evaluateDerivatives(u, v, 1, ders);
```

PR #1778, opened 2026-01-21
Finish 2-pass candidate search

gunney1:
# Summary

- This PR is a refactoring
- It does the following:
  - Changes the two `computeClipVolumes3D` methods to call the workhorse `computeClipVolumes3DTet`, which implements the two-pass candidate search (see PR https://github.com/llnl/axom/pull/1756).  This is facilitated by two temporary index lists that `computeClipVolumes3DTet` require.
  - Adds a few comments and error checks that didn't make it into previous PRs in this series.

This PR, along with https://github.com/llnl/axom/pull/1756, resolves issue https://github.com/llnl/axom/issues/1704.

# Performance data
This should be the last of the performance-related changes in the mesh clipper, so here are some performance data.

## Test problem
I collected clipping times using a 100^3 cubic computational mesh.  Each test geometry occupied between 6.1% and 9.7% of the domain.  I added a "cupmesh," which is a tetrahedral mesh representation of a cup, from the `data/quest/cup.proe` file.  This image is for illustration only.  The actual shapes are bigger and centered on the mesh.

qu6 shapes

## Timers
The benchmarks were run on 1 rank, using OpenMP on rzwhippet and HIP on rzadams.  Here are the Caliper-reported times spent clipping each geometry.  The values are the medians of 5 repetitions.  The "sl" column headings below indicate the level of screening that was used to avoid calls to the clipping function where possible.
- sl0: no screening
- sl1: screen out cells using geometry-specific algorithms
- sl2: decompose each remaining cell from sl1 into tets and screen the tets.
- sl3: sl2 plus try to determine in/out relationship between 2 primitives before computing their overlap volumes.

On rzwhippet:
```
                                  sl0    sl1    sl2    sl3
ShapeMesh::precomputeMeshData   0.222  0.197  0.213  0.257
clipping:cone.0                 2.012  1.395  1.255  1.262
clipping:cupmesh.0             18.521  2.346  0.916  0.801
clipping:cyl.0                  2.177  0.518  0.453  0.447
clipping:hex.0                 11.436  0.910  0.291  0.238
clipping:plane.0                0.028  0.003  0.003  0.003
clipping:sor.0                  5.056  1.111  0.982  0.969
clipping:tet.0                  9.033  0.354  0.114  0.114
clipping:tetmesh.0             16.058  1.148  0.414  0.359
```

On rzadams:
```
                                 sl0    sl1    sl2    sl3
ShapeMesh::precomputeMeshData  0.129  0.130  0.129  0.132
clipping:cone.0                0.658  0.278  0.258  0.259
clipping:cupmesh.0             0.811  0.284  0.281  0.306
clipping:cyl.0                 0.662  0.193  0.207  0.208
clipping:hex.0                 0.669  0.176  0.201  0.231
clipping:plane.0               0.161  0.144  0.141  0.143
clipping:sor.0                 0.971  0.221  0.236  0.235
clipping:tet.0                 0.646  0.170  0.164  0.163
clipping:tetmesh.0             0.696  0.245  0.246  0.246
```

For comparison, the time currently taken by `IntersectionShaper` on rzwhippet are
```
cone    487
cupmesh 32.0
cyl     672
hex     21.4
plane   118
sor     62.3
tet     15.0
tetmesh 28.4
```

## Remarks about the performance times
- The largest gain came from sl1, the simplest screening.  The higher screening levels built on this one.
- The sl2 and sl3 screening levels had varied positive gains on rzwhippet.  On rzadams, the performance gains were small and even negative for half of the test geometries.  On higher-res computational meshes, there would be more cell/tets that can be labeled in or out, and more gains from screening, so the result trends may differ.  As always, keep in mind that performance is highly configuration-dependent.
- The gains from sl1 to sl2 were better before we switched from the 24-tet decomposition to the 18-tet decomposition (which was really a 6-tet decomposition on the cartensian mesh).  This makes sense because the larger tets are more likely to intersect the geometry's boundary.  sl2 screening was helpful on rzwhippet but not rzadams.  For hex cells with non-planar faces, we would have 18 non-degenerate tets per cell, so I think we would see more benefits from the sl2's tet screening.
- Detailed timers show that the time spent screening was little to negligible.
- All screen levels benefited from the 2-pass candidate search.  These results are not a comparison of the 2-pass vs the single-pass.  The original implementation uses single-pass and was significantly slower.


Arlie-Capps (approved):
Thanks for the work on this, Brian.

src/axom/quest/tests/quest_mesh_clipper.cpp
```diff
@@ -1418,53 +1418,50 @@ int main(int argc, char** argv)
...
-  /*
-    Copy mesh to host check results and plot.
-  */
-  SLIC_INFO(axom::fmt::format("{:-^80}", "Copying mesh to host and write out"));
+  if(!params.outputFile.empty())
+  {
+    SLIC_INFO(axom::fmt::format("{:-^80}", "Copying mesh to host and write out"));
 
-  AXOM_ANNOTATE_BEGIN("Copy results to host and write out");
+    AXOM_ANNOTATE_BEGIN("Copy results to host and write out");
```
kennyweiss: It's hard to tell on github, but it looks like this `AXOM_ANNOTATE_BEGIN` is inside the conditional, and eht matching `*_END()` is outside the conditional.

Should this be an `AXOM_ANNOTATE_SCOPE`  (and remove the `*_END()` below )?

kennyweiss: You can check that the annotations are correct by running the code through caliper's `validator` service:

```
> CALI_LOG_VERBOSITY=2 CALI_SERVICES_ENABLE=validator  <executable> <args>
```

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
```diff
@@ -753,6 +285,8 @@ class MeshClipperImpl : public MeshClipper::Impl
                                 conduit::Node& statistics) override
 
   {
+    AXOM_ANNOTATE_SCOPE("MeshClipperImpl::computeClipVolumes3DTets");
```
kennyweiss, line 288: Suggestion: Looks like you're trying to use only one colon in the annotations.
If so:
```suggestion
    AXOM_ANNOTATE_SCOPE("MeshClipperImpl:computeClipVolumes3DTets");
```

src/axom/quest/tests/quest_mesh_clipper.cpp
gunney1: Thank you!

src/axom/quest/detail/clipping/MeshClipperImpl.hpp
gunney1: Actually, I use `::` before a method name and `:` for others.  Just my convention.  But good eyes!

Issue #1783, opened 2026-01-28
Axom CUDA build failure


dtaller:
I got the following error on a GPU CUDA build
```

MeshClipper.cpp(284): error: no instance of function template "axom::quest::experimental::globalReduce" matches the argument list
            argument types are: (axom::Array<int64_t, 1, axom::MemorySpace::Dynamic, axom::detail::DefaultStoragePolicy<int64_t>>, MPI_Op)
    globalReduce(sums,
    ^
MeshClipper.cpp(232): note #3326-D: function template "axom::quest::experimental::globalReduce" does not match because argument #2 does not match parameter
  void globalReduce(axom::Array<T>& values, int reduceOp)
       ^
 
make[5]: *** [axom/quest/CMakeFiles/quest.dir/build.make:339: axom/quest/CMakeFiles/quest.dir/MeshClipper.cpp.o] Error 2
make[5]: *** Waiting for unfinished jobs....
Warning #20208-D: 'long double' is treated as 'double' in device code
```

I am attempting to fix it by changing the function signature of globalReduce from
`
void globalReduce(axom::Array<T>& values, int reduceOp)`

to 

`void globalReduce(axom::Array<T>& values, MPI_Op reduceOp)`

dtaller:
That fix worked!

the axom tutorials also seem to crash, but I disabled that for now.

PR #1784, opened 2026-01-28
Fix compiler error when `MPI_Op` is not `int`

gunney1:
# Summary
Fix a build error that happens when `MPI_Op` is not `int`.
Fixes https://github.com/llnl/axom/issues/1783.  Thanks @dtaller for this fix.


dtaller (approved):

kennyweiss (approved):
Thanks @gunney1

PR #1786, opened 2026-01-28
Fix Bump dependencies for Quest

### LINKED ISSUE #1782: Fix quest dependency on bump
A user reported the following error in a config that disabled bump

```
<axom>/src/axom/quest/detail/clipping/TetMeshClipper.cpp:12:10: fatal error: 'axom/bump.hpp' file not found
   12 | #include "axom/bump.hpp"
      |          ^~~~~~~~~~~~~~~
1 error generated.
```

Looks like quest now depends on bump, but we don't yet indicate that here:
https://github.com/llnl/axom/blob/80bf45ee2bb79817f8b709dc6a88130f1fce2a67/src/axom/quest/CMakeLists.txt#L10-L16

kennyweiss:
# Summary

This PR is a bugfix for configurations that build with `quest` but without `bump`
- `quest` has a soft dependency on `bump`, and `bump` has a hard dependency on `conduit`
- we don't want `quest` to have a hard dependency on `conduit` since that would exclude it from builds that do not have TPLs
- instead, update the logic for including files, tests and examples that depend on `bump`
- Resolves #1782 

Misc:
- I fixed some guards in the CMake and source files related to quest applications
- I removed the `AXOM_ENABLE_MFEM_SIDRE_DATACOLLECTION` CMake config option, which we no longer need
- I updated the copyright header for a bunch of files that were missed in the first round

### DIFF
--- a/src/thirdparty/axom/fmt/README
+++ b/src/thirdparty/axom/fmt/README
@@ -1,4 +1,4 @@
-This directory holds the source files for MIT-licensed fmt libary (version 9.1.0)
+This directory holds the source files for MIT-licensed fmt libary (version 12.1.0)
 retrieved from github.com/fmtlib/fmt
 
 Note: Axom adds a custom header file ('fmt.hpp') to simplify usage of fmt configured as a header only library.


Arlie-Capps (approved):
Thanks, @kennyweiss .

src/axom/quest/CMakeLists.txt
kennyweiss, line 1: This file probably has the most important changes in this branch

PR #1788, opened 2026-01-29
Util to instantiate `MeshClipperStrategy` from `klee::Geometry` spec

gunney1:
# Summary

- This PR is a feature
- It does the following:
  - Provide a utility to instantiate the correct `MeshClipperStrategy` using the data in the `klee::Geometry`.  Apps should use this method to avoid having to write a chain of if-else blocks and thinking about the details of the strategies that Axom provides.
  - Alters the mesh clipper test to exercise this utility.

### DIFF
--- a/src/axom/quest/util/make_clipper_strategy.cpp
+++ b/src/axom/quest/util/make_clipper_strategy.cpp
@@ -67,7 +67,8 @@ std::shared_ptr<MeshClipperStrategy> make_clipper_strategy(const axom::klee::Geo
   }
   else
   {
-    SLIC_WARNING(axom::fmt::format("klee::Geometry format {} is not supported by MeshClipper.", format));
+    SLIC_WARNING(
+      axom::fmt::format("klee::Geometry format {} is not supported by MeshClipper.", format));
   }
 
   return strategy;


src/axom/core/memory_management.cpp
```diff
@@ -76,6 +76,8 @@ int getSharedMemoryAllocatorID(std::size_t minSegmentSize)
     }
     allocator_id = alloc.getId();
   }
+#else
+  AXOM_UNUSED_VAR(minSegmentSize);
```
kennyweiss, line 80: Thanks!

src/axom/quest/util/make_clipper_strategy.hpp
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2017-2025, Lawrence Livermore National Security, LLC and
```
kennyweiss: Please update to the new copyright statement

src/axom/quest/util/make_clipper_strategy.cpp
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2017-2025, Lawrence Livermore National Security, LLC and
```
kennyweiss: Please update to the new copyright header

gunney1: OK.  Thanks!

PR #1789, opened 2026-01-30
Update to RAJA v2025.12.1, caliper 2.14.0

bmhan12:
This PR:
- Updates TPLs to use  RAJA v2025.12.1 and caliper 2.14.0
- Re-enables OpenMP for the Windows CI testing post-RAJA patch release fix.
  - Last Windows CI run with x64 builds passing: https://github.com/llnl/axom/actions/runs/21497397962

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -12,8 +12,8 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/develop' }}
 
 env:
-  CLANG_DOCKER_IMAGE: axom/tpls:clang-19_01-09-26_18h-07m
-  GCC_DOCKER_IMAGE: axom/tpls:gcc-13_01-09-26_18h-06m
+  CLANG_DOCKER_IMAGE: axom/tpls:clang-19_01-29-26_21h-32m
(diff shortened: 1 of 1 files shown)


scripts/spack/configs/versions.yaml
```diff
@@ -6,10 +6,10 @@ packages:
   adiak:
     require: 
       - spec: "@0.4.0"
-  # 7/23/25 - Commit after 2.13.1 release, has fix for CUDA 12.9.1
+  # 2.14.0
   caliper:
     require:
-      - spec: "@git.527e1ae3a16487e425707e928e05c43d372e144f=master"
+      - spec: "@git.9ab0e64d2e328d1fdff2ffebb53f30a899d6a598=master"
```
kennyweiss, line 12: Assumption check -- This is the SHA for caliper@2.14.0, but the version hasn't yet been added to the spack packages repo, right?  

And the same is true for raja@2025.12.1 a few lines down, right?

bmhan12: > And the same is true for raja@2025.12.1 a few lines down, right?

Yes, for RAJA the version isn't available yet.

Technically, the newer caliper version is available in upstream spack, but not in `radiuss-spack-configs` which has higher precedence over upstream spack. Once recipes are removed from `radiuss-spack-configs`, this will no longer be an issue.

PR #1791, opened 2026-02-02
Implement first set of fast approximation GWN methods

jcs15c:
# Summary

- This PR enhances the GWN capabilities of `quest_winding_number_2d_ex` (previously `quest_winding_number_ex`) and `quest_winding_number_3d_ex` by adding options for fast approximations of the GWN field. These methods are currently implemented for collections of 3D triangles and 2D lines as described in [this paper](https://www.dgp.toronto.edu/projects/fast-winding-numbers/). 
- To support this, the `quest_winding_number_3d_ex` example now has command line arguments to triangulate the input STEP file and evaluate its GWN field, and `quest_winding_number_2d_ex` now has command line arguments to linearize input curves. Subsequent containment decisions will faithfully reproduce the boundary of the discrete shape, but this will not exactly match the boundary of the curved surface. This leads to classification errors for points close to the surface.
- New tests for 2D and 3D fast-approximate GWN methods have been added.

## Algorithm Details
- Briefly, these fast approximations work by adding each piece of the shape to a BVH spatial index, which is traversed at the given query point. During traversal:
  - If the query point is far from the bounding box of an internal node, the contribution of all its children to the GWN is be approximated rapidly using a Taylor expansion. 
  - If the query point is near to or within the bounding box of an internal node, recurse to its children.
  - If the query point is near to or within the bounding box of a leaf node, it's contribution to the GWN is computed exactly using a direct formula. 
  
## Implementation Details
 - Utilizing the existing BVH tree for this purpose required modifying it in the following ways:
   - Adding a `LinearBVHTraverser::reduce_tree(<LeafAction>)` method which evaluates the `LeafAction` lambda on each leaf node to produce a value, then adding those values up the tree so that each internal node is associated with the sum of values for all its children. This returns an array of these values for each internal node.
   - Adding an optional third positional argument to the search predicate lambda (i.e. the function which determines if a traverser should recurse at a given internal node) which contains an index to the internal node as laid out within the BVH.

~~This PR should consider the following before being un-drafted:~~ Done!
- [x] Extend `quest_winding_number_3d_ex` to take in `.stl` files directly as input.
- [x] Add more command line options to the 2D and 3D examples to permit changing Taylor approximation order
- [x] Add more command line options to the 3D example to permit direct evaluation of a triangulated STEP file instead of fast approximation, even if this would be significantly slower without a difference in quality.
- [x] Add a few extra tests to make sure containment decisions between the current direct methods and fast approximation methods agree.
- [x] Add a method to the quest STEP reader to get a bounding box directly from the .STEP file using OpenCascade.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -26,12 +26,18 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 
 ### Changed
 - Primal: Axom's polygon clipping was modified to handle some corner cases.
+- Bump: Several Bump algorithms were improved so they perform better with OpenMP.
+- Mir: The `EquiZAlgorithm was improved so it performs better with OpenMP.
+- Many algorithms in Bump and Mir have been enhanced with `setAllocatorID()` methods to permit use of custom allocators.
 - Uberenv's spack updated to v1.1.1
 - `radiuss-spack-configs` submodule removed.
 
 ### Fixed
 - Primal: Fixed `NURBSPatch::evaluateDerivatives` returning a matrix with empty values. The old behavior 
   can be recovered with an optional flag, which is more efficient if the previously missing values are unneeded
+- Core: the `ArrayView` class was modified so it defers initializing an internal
+  allocator id via Umpire, if present. This prevents excessive calls to Umpire,
+  which are not needed in all use cases.
 
 ### Removed
 
--- a/scripts/benchmarks/mir_concentric_circles.py
+++ b/scripts/benchmarks/mir_concentric_circles.py
@@ -13,14 +13,23 @@ def lrun(np):
 
 # Parallel options use 4 ranks per node since that is how many GPUs there are.
 runs = {
-  "build-rzansel-blueos_3_ppc64le_ib_p9-clang@10.0.1.2_cuda-release" : {"policies":["seq", "omp", "cuda"], "launch":lrun},
-  "build-rzwhippet-toss_4_x86_64_ib-clang@14.0.6-release" : {"policies":["seq", "omp"], "launch":srun},
-  "build-rzwhippet-toss_4_x86_64_ib-gcc@10.3.1-release" : {"policies":["seq", "omp"], "launch":srun},
-  "build-rzwhippet-toss_4_x86_64_ib-intel@2022.1.0-release" : {"policies":["seq", "omp"], "launch":srun},
-  "build-rzvernal-toss_4_x86_64_ib_cray-rocmcc@6.2.1_hip-release" : {"policies":["seq", "hip"], "launch":srun},
-  "build-rzvernal-toss_4_x86_64_ib_cray-clang@17.0.0_hip-release" : {"policies":["seq", "hip"], "launch":srun},
-  "build-rzadams-toss_4_x86_64_ib_cray-cce@18.0.0_hip-release" :{"policies":["seq", "hip"], "launch":flux_run},
-  "build-rzadams-toss_4_x86_64_ib_cray-rocmcc@6.2.1_hip-release" :{"policies":["seq", "hip"], "launch":flux_run}
+  "build-dane-toss_4_x86_64_ib-gcc@13.3.1-release" : {"policies":["seq", "omp"], "launch":srun},
+  "build-dane-toss_4_x86_64_ib-intel-oneapi-compilers@2025.2.0-release" : {"policies":["seq", "omp"], "launch":srun},
+  "build-dane-toss_4_x86_64_ib-llvm@19.1.3-release" : {"policies":["seq", "omp"], "launch":srun},
+  "build-matrix-toss_4_x86_64_ib-gcc@13.3.1_cuda-release" : {"policies":["seq", "omp", "cuda"], "launch":srun},
+  "build-matrix-toss_4_x86_64_ib-llvm@19.1.3_cuda-release" : {"policies":["seq", "omp", "cuda"], "launch":srun},
+  "build-rzwhippet-toss_4_x86_64_ib-llvm@19.1.3-release" : {"policies":["seq", "omp"], "launch":srun},
+  "build-rzwhippet-toss_4_x86_64_ib-gcc@13.3.1-release" : {"policies":["seq", "omp"], "launch":srun},
+  "build-rzwhippet-toss_4_x86_64_ib-intel-oneapi-compilers@2025.2.0-release" : {"policies":["seq", "omp"], "launch":srun},
+  "build-rzvernal-toss_4_x86_64_ib_cray-cce@20.0.0_hip-release" : {"policies":["seq", "hip"], "launch":srun},
+  "build-rzvernal-toss_4_x86_64_ib_cray-llvm-amdgpu@6.3.1_hip-release" : {"policies":["seq", "hip"], "launch":srun},
+  "build-rzvernal-toss_4_x86_64_ib_cray-llvm-amdgpu@6.4.2_hip-release" : {"policies":["seq", "hip"], "launch":srun},
+  "build-rzadams-toss_4_x86_64_ib_cray-cce@20.0.0_hip-release" :{"policies":["seq", "hip"], "launch":flux_run},
+  "build-rzadams-toss_4_x86_64_ib_cray-llvm-amdgpu@6.3.1_hip-release" :{"policies":["seq", "hip"], "launch":flux_run},
+  "build-rzadams-toss_4_x86_64_ib_cray-llvm-amdgpu@6.4.2_hip-release" :{"policies":["seq", "hip"], "launch":flux_run},
+  "build-tioga-toss_4_x86_64_ib_cray-cce@20.0.0_hip-release" :{"policies":["seq", "hip"], "launch":flux_run},
+  "build-tioga-toss_4_x86_64_ib_cray-llvm-amdgpu@6.3.1_hip-release" :{"policies":["seq", "hip"], "launch":flux_run},
+  "build-tioga-toss_4_x86_64_ib_cray-llvm-amdgpu@6.4.2_hip-release" :{"policies":["seq", "hip"], "launch":flux_run}
 }
 
 def generate(params):
@@ -39,6 +48,10 @@ def generate(params):
     f.write("#!/bin/bash\n\n")
     f.write("CONCENTRIC_CIRCLES=./examples/mir_concentric_circles\n")
     f.write("CONCENTRIC_CIRCLES_MPI=./examples/mir_concentric_circles_mpi\n\n")
+
+    f.write("export OMP_PLACES=cores\n")
+    f.write("export OMP_PROC_BIND=spread\n")
+    f.write("export OMP_DYNAMIC=FALSE\n")
 
     dimension = params["dimension"]
     trials = params["trials"]
@@ -68,7 +81,10 @@ def read_timings(filename, searchKey):
   retval = "" # no data
   try:
     lines = open(filename, "rt").readlines()
-    print(f"Reading {filename}")
+    msg = ""
+    if searchKey.find("Algorithm") == -1:
+      msg = f" Getting '{searchKey}' value."
+    print(f"Reading {filename}.{msg}")
     for line in lines:
       pos = line.find(searchKey)
       if pos != -1:
@@ -113,10 +129,7 @@ def make_columns(params):
       avg = value
     return avg
 
-  # Measure just the MIR algorithm
-  searchKey = "EquizAlgorithm"
-  if params["method"] == "elvira":
-    searchKey = "ElviraAlgorithm"
+  searchKey = params["searchKey"]
 
   columns = []
   # Add NumZones column (either square or cube of s, depending on dimension)
@@ -184,9 +197,17 @@ def make_csv(params, outputfile):
     f.write(f"{line}\n")
   f.close()
 
-def seriesName(name):
+def seriesName(name, includeCompiler = False):
   policies = {"SEQ" : "Serial", "OMP": "OpenMP", "CUDA" : "CUDA", "HIP" : "HIP"}
-  newName = name[:name.find("-")]
+  if includeCompiler:
+    d0 = name.find("-")
+    d1 = name.rfind("-")
+    s0 = name.rfind(" ")
+    hostName = name[:d0]
+    compName = name[d1 + 1:s0]
+    newName = f"{hostName} {compName}"
+  else:
+    newName = name[:name.find("-")]
   for p in policies.keys():
     if name[-len(p) - 1:] == " " + p:
       newName = newName + " " + policies[p]
@@ -197,8 +218,15 @@ def lineProps(name):
   """
   Make line properties when given a series name.
   """
-  hostColor = {"rzansel" : "g", "rzwhippet" : "r", "rzvernal" : "o", "rzadams" : "b"}
-  policyStyle = {"seq" : "--", "omp": ":", "cuda" : "-", "hip" : "-"}
+  red = "#ff0000"
+  green = "#00aa00"
+  blue = "#0000ff"
+  orange = "#ff6622"
+  #magenta = "#ff00ff"
+  #purple = "#8811aa"
+  hostColor = {"rzwhippet" : red, "rzvernal" : green, "rzadams" : blue,
+               "dane" : red, "matrix" : green, "tioga": orange, "tuolumne" : blue}
+  policyStyle = {"seq" : "-", "omp": ":", "cuda" : "--", "hip" : "--"}
   policyMark = {"seq" : "o", "omp": "s", "cuda" : "^", "hip" : "^"}
   color = "b"
   style = "-"
@@ -213,6 +241,25 @@ def lineProps(name):
            break
        break
   return color, style, mark
+
+def hex_to_rgb(hex_color: str):
+    """
+    Convert a color in "#rrggbb" (or "rrggbb") format to an (r, g, b) tuple.
+    """
+    s = hex_color.strip()
+    if s.startswith("#"):
+        s = s[1:]
+    if len(s) != 6:
+        raise ValueError(f"Expected 6 hex digits, got {len(s)}: {hex_color!r}")
+
+    try:
+        r = int(s[0:2], 16)
+        g = int(s[2:4], 16)
+        b = int(s[4:6], 16)
+    except ValueError as e:
+        raise ValueError(f"Invalid hex color: {hex_color!r}") from e
+
+    return (r, g, b)
 
 def plot(params):
   """
@@ -233,22 +280,78 @@ def plot(params):
         break
     return x,y
 
+  def rgbToColor(r,g,b):
+     return "#%02x%02x%02x" % (r,g,b)
+
+  def lightenColor(r,g,b):
+     def clamp255(value):
+        if value > 255:
+          return 255
+        return int(value)
+     scale = 1.3
+     return rgbToColor(clamp255(r * scale), clamp255(g * scale), clamp255(b * scale))
+
+  def darkenColor(r,g,b):
+     scale = 0.7
+     return rgbToColor(int(r * scale), int(g * scale), int(b * scale))
+
+  def modifyColor(color, index):
+    if index == 0:
+      return color
+    r,g,b = hex_to_rgb(color)
+    if index == 1:
+      return darkenColor(r,g,b)
+    return lightenColor(r,g,b)
+
+  def valid_column(column):
+    valid = False
+    for i in range(1, len(column)):
+      if column[i] != "":
+        valid = True
+    return valid
+
+  # Make columns from the data
   columns = make_columns(params)
 
+  # Count how many plots would use the normal legend name
+  counts = {}
+  for c in range(1, len(columns)):
+    if not valid_column(columns[c]):
+      continue
+    sn = seriesName(columns[c][0])
+    if sn in counts:
+      count, _ = counts[sn]
+      counts[sn] = (count + 1, 0)
+    else:
+      counts[sn] = (1, 0)
+
   import matplotlib.pyplot as plt
+  seriesLegendNames = {}
   for c in range(1, len(columns)):
     x, y = make_series(columns[0][1:], columns[c][1:])
     if len(x) > 0:
       color, style, mark = lineProps(columns[c][0])
-      plt.plot(x, y, marker=mark, linestyle=style, color=color, linewidth=2., label=seriesName(columns[c][0]))
+      sn = seriesName(columns[c][0])
+      if counts[sn][0] > 1:
+        oldsn = sn
+        sn = seriesName(columns[c][0], True)
+        color = modifyColor(color, counts[oldsn][1])
+        counts[oldsn] = (counts[oldsn][0], counts[oldsn][1] + 1)
+      seriesLegendNames[c] = sn
+      plt.plot(x, y, marker=mark, linestyle=style, color=color, linewidth=2., label=sn)
 
   dimension = params["dimension"]
   method = params["method"]
+  searchKey = params["searchKey"]
+  if searchKey.find("Algorithm") != -1:
+    # more concise
+    searchKey = method
+  doLabels = params["labels"]
 
   # Add labels and title
   plt.xlabel('Number of Zones', fontsize=24)
   plt.ylabel('Time (s)', fontsize=24)
-  plt.title(f'{dimension}D MIR Timings ({method})', fontsize=28)
+  plt.title(f'{dimension}D MIR Timings ({searchKey})', fontsize=28)
   xlabels = make_series(columns[0][1:], columns[0][1:])[0]
   plt.xticks(ticks=xlabels, labels=xlabels, fontsize=18)
   plt.yticks(fontsize=18)
@@ -267,6 +370,65 @@ def plot(params):
   except ValueError:
     print(f"There was an error, probably because the benchmark is still running. {columns}")
     raise
+
+  # Add labels
+  if doLabels and len(columns) >= 2:
+    def findIndex(seq, value):
+      eps = 1.e-8
+      for i in range(len(seq)):
+        if math.fabs(value - seq[i]) < eps:
+          return i
+      return -1
+
+    # Group the related series.
+    seriesGroups = {}
+    seriesSer = {}
+    for c in seriesLegendNames.keys():
+      sn = seriesLegendNames[c]
+      host = sn[:sn.rfind(" ")]
+      if host in seriesGroups:
+        seriesGroups[host].append(c)
+      else:
+        seriesGroups[host] = [c]
+      if sn.find("Serial") != -1:
+        seriesSer[host] = c
+
+    # Go through groups
+    for host in seriesGroups:
+      if not host in seriesSer:
+        continue
+
+      # The serial column
+      cser = seriesSer[host]
+
+      # Compare other columns against the serial column
+      for c in seriesGroups[host]:
+        if cser == c:
+          continue
+
+        x1, y1 = make_series(columns[0][1:], columns[cser][1:])
+        x2, y2 = make_series(columns[0][1:], columns[c][1:])
+        xa = []
+        ya = []
+        labels = []
+        idx = 0
+        for i in range(len(x1)):
+          # Where series 1 (SEQ) > series 2 (OMP)
+          j = findIndex(x2, x1[i])
+          if j != -1 and y1[i] > y2[j]:
+            xa.append(x2[j])
+            ya.append(y2[j])
+            labels.append(f"{math.trunc(100. * y1[i] / y2[i]) / 100.:.2f}x")
+            plt.annotate(
+              labels[idx],
+              (xa[idx], ya[idx]),
+              xytext=(6, -6), textcoords="offset points",  # pixel offset
+              ha="left", va="bottom",
+              fontsize=10,
+              bbox=dict(boxstyle="round,pad=0.2", fc="white", ec="none", alpha=0.8),
+              arrowprops=dict(arrowstyle="-", color="0.3", lw=0.8)
+            )
+            idx = idx + 1
 
   # Show the plot
   plt.grid(True)
@@ -320,9 +482,23 @@ def get_params():
   )
 
   parser.add_argument(
+    "--key",
+    type=str,
+    help="Name of the caliper timings to extract from the logs",
+    required=False
+  )
+
+  parser.add_argument(
     "--trials",
     type=int,
     help="Number of times to run MIR",
+    required=False
+    )
+
+  parser.add_argument(
+    "--labels",
+    action="store_true",
+    help="Boolean flag to indicate whether to draw speedup labels",
     required=False
     )
 
@@ -368,6 +544,20 @@ def get_params():
     params["trials"] = max(1, args.trials)
   else:
     params["trials"] = 1
+
+  if args.labels is not None:
+    params["labels"] = args.labels
+  else:
+    params["labels"] = False
+
+  if args.key is not None:
+    params["searchKey"] = args.key
+  else:
+    # Default to MIR algorithm timings
+    if params["method"] == "elvira":
+      params["searchKey"] = "ElviraAlgorithm"
+    else:
+      params["searchKey"] = "EquizAlgorithm"
 
   # Generate some sizes.
   sides = (50, 100, 200, 500, 1000, 1500, 2000, 4000, 8000)
--- a/scripts/spack/packages/axom/package.py
+++ b/scripts/spack/packages/axom/package.py
@@ -283,11 +283,11 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
     for c in _AXOM_COMPONENTS:
         conflicts(f"components={c}", when="components=none")
 
-    requires("components=slic,spin,primal", when="components=bump")
+    requires("components=sidre,slic,spin,primal", when="components=bump")
     requires("components=sidre,slic,primal", when="components=inlet")
     requires("components=sidre,slic,inlet,primal", when="components=klee")
     requires("components=slic,slam", when="components=mint")
-    requires("components=bump,slic,slam,primal", when="components=mir")
+    requires("components=bump,sidre,slic,slam,primal", when="components=mir")
     requires("components=slic,slam", when="components=multimat")
     requires("components=slic", when="components=primal")
     requires("components=slic,slam,primal,mint,spin", when="components=quest")
--- a/src/axom/bump/BlendData.hpp
+++ b/src/axom/bump/BlendData.hpp
@@ -64,6 +64,19 @@ struct BlendData
   CoeffView m_blendCoeffView {};       //!< Weights that make up blend groups.
 };
 
+/*!
+ * \brief Return the number of blend groups in the BlendData (the number of values produced from the BlendData)
+ *
+ * \param blend The BlendData we're querying.
+ *
+ * \return The number of blend groups in the BlendData.
+ */
+AXOM_HOST_DEVICE
+inline axom::IndexType numberOfValues(const BlendData &blend)
+{
+  return blend.m_blendGroupSizesView.size();
+}
+
 }  // end namespace bump
 }  // end namespace axom
 
--- a/src/axom/bump/CMakeLists.txt
+++ b/src/axom/bump/CMakeLists.txt
@@ -11,7 +11,7 @@
 # Check necessary dependencies
 #------------------------------------------------------------------------------
 axom_component_requires(NAME       BUMP
-                        COMPONENTS SLIC SPIN PRIMAL
+                        COMPONENTS SLIC SIDRE SPIN PRIMAL
                         TPLS       Conduit)
 
 #------------------------------------------------------------------------------
@@ -37,6 +37,7 @@ set(bump_headers
     extraction/TableManager.hpp
     data/MeshTester.hpp
     io/save.hpp
+    utilities/conduit_array_view.hpp
     utilities/blueprint_utilities.hpp
     utilities/conduit_memory.hpp
     utilities/conduit_traits.hpp
@@ -123,6 +124,7 @@ set(bump_sources
     extraction/tables/cutting/CutCasesWedge.cpp
     data/MeshTester.cpp
     io/save.cpp
+    utilities/conduit_memory.cpp
     utilities/conduit_traits.cpp
     utilities/blueprint_utilities.cpp
     views/MaterialView.cpp
@@ -133,7 +135,7 @@ set(bump_sources
 #------------------------------------------------------------------------------
 # Build and install the library
 #------------------------------------------------------------------------------
-set(bump_depends_on core slic spin primal conduit::conduit)
+set(bump_depends_on core slic spin sidre primal conduit::conduit)
 blt_list_append( TO bump_depends_on ELEMENTS RAJA IF RAJA_FOUND )
 
 axom_add_library(
@@ -155,4 +157,3 @@ axom_install_component(NAME    bump
 if (AXOM_ENABLE_TESTS)
   add_subdirectory(tests)
 endif()
-
--- a/src/axom/bump/CoordsetBlender.hpp
+++ b/src/axom/bump/CoordsetBlender.hpp
@@ -13,6 +13,7 @@
 #include "axom/bump/BlendData.hpp"
 #include "axom/primal/geometry/Point.hpp"
 #include "axom/primal/geometry/Vector.hpp"
+#include "axom/sidre/core/ConduitMemory.hpp"
 #include "axom/slic.hpp"
 
 #include <conduit/conduit.hpp>
@@ -53,7 +54,8 @@ public:
   void execute(const BlendData &blend,
                const CoordsetViewType &view,
                const conduit::Node &n_input,
-               conduit::Node &n_output) const
+               conduit::Node &n_output,
+               int allocator_id = axom::execution_space<ExecSpace>::allocatorID()) const
   {
     using value_type = typename CoordsetViewType::value_type;
     using PointType = typename CoordsetViewType::PointType;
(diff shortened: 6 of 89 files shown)


src/axom/quest/LinearizeCurves.hpp
```diff
@@ -25,7 +25,7 @@ class LinearizeCurves
 {
 public:
   using NURBSCurve = axom::primal::NURBSCurve<double, 2>;
-  using CurveArrayView = axom::ArrayView<NURBSCurve>;
+  using CurveArrayView = axom::ArrayView<const NURBSCurve>;
```
jcs15c, line 28: I had to make this change for this to work more straight-forwardly with `LinearizeCurves`, since otherwise (to my knowledge) you can't use it to linearize a view of const objects without making a copy of all of them. But I think this is more close to how the class is actually implemented everywhere else, right? It doesn't seem to me like there's anywhere that a `LinearizeCurves` object ever needs to change a `NURBSCurve` in it's view.

src/axom/primal/geometry/detail/analytic_test_surfaces.hpp
```diff
@@ -0,0 +1,294 @@
...
+#include "axom/primal.hpp"
+
+namespace axom
+{
+namespace primal
+{
+namespace detail
+{
+
+axom::Array<primal::NURBSPatch<double, 3>> make_sphere_biquartic()
```
kennyweiss: Minor: Would it make sense for this function to take a radius parameter? 

Alternatively, the user can scale/transform the returned patches, so perhaps rename this to `make_unit_sphere_biquartic()` ?  (Or at least add a doxy comment that it's the unit sphere)

src/axom/primal/geometry/detail/analytic_test_surfaces.hpp
```diff
@@ -0,0 +1,294 @@
...
+  using Point3D = primal::Point<double, 3>;
+  using NPatch = primal::NURBSPatch<double, 3>;
+
+  double rt2 = sqrt(2), rt3 = sqrt(3), rt6 = sqrt(6);
+
+  // Define the nodes and weights for one of six rational, biquartic Bezier patches
+  //  that compose the unit sphere. These will be rotated to form the other 5.
+  // Nodes and weights obtained from the technical report
+  // "Tiling the Sphere with Rational Bezier Patches",
+  //  James E. Cobb, University of Utah, 1988
```
kennyweiss: Please move some of this up to the doxygen comment for the function

src/axom/primal/geometry/detail/analytic_test_surfaces.hpp
```diff
@@ -0,0 +1,294 @@
...
+      sphere_faces[5](i, j)[1] = -node_data[idx][0];
+      sphere_faces[5](i, j)[2] = -node_data[idx][1];
+      sphere_faces[5](i, j).array() /= weight_data[idx];
+    }
+  }
+
+  return sphere_faces;
+}
+
+axom::Array<primal::NURBSPatch<double, 3>> make_sphere_bicubic()
```
kennyweiss: Please add a brief doxygen comment

src/axom/primal/geometry/detail/analytic_test_surfaces.hpp
```diff
@@ -0,0 +1,294 @@
...
+      sphere_faces[4*n + 2].setWeight(i, 2, the_weight);
+      sphere_faces[4*n + 3].setWeight(i, 2, the_weight);
+      // clang-format on
+    }
+  }
+
+  return sphere_faces;
+}
+
+axom::Array<primal::NURBSPatch<double, 3>> make_teardrop()
```
kennyweiss, line 214: Doxygen comment for this one too, please.

src/axom/primal/geometry/detail/analytic_test_surfaces.hpp
kennyweiss, line 1: 👍

src/axom/primal/tests/primal_solid_angle.cpp
```diff
@@ -732,43 +573,108 @@ TEST(primal_integral, nurbspatch_sphere)
...
+  using Point3D = primal::Point<double, 3>;
+  using Vector3D = primal::Vector<double, 3>;
+
+  const double edge_tol = 1e-6;
+  const double ls_tol = 1e-10;
+  const double quad_tol = 1e-5;
+  const double disk_size = 0.01;
+  const double EPS = 1e-11;
+
+  // Test the points on a teardrop madefor which
```
kennyweiss: Please fix the typo in this comment line. Perhaps:

```suggestion
  // Test the points on a teardrop for which
```

src/axom/primal/tests/primal_solid_angle.cpp
```diff
@@ -732,43 +573,108 @@ TEST(primal_integral, nurbspatch_sphere)
...
-  EXPECT_NEAR(difficult_gwn[4], 0.5, 8 * quad_tol); // Coincident point
+  EXPECT_NEAR(difficult_gwn[0], 1.0, 8 * quad_tol);  // Inner point
+  EXPECT_NEAR(difficult_gwn[1], 1.0, 8 * quad_tol);  // Inner point
+  EXPECT_NEAR(difficult_gwn[2], 0.0, 8 * quad_tol);  // Outer point
+  EXPECT_NEAR(difficult_gwn[3], 0.0, 8 * quad_tol);  // Outer point
+  EXPECT_NEAR(difficult_gwn[4], 0.5, 8 * quad_tol);  // Coincident point
+}
+
+//------------------------------------------------------------------------------
+TEST(primal_solid_angle, teardrop_regression_test)
```
kennyweiss, line 616: Thanks for adding this!

src/axom/quest/examples/quest_winding_number_2d.cpp
```diff
@@ -0,0 +1,356 @@
...
+// files for dates and other details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/*!
+ * \file quest_winding_number.cpp
+ * \brief Example that computes the winding number of a grid of points
+ * against a collection of 2D parametric rational curves.
+ * Supports MFEM meshes in the cubic positive Bernstein basis or the (rational)
+ * NURBS basis.
```
kennyweiss, line 12: Note: This example should also read in c2c files in configs built against C2C. 
(But let's handle that in a follow-up PR).

src/axom/quest/examples/quest_winding_number_2d.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+//------------------------------------------------------------------------------
+// CLI input
+//------------------------------------------------------------------------------
+
+class Input
+{
+public:
+  std::string inputFile;
+  std::string outputPrefix = {"winding"};
```
kennyweiss: suggestion: 

```suggestion
  std::string outputPrefix = {"winding2d"};
```

src/axom/quest/examples/quest_winding_number_2d.cpp
```diff
@@ -0,0 +1,356 @@
...
+  }
+  catch(const axom::CLI::ParseError& e)
+  {
+    return app.exit(e);
+  }
+
+  axom::utilities::raii::AnnotationsWrapper annotation_raii_wrapper(input.annotationMode);
+  AXOM_ANNOTATE_SCOPE("winding number example");
+
+  // Read curves from the MFME mesh
```
kennyweiss: Typo:

```suggestion
  // Read curves from the MFEM mesh
```

src/axom/spin/policy/LinearBVH.hpp
kennyweiss, line 1: @publixsubfan -- This PR adds the ability to apply reductions along interior nodes of the tree. 
Can you please focus on these changes as part of your review?

kennyweiss, line 66: Please clarify/improve the doxygen here to indicate that there are two operations: `traverse_tree` and `reduce_tree` and highlight the distinction between them.

src/axom/spin/policy/LinearBVH.hpp
```diff
@@ -103,7 +110,89 @@ class LinearBVHTraverser
...
+
+      // Return the precomputed values in the reduction.
+      auto returnLeafValue =
+        AXOM_LAMBDA(std::int32_t currentNode, const std::int32_t* leafNodes)->FieldType
+      {
+        const auto idx = leafNodes[currentNode];
+        return leafFieldView[idx];
+      };
+
+      // TODO: Replace this with GPU-compatible code.
```
kennyweiss, line 175: If this isn't currently compatible w/ GPU policies, should we add a static_assert?

src/axom/spin/policy/LinearBVH.hpp
```diff
@@ -103,7 +110,89 @@ class LinearBVHTraverser
...
+   *        using a "+" reduction. Return the Array that contains values for
+   *        all tree nodes.
+   *
+   * \param lf The function to invoke on a leaf node to make its data.
+   * \param allocatorID The allocator to use to allocate array data.
+   *
+   * \return An Array that contains the reduced data for all nodes in the BVH.
+   */
+  template <typename ExecSpace, typename FieldType, typename LeafAction>
+  axom::Array<FieldType> reduce_tree(LeafAction&& lf,
```
kennyweiss: Suggestion (minor): `leaf_action` or `leafAction` might be easier to read than `lf`

src/axom/spin/policy/LinearBVH.hpp
```diff
@@ -103,7 +110,89 @@ class LinearBVHTraverser
...
+      reduce_recursion(std::forward<LeafAction>(lf), reducedField.view(), 1);
+    }
+    else
+    {
+      // Do it in 2 stages.
+
+      // Make a field for just the leaf data. Compute it in parallel.
+      axom::Array<FieldType> leafField(m_leaf_nodes.size(), m_leaf_nodes.size(), allocatorID);
+      auto leafFieldView = leafField.view();
+      const std::int32_t* leaf_nodes_data = m_leaf_nodes.data();
```
kennyweiss, line 161: Please add a comment about why we're using a raw pointer here instead of the `ArrayView`

src/axom/spin/policy/LinearBVH.hpp
```diff
@@ -103,7 +110,89 @@ class LinearBVHTraverser
...
+   *        produce a value and then iterate back up the tree, combining nodes
+   *        using a "+" reduction. Return the Array that contains values for
+   *        all tree nodes.
+   *
+   * \param lf The function to invoke on a leaf node to make its data.
+   * \param allocatorID The allocator to use to allocate array data.
+   *
+   * \return An Array that contains the reduced data for all nodes in the BVH.
+   */
+  template <typename ExecSpace, typename FieldType, typename LeafAction>
```
kennyweiss: Minor: Would `ValueType` be better than `FieldType`? 
When I read it below, I had to check that `FieldType` wasn't defined as `Array<ValueType>`

src/axom/spin/policy/LinearBVH.hpp
```diff
@@ -103,7 +110,89 @@ class LinearBVHTraverser
...
+    axom::Array<FieldType> reducedField(m_inner_nodes.size(), m_inner_nodes.size(), allocatorID);
+
+    if constexpr(std::is_same_v<ExecSpace, axom::SEQ_EXEC>)
+    {
+      reduce_recursion(std::forward<LeafAction>(lf), reducedField.view(), 0);
+      reduce_recursion(std::forward<LeafAction>(lf), reducedField.view(), 1);
+    }
+    else
+    {
+      // Do it in 2 stages.
```
kennyweiss, line 156: Can you please expand on the two steps and why we need them for the non-seq policy?
Looks like step one is to compute on the leaf nodes and step 2 is to move it up the tree.
(whereas the seq policy branch does both)

src/axom/spin/internal/linear_bvh/bvh_traverse.hpp
kennyweiss, line 1: 👍

src/axom/spin/policy/LinearBVH.hpp
kennyweiss, line 1: @jcs15c -- is there a simple example or unit test we can add to spin for the new `reduce_tree` traverser? E.g. adding a random collection of points to a BVH by their bounding boxes and the counting the number of points contained within each inner node?

src/axom/primal/geometry/KnotVector.hpp
```diff
@@ -758,6 +758,273 @@ class KnotVector
     return ders;
   }
 
+  struct DerivativeBasisWorkspace
```
kennyweiss: It's hard to tell on github, but this looks like a bad merge conflict and duplicates the `DerivativeBasisWorkspace` struct below

src/axom/quest/examples/quest_winding_number_2d.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+    // Options for triangulation of the input STEP file
+    auto* linearize_curves_subcommand = app.add_subcommand("linearize_curves")
+                                          ->description("Options for linearizing NURBS curves")
+                                          ->fallthrough();
+
+    linearize_curves_subcommand->add_option("--num-segments", segmentsPerKnotSpan)
+      ->description("Number of segments for each knot span of each input curve.")
+      ->check(axom::CLI::PositiveNumber)
+      ->capture_default_str();
```
kennyweiss: Should this also take in a linearization error percentage to call `LinearizeCurves::getLinearMeshNonUniform()` instead of `LinearizeCurves::getLinearMeshUniform()`
(If so, that option would be mutually exclusive w/ the `--num-segments` option)

See: 
https://github.com/llnl/axom/blob/develop/src/axom/quest/LinearizeCurves.hpp#L49-L57

src/axom/quest/examples/quest_winding_number_2d.cpp
```diff
@@ -0,0 +1,356 @@
...
+    linearize_curves_subcommand
+      ->add_flag(
+        "--direct,!--fast-approximation",
+        directPolylineEval,
+        "Use direct evaluation instead of fast, heirarchical approximation? (significantly "
+        "slower, slightly more precise)")
+      ->capture_default_str();
```
kennyweiss: Can this use a string option instead of a bool flag, e.g.

```suggestion
    std::string algorithm {"fast-approximation"};
    const std::array<std::string, 2> valid_algorithms {"direct", "fast-approximation"};
    linearize_curves_subcommand.add_option("--algorithm", algorithm)
      ->description("Use fast approximated WN or direct approach (slower)")
      ->capture_default_str()
      ->check(axom::CLI::IsMember(valid_algorithms));
```

src/axom/quest/examples/quest_winding_number_3d.cpp
```diff
@@ -220,6 +93,41 @@ int main(int argc, char** argv)
...
+        deflection_is_relative,
+        "Is linear deflection in relative to local edge lengths (true) or mesh units (false)")
+      ->capture_default_str();
+
+    triangulate_step_subcommand
+      ->add_flag("--direct,!--fast-approximation",
+                 directTriangleEval,
+                 "Use direct evaluation instead of fast, heirarchical approximation? "
+                 "(significantly slower, slightly more precise)")
+      ->capture_default_str();
```
kennyweiss: Can we use an `--algorithm` option here instead of a boolean flag (see my comment on the 2D example)?

src/axom/quest/io/STEPReader.hpp
```diff
@@ -66,9 +66,15 @@ class STEPReader
   PatchArray& getPatchArray() { return m_patches; }
   const PatchArray& getPatchArray() const { return m_patches; }
 
+  /// Get the number of patches in the read file
+  int numPatches() { return m_patches.size(); }
+
   /// Returns some information about the loaded BRep
   std::string getBRepStats() const;
 
+  /// Returns an AABB for the loaded BRep
```
kennyweiss: Please update this comment to indicate that Open Cascade is computing the bounds of the AABB in this function.

src/axom/quest/tests/quest_gwn_methods.cpp
```diff
@@ -0,0 +1,228 @@
...
+std::string pjoin(const std::string &str, Args... args)
+{
+  return axom::utilities::filesystem::joinPath(str, pjoin(args...));
+}
+
+template <typename... Args>
+std::string pjoin(const char *str, Args... args)
+{
+  return axom::utilities::filesystem::joinPath(std::string(str), pjoin(args...));
+}
```
kennyweiss, line 45: @BradWhitlock -- should we move these `pjoin` functions to core/utilities?  Perhaps in a separate PR?

src/axom/quest/tests/quest_gwn_methods.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+  // Compare the in-out values between the three fields
+  const auto *query_mesh = dc[0].GetMesh();
+  const auto num_query_points = query_mesh->GetNodalFESpace()->GetNDofs();
+
+  auto &inout_direct = *dc[0].GetField("inout");
+  auto &inout_polyline = *dc[1].GetField("inout");
+  auto &inout_polyline_fast = *dc[2].GetField("inout");
+
+  for(int i = 0; i < num_query_points; ++i)
```
kennyweiss, line 239: Please add a comment that we're expecting the in/out values to match in this case because of the discretization and sampling locations, but that in general, the WN on the discretized curves might differ from the ground truth on the curves.

src/axom/quest/tests/quest_gwn_methods.cpp
kennyweiss, line 1: Nice tests!

Should this file be reorganized so that it is is always added, but the mfem usage is guarded by `AXOM_USE_MFEM` in the same way that open cascascade usage is guarded by `AXOM_USE_OPENCASCADE` ?  Or maybe not b/c we need the MFEMSidreDataCollections ?

src/axom/quest/tests/quest_gwn_methods.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+template <typename... Args>
+std::string pjoin(const char *str, Args... args)
+{
+  return axom::utilities::filesystem::joinPath(std::string(str), pjoin(args...));
+}
+//------------------------------------------------------------------------------
+
+//------------------------------------------------------------------------------
+TEST(quest_winding_number_approximations, mfem_mesh_linearization)
```
kennyweiss: Please rename the first parameter to match the filename (`quest_gwn_methods`), to help gtest filter the tests

src/axom/quest/tests/quest_gwn_methods.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+  // Compare the in-out values between the three fields
+  const auto *query_mesh = dc[0].GetMesh();
+  const auto num_query_points = query_mesh->GetNodalFESpace()->GetNDofs();
+
+  auto &inout_direct = *dc[0].GetField("inout");
+  auto &inout_tri = *dc[1].GetField("inout");
+  auto &inout_tri_fast = *dc[2].GetField("inout");
+
+  for(int i = 0; i < num_query_points; ++i)
```
kennyweiss, line 325: Similarly to the 2D case, please note that we expect the in/out values for curved and discretize to match here due to the sampling and discretization params, but this will not generally be the case

src/axom/quest/tests/quest_step_reader.cpp
```diff
@@ -0,0 +1,212 @@
...
+//------------------------------------------------------------------------------
+bool isNearInteger(double value, double eps)
+{
+  const double nearest = std::round(value);
+  return std::abs(value - nearest) <= eps;
+}
+
+//------------------------------------------------------------------------------
+// Use a pared-down version of the TriangleGWN3D method to directly evaluate 
+//  the for a triangle mesh
```
kennyweiss: Please fix this sentence

src/axom/quest/tests/quest_step_reader.cpp
```diff
@@ -0,0 +1,212 @@
...
+  EXPECT_EQ(gwn_arr.size(), query_arr.size());
+
+  constexpr double integer_eps = 1e-3;
+  for(int i = 0; i < gwn_arr.size() && i < query_arr.size(); ++i)
+  {
+    const double wn = gwn_arr[i];
+    EXPECT_NEAR(wn, std::round(wn), integer_eps);
+  }
+
+  mint::UnstructuredMesh<mint::SINGLE_SHAPE> triMesh(3, mint::TRIANGLE);
```
kennyweiss: Please add a comment here that we are triangulating the STEP file and checking the triangle winding numbers

src/axom/quest/GWNMethods.hpp
```diff
@@ -0,0 +1,827 @@
...
+        for(const auto& curv : input_curves)
+        {
+          m_nurbs_caches.emplace_back(curv);
+        }
+      }
+    }
+    timer.stop();
+    AXOM_ANNOTATE_METADATA("preprocessing_time", timer.elapsed(), "");
+    SLIC_INFO(axom::fmt::format("Direct query preprocessing (memoization caches): {} s",
+                                timer.elapsedTimeInSec()));
```
kennyweiss: Please update the log message -- this is only memoized if `use_memoization`

src/axom/quest/GWNMethods.hpp
```diff
@@ -0,0 +1,827 @@
...
+  }
+
+  void query(mfem::DataCollection& dc, const primal::WindingTolerances& tol)
+  {
+    auto* query_mesh = dc.GetMesh();
+    auto& winding = *dc.GetField("winding");
+    auto& inout = *dc.GetField("inout");
```
kennyweiss, line 192: We should check that `HasField` on both of these and error out if they are missing.

Please add a doxy comment that this is for a query grid at the DOFs of the `dc` query mesh.

src/axom/primal/geometry/detail/analytic_test_surfaces.hpp
jcs15c: Will do! I'm also just now realizing that this shape is actually biquadratic, not bicubic. It's order 2 in each axis. Whoops!

src/axom/quest/GWNMethods.hpp
```diff
@@ -0,0 +1,827 @@
...
+          winding[static_cast<int>(nidx)] = wn;
+          inout[static_cast<int>(nidx)] = std::lround(wn);
+        });
+      }
+    }
+    query_timer.stop();
+
+    const double query_time_s = query_timer.elapsed();
+    const double ms_per_query = query_timer.elapsedTimeInMilliSec() / num_query_points;
+    SLIC_INFO(axom::fmt::format(
```
kennyweiss, line 239: Note: We'll need to update this message when we test this with MPI

src/axom/quest/GWNMethods.hpp
```diff
@@ -0,0 +1,827 @@
...
+      }
+      stage_timer.stop();
+      SLIC_INFO(
+        axom::fmt::format("  Preprocessing stage (bvh): {} s", stage_timer.elapsedTimeInSec()));
+
+      stage_timer.reset();
+      stage_timer.start();
+      {
+        AXOM_ANNOTATE_SCOPE("moments");
+        const auto triangles_view = m_segments.view();
```
kennyweiss:
```suggestion
        const auto segments_view = m_segments.view();
```

src/axom/quest/GWNMethods.hpp
```diff
@@ -0,0 +1,827 @@
...
+  axom::Array<SegmentType> m_segments;
+
+  // Only needed for fast approximation method
+  axom::Array<GWNMoments> m_internal_moments;
+  axom::spin::BVH<2, axom::SEQ_EXEC> m_bvh;
+};
+///@}
+
+///@{
+/// \name Query methods for 2D GWN applications
```
kennyweiss:
```suggestion
/// \name Query methods for 3D GWN applications
```

src/axom/quest/GWNMethods.hpp
```diff
@@ -0,0 +1,827 @@
...
+    }
+    timer.stop();
+    AXOM_ANNOTATE_METADATA("preprocessing_time", timer.elapsed(), "");
+    SLIC_INFO(axom::fmt::format("Direct query preprocessing (memoization caches): {} s",
+                                timer.elapsedTimeInSec()));
+  }
+
+  void query(mfem::DataCollection& dc,
+             const primal::WindingTolerances& tol,
+             const double slice_z = 0.0) const
```
kennyweiss, line 499: Please add a comment about `slice_z`

src/axom/quest/GWNMethods.hpp
```diff
@@ -0,0 +1,827 @@
...
+      BoxType* shape_bbox_ptr = &shape_bbox;
+      axom::mint::for_all_nodes<axom::SEQ_EXEC, axom::mint::xargs::xyz>(
+        tri_mesh,
+        AXOM_LAMBDA(axom::IndexType nodeIdx, double x, double y, double z) {
+          shape_bbox_ptr->addPoint(Point3D {x, y, z});
+        });
+
+      // Extract the triangles from the mesh into axom primitives,
+      //  scaled and translated so that `shape_box` is centered at the origin
+      //  and has roughly unit volume. Otherwise, small triangles introduce numerical issues
```
kennyweiss, line 636: Do we always have to center and scale the triangle mesh and query points?

Please add a class doxygen comment about this.

src/axom/quest/FastApproximateGWN.hpp
```diff
@@ -0,0 +1,436 @@
...
+  /// TODO: Technically, the raw moments for b1 and b2 can be deallocated after this
+  ///  function is called, which would decrease the memory footprint
+  friend GWNMomentData operator+(GWNMomentData& b1, GWNMomentData& b2)
+  {
+    GWNMomentData<T, NDIMS, ORD> b_out;
+
+    b_out.a = b1.a + b2.a;
+    b_out.ap = b1.ap + b2.ap;
+
+    for(int i = 0; i < NumberOfEntries; ++i) b_out.rm[i] = b1.rm[i] + b2.rm[i];
```
kennyweiss: Please wrap the one-liner with curly brackets.

src/axom/quest/FastApproximateGWN.hpp
```diff
@@ -0,0 +1,436 @@
...
+      for(int i = 0; i < 9; ++i, ++m)
+      {
+        // In tensor product notation, equal to
+        // centroid \otimes normal
+        rm[m] = normal[i % 3] * centroid[i / 3];
+      }
+
+      if constexpr(ORD >= 2)
+      {
+        constexpr auto twlv = 0.083333333333333333333333333333;
```
kennyweiss: Let's let the compiler do the division:

```suggestion
        constexpr auto twlv = 1./12.;
```

src/axom/quest/FastApproximateGWN.hpp
```diff
@@ -0,0 +1,436 @@
...
+      if constexpr(ORD >= 1)
+      {
+        const axom::primal::Vector<T, 9>
+          F2 {ec[3], ec[4], ec[5], ec[6], ec[7], ec[8], ec[9], ec[10], ec[11]};
+        // clang-format off
+				axom::primal::Vector<T, 9> G2_1{ 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0 };
+				axom::primal::Vector<T, 9> G2_2{ pq[0] * pq[0], pq[1] * pq[0], pq[2] * pq[0],
+																				 pq[0] * pq[1], pq[1] * pq[1], pq[2] * pq[1],
+																				 pq[0] * pq[2], pq[1] * pq[2], pq[2] * pq[2] };
+        // clang-format on
```
kennyweiss, line 255: The formatting here looks weird. Is it possible that tabs snuck in?

src/axom/quest/FastApproximateGWN.hpp
```diff
@@ -0,0 +1,436 @@
...
+                                              0.0,
+                                              0.0,
+                                              0.0,
+                                              0.0};
+          G3[2] = G3[1];
+          G3[3] = -G3[0];
+          G3[4] = G3[1];
+          G3[5] = -G3[0];
+          G3[6] = -G3[0];
+          G3[7] = -G3[1];
```
kennyweiss: Suggestion(minor):
This might be a bit easier to read as:
```suggestion
          const double g0 = 2 * pq[0] * (pq[0] * pq[0] - 3 * pq[1] * pq[1]);
          const double g1 = 2 * pq[1] * (3 * pq[0] * pq[0] - pq[1] * pq[1]);
          axom::primal::Vector<double, 8> G3 {g0, g1, g1, -g0, g1, -g0, -g0, -g1};
```

src/axom/primal/geometry/detail/analytic_test_surfaces.hpp
jcs15c: I like just renaming it here. If the caller wants to change the scale, it's possible they want to shift/rotate the thing too, and that seems like a lot of legwork to do in this function.

src/axom/quest/FastApproximateGWN.hpp
```diff
@@ -0,0 +1,436 @@
...
+            ec[21], ec[22], ec[23], ec[24], ec[25], ec[26], ec[27], ec[28], ec[29],
+            ec[30], ec[31], ec[32], ec[33], ec[34], ec[35], ec[36], ec[37], ec[38]};
+
+          // clang-format off
+				  axom::primal::Vector<T, 27> G3_1{ 3 * pq[0], pq[1], pq[2], pq[1], pq[0], 0, pq[2], 0, pq[0], pq[1], pq[0], 0, pq[0],3 * pq[1], pq[2], 0, pq[2], pq[1], pq[2], 0, pq[0], 0, pq[2], pq[1], pq[0], pq[1], 3 * pq[2] };
+				  axom::primal::Vector<T, 27> G3_2{ pq[0] * pq[0] * pq[0], pq[0] * pq[0] * pq[1], pq[0] * pq[0] * pq[2], pq[0] * pq[1] * pq[0], pq[0] * pq[1] * pq[1], pq[0] * pq[1] * pq[2], pq[0] * pq[2] * pq[0], pq[0] * pq[2] * pq[1], pq[0] * pq[2] * pq[2],
+																					  pq[1] * pq[0] * pq[0], pq[1] * pq[0] * pq[1], pq[1] * pq[0] * pq[2], pq[1] * pq[1] * pq[0], pq[1] * pq[1] * pq[1], pq[1] * pq[1] * pq[2], pq[1] * pq[2] * pq[0], pq[1] * pq[2] * pq[1], pq[1] * pq[2] * pq[2],
+																					  pq[2] * pq[0] * pq[0], pq[2] * pq[0] * pq[1], pq[2] * pq[0] * pq[2], pq[2] * pq[1] * pq[0], pq[2] * pq[1] * pq[1], pq[2] * pq[1] * pq[2], pq[2] * pq[2] * pq[0], pq[2] * pq[2] * pq[1], pq[2] * pq[2] * pq[2] };
```
kennyweiss: This might also have tabs (?)

src/axom/quest/FastApproximateGWN.hpp
```diff
@@ -0,0 +1,436 @@
...
+ * 
+ * Stores arrays `rm` of raw moments and `ec` of Taylor expansion coefficients, as only the 
+ * latter is used to compute the approximated GWN. Transforming raw moments to expansion coefficients
+ * requires knowing the center of the expansion, which we take as the centroid of the collection.
+ * 
+ * Stores double `a` and vector `ap` which are the total unsigned area and weighted centroid of
+ * the collection of geometric objects. 
+ */
+template <typename T, int NDIMS, int ORD>
+class GWNMomentData
```
kennyweiss, line 53: Would it make sense to add some unit tests for the moment data computations? Or is this covered by the segment/triangle tests?

src/axom/quest/FastApproximateGWN.hpp
```diff
@@ -0,0 +1,436 @@
...
+
+  // Raw moments
+  axom::StackArray<T, NumberOfEntries> rm {};
+
+  // Expansion coefficients
+  axom::StackArray<T, NumberOfEntries> ec {};
+};
+
+/*!
+ * \brief Evaluate a heirarchical approximation of the GWN for the shape in
```
kennyweiss:
```suggestion
 * \brief Evaluate a hierarchical approximation of the GWN for the shape in 
```

src/axom/quest/FastApproximateGWN.hpp
```diff
@@ -0,0 +1,436 @@
...
+  double gwn = 0.0;
+
+  const auto leaf_objects_view = leaf_objects;
+  const auto internal_moments_view = internal_moments;
+  using LeafGeom = std::decay_t<LeafGeometry>;
+
+  auto bbContain = [&gwn, &internal_moments_view](const primal::Point<T, NDIMS>& query,
+                                                  const primal::BoundingBox<T, NDIMS>& bvhBbox,
+                                                  std::int32_t node_index) -> bool {
+    constexpr double beta = 2.0;
```
kennyweiss: Should we expose `beta` as a parameter to `fast_approximate_winding_number`?

src/axom/quest/FastApproximateGWN.hpp
```diff
@@ -0,0 +1,436 @@
...
+        const axom::primal::Vector<T, 9>
+          F2 {ec[3], ec[4], ec[5], ec[6], ec[7], ec[8], ec[9], ec[10], ec[11]};
+        // clang-format off
+				axom::primal::Vector<T, 9> G2_1{ 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0 };
+				axom::primal::Vector<T, 9> G2_2{ pq[0] * pq[0], pq[1] * pq[0], pq[2] * pq[0],
+																				 pq[0] * pq[1], pq[1] * pq[1], pq[2] * pq[1],
+																				 pq[0] * pq[2], pq[1] * pq[2], pq[2] * pq[2] };
+        // clang-format on
+
+        const auto norm_to_5 = norm_to_3 * norm * norm;
```
kennyweiss, line 257: Was this the line w/ the typo? If so, consider adding a comment.

src/axom/spin/policy/LinearBVH.hpp
jcs15c: I think the main benefit is just to improve the efficiency of operating over all the leaves, and I'm not certain that needs to be an exclusively parallel thing. I could probably rewrite `reduce_recursion` to not even take a lambda function, and just pass a view to the leaf array. I think @BradWhitlock originally wrote this, do you know of any particular reason it can't be done in two stages for both policies?

src/axom/quest/examples/quest_winding_number_2d.cpp
jcs15c: Sure, I can see that working with maybe with another subcommand `linearize_curves_nonuniform` that accepts a separate option for percent error. 

Or alternatively, we can impose that the percentage gets passed in as a decimal so that if the input is an integer 1 or greater, it's a number of segments for a uniform refinement, and if the input is in the range (0, 1), it's a percentage for nonuniform refinement. 

What do you think?

src/axom/quest/examples/quest_winding_number_2d.cpp
jcs15c: Yes! Good idea, and much simpler. Thank you!

src/axom/quest/examples/quest_winding_number_2d.cpp
kennyweiss: I was thinking there would be two separate options, and only one of them can be supplied. 

CLI11 makes this relatively easy if you save the reference/pointer to each option. We do similar things for the bounding box to ensure that if the user supplies a min/max parameter, both are necessary.

src/axom/spin/policy/LinearBVH.hpp
jcs15c: I added some more detail to the first `\brief`, and made things better organized. Is there something specific you have in mind that isn't covered by the other `\brief`'s?

kennyweiss: I think doxygen only supports a single `\brief` (the last one).

I was thinking the \brief could be something like:
```cpp
/**
 * \brief Interface for traversing a BVH tree with an action at leaf nodes. 
 * `traverse_tree` only applies the action to the leaf nodes, 
 * while `reduce_tree` sums terms to a function defined at all nodes of the tree (leaf and internal)
 */
```
and then the other `\brief` annotations can be regular doxygen comments.

src/axom/spin/policy/LinearBVH.hpp
jcs15c: To be honest I don't know why a raw pointer is used, that's how it is in the internal `bvh_traverse` function, and all of the other instances of leaf action lambda functions. In that function an `ArrayView` is passed to `bvh_traverse`, but then when the leaf action is called, the data is accessed with `ArrayView::data()`.

src/axom/spin/policy/LinearBVH.hpp
jcs15c: Yes, I'm sure I can come up with something. One way that's agnostic to the way the tree is constructed is that, if you have 8 points added to the BVH, then 8 inner nodes should contain one point (the leaves), 4 contain two points, 2 contain four points, and 1 contains all eight points, right?

src/axom/spin/policy/LinearBVH.hpp
```diff
@@ -103,7 +112,89 @@ class LinearBVHTraverser
...
+   *
+   * \param lf The function to invoke on a leaf node to make its data.
+   * \param allocatorID The allocator to use to allocate array data.
+   *
+   * \return An Array that contains the reduced data for all nodes in the BVH.
+   */
+  template <typename ExecSpace, typename ValueType, typename LeafAction>
+  axom::Array<ValueType> reduce_tree(LeafAction&& lf,
+                                     int allocatorID = axom::getDefaultAllocatorID()) const
+  {
```
publixsubfan, line 145: Is the eventual plan for this function to be invokable from inside a GPU kernel as a device function? If not, I'd suggest moving it into the standard BVH interface.

kennyweiss: We'd like to, but are not there yet

src/axom/quest/tests/quest_gwn_methods.cpp
jcs15c: At the moment, all of these `GWNMethod` objects rely on the `MFEMSidreDataCollections` (mostly for convenience and arbitrary historical reasons) but maybe we can reassess if that changes in the future.

src/axom/quest/tests/quest_step_reader.cpp
jcs15c: Oh whoops, this entire file/test is supposed to be a part of my entirely separate PR, genuinely not even sure how those changes made it to this branch. I'm going to excise them here and leave it for later. I'll be sure to fix it in the other branch though!

src/axom/quest/GWNMethods.hpp
jcs15c: Added some more comments here and on the `preprocess()` methods, as well as a bunch of early returns for bad/incomplete input.

src/axom/quest/GWNMethods.hpp
jcs15c: Technically no, you don't have to always do this, since it's really only an issue if you have really small triangles for which the GWN gets too-conservatively set to zero. But in my experience, it's hard to know ahead of time if you're going to run into that issue, since it doesn't directly correlate with either deflection parameter. I don't think it adds too much overhead, since its all preprocessing and (at the moment) you have to store an array of triangles either way.

src/axom/quest/FastApproximateGWN.hpp
jcs15c: Very much better, yes!

src/axom/quest/FastApproximateGWN.hpp
jcs15c: We might as well, right? Probably at the end as a default parameter, and explain we take `beta=2` from the Barill paper.

src/axom/quest/FastApproximateGWN.hpp
jcs15c: It was actually the second order approximation, but yes I should point it out

src/axom/spin/policy/LinearBVH.hpp
jcs15c: As I understand it, a method like this couldn't go directly into the standard BVH interface because only `LinearBVH` describes how the data at internal nodes is actually arranged into arrays and arrayviews. This output of this function needs to align with that internal layout, and so if there was ever another implementation type for BVH, the actual logic of the function would have to change to reflect the internal layout of that other implementation. Does that seem consistent with how `BVH.hpp` and `LinearBVH.hpp` relate to one another?

src/axom/spin/policy/LinearBVH.hpp
jcs15c: Added!

src/axom/quest/FastApproximateGWN.hpp
jcs15c: Added some for 2D segments and 3D triangles!

src/axom/quest/GWNMethods.hpp
```diff
@@ -0,0 +1,923 @@
...
+      AXOM_ANNOTATE_SCOPE("extract_segments");
+
+      m_segments.resize(poly_mesh->getNumberOfCells());
+      auto segments_view = m_segments.view();
+
+      axom::mint::for_all_cells<axom::SEQ_EXEC, axom::mint::xargs::coords>(
+        poly_mesh,
+        AXOM_LAMBDA(axom::IndexType cellIdx,
+                    const axom::numerics::Matrix<double>& coords,
+                    [[maybe_unused]] const axom::IndexType* nodeIds) {
```
kennyweiss: hip is not happy w/ this attribute. Looks like we should change it to something like

```suggestion
                    const axom::IndexType* /*nodeIds*/) {
```

... and similarly for the one below

jcs15c: No complaints here!

src/axom/quest/examples/CMakeLists.txt
```diff
@@ -687,7 +687,7 @@ if(MFEM_FOUND AND OPENCASCADE_FOUND)
                 set(_res_k 5)
                 string(CONCAT _pass_regex
                     "INOUT_STATS:.*"
-                    "pos_dofs=16.*"
+                    "pos_dofs=14.*"
```
jcs15c, line 690: Just for future reference, the new OCC-derived AABB method has slightly different bounds, so there are different numbers of contained points.

PR #1792, opened 2026-02-03
Bugfix for regressions when loading sidre files in VisIt

kennyweiss:
# Summary

This PR contains bugfixes for regressions we noticed when saving visualization files from Sidre following the mesh blueprint convention.

- In https://github.com/llnl/axom/pull/1510, Sidre's View class changed its internal state representation for non-array types from "SCALAR" to "TUPLE" and removed an internal `sidre::View::State::SCALAR` type.
- This caused a problem loading these files into [VisIt](https://visit-dav.github.io/visit-website/)
- This PR adds a config variable `AXOM_SIDRE_IO_USE_SCALAR_STATE_STRING` for backward compatibility until VisIt can load Sidre files with "TUPLE" types.  This option is ~`OFF`~ `ON` by default. When enabled, it will use "SCALAR" instead of "TUPLE" for views containing a single item.
   - **Update:** Based on PR review feedback, I have enabled this option by default which will hopefully avoid problems until there is greater update for the `TUPLE` construct.
- The following VisIt PR will add support for "TUPLE": https://github.com/visit-dav/visit/pull/20811  (thanks @cyrush !)

- This PR also updates the Sidre Fortran interface for `sidre::Group::createViewScalar()` and `sidre::Group::createViewString()` to allow users to specify an allocator ID for the view's memory. These overloads were added in the same PR (#1510)

### DIFF
--- a/src/axom/sidre/core/SidreTypes.hpp
+++ b/src/axom/sidre/core/SidreTypes.hpp
@@ -76,14 +76,14 @@ inline bool nameIsValid(const std::string &name)
 }
 
 /*!
- * \brief Return Axom's invalid allocator id sentinel.
+ * \brief Return Axom's identifier for an invalid allocator
  *
  * This returns the value of \c axom::INVALID_ALLOCATOR_ID.
  */
 inline constexpr int getInvalidAllocatorID() noexcept { return axom::INVALID_ALLOCATOR_ID; }
 
 /*!
- * \brief Return Axom's malloc allocator id sentinel.
+ * \brief Return Axom's identifier for the malloc allocator
  *
  * This returns the value of \c axom::MALLOC_ALLOCATOR_ID.
  */
--- a/src/axom/sidre/core/View.cpp
+++ b/src/axom/sidre/core/View.cpp
@@ -1480,8 +1480,8 @@ char const* View::getStateStringName(State state)
 
 const char* View::getIoStateStringName() const
 {
-// Backward compatibility: prior to removing State::SCALAR, singleton tuple
-// data (setScalar()) was serialized as "SCALAR". Some downstream readers
+// Backward compatibility: prior to removing State::SCALAR, and adding State::TUPLE,
+// tuple types with one value were serialized as "SCALAR". Some downstream readers
 // (e.g. VisIt's Blueprint database plugin) still expect that string.
 #if defined(AXOM_SIDRE_IO_USE_SCALAR_STATE_STRING)
   if(m_state == TUPLE && isScalar()) return "SCALAR";
@@ -1491,12 +1491,11 @@ const char* View::getIoStateStringName() const
 }
 
 /*
- *************************************************************************
+ ******************************************************************************
  *
- * PRIVATE method returns state enum value when given string with a
- * state name.
+ * PRIVATE method returns state enum value when given string with a state name.
  *
- *************************************************************************
+ ******************************************************************************
  */
 View::State View::getStateId(const std::string& name) const
 {
@@ -1516,7 +1515,6 @@ View::State View::getStateId(const std::string& name) const
   else if(name == "SCALAR")
   {
     // Backward compatibility with files written before removing State::SCALAR
-    // (singleton tuple data).
     res = TUPLE;
   }
   else if(name == "TUPLE")
--- a/src/cmake/AxomOptions.cmake
+++ b/src/cmake/AxomOptions.cmake
@@ -26,13 +26,12 @@ option(AXOM_ENABLE_ALL_COMPONENTS "Enables all components by default" ON)
 option(AXOM_USE_64BIT_INDEXTYPE "Use 64-bit integers for axom::IndexType" ON)
 
 
-# When enabled, Sidre will serialize singleton tuple views (scalars) with
-# state="SCALAR" in its I/O metadata for compatibility with older downstream
-# readers (e.g. VisIt's Blueprint database plugin). When disabled (default),
-# Sidre will serialize these views with state="TUPLE".
+# When enabled (default), Sidre will serialize tuple views of size 1 with state="SCALAR" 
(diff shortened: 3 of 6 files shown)


src/axom/sidre/core/SidreTypes.hpp
```diff
@@ -74,6 +75,20 @@ inline bool nameIsValid(const std::string &name)
...
 
+/*!
+ * \brief Return Axom's invalid allocator id sentinel.
+ *
+ * This returns the value of \c axom::INVALID_ALLOCATOR_ID.
+ */
+inline constexpr int getInvalidAllocatorID() noexcept { return axom::INVALID_ALLOCATOR_ID; }
+
+/*!
+ * \brief Return Axom's malloc allocator id sentinel.
```
rhornung67: IIRC, the way allocator's work in Sidre is that you can attach one to a Group or View in the hierarchy and then it will be used for data allocation in the entire subtree rooted at the group (unless another allocator is attached to Group in the subtree) or View, respectively.  Is this sentinel the default allocator if none is associated with a Group or View as I've described?

kennyweiss: The default is `INVALID_ALLOCATOR_ID`  which eventually maps to Umpire's default (when using Umpire) or malloc (when not using Umpire.

See:
https://github.com/llnl/axom/blob/676ca97259735ce43c2446b02b6c80aa87772efe/src/axom/core/memory_management.hpp#L114-L127
and 
https://github.com/llnl/axom/blob/676ca97259735ce43c2446b02b6c80aa87772efe/src/axom/sidre/core/Group.cpp#L2285-L2311

src/axom/sidre/core/View.cpp
```diff
@@ -1501,6 +1513,12 @@ View::State View::getStateId(const std::string& name) const
   {
     res = EXTERNAL;
   }
+  else if(name == "SCALAR")
+  {
+    // Backward compatibility with files written before removing State::SCALAR
+    // (singleton tuple data).
```
cyrush: singleton implies other semantics  - there can be only one (highlander) vs there is only one (home alone)

src/axom/sidre/core/View.cpp
```diff
@@ -1478,6 +1478,18 @@ char const* View::getStateStringName(State state)
   return ret_string;
 }
 
+const char* View::getIoStateStringName() const
+{
+// Backward compatibility: prior to removing State::SCALAR, singleton tuple
+// data (setScalar()) was serialized as "SCALAR". Some downstream readers
+// (e.g. VisIt's Blueprint database plugin) still expect that string.
+#if defined(AXOM_SIDRE_IO_USE_SCALAR_STATE_STRING)
+  if(m_state == TUPLE && isScalar()) return "SCALAR";
```
cyrush, line 1487: If we save a tuple that is actually an array - VisIt's released logic would still likely work if you also use "SCALAR". 

TUPLE as array could happen for some Mesh metadata -- not sure the probability of it occurring.

kennyweiss: @gunney1 -- what did we do for small tuple arrays before your changes in #1510 -- I think we treated those as BUFFER arrays, so we don't have those cases. (I'll check)

src/axom/sidre/core/View.cpp
kennyweiss: :) -- thanks -- I'll fix the comment

src/axom/sidre/core/View.cpp
cyrush: I think that is true, it wasn't possible before -- but it is possible now -- in the past they would be attached to a buffer, but now they could be tagged TUPLE.

If tagged with SCALAR for the output file -- it may work in VisIt b/c that case we simply `set` using the inline values. 

However the visit logic does not apply a schema, like it would the buffer case.

src/cmake/AxomOptions.cmake
```diff
@@ -25,6 +25,16 @@ option(AXOM_ENABLE_SPARSEHASH "Enables Sparsehash." ON)
...
 option(AXOM_USE_64BIT_INDEXTYPE "Use 64-bit integers for axom::IndexType" ON)
 
+
+# When enabled, Sidre will serialize singleton tuple views (scalars) with
+# state="SCALAR" in its I/O metadata for compatibility with older downstream
+# readers (e.g. VisIt's Blueprint database plugin). When disabled (default),
+# Sidre will serialize these views with state="TUPLE".
+option(AXOM_SIDRE_IO_USE_SCALAR_STATE_STRING
+       "Write sidre View scalars with state='SCALAR' (legacy compatibility); default writes state='TUPLE'."
+       OFF)
```
Arlie-Capps: I prefer default ON because the need for OFF (that is, using "TUPLE" instead of "SCALAR") is a small new population.  The old default should keep working for everyone else, correct?  Keeping the old default but letting those who need it set the new behavior is the gentlest way forward.  Plus it allows us to test in a somewhat more controlled manner.

Arlie-Capps (approved):
@kennyweiss , thank you for this expeditious and thorough PR.  Please consider having `AXOM_SIDRE_IO_USE_SCALAR_STATE_STRING` default to ON: more discussion at AxomOptions.cmake.

src/axom/sidre/core/View.cpp
gunney1: I think that before https://github.com/llnl/axom/pull/1510 anything with more than 1 element is an array (unless it's a string), so it's stored in a BUFFER (unless it's external).

src/axom/sidre/core/View.cpp
kennyweiss: done

src/cmake/AxomOptions.cmake
kennyweiss: Done.

PR #1796, opened 2026-02-06
Reduce compiler warnings for CUDA

BradWhitlock:
This PR takes a stab at reducing the number of compiler warnings issued on CUDA machines. There are still various warnings about using host functions in host device code that have not yet been addressed.

* Various fixes for compiler warnings
* Suppress warning for host `constexpr` functions in host device code in one of the matrix host-configs
* Work around a test failure where Conduit was unable to read a YAML baseline _(we convert to JSON on the fly and use that)_. It seems like a Conduit problem with that particular build.

### DIFF
--- a/COPYRIGHT
+++ b/COPYRIGHT
@@ -1,6 +1,20 @@
 Intellectual Property Notice
 ------------------------------
 
+Axom is produced at the Lawrence Livermore National Laboratory
+
+Unlimited Open Source - BSD Distribution
+LLNL-CODE-825966
+CP02462
+
+Previous open source releases:
+Unlimited Open Source - BSD Distribution
+LLNL-CODE-741217
+OCEC-17-187
+
+This file is part of Axom
+For details, see https://github.com/llnl/axom
+
 Axom is licensed under the BSD 3-Clause license
 (BSD-3-Clause or https://opensource.org/licenses/BSD-3-Clause).
 
--- a/LICENSE
+++ b/LICENSE
@@ -1,23 +1,7 @@
+BSD 3-Clause License
+
 Copyright (c) 2017-2026, Lawrence Livermore National Security, LLC.
-
-Produced at the Lawrence Livermore National Laboratory
-
-Unlimited Open Source - BSD Distribution
-LLNL-CODE-825966
-CP02462
-
-Previous open source releases:
-Unlimited Open Source - BSD Distribution
-LLNL-CODE-741217
-OCEC-17-187
-
 All rights reserved.
-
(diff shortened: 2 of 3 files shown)


kennyweiss (approved):
Thanks @BradWhitlock !

src/axom/bump/tests/blueprint_testing_helpers.hpp
```diff
@@ -448,16 +451,166 @@ void saveBaseline(const std::vector<std::string> &baselinePaths,
...
+ * \brief This routine converts a YAML file to a JSON file. Ordinarily, we could
+ *        use Conduit for this but this works around a case where Conduit is
+ *        failing to read certain YAML baselines.
+ *
+ * \param The YAML filename to convert.
+ * \param The output JSON filename.
+ *
+ * \return True on success; False otherwise.
+ */
+bool convert_yaml_json(const std::string &yaml_filename, const std::string &json_filename)
```
gunney1, line 465: Is this a temporary work-around until conduit is fixed?  Is the idea to remove this at that point?

src/axom/bump/tests/blueprint_testing_helpers.hpp
```diff
@@ -448,16 +451,166 @@ void saveBaseline(const std::vector<std::string> &baselinePaths,
...
+#endif
+
+bool loadBaseline(const std::string &filename, const std::string &protocol, conduit::Node &n)
 {
   bool loaded = false;
-  std::string file_with_ext(filename + ".yaml");
+  std::string file_with_ext(filename + "." + protocol);
   //SLIC_INFO(axom::fmt::format("Load baseline {}", file_with_ext));
-  if(axom::utilities::filesystem::pathExists(file_with_ext))
+  const int MAX_ATTEMPTS = 3;
```
gunney1, line 534: Please add a comment explaining why we would expect different results from multiple attempts.

src/axom/core/FlatMapView.hpp
```diff
@@ -46,7 +46,8 @@ class FlatMapView : detail::flat_map::SequentialLookupPolicy<typename Hash::resu
   using FlatMapType =
     std::conditional_t<IsConst, const FlatMap<KeyType, ValueType, Hash>, FlatMap<KeyType, ValueType, Hash>>;
 
-  AXOM_HOST_DEVICE FlatMapView() = default;
+  // NOTE: Defaulted constructors do not need to be marked AXOM_HOST_DEVICE
```
gunney1, line 49: Thanks for this comment.

Issue #1797, opened 2026-02-10
`TetMeshClipper` needs the `fixOrientation` implemented


gunney1:
If user's tets have the wrong orientation, will get bad clip volumes.  The code dox mentions using the `fixOrientation` flag, but I forgot to implement it.

PR #1800, opened 2026-02-12
Check and fix tet orientations in `TetMeshClipper`

gunney1:
# Summary

- This PR is a feature
- It does the following:
  - Check and fix orientation of tets for clipping a tet mesh.  This feature was advertised in the `TetMeshClipper` class comments, but I had forgotten to implement it.
  - Resolves issue https://github.com/llnl/axom/issues/1797


src/axom/quest/detail/clipping/TetMeshClipper.cpp
```diff
@@ -656,10 +656,98 @@ void TetMeshClipper::extractClipperInfo()
...
+        else
+        {
+          SLIC_ERROR(axom::fmt::format(
+            "TetMeshClipper: input tet[{}] is inverted, with volume {}.  Please fix or add "
+            "'fixOrientation: 1' to the input klee::Geometry's hierarchical data.",
+            iTet,
+            signedVol));
+        }
+      }
+    });
```
kennyweiss, line 750: This approach, modifies the underlying connectivity arrays, and runs even when the user does not request `fixOrientation`. It also always throws an error if the user does not modify the connectivity if there is a negatively oriented tetrahedron.

Another approach might be to optionally add a pass that calls `Tetrahedron::checkAndFixOrientation()` on each tet in the array before running the clipping algorithm (without modifying the connectivity array).
And if the user doesn't ask for `fixOrientation`, they can get negative volumes.

Is there a reason to prefer the former to the latter?

gunney1: What I do is the same as `checkAndFixOrientation`.  I just do it manually because I want to error out if the user sets `fixOrientation` to false, effectively saying that they know what they're doing.  I can't count on negative volumes working out in any logical way.

Negative volumes don't behave the way one may think it should because of the various checks we have.  It actually could yield zero instead of a negative number.  To avoid such unexpected behavior, I thought it best to fix or reject the input based on `fixOrientation` right when we get it.  So that negative volumes are avoided.

kennyweiss: I was suggesting to (a) modify the primal::Tetrahedra after reading them in, but before clipping, rather than modifying the conduit connectivity array, which would allow you to use the `checkAndFix` function  (b) default to `fixOrientation = true`  since the code won't work without it (but, maybe let the user override that

gunney1: Yeah, either would work.  Note that the user's connectivity array isn't changed.  It's the internal one that gets fixed.

kennyweiss: Working on the tetrahedra and calling `tet.checkAndFixOrientation()` seems cleaner to me

kennyweiss: Also, if we're always checking the orientation (to fix or warn/error) and the code won't work with bad orientations, we might as well always fix it and skip the extra option, right?

gunney1: Oh, another reason to fix the connectivity is to do it right when we get the input.  The array of tets isn't generated until later, when we know what memory space to put it in.  Not a huge reason, but that's how my brain sorted out the steps.

src/axom/quest/tests/quest_mesh_clipper.cpp
```diff
@@ -616,7 +616,7 @@ axom::klee::Geometry createGeom_TetMesh(sidre::DataStore& ds, const std::string&
   tetMesh.appendNode(+lll, -lll, +lll);
   axom::IndexType conn0[4] = {0, 1, 2, 3};
   tetMesh.appendCell(conn0);
-  axom::IndexType conn1[4] = {4, 5, 6, 7};
+  axom::IndexType conn1[4] = {4, 5, 7, 6};
```
bmhan12: Can you mark here and elsewhere where the orientation is purposefully set incorrectly for testing purposes?

src/axom/quest/detail/clipping/TetMeshClipper.hpp
```diff
@@ -113,6 +113,12 @@ class TetMeshClipper : public MeshClipperStrategy
   // Extract clipper info from MeshClipperStrategy::m_info.
   void extractClipperInfo();
 
+  //! @brief Check tet orientations, and possibly fix bad ones.
```
bmhan12: Based on direction of discussion here: https://github.com/llnl/axom/pull/1800/changes#r2823897343, should document here if this function fixes the issue by correcting the connectivity at the conduit Node level or at the data/Tetrahedron level.

src/axom/quest/tests/quest_mesh_clipper.cpp
gunney1: Yes.  Good idea!

src/axom/quest/detail/clipping/TetMeshClipper.hpp
gunney1: Done.  Thanks!

PR #1805, opened 2026-02-18
Feature/bergel1/lumberjack stream ljvec

gberg617:
# Summary
The goal of this PR is to:

- Use multiple Lumberjack objects within the same LumberjackStream instance, each with its own combiner criteria.
- Log messages from those Lumberjack objects to the same file.
- Preserve a global sort by timestamp across all messages for a given file managed by a LumberjackStream instance.

The key ideas are:

- LumberjackStream now holds a vector of Lumberjack objects instead of a single one.
- Each Lumberjack object can be configured to only process messages with certain tags, if tags are provided.
- The sorting responsibility moves from Lumberjack to LumberjackStream, which gathers messages from all of its Lumberjack instances and performs a single sort by timestamp before writing.

This avoids the expensive duplicate checking for tagged messages that are not intended to be combined, and still allows:

- Sorting all messages that go to a given file.
- Using different combiner strategies for different tags.

### DIFF
--- a/src/axom/slic/streams/LumberjackStream.cpp
+++ b/src/axom/slic/streams/LumberjackStream.cpp
@@ -266,43 +266,48 @@ void LumberjackStream::write(bool local)
     if(lj->isOutputNode() || local) {
       std::vector<lumberjack::Message*> lj_messages = lj->getMessages();
       messages.insert(messages.begin(), lj_messages.begin(), lj_messages.end());
+    }
+  }
+
+  std::sort(messages.begin(), messages.end(), [](lumberjack::Message* const a, lumberjack::Message* const b) {
+    return a->creationTime() < b->creationTime();
+  });
+
+  for(const auto* curr_message : messages)
+  {
+    if(curr_message == nullptr)
+    {
+      continue;
+    }
+
+    if(m_isOstreamOwnedBySLIC && !m_opened)
+    {
+      std::ofstream* ofs = dynamic_cast<std::ofstream*>(m_stream);
+      if(ofs != nullptr)
+      {
+        ofs->open(m_file_name);
+        m_opened = true;
+      }
+    }
+
+    (*m_stream) << this->getFormatedMessage(
+      message::getLevelAsString(static_cast<message::Level>(curr_message->level())),
+      curr_message->text(),
+      curr_message->tag(),
+      curr_message->stringOfRanks(),
+      std::to_string(curr_message->count()),
+      curr_message->fileName(),
+      curr_message->lineNumber());
+  }
+
+  m_stream->flush();
+
+  for (auto* lj : m_lj) {
+    if(lj->isOutputNode() || local) {
       lj->clearMessages();
     }
-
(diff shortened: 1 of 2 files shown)


src/axom/slic/streams/LumberjackStream.cpp
```diff
@@ -205,97 +224,125 @@ void LumberjackStream::outputLocal()
...
+    if(lj->isOutputNode() || local) {
+      lj->clearMessages();
+    }
+  }
+
 }
 
 //------------------------------------------------------------------------------
-bool LumberjackStream::hasPendingMessages() { return m_lj->getMessages().size() > 0; }
+bool LumberjackStream::hasPendingMessages() {
```
gberg617, line 314: This needs to be fixed.

src/axom/slic/streams/LumberjackStream.cpp
```diff
@@ -205,97 +224,125 @@ void LumberjackStream::outputLocal()
...
+    return lj->getMessages().size() > 0; 
+  }
+  return false;
+}
 
 //------------------------------------------------------------------------------
 bool LumberjackStream::isUsingMPI() { return true; }
 
-MPI_Comm LumberjackStream::comm() { return m_lj->getCommunicator()->comm(); }
+MPI_Comm LumberjackStream::comm() { return m_lj[0]->getCommunicator()->comm(); }
```
gberg617, line 324: This also needs to be fixed.

gberg617:
> > Thanks for clearly laying out the issue, @gberg617 !
> > I wonder if instead of having multiple LumberjackStream objects each corresponding to different tags that filter out different messages to queue by `queueMessage()`, we can stick with a single LumberjackStream object and have a blacklist vector of tag(s) that do not get combined in `combineMessages()`? ([link](https://github.com/llnl/axom/blob/5ce7042f6179c643c522f1b7456bfc2bcd6eedf7/src/axom/lumberjack/Lumberjack.cpp#L234-L287) to combineMessages() function)
> > So instead of these non-combinable messages being added to `finalMessages` and iterated over repeatedly, they get added to a separate vector i.e.`uncombinableMessages` and get appended back to `m_messages` at the very end?
> 
> @bmhan12 That's a good point. I think your idea can work. The advantage of your approach is that we don't change the behavior in LumberjackStream, and all the changes stay localized within Lumberjack. However, this would require me to keep the custom Combiner class I wrote in my host code that inspects the tag to determine whether messages should be combined. At first, this seemed somewhat undesirable, but perhaps it's still better than modifying LumberjackStream.

@bmhan12 After thinking about this, I slightly favor the approach in the current PR. I think it is generally useful to let a single LumberjackStream instance cleanly handle different combiner strategies for one file, instead of creating multiple LumberjackStream objects that all own the same file but use different combiners, or relying on custom Combiner classes like I have been doing.  Moving the message sorting responsibility from Lumberjack into LumberjackStream in slic also improves the separation of responsibilities. LumberjackStream is now responsible for formatting and writing sorted messages to a file, and Lumberjack focuses on combining and passing messages between ranks.  if you feel this isn't the best approach, I'm happy to discuss it further.

PR #1808, opened 2026-02-24
Add Conduit's python interface, yapf python formatter, Spack updates

### LINKED ISSUE #1742: Spack - Remove conduit+python dependency on devtools
As part of pysidre development, workaround related to conduit/python/devtools was added.

This should be corrected so axom only builds conduit+python with `+python` variant (devtools variant should not matter).

_Originally posted by @kennyweiss in https://github.com/LLNL/axom/pull/1740#discussion_r2615428818_

bmhan12:
This PR:
- Add Conduit's python interface to pass Conduit Node between C++ and python.
- Add yapf, a [Python formatter,](https://github.com/google/yapf)  to format `*.py` files. Config file is `.style.yapf`
- Spack updates:
  - Update to Spack v1.1.1
    - Move caliper commit forward for `rocprofiler-sdk` dependency
  - Removes `radiuss-spack-configs` as a git submodule and from our  `spack_packages_path`
  - Removes conduit+python from +devtools dependency
    -  Resolves #1742

### DIFF
--- a/src/axom/multimat/multimat.cpp
+++ b/src/axom/multimat/multimat.cpp
@@ -1502,9 +1502,9 @@ axom::Array<DataType> TransposeDenseImpl(const MultiMat::DenseField2D<DataType> 
 
   // Note: even though this is a dense field, we iterate over the relation set
   // in order to only copy over filled-in slots.
-  ExecLambdaForMemory(
+  // Note: Execute sequentially to avoid intermittent error with HIP
+  axom::for_all<axom::SEQ_EXEC>(
     relationSet->totalSize() * stride,
-    allocatorId,
     AXOM_LAMBDA(int index) {
       int flatIdx = index / stride;
       int comp = index % stride;


src/axom/sidre/nanobind_sidre.cpp
```diff
@@ -151,6 +154,60 @@ void bindIterator(nb::module_& m, const char* iterator_name)
...
 
+/*!
+ * \brief Helper function that converts a C++ Conduit Node into a python nanobind object
+ *        that is of type Node in python.
+ *
+ * \param [in] node C++ Conduit Node
+ *
+ * \return A python nanobind object
+ */
+nb::object nodeToNbObject(conduit::Node& node)
```
bmhan12, line 165: The two functions added here, `nodeToNbObject` and `nbObjectToNode`, converts a C++ Conduit::Node to a python nanobind object and vice versa.

src/axom/bump/extraction/tables/split/convert_clip_cases.py
```diff
@@ -7,44 +7,47 @@
 #       clipping algorithms to produce fewer zones and fewer artificially
 #       introduced mesh lines.
 
-This script converts the polygonal clipping tables in this directory 
+
+# This script converts the polygonal clipping tables in this directory
```
bmhan12, line 11: Neat - When calling `make style`, `yapf` threw an error for this uncommented comment.

scripts/spack/specs.json
```diff
@@ -15,31 +15,34 @@
     "__comment__":"##############################################################################",
 
     "__comment__":"# mfem disabled for intel (icpx compiler error)",
+    "__comment__":"# python disabled for intel (mixed compiler conduit/caliper MPI spack compiler error)",
```
bmhan12, line 18: Documented in `specs.json` variant changes I needed to make.
- For toss_4 `intel` compiler, was seeing compiler mixing with Spack v1.1.1 + `+python` variant. Set to `~python`.
- For toss_4_cray, needed to `~caliper` for `rocm@6.3.1`, as caliper couldn't compile with `rocm@6.3.1`.

scripts/spack/configs/toss_4_x86_64_ib_cray/spack.yaml
```diff
@@ -15,8 +15,13 @@ spack:
...
 
-  # Regular TPLs do not need views
-  view: false
+  # View to get python modules
+  view:
+    default:
+      root: ../view
+      select: [python, ^python]
+      projections:
+        all: '{name}-{version}'
```
bmhan12: Necessary view as part of moving `conduit+python` dependency to axom's `+python` variant instead of building devtools prior to building axom.

scripts/spack/packages/axom/package.py
```diff
@@ -229,14 +226,14 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
         ext_cuda_dep = f"+cuda cuda_arch={val}"
         depends_on(f"raja {ext_cuda_dep}", when=f"+raja {ext_cuda_dep}")
         depends_on(f"umpire {ext_cuda_dep}", when=f"+umpire {ext_cuda_dep}")
-        depends_on(f"caliper {ext_cuda_dep}", when=f"^caliper {ext_cuda_dep}")
+        depends_on(f"caliper {ext_cuda_dep}", when=f"{ext_cuda_dep} ^caliper")
```
bmhan12, line 234: Subtle bugfix here.

Previously, would add `cuda_arch` if axom spec was `+variants %compiler ^caliper+cuda cuda_arch=XX` (no-op).
Change will now match axom spec when `+variants+cuda cuda_arch=XX %compiler ^caliper` (intended behavior).

.uberenv_config.json
```diff
@@ -4,10 +4,10 @@
 "package_final_phase": "initconfig",
 "package_source_dir": "../..",
 "spack_url": "https://github.com/spack/spack.git",
-"spack_commit": "0c2be44e4ece21eb091ad5de4c97716b7c6d4c87",
+"spack_commit": "2e2169d5282d166f63e3ee4db8d4446c43cefa8a",
 "spack_configs_path": "scripts/spack/configs",
-"spack_packages_commit": "ba7ec6e6121679029b2f4ccff8c3806f0450e563",
-"spack_packages_path": ["scripts/spack/radiuss-spack-configs/spack_repo/llnl_radiuss/packages", "scripts/spack/packages"],
+"spack_packages_commit": "17ea6a6d483fc0bde1b2cca96ddaad0a10f2ef69",
+"spack_packages_path": "scripts/spack/packages",
```
kennyweiss, line 10: 👍

RELEASE-NOTES.md
kennyweiss, line 1: Please note that the `radiuss-spack-configs` submodule was removed.

.uberenv_config.json
```diff
@@ -4,10 +4,10 @@
 "package_final_phase": "initconfig",
 "package_source_dir": "../..",
 "spack_url": "https://github.com/spack/spack.git",
-"spack_commit": "0c2be44e4ece21eb091ad5de4c97716b7c6d4c87",
+"spack_commit": "2e2169d5282d166f63e3ee4db8d4446c43cefa8a",
```
kennyweiss, line 7: What version of spack does this correspond to?

host-configs/dane-toss_4_x86_64_ib-gcc@13.3.1.cmake
```diff
@@ -73,72 +73,74 @@ set(ENABLE_GTEST_DEATH_TESTS ON CACHE BOOL "")
...
 # Devtools
 #------------------------------------------------------------------------------
 
-set(DEVTOOLS_ROOT "/collab/usr/gapps/axom/devtools/toss_4_x86_64_ib/2025_09_25_11_33_37" CACHE PATH "")
+set(DEVTOOLS_ROOT "/collab/usr/gapps/axom/devtools/toss_4_x86_64_ib/2026_02_17_14_42_14" CACHE PATH "")
 
 set(CLANGFORMAT_EXECUTABLE "/usr/tce/packages/clang/clang-19.1.3/bin/clang-format" CACHE PATH "")
 
-set(Python_EXECUTABLE "${DEVTOOLS_ROOT}/._view/qoeiu74kfjckl7xkcpsbmskffr2fq2tm/python-3.13.5/bin/python3" CACHE PATH "")
+set(Python_EXECUTABLE "/usr/WS1/axom/libs/toss_4_x86_64_ib/2026_02_20_07_57_59/view/python-3.13.11/bin/python3" CACHE PATH "")
```
kennyweiss: This is unexpected. I wonder why it's picking up a different python?  (from `/usr/workspace` instead of `/collab`

kennyweiss: They all seem to be python-3.13.11

host-configs/dane-toss_4_x86_64_ib-intel-oneapi-compilers@2025.2.0.cmake
```diff
@@ -75,52 +75,54 @@ set(ENABLE_GTEST_DEATH_TESTS ON CACHE BOOL "")
...
 # Devtools
 #------------------------------------------------------------------------------
 
-set(DEVTOOLS_ROOT "/collab/usr/gapps/axom/devtools/toss_4_x86_64_ib/2025_09_25_11_33_37" CACHE PATH "")
+set(DEVTOOLS_ROOT "/collab/usr/gapps/axom/devtools/toss_4_x86_64_ib/2026_02_17_14_42_14" CACHE PATH "")
 
 set(CLANGFORMAT_EXECUTABLE "/usr/tce/packages/clang/clang-19.1.3/bin/clang-format" CACHE PATH "")
 
-set(Python_EXECUTABLE "${DEVTOOLS_ROOT}/._view/qoeiu74kfjckl7xkcpsbmskffr2fq2tm/python-3.13.5/bin/python3" CACHE PATH "")
+set(Python_EXECUTABLE "${DEVTOOLS_ROOT}/._view/kky5oskfhacihjb5gpy7qkeh6edlrtcb/python-3.13.11/bin/python3" CACHE PATH "")
```
kennyweiss, line 112: This one has the right python

host-configs/dane-toss_4_x86_64_ib-llvm@19.1.3.cmake
```diff
@@ -75,72 +75,74 @@ set(ENABLE_GTEST_DEATH_TESTS ON CACHE BOOL "")
...
 # Devtools
 #------------------------------------------------------------------------------
 
-set(DEVTOOLS_ROOT "/collab/usr/gapps/axom/devtools/toss_4_x86_64_ib/2025_09_25_11_33_37" CACHE PATH "")
+set(DEVTOOLS_ROOT "/collab/usr/gapps/axom/devtools/toss_4_x86_64_ib/2026_02_17_14_42_14" CACHE PATH "")
 
 set(CLANGFORMAT_EXECUTABLE "/usr/tce/packages/clang/clang-19.1.3/bin/clang-format" CACHE PATH "")
 
-set(Python_EXECUTABLE "${DEVTOOLS_ROOT}/._view/qoeiu74kfjckl7xkcpsbmskffr2fq2tm/python-3.13.5/bin/python3" CACHE PATH "")
+set(Python_EXECUTABLE "/usr/WS1/axom/libs/toss_4_x86_64_ib/2026_02_20_07_57_59/view/python-3.13.11/bin/python3" CACHE PATH "")
```
kennyweiss: This one has the local python

scripts/spack/packages/axom/package.py
kennyweiss: Should that also be applied to the raja and umpire lines right above?

kennyweiss: ... and in the rocm section right below?

scripts/spack/packages/axom/package.py
```diff
@@ -672,10 +676,17 @@ def initconfig_package_entries(self):
             entries.append("# ClangFormat disabled since llvm@19 and devtools not in spec\n")
             entries.append(cmake_cache_option("ENABLE_CLANGFORMAT", False))
 
-        if spec.satisfies("+python") or spec.satisfies("+devtools"):
+        if spec.satisfies("+python"):
+            # Get path to python view with modules
+            python_view_exe = pjoin(os.path.dirname(os.path.dirname(self.prefix)), "view",
+                spec["python"].name + "-" + str(spec["python"].version), "bin/python3")
+            entries.append(cmake_cache_path("Python_EXECUTABLE", python_view_exe))
```
kennyweiss: Presumably, these line are responsible for using the local python rather than the devtools one. 
Is this the right approach?

Tag: @white238 @cyrush

scripts/spack/specs.json
```diff
@@ -15,31 +15,34 @@
...
     "__comment__":"# Configs are for dane/rzwhippet",
+    "__comment__":"# Order matters - if ~python spec is built last this removes view of python",
     "toss_4_x86_64_ib":
-    [ "+devtools+hdf5+mfem+c2c+scr+adiak+caliper %gcc_13",
-      "+devtools+hdf5+mfem+c2c+scr+adiak+caliper+opencascade %clang_19",
-      "+devtools+hdf5~mfem+c2c+adiak+caliper %intel_25" ],
+    [ "~python+devtools+hdf5~mfem+c2c+adiak+caliper %intel_25",
```
kennyweiss, line 22: Do you remember if we disabled mfem w/ intel_25 for a specific reason? 
Should we create an issue to fix/re-enable?

scripts/spack/specs.json
kennyweiss: Should we report this to the spack team?  Perhaps they're already aware?

RELEASE-NOTES.md
kennyweiss: Please also mention which version of spack our uberenv was updated to.

scripts/spack/packages/axom/package.py
```diff
@@ -259,8 +263,8 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
         depends_on("python")
         depends_on("py-sphinx")
         depends_on("py-shroud")
-        depends_on("py-pytest")
         depends_on("py-jsonschema")
+        depends_on("py-yapf")
```
kennyweiss, line 272: Since devtools depends on python, perhaps we should use `^python` rather than `+python` when checking the spec for python dependencies.  (i.e. use the dependency rather than the variant, except in cases where we specifically care about the variant)

src/axom/bump/extraction/tables/split/convert_clip_cases.py
```diff
@@ -321,52 +331,88 @@ def make_cut_cases(sizes, offsets, shapes):
...
-   outSizes, outOffsets, outShapes = convert_clip_cases("ClipCasesPoly6.cpp", "Poly6", 6, max_points=6)
-   write_new_tables("../clipping/ClipCasesPoly6.cpp", "Poly6", tableNames, outSizes, outOffsets, outShapes)
-
-   outSizes, outOffsets, outShapes = convert_clip_cases("ClipCasesPoly7.cpp", "Poly7", 7, max_points=7)
-   write_new_tables("../clipping/ClipCasesPoly7.cpp", "Poly7", tableNames, outSizes, outOffsets, outShapes)
+def make_polygonal_clip_tables():
+    tableNames = ["Clip", "clip", "clipping"]
+
+    outSizes, outOffsets, outShapes = convert_clip_cases("ClipCasesQua.cpp", "Qua", 4, max_points=4)
+    write_new_tables("../clipping/ClipCasesQua.cpp", "Qua", tableNames, outSizes, outOffsets,
```
kennyweiss: Does yapf have an equivalent to `clang-format {off,on}`?

kennyweiss: Looks like:
```python
# yapf: disable

# yapf: enable
```
https://github.com/google/yapf?tab=readme-ov-file#why-does-yapf-destroy-my-awesome-formatting

Should we add it here (or elsewhere) as an example of disabling/enabling formatting within python?

src/tools/gen-multidom-structured-mesh.py
```diff
@@ -7,109 +7,117 @@
...
-  import conduit
-  import conduit.blueprint
-  import conduit.relay
+    import conduit
+    import conduit.blueprint
+    import conduit.relay
 except ModuleNotFoundError as e:
-  print(f'{e}\nMake sure your PYTHONPATH includes /path/to/conduit/install/python-modules\nConduit must be configured with python and hdf5.')
-  exit(-1)
+    print(
```
kennyweiss, line 16: Does this line need to be updated? Hopefully, it should work with our `run_python_with_axom` script (please confirm).

src/cmake/thirdparty/SetupAxomThirdParty.cmake
```diff
@@ -326,6 +326,10 @@ if(EXISTS ${Python_EXECUTABLE})
     endif()
 endif()
 
+if(nanobind_ROOT AND NOT CONDUIT_PYTHON_ENABLED)
+    message(FATAL_ERROR "Conduit python module is required if nanobind python module is provided.")
+endif()
```
kennyweiss: Should conduit python only be required when using sidre?  I.e. can we enable nanobind when `nanobind_ROOT`, but disable the sidre python bindings if conduit python is not available?

Perhaps this is fine for now, and we can hold off on this extra logic until we actually have nanobind python bindings for other components.

scripts/spack/configs/versions.yaml
```diff
@@ -6,11 +6,11 @@ packages:
   adiak:
     require: 
       - spec: "@0.4.0"
-  # 2.14.0
+  # Newer commit than 2.14.0 that fixes rocprofiler dependency
   caliper:
     require:
-      - spec: "@git.9ab0e64d2e328d1fdff2ffebb53f30a899d6a598=master"
-  # v2025.12.0
+       - spec: "@git.7e5b7a5c0eacc077f9b842abf41c9fc7b996ce0c=master"
```
Arlie-Capps: added initial space?

src/axom/sidre/tests/sidre_attribute_Py.py
```diff
@@ -34,918 +35,913 @@
...
 
-	node1 = view1a.getAttributeNodeRef(attr_color)
-	assert node1.as_string() == g_color_red
+    node1 = view1a.getAttributeNodeRef(attr_color)
+    assert node1.value() == g_color_red
 
-	node2 = view1a.getAttributeNodeRef(attr_dump)
-	assert node2.as_int() == g_dump_no
+    node2 = view1a.getAttributeNodeRef(attr_dump)
+    assert node2.value() == g_dump_no
```
kennyweiss, line 410: Nice!

src/axom/sidre/tests/sidre_attribute_Py.py
```diff
@@ -34,918 +35,913 @@
...
 
-	node3 = view1a.getAttributeNodeRef(g_attr_null)
-	assert node3.schema().dtype().is_empty()
-"""
+    # "None" is not type Attribute
+    # node3 = view1a.getAttributeNodeRef(g_attr_null)
+    # assert node3.schema().dtype() == DataType.empty()
```
kennyweiss: Are we expecting this one to work but need to fix something to make it work? 
Or is it a comment that we don't expect this work?

kennyweiss (approved):
Thanks @bmhan12 -- this is a nice contribution!

My only real question is about if we want the local python as the `Python_EXECUTABLE` as opposed to the one from `devtools`, and if that might lead to incompatibilities. I don't think that should hold this PR back, but we should probably reach consensus on it.

scripts/spack/specs.json
bmhan12: The first issue I need to narrow down more. Not sure exactly if it's because of the compiler's recipe, Spack version, python, or something else.

The caliper issue I created a posted here: https://github.com/llnl/Caliper/issues/725

scripts/spack/packages/axom/package.py
bmhan12: > Should that also be applied to the raja and umpire lines right above?

I think `+raja/umpire` should be fine as-is. The `^caliper` was to catch the fact that there are two accepted variants/ways of getting the dependency , raja/umpire just have the one-way through their respective variants.

scripts/spack/configs/versions.yaml
bmhan12: Fixed!

RELEASE-NOTES.md
bmhan12: notes updated!

.uberenv_config.json
bmhan12: > What version of spack does this correspond to?

v1.1.1 ; I've updated release notes to reflect it here.

scripts/spack/packages/axom/package.py
bmhan12: > Presumably, these line are responsible for using the local python rather than the devtools one.
Is this the right approach?

Yes. 
The devtools python has its own environment/view/??? (not sure the right terminology) that does not include the nanobind-related dependencies, while the local one to the Axom TPL installation does.

Very curious to know if there's a better approach.

scripts/spack/specs.json
bmhan12: > Do you remember if we disabled mfem w/ intel_25 for a specific reason?

Yes, I left a comment higher up here, and just opened an mfem issue (confirmed that the error is still occurring): https://github.com/mfem/mfem/issues/5250

scripts/spack/packages/axom/package.py
bmhan12: To address #1742, I think we want to keep it to `+python`, otherwise we get the current behavior where specifying `+devtools` will build nanobind, `conduit+python`, and the rest of those pysidre dependencies.

src/axom/bump/extraction/tables/split/convert_clip_cases.py
bmhan12: Added to the function here!

src/tools/gen-multidom-structured-mesh.py
bmhan12: > Does this line need to be updated? Hopefully, it should work with our run_python_with_axom script (please confirm).

Confirmed that python file works with `run_python_with_axom.sh`, and updated error message to include the script as an alternative run option to specifying PYTHONPATH.

src/cmake/thirdparty/SetupAxomThirdParty.cmake
bmhan12: > Perhaps this is fine for now, and we can hold off on this extra logic until we actually have nanobind python bindings for other components.

Documented this in #1609!

src/axom/sidre/tests/sidre_attribute_Py.py
bmhan12: Expectation is for this not to work - updated comment to reflect the difference between C++/python.

scripts/spack/packages/axom/package.py
bmhan12: Recapping: @white238 took a look, pointed out this logic is fragile, relying on a hard-coded path to python  specified in the `spack.yaml` configuration.

Tagging @becker33 to explore a better approach.

cyrush: Note: To have a usable host-config, I have to patch my host-config to use the python within the final spack view. That is part of this logic:

https://github.com/llnl/uberenv/pull/155

The issue for me is that I couldn't record the proper path to the view at spack build time. 

Separate devtools likely makes this even more complicated, or maybe it's unrelated. 

As you move to use python as a core dependency, maybe we should rethink it as part of separate devtools

scripts/spack/packages/axom/package.py
```diff
@@ -701,6 +717,14 @@ def initconfig_package_entries(self):
...
                 )
 
+        if spec.satisfies("+python"):
+            # pytest requires pluggy and iniconfig
+            for dep in ("py-nanobind", "py-pytest", "py-numpy", "py-pluggy", "py-iniconfig"):
+                if spec.satisfies("^{0}".format(dep)):
+                    dep_dir = get_spec_path(spec, dep, path_replacements, use_lib=True)
+                    py_libdir = join_path(dep_dir, f"python{spec['python'].version.up_to(2)}", "site-packages")
+                    entries.append(cmake_cache_path("%s_DIR" % dep.upper().replace("-", "_"), py_libdir))
+
```
bmhan12, line 727: The python module installations paths are added here to the host-config. 

Naming wise, the format is `py-pytest` --> `PY_PYTEST_DIR`. Open to suggestions for naming!

src/cmake/thirdparty/SetupAxomThirdParty.cmake
```diff
@@ -311,19 +311,55 @@ if(EXISTS ${Python_EXECUTABLE})
...
+   OR NOT PY_NUMPY_DIR
+   OR NOT PY_PYTEST_DIR
+   OR NOT PY_PLUGGY_DIR
+   OR NOT PY_INICONFIG_DIR))
+    message(FATAL_ERROR
+      "Axom's python extensions require nanobind, numpy, pytest, and conduit."
+      "\nThe python library installation paths "
+      "(and pytest's dependencies pluggy and iniconfig) "
+      "can be specified with CMake variables: "
+      "PY_NANOBIND_DIR, CONDUIT_PYTHON_MODULE_DIR, PY_NUMPY_DIR, PY_PYTEST_DIR, PY_PLUGGY_DIR, PY_INICONFIG_DIR ")
```
bmhan12, line 362: Added a check here to verify if either:
- The given python environment/executable has the required modules
- all the necessary python installation `XXX_DIR` paths were provided

host-configs/rzwhippet-toss_4_x86_64_ib-gcc@13.3.1.cmake
```diff
@@ -73,72 +73,84 @@ set(ENABLE_GTEST_DEATH_TESTS ON CACHE BOOL "")
...
+set(YAPF_EXECUTABLE "${DEVTOOLS_ROOT}/._view/kky5oskfhacihjb5gpy7qkeh6edlrtcb/python-3.13.11/bin/yapf" CACHE PATH "")
 
 set(SHROUD_EXECUTABLE "/collab/usr/gapps/shroud/public/toss_4_x86_64_ib/shroud-0.14.0/bin/shroud" CACHE PATH "")
 
-set(CPPCHECK_EXECUTABLE "${DEVTOOLS_ROOT}/gcc-13.3.1/cppcheck-2.18.0-d2tmb7ls2d5vgh6vhcyohecwcu6msopb/bin/cppcheck" CACHE PATH "")
+set(CPPCHECK_EXECUTABLE "${DEVTOOLS_ROOT}/gcc-13.3.1/cppcheck-2.18.0-i3p56kpf66xzj4wm5kfqxyc4vgox7gb3/bin/cppcheck" CACHE PATH "")
+
+set(DOXYGEN_EXECUTABLE "${DEVTOOLS_ROOT}/._view/kky5oskfhacihjb5gpy7qkeh6edlrtcb/doxygen-1.15.0/bin/doxygen" CACHE PATH "")
+
+set(PY_NANOBIND_DIR "${TPL_ROOT}/py-nanobind-2.7.0-szozdpwzge7ah4bsilit6fijezxqbfop/lib/python3.13/site-packages" CACHE PATH "")
```
bmhan12, line 146: Example host-config here with the python installation path logic.

The TPLs/host-configs will be regenerated with the new paths.

scripts/spack/packages/axom/package.py
rhornung67: I think the naming you describe here is good.

scripts/spack/specs.json
rhornung67: @bmhan12 does the caliper compilation issue go away with rocm 6.4.x? I think most/all of our users are on 6.4.3 and are waiting for the green light to move ahead to rocm 7.x.

bmhan12: Yes, next caliper release should make it explicit that rocm >= 6.4.X is required.

host-configs/rzwhippet-toss_4_x86_64_ib-gcc@13.3.1.cmake
```diff
@@ -4,28 +4,28 @@
...
 #------------------------------------------------------------------------------
 # Compilers
 #------------------------------------------------------------------------------
-# Compiler Spec: gcc@13.3.1/75lfrhqb3monlgkmkokrboaqienagc6u
+# Compiler Spec: gcc@13.3.1/xrm3n2pex2en2wvm6p4ap62nieoz7kne
 #------------------------------------------------------------------------------
 if(DEFINED ENV{SPACK_CC})
 
-  set(CMAKE_C_COMPILER "/usr/WS1/axom/libs/toss_4_x86_64_ib/2026_01_29_13_39_33/none-none/compiler-wrapper-1.0-32dgmtd4fhqezj4ib4pbwjnwrxcuzxs7/libexec/spack/gcc/gcc" CACHE PATH "")
+  set(CMAKE_C_COMPILER "/usr/WS1/han12/axom/toss_4_x86_64_ib/2026_03_04_16_02_53/none-none/compiler-wrapper-1.0-uesxbqunebynqp5g2i6r3corj4oval5q/libexec/spack/gcc/gcc" CACHE PATH "")
```
kennyweiss: These will need to be updated to point to /usr/WS1/axom/libs

src/axom/sidre/tests/CMakeLists.txt
```diff
@@ -125,14 +125,12 @@ if(NANOBIND_FOUND AND AXOM_ENABLE_PYTHON_TESTS)
...
-        # https://docs.pytest.org/en/latest/explanation/pythonpath.html#invoking-pytest-versus-python-m-pytest
+        # Run unit test with pytest ("python3 -m pytest").
+        # Use convenience script that has
+        # pytest, pysidre, and conduit added to PYTHONPATH.
+        # "-p no:cacheprovider" disables caching.
         add_test (NAME    ${test_name}
-          COMMAND ${Python_EXECUTABLE} -m pytest -s -p no:cacheprovider ${TEST_OUTPUT_DIRECTORY}/${test}
-          WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/lib
+          COMMAND ${PROJECT_BINARY_DIR}/bin/run_python_with_axom.sh -m pytest -s -p no:cacheprovider ${TEST_OUTPUT_DIRECTORY}/${test}
         )
```
kennyweiss: Would it make sense to have an `axom_add_python_test` function/macro to avoid copy/pasting the params to the run_python_with_axom script?

src/axom/sidre/tests/sidre_attribute_Py.py
kennyweiss: Thanks -- please consider adding a comment to the top of this block to clarify that we do not expect the following to work b/c: "None" is not type Attribute

kennyweiss: (It \*is\* all there, but could be a bit clearer)

src/tools/CMakeLists.txt
kennyweiss: Minor:

```diff
-     # Add unit test to check script
+     # Add smoke test to check script
```

kennyweiss (approved):
Thanks @bmhan12

src/axom/sidre/tests/CMakeLists.txt
bmhan12: `axom_add_python_test` CMake macro added!

src/cmake/AxomMacros.cmake
```diff
@@ -565,6 +566,38 @@ macro(axom_configure_file _source _target)
     execute_process(COMMAND ${CMAKE_COMMAND} -E remove ${_tmp_target})
 endmacro(axom_configure_file)
 
+##------------------------------------------------------------------------------
```
kennyweiss, line 569: Thanks for adding this -- really clean!

PR #1812, opened 2026-03-03
Lumberjack Combiner message candidates

bmhan12:
This PR:
- An alternate design to solve issue described in PR #1805 
- Adds a virtual function `isMessageCandidateForCombiner` to the Combiner class that checks if a single given Message is a candidate for the Combiner. 
  - Can be optionally overwritten, default behavior is every Message is a candidate (preserves current Combiner behavior).
  -  Inside Lumberjack's `combineMessages()` call, `isMessageCandidateForCombiner` function is called on each Message prior to `shouldMessagesBeCombined` comparison. This filters out non-combinable messages from being repeatedly checked.

Relevant ReadTheDocs changes:
- https://axom.readthedocs.io/en/feature-han12-lj_combiner_tags/axom/lumberjack/docs/sphinx/core_concepts.html
- https://axom.readthedocs.io/en/feature-han12-lj_combiner_tags/axom/lumberjack/docs/sphinx/combiner_class.html#combiner-class-label

### DIFF
--- a/src/axom/lumberjack/Combiner.hpp
+++ b/src/axom/lumberjack/Combiner.hpp
@@ -17,7 +17,6 @@
 #define COMBINER_HPP
 
 #include "axom/lumberjack/Message.hpp"
-#include "axom/core/Macros.hpp"
 
 namespace axom
 {
@@ -62,10 +61,7 @@ public:
    * but likely used in derived classes.   
    *****************************************************************************
    */
-  virtual bool isMessageCandidateForCombiner(const Message&)
-  {
-    return true;
-  }
+  virtual bool isMessageCandidateForCombiner(const Message&) { return true; }
 
   /*!
    *****************************************************************************
--- a/src/axom/lumberjack/tests/lumberjack_Lumberjack.hpp
+++ b/src/axom/lumberjack/tests/lumberjack_Lumberjack.hpp
@@ -6,6 +6,8 @@
 
 #include "gtest/gtest.h"
 
+#include "axom/core/Macros.hpp"
+
 #include "axom/lumberjack/Lumberjack.hpp"
 
 #include "axom/lumberjack/Communicator.hpp"


gberg617:
Would it be worth adding a performance test to capture why we need the pre-filter for certain cases?

src/axom/lumberjack/docs/sphinx/core_concepts.rst
```diff
@@ -21,7 +21,7 @@ Combiners and register them with Lumberjack.  The idea is that each Combiner wou
 its own criteria for whether a Message should be combined and how to combine that specific
 Message with another of the same type.
 
-Combiner's have two main functions, shouldMessagesBeCombined and combine.
+Combiner's have three main functions: shouldMessagesBeCombined, combine, and optionally isMessageCandidateForCombiner.
```
Arlie-Capps:
```suggestion
Combiners have three main functions: shouldMessagesBeCombined, combine, and optionally isMessageCandidateForCombiner.
```

src/axom/lumberjack/Lumberjack.cpp
```diff
@@ -240,6 +240,8 @@ void Lumberjack::combineMessages()
   }
 
   std::vector<Message*> finalMessages;
```
kennyweiss, line 245: Not strictly required for this PR, but a few short comments in this function would help a lot w/ readability.

src/axom/lumberjack/Combiner.hpp
```diff
@@ -52,6 +53,20 @@ class Combiner
...
+   * should be considered for this combiner. Default is always true.
+   *
+   * \param [in] m The Message to be considered
+   *****************************************************************************
+   */
+  virtual bool isMessageCandidateForCombiner(const Message& m)
+  {
+    AXOM_UNUSED_VAR(m);
+    return true;
+  }
```
kennyweiss: Your solution w/ `AXOM_UNUSED_VAR` is fine.
If we're always ignoring the parameter, we could leave it unnamed:

```suggestion
  /*!
   *****************************************************************************
   * \brief Function used by Lumberjack to indicate whether a Message class
   * should be considered for this combiner. Default is true in the base class.
   *
   * \param [in] The Message to be considered. Unused in the base class
   * but likely used in derived classes.   
   *****************************************************************************
   */
  virtual bool isMessageCandidateForCombiner(const Message&)
  {
    return true;
  }
```

src/axom/lumberjack/tests/lumberjack_speedTestSkipNoncombinable.cpp
```diff
@@ -0,0 +1,104 @@
...
+  {
+    lumberjack.queueMessage("Unique message " + std::to_string(i) + " is not combinable ",
+                            "",
+                            0,
+                            0,
+                            1.0,
+                            "ignoreTag");
+  }
+  lumberjack.queueMessage("Should be combined.", "", 0, 0, 0.0, "tagA");
+
```
kennyweiss, line 95: I think you added this to visually check the runtime. Please add a comment about this.

Is there a check that we can add to see how many messages were sent/received?

src/axom/lumberjack/tests/lumberjack_Lumberjack.hpp
```diff
@@ -82,6 +82,57 @@ class TestCombiner : public axom::lumberjack::Combiner
...
+{
+public:
+  const std::string id() { return m_id; }
+
+  bool isMessageCandidateForCombiner(const axom::lumberjack::Message& m)
+  {
+    return m.text().compare("foo") == 0 || m.text().compare("bar") == 0;
+  }
+
+  bool shouldMessagesBeCombined(const axom::lumberjack::Message& leftMessage,
```
kennyweiss, line 97: Should we use `override` on the functions that are being overridden?

src/axom/lumberjack/tests/lumberjack_speedTestSkipNoncombinable.cpp
bmhan12: Done, and message count printed and checked as regex output!

PR #1813, opened 2026-03-12
EquiZ MIR performance improvements

BradWhitlock:
This PR includes various performance improvements for EquiZ MIR, whose OpenMP performance prior to this was 10x slower than serial. This was largely due to ArrayView calling Umpire in its constructor. The Bump library uses ArrayView all over the place for array slices that represent zones.

* Change ArrayView so calling any Umpire functionality to set the allocator id is deferred until needed. This lets Bump use ArrayView more cheaply.
* I added various optimizations through Bump that improve performance, sometimes adding `axom::OMP_EXEC` special cases.
* SEQ gets about 10x faster (with Umpire enabled)
* OMP gets up to 320x faster than previous OMP
* Changed scans so they accumulate using output array precision (when possible - to allow char mask)
* Added tests for scans
* Improved MIR benchmark plotting

### DIFF
--- a/src/axom/lumberjack/tests/lumberjack_Lumberjack.hpp
+++ b/src/axom/lumberjack/tests/lumberjack_Lumberjack.hpp
@@ -63,7 +63,7 @@ private:
 class TestCombiner : public axom::lumberjack::Combiner
 {
 public:
-  const std::string id() { return m_id; }
+  const std::string id() override { return m_id; }
 
   bool shouldMessagesBeCombined(const axom::lumberjack::Message& leftMessage,
                                 const axom::lumberjack::Message& rightMessage) override
@@ -87,7 +87,7 @@ private:
 class TestCandidateCombinerA : public TestCombiner
 {
 public:
-  const std::string id() { return m_id; }
+  const std::string id() override { return m_id; }
 
   bool isMessageCandidateForCombiner(const axom::lumberjack::Message& m) override
   {
@@ -108,7 +108,7 @@ private:
 
 class TestCandidateCombinerB : public axom::lumberjack::Combiner
 {
-  const std::string id() { return m_id; }
+  const std::string id() override { return m_id; }
 
   bool isMessageCandidateForCombiner(const axom::lumberjack::Message& m) override
   {


src/axom/core/ArrayBase.hpp
```diff
@@ -731,7 +731,8 @@ inline std::ostream& print(std::ostream& os, const ArrayBase<T, DIM, ArrayType>&
   os << "[ ";
   for(IndexType i = 0; i < static_cast<const ArrayType&>(array).size(); i++)
   {
-    os << data[i] << " ";
+    if(i > 0) os << ", ";
```
kennyweiss: Please wrap this one-liner in curly braces

```suggestion
    if(i > 0) { os << ", "; }
```

src/axom/core/execution/timed_for_all.hpp
```diff
@@ -0,0 +1,133 @@
...
+    ompEnd[nthreads - 1] =
+      std::chrono::duration<double>(std::chrono::high_resolution_clock::now().time_since_epoch()).count();
+    AXOM_ANNOTATE_END(name);
+
+    std::cout << name << ":\n";
+    std::cout << "\tn: " << n << "\n";
+    std::cout << "\tnthreads: " << (nthreads - 1) << "\n";
+    std::cout << std::setprecision(20) << "\tstart=" << ompStart << std::endl;
+    std::cout << "\tend=" << ompEnd << std::endl;
+  }
```
kennyweiss, line 111: Should this print to `cout`? I suppose it's fine if the user explicitly asks for the `timed_for_all`.

src/axom/bump/NodeToZoneRelationBuilder.hpp
```diff
@@ -50,102 +52,178 @@ struct BuildRelation
...
+   *        that points to the start of each list of zones.
+   *
+   * \param[in]    nodesView   A view that contains the set of all of the nodes in the topology (the connectivity)
+   * \param[inout] zonesView   A view (same size as \a nodesView) that contains the zone number of each node.
+   * \param[out]   sizesView A view that we fill with sizes.
+   * \param[out]   offsetsView A view that we fill with offsets so offsetsView[i] points to the start of the i'th list in \a zonesView.
+   */
+  static void execute(ViewType nodesView, ViewType zonesView, ViewType sizesView, ViewType offsetsView)
+  {
+    // Call serial implementation for now because it is faster.
```
kennyweiss, line 222: Is this always true?

kennyweiss (approved):
Thanks @BradWhitlock

src/axom/bump/NodeToZoneRelationBuilder.hpp
BradWhitlock: It was true for CPU on the LC platforms I timed. I think it is the `unordered_map` in the serial version vs the sort and mask approach in the general case.

src/axom/core/execution/timed_for_all.hpp
BradWhitlock: I guess it could take a std::ostream and print to that. This was mostly to be able to get per-thread kernel execution timing so I could plot it as a timeline. It seemed useful enough to leave in the code for future OMP debugging but not something I'd necessarily want enabled all the time.

src/axom/core/ArrayBase.hpp
BradWhitlock: Done.

PR #1817, opened 2026-03-12
Allow `insert()`ing nullptr if `n == 0` in `axom::Array`

ebchin:
# Summary

- This PR is a bugfix
- It does the following:
  - Fixes the `assert()` to allow `nullptr` to be `insert()`ed if the number of elements to insert (`n`) is 0.

### DIFF
--- a/scripts/github-actions/linux-build_and_test.sh
+++ b/scripts/github-actions/linux-build_and_test.sh
@@ -45,7 +45,7 @@ if [[ "$DO_BUILD" == "yes" ]] ; then
     fi
 
     echo "~~~~~~ RUNNING TESTS ~~~~~~~~"
-    make CTEST_OUTPUT_ON_FAILURE=1 test ARGS='-T Test -VV -j$NUM_BUILD_PROCS'
+    make CTEST_OUTPUT_ON_FAILURE=1 test ARGS='-T Test --output-on-failure -j$NUM_BUILD_PROCS'
 
     if [[ "${DO_BENCHMARKS}" == "yes" ]] ; then
         echo "~~~~~~ RUNNING BENCHMARKS ~~~~~~~~"
--- a/src/axom/bump/TopologyMapper.hpp
+++ b/src/axom/bump/TopologyMapper.hpp
@@ -26,6 +26,27 @@
 // Uncomment to emit debugging messages
 //#define AXOM_DEBUG_TOPOLOGY_MAPPER
 
+#if defined(AXOM_DEVICE_CODE) && defined(AXOM_USE_HIP)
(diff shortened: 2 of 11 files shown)


src/axom/core/Array.hpp
```diff
@@ -1428,7 +1428,7 @@ Array<T, DIM, SPACE, StoragePolicy>::insert(Array<T, DIM, SPACE, StoragePolicy>:
 template <typename T, int DIM, MemorySpace SPACE, typename StoragePolicy>
 inline void Array<T, DIM, SPACE, StoragePolicy>::insert(IndexType pos, IndexType n, const T* values)
 {
-  assert(values != nullptr);
+  assert(n == 0 || values != nullptr);
   reserveForInsert(n, pos);
   m_arrayOps.fill_range(m_data, pos, n, values, MemorySpace::Dynamic);
```
kennyweiss, line 1433: Gut check:  Do `reserveForInsert(0, ...)` and `fill_range(..., 0, nullptr, ...)` both do the right thing for `0` and `nullptr`?

Would this be better as:

```suggestion
  if(n>0)
  {
    assert(values != nullptr);
    reserveForInsert(n, pos);
    m_arrayOps.fill_range(m_data, pos, n, values, MemorySpace::Dynamic);
  }
```

ebchin: I looked at the methods: `reserveForInsert()` has code which explicitly handles the case `n == 0`.  `fill_range()` is more complicated (depends whether the type is trivially copyable and if Umpire is available), though if it didn't work with `n == 0` and `nullptr`, then things like [copy assignment](https://github.com/llnl/axom/blob/develop/src/axom/core/Array.hpp#L363) of empty `axom::Array`s wouldn't work either.
 
 Maybe it's best to leave the assert as is, so `fill_range()` can be tested implicitly by the new test I added (though admittedly, not for all possible cases)?

kennyweiss: Thoughts @publixsubfan ?

publixsubfan: That does seem like a reasonable explanation. Looking at `fill_range` it looks like we either call to `axom::copy` or `std::uninitialized_copy` and I don't expect either of them to have issues with `n == 0`.

src/axom/core/tests/core_array.hpp
```diff
@@ -12,6 +12,7 @@
 #include "gtest/gtest.h"
 
 #include <algorithm>
+#include <cstdlib>
```
publixsubfan: Just curious, why was this needed?

ebchin: It's not. It's was part of a test I've since removed, and it's been taken out. Thanks for noticing this!

PR #1818, opened 2026-03-16
Improves testing for `quest::STEPReader` for surface orientation

jcs15c:
# Summary

- This PR is a slight refactoring of the `quest::STEPReader` to handle additional edge cases stemming from different canonical representations of OpenCascade STEP files and `primal`'s trimmed NURBS primitives. In general, OCC stores surface orientation through both an internal flag and the orientation of loops of trimming curves, while `primal` encodes orientation information only through patch control points and requires CCW trimming curve loops.

An issue was identified and partially fixed in the closed PR #1765, but since then additional edge cases have popped up. To better test for these issues as the reader is exposed to more in-the-wild meshes, a dedicated test has been added to `quest`. This test is supported by additional STEP files added to [Axom's data repo](https://github.com/llnl/axom_data/pull/37).

Below is a table which shows which tested files fail for each configuration of the STEP reader. Files marked with * are new to `axom_data`.

| Filename | Pre #1765 | develop | current PR |
| -------- | ---------------- | ---------------- | ---------------- |
| tet        |            ✓              |          ✓          |          ✓         |
| sliced_cylinder |       ✓     |          ✓        |          ✓       |
| nut |           ✓      |       ✓        |          ✓         |
| bearings |      ✓     |        ✓        |     ✓      |
| boxed_sphere | |  ✓   |✓  | ✓     |
| plate* |    |   |✓ |
| bolt_clip* | ✓  | | ✓|

Additionally, this PR addresses an incorrect overload in `winding_number.hpp` for 2D NURBS curves.

### DIFF
--- a/src/axom/quest/tests/quest_step_reader.cpp
+++ b/src/axom/quest/tests/quest_step_reader.cpp
@@ -61,7 +61,7 @@ struct MiniTriangleGWN3D
     BBox3D* shapeBBoxPtr = &shapeBBox;
     mint::for_all_nodes<axom::SEQ_EXEC, mint::xargs::xyz>(
       &mesh,
-      AXOM_LAMBDA(axom::IndexType /*nodeIdx*/, double x, double y, double z) {
+      AXOM_LAMBDA(axom::IndexType AXOM_UNUSED_PARAM(nodeIdx), double x, double y, double z) {
         shapeBBoxPtr->addPoint(Point3D {x, y, z});
       });
 
@@ -76,7 +76,7 @@ struct MiniTriangleGWN3D
       &mesh,
       AXOM_LAMBDA(axom::IndexType cellIdx,
                   const axom::numerics::Matrix<double>& coords,
-                  const axom::IndexType* /*nodeIds*/) {
+                  const axom::IndexType* AXOM_UNUSED_PARAM(nodeIds)) {
         // clang-format off
         trisView[cellIdx] = Triangle3D {Point3D {(coords(0, 0) - ctr[0]) / scl, (coords(1, 0) - ctr[1]) / scl, (coords(2, 0) - ctr[2]) / scl},
                                         Point3D {(coords(0, 1) - ctr[0]) / scl, (coords(1, 1) - ctr[1]) / scl, (coords(2, 1) - ctr[2]) / scl},
--- a/src/axom/quest/GWNMethods.hpp
+++ b/src/axom/quest/GWNMethods.hpp
@@ -295,7 +295,7 @@ public:
         poly_mesh,
         AXOM_LAMBDA(axom::IndexType cellIdx,
                     const axom::numerics::Matrix<double>& coords,
-                    const axom::IndexType* /*nodeIds*/) {
+                    const axom::IndexType* AXOM_UNUSED_PARAM(nodeIds)) {
           segments_view[cellIdx] =
             SegmentType {Point2D {coords(0, 0), coords(1, 0)}, Point2D {coords(0, 1), coords(1, 1)}};
         });
@@ -649,7 +649,7 @@ public:
         tri_mesh,
         AXOM_LAMBDA(axom::IndexType cellIdx,
                     const axom::numerics::Matrix<double>& coords,
-                    const axom::IndexType* /*nodeIds*/) {
+                    const axom::IndexType* AXOM_UNUSED_PARAM(nodeIds)) {
           triangles_view[cellIdx] =
             TriangleType {Point3D {(coords(0, 0) - shape_center[0]) / scale,
                                    (coords(1, 0) - shape_center[1]) / scale,


src/axom/primal/operators/winding_number.hpp
```diff
@@ -383,13 +383,19 @@ axom::Array<double> winding_number(const axom::Array<Point<T, 2>>& query_arr,
   {
     ret_val[n] = 0.0;
 
-    for(int i = 0; i < nurbs_curve_arr.size(); ++i)
+    for(int i = 0; i < nurbs_cache_arr.size(); ++i)
```
kennyweiss: Minor: should we be using the iterator interface on any/all/some of these loops to avoid worrying about indexing?

src/axom/quest/io/STEPReader.cpp
```diff
@@ -66,6 +69,7 @@ struct PatchData
   int patchIndex {-1};
   bool wasOriginallyPeriodic_u {false};
   bool wasOriginallyPeriodic_v {false};
+  bool reversed_u {false};
```
kennyweiss: Suggestion on variable name (minor): 

Would `was_reversed_u` be clearer?

src/axom/quest/io/STEPReader.cpp
```diff
@@ -867,14 +886,12 @@ class StepFileProcessor
         const TopoDS_Wire& wire = TopoDS::Wire(wireExp.Current());
 
         int edgeIndex = 0;
-        for(TopExp_Explorer edgeExp(wire, TopAbs_EDGE); edgeExp.More(); edgeExp.Next(), ++edgeIndex)
+        BRepTools_WireExplorer edgeExp(wire, TopoDS::Face(faceExp.Current()));
+        for(; edgeExp.More(); edgeExp.Next(), ++edgeIndex)
```
kennyweiss, line 890: Suggestion variable name (minor):

Since this is a `WireExplorer`, should the variable be `wireExp` instead of `edgeExp`?

src/axom/quest/tests/quest_step_reader.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+template <typename... Args>
+std::string pjoin(const char* str, Args... args)
+{
+  return fs::joinPath(std::string(str), pjoin(args...));
+}
+
+//------------------------------------------------------------------------------
+// Use a pared-down version of the TriangleGWN3D method to directly evaluate
+//  the for a triangle mesh
```
kennyweiss: Missing word(s) in this comment. 

Perhaps:

```suggestion
//  the winding number for a triangle mesh
```

src/axom/quest/tests/quest_step_reader.cpp
```diff
@@ -0,0 +1,263 @@
...
+    m_scale = shapeBBox.getMax()[longestDim] - shapeBBox.getMin()[longestDim];
+
+    const auto& ctr = m_shapeCenter;
+    const auto scl = m_scale;
+    auto trisView = m_triangles.view();
+    mint::for_all_cells<axom::SEQ_EXEC, mint::xargs::coords>(
+      &mesh,
+      AXOM_LAMBDA(axom::IndexType cellIdx,
+                  const axom::numerics::Matrix<double>& coords,
+                  [[maybe_unused]] const axom::IndexType* nodeIds) {
```
kennyweiss: I think `[[maybe_unused]]` in a lambda was an error in some configs (e.g. hip) and the following fixed it:

```suggestion
                  const axom::IndexType* /* nodeIds */) {
```

src/axom/quest/tests/quest_step_reader.cpp
```diff
@@ -0,0 +1,263 @@
...
+  EXPECT_GT(patches.size(), 0) << "No NURBS patches were extracted from '" << fileName << "'";
+  if(patches.empty())
+  {
+    return;
+  }
+
+  const auto shapeBbox = stepReader.getBRepBoundingBox().scale(1.1);
+  auto bboxMin = shapeBbox.getMin();
+  auto bboxMax = shapeBbox.getMax();
+  const auto bboxDiag = bboxMax.array() - bboxMin.array();
```
kennyweiss: No need to change it, and the semantics are slightly different, but `primal::BoundingBox::range()` would have the same effect.

https://github.com/llnl/axom/blob/c84919a35b81631efdbe1dfdca1d4aef3a7accc3/src/axom/primal/geometry/BoundingBox.hpp#L156-L161

src/axom/quest/io/STEPReader.cpp
jcs15c: Maybe, but `wireExp` is already defined above as an object of type `TopExp_Explorer` since that iterates over the topology, producing wires. Then the object of type `BRepTools_WireExplorer` iterates over the wires, producing edges. It's a little complicated, but I think there is logic to it. Wire explorers explore wires, and edge explorers explore edges.

src/axom/quest/tests/quest_step_reader.cpp
rhornung67: We also have this macro, which is preferred because it's easy to search for. https://github.com/llnl/axom/blob/develop/src/axom/core/Macros.hpp#L168

PR #1819, opened 2026-03-17
Improve support for allocators in Bump and Mir components

### LINKED ISSUE #1780: Resolve conduit memory handling
We currently have 2 ways to use Axom memory in Conduit.  They were developed as independent utilities in different efforts.  This issue is for discussing how to combine them to remove redundancies.

Both methods create call-back functions for Conduit to use in allocation/deallocation.  Both register them with Conduit and receive a Conduit allocator ID indicating the functions to use.  This is standard for Conduit.

## First method, use template class `ConduitAllocateThroughAxom` in `bump`
Sample use:
```
template <typename ExecSpace>
void foo(conduit::Node& n) {
  utils::ConduitAllocateThroughAxom<ExecSpace> c2a;
  n.set_allocator(c2a.getConduitAllocatorID());  // n will use the default Axom allocator for ExecSpace.
}
```
This method uses the default allocator id for `ExecSpace`.

## Second method, use `ConduitMemory` class in `sidre`
Sample use:
```
void bar(conduit::Node& n, int axomAllocId)
{
  auto conduitAllocId = sidre::ConduitMemory::axomAllocIdToConduit(axomAllocId);
  n.set_allocator(conduitAllocId);  // n will use axomAllocId.
}
```
This method takes an Axom allocator id.

BradWhitlock:
This PR resolves #1780 by enhancing Bump and MIR algorithms so they can take a user-supplied memory allocator id.

* Added `setAllocatorID()` and `getAllocatorID()` methods to many Bump and MIR algorithms and used the allocator for memory allocation.
* Removed Bump's `ConduitAllocateThroughAxom` class and used the Conduit memory infrastructure from Sidre, which supports user-supplied allocators.
* Enhanced the "mir_concentric_circles" application so it creates a memory pool for HIP and CUDA backends to reuse memory and reduce memory allocation overhead. For HIP, this made the application timings 10-20x faster.
* Fixed various host/device memory consistency problems that manifested on CUDA

NOTE: There is a follow-on (already approved) PR to this that applies some Axom coding style updates.

### DIFF
--- a/src/axom/bump/extraction/PlaneIntersector.hpp
+++ b/src/axom/bump/extraction/PlaneIntersector.hpp
@@ -168,35 +168,25 @@ private:
 #endif
 
   /*!
-   * \brief Get fixed number of array values from a Conduit node in a way that should
-   *        not matter whether the data are in host or device memory.
+   * \brief Get fixed number of floating point array values from a Conduit node in a
+   *        way that should not matter whether the data are in host or device memory.
    *
    * \param n The Conduit node with the data.
    * \param[out] values An output array of the data on the host, converted to value_type.
    */
   void getArrayValues(const conduit::Node &n, value_type values[NDIMS]) const
   {
-    if(n.dtype().is_float32())
+    SLIC_ERROR_IF(n.dtype().number_of_elements() != NDIMS, "Incompatible sizes.");
+
+    // Make sure the data are on the host.
+    conduit::Node hostNode;
+    axom::bump::utilities::copy<axom::SEQ_EXEC>(hostNode, n);
+
+    // Copy the values into output values
+    const auto hostAccessor = hostNode.as_double_accessor();
+    for(int i = 0; i < NDIMS; i++)
     {
-      conduit::float32 hostData[NDIMS];
-      axom::copy(hostData, n.data_ptr(), NDIMS * sizeof(conduit::float32));
-      for(int i = 0; i < NDIMS; i++)
-      {
-        values[i] = static_cast<value_type>(hostData[i]);
-      }
-    }
-    else if(n.dtype().is_float64())
-    {
-      conduit::float64 hostData[NDIMS];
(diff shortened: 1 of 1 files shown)


src/axom/bump/extraction/PlaneIntersector.hpp
```diff
@@ -162,6 +167,39 @@ class PlaneIntersector
 private:
 #endif
 
+  /*!
+   * \brief Get fixed number of array values from a Conduit node in a way that should
+   *        not matter whether the data are in host or device memory.
```
Arlie-Capps: Please make the type restriction clear.  I'm thinking something like "Get fixed number of values from a Conduit node that contains a floating-point array in a way..."  and adding a sentence at the end saying something like "It is an error if the Conduit node is not float32 or float64."

src/axom/sidre/core/ConduitMemory.cpp
```diff
@@ -20,13 +20,19 @@ void ConduitMemory::privateRegisterAllocator()
   using conduit::utils::register_allocator;
   auto deallocator = [](void* ptr) -> void {
     char* cPtr = (char*)(ptr);
+    //std::cout << axom::execution_space<ExecSpace>::name()
+    //  << ": Dellocating for Conduit via axom: allocator=" << axomId
+    //  << ", ptr=" << ptr << std::endl;
```
Arlie-Capps: Please remove these if there's no need for them.  Or perhaps make a log entry if we're a DEBUG build.

src/axom/bump/extraction/PlaneIntersector.hpp
BradWhitlock: I changed it to support more data array types via an accessor.

src/axom/sidre/core/ConduitMemory.cpp
BradWhitlock: Removed

PR #1823, opened 2026-03-17
Fix slic id warning message

bmhan12:
This PR:
- Adds `override` keyword to remove warning introduced by me/PR #1812


Arlie-Capps (approved):
Thanks for this, Brian.

PR #1824, opened 2026-03-18
Apply Axom coding conventions to bump and mir components

BradWhitlock:
Apply coding style fixes to bump and mir to address:

> Aside observation and question: it looks like there are multiple naming style conventions applied in bump for methods, local variables, method args, etc. For example, some methods are named doSomething() and others are DoSomethingElse() (different capitalization). Some variables are named my_thing and others are named myOtherThing (different use of capitalization and underscores). Have we abandoned consistent conventions?

* Codex read the coding style guide and applied changes to bump and mir.
* Codex recommended changes to `NodeArrayView` code in bump.

### DIFF
--- a/src/axom/bump/tests/bump_utilities.cpp
+++ b/src/axom/bump/tests/bump_utilities.cpp
@@ -59,19 +59,19 @@ struct test_conduit_allocate
   }
 };
 
-TEST(bump_blueprint_utilities, allocate_seq) { test_conduit_allocate<seq_exec>::test(); }
+TEST(bump_utilities, allocate_seq) { test_conduit_allocate<seq_exec>::test(); }
 #if defined(AXOM_RUNTIME_POLICY_USE_OPENMP)
-TEST(bump_blueprint_utilities, allocate_omp) { test_conduit_allocate<omp_exec>::test(); }
+TEST(bump_utilities, allocate_omp) { test_conduit_allocate<omp_exec>::test(); }
 #endif
 #if defined(AXOM_RUNTIME_POLICY_USE_CUDA)
-TEST(bump_blueprint_utilities, allocate_cuda) { test_conduit_allocate<cuda_exec>::test(); }
+TEST(bump_utilities, allocate_cuda) { test_conduit_allocate<cuda_exec>::test(); }
 #endif
 #if defined(AXOM_RUNTIME_POLICY_USE_HIP)
-TEST(bump_blueprint_utilities, allocate_hip) { test_conduit_allocate<hip_exec>::test(); }
+TEST(bump_utilities, allocate_hip) { test_conduit_allocate<hip_exec>::test(); }
 #endif
 
 //------------------------------------------------------------------------------
-TEST(bump_blueprint_utilities, make_array_view_interleaved_seq)
+TEST(bump_utilities, make_array_view_interleaved_seq)
 {
   constexpr conduit::index_t n = 4;
   axom::Array<double> interleaved {{-1., 10., -2., 20., -3., 30., -4., 40., -5.}};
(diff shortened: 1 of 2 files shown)


src/axom/bump/extraction/FieldIntersector.hpp
```diff
@@ -68,21 +68,21 @@ class FieldIntersector
...
      *
-     * \param zoneIndex The zone index.
-     * \param nodeIds A view containing node ids for the zone.
+     * \param zone_index The zone index.
+     * \param node_ids A view containing node ids for the zone.
      */
     AXOM_HOST_DEVICE
-    axom::IndexType determineTableCase(axom::IndexType AXOM_UNUSED_PARAM(zoneIndex),
-                                       const ConnectivityView &nodeIds) const
+    axom::IndexType determineTableCase(axom::IndexType AXOM_UNUSED_PARAM(zone_index),
```
kennyweiss, line 75: I'm surprised (but unconcerned) about this change. I thought our style is ok with either camelCase or pothole_case (but presumably we'd want to be consistent w/in the same class/file/component)

fwiw, I like the idea of camelCase for function names and pothole_case for variable names

rhornung67 (approved):
It was sort of a nit-pick comment. At any rate, thank you @BradWhitlock

src/axom/bump/tests/bump_utilities.cpp
```diff
@@ -70,6 +70,38 @@ TEST(bump_blueprint_utilities, allocate_cuda) { test_conduit_allocate<cuda_exec>
 TEST(bump_blueprint_utilities, allocate_hip) { test_conduit_allocate<hip_exec>::test(); }
 #endif
 
+//------------------------------------------------------------------------------
+TEST(bump_blueprint_utilities, make_array_view_interleaved_seq)
```
kennyweiss: Minor, and unrelated to changes in this PR -- I think the way CTest filters tests, it would be best to have the first parameter match the name of the file (e.g. `bump_utilities` instead of `bump_blueprint_utilities`).

(I tried looking this up, but didn't find a good resource -- I'll update when I do)

src/axom/bump/utilities/conduit_memory.cpp
```diff
@@ -9,7 +9,7 @@
 namespace axom::bump::utilities
 {
 
-bool isDeviceAllocated(const conduit::Node &n)
+bool isDeviceAllocated([[maybe_unused]] const conduit::Node &n)
```
kennyweiss: Minor: Since `n` is always used in the `then` case, but not in the `else` case, 
I wonder if having `AXOM_UNUSED_VAR(n)` in the else case would better convey the expectations.

src/axom/bump/views/MaterialView.hpp
```diff
@@ -30,8 +30,8 @@ namespace views
  */
 struct Material
 {
-  int number {};
-  std::string name {};
+  int m_number {};
+  std::string m_name {};
 };
```
kennyweiss, line 35: I wonder if exposed public members of a struct should have the `m_` prefix.
(I'm pretty sure our style guide would say to do this, but it might be good to revisit this).

kennyweiss (approved):
Thanks @BradWhitlock -- I had some minor observations along the way that we might want to consider, but that should not block this PR.

src/axom/bump/extraction/FieldIntersector.hpp
BradWhitlock: I made Codex read the style guide and apply it literally. It said there was a preference for pothole_case for variable names, though both were allowed.

src/axom/bump/tests/bump_utilities.cpp
BradWhitlock: I had codex write the test so I'll change that prefix.

BradWhitlock: I fixed the test prefix name here and in one of the mir tests.

src/axom/bump/utilities/conduit_memory.cpp
BradWhitlock: Done

PR #1826, opened 2026-03-20
Fix quest compilation problem with AXOM_NO_INT64_T.

### LINKED ISSUE #1822: Build problem with -DAXOM_NO_INT64_T definition
When `AXOM_NO_INT64_T` is defined, some `mpi_traits` in axom/core/Types.hpp are not defined. This later causes a compilation problem in quest where some MPI reductions rely on `int64` support.

```
.../spack-src/src/axom/quest/MeshClipper.cpp:248: undefined reference to `axom::mpi_traits<long>::type'
```

See [where it is used](https://github.com/llnl/axom/blob/develop/src/axom/quest/MeshClipper.cpp#L285-L301).
```
  axom::Array<std::int64_t> maxs(sums);
  globalReduce(maxs, MPI_MAX);
```

It would seem that either the fix is to enable the mpi_traits, probably defeating the purpose of the macro. Or, to change `MeshClipper.cpp` so it can use `int32` when the macro is defined.

BradWhitlock:
This PR resolves #1822, which is a build problem that arises when `AXOM_NO_INT64_T` is defined.

* Fix problem by switching from explicit `int64_t` to `axom::IndexType`
* Updated release notes
* Changed one of the CI builds that used 32-bit axom::IndexType so it sets `AXOM_NO_INT64_T`

Other changes:
* Changed some quest code from using `RAJA::` functions to `axom::` wrappers
* Changed a few methods to static in `MeshClipper` code to prevent capturing `this` pointer in device lambdas.


src/axom/quest/detail/clipping/HexClipper.cpp
```diff
@@ -206,7 +206,7 @@ AXOM_HOST_DEVICE inline MeshClipperStrategy::LabelType HexClipper::polyhedronToL
   const BoundingBox3DType& vertsBb,
   const BoundingBox3DType& hexBb,
   const axom::ArrayView<const TetrahedronType>& hexTets,
-  const axom::StackArray<Triangle3DType, 24>& surfaceTriangles) const
+  const axom::StackArray<Triangle3DType, 24>& surfaceTriangles)
```
rhornung67, line 209: Why is the `const` qualifer on the method removed?

src/axom/quest/detail/clipping/HexClipper.hpp
```diff
@@ -101,12 +101,12 @@ class HexClipper : public MeshClipperStrategy
...
    */
   template <typename Polyhedron>
-  AXOM_HOST_DEVICE inline LabelType polyhedronToLabel(
+  AXOM_HOST_DEVICE static inline LabelType polyhedronToLabel(
     const Polyhedron& verts,
     const BoundingBox3DType& vertsBb,
     const BoundingBox3DType& hexBb,
     const axom::ArrayView<const TetrahedronType>& hexTets,
-    const axom::StackArray<Triangle3DType, 24>& surfaceTriangles) const;
+    const axom::StackArray<Triangle3DType, 24>& surfaceTriangles);
```
rhornung67, line 109: Same question about const qualifier here.

src/axom/quest/detail/clipping/SphereClipper.cpp
```diff
@@ -177,7 +177,7 @@ void SphereClipper::labelTetsInOutImpl(quest::experimental::ShapeMesh& shapeMesh
 template <typename Polyhedron>
 AXOM_HOST_DEVICE inline MeshClipperStrategy::LabelType SphereClipper::polyhedronToLabel(
   const Polyhedron& verts,
-  const SphereType& sphere) const
+  const SphereType& sphere)
```
rhornung67, line 180: Another const qualifier removal?  

There are others. I have the same question about them. But I will add no more comments.

src/axom/quest/detail/clipping/HexClipper.cpp
BradWhitlock: I removed it because in the header, the method is now marked `static`.

PR #1827, opened 2026-03-23
Evaluate surface and volume integrals over Bezier and NURBS patches

kennyweiss:
# Summary

- This PR is a feature
- It add `primal::evaluate_surface_integral` and `primal::evaluate_volume_integral` functions for integrands over collections of `BezierPatch` and `NURBSPatch` objects
- This is an implementation of the algorithm from "High-Accuracy Mesh-Free Quadrature for Trimmed Parametric Surfaces and Volumes" by D. Gunderman, K. Weiss and J. Evans in Computer-Aided Design 2021. https://doi.org/10.1016/j.cad.2021.103093.
- Also adds an example that evaluates surface- and volume- moments over a BRep defined in a STEP model, and computes a best-fit ellipsoid and oriented bounding box to the model based on its mass, centroid and moments of inertia.
    - Here is an inertia-matched ellipsoid to the `connector.step` model
      image

### DIFF
--- a/src/axom/quest/tests/quest_step_quadrature.cpp
+++ b/src/axom/quest/tests/quest_step_quadrature.cpp
@@ -62,31 +62,6 @@ void transform_patches(PatchArray& patches, const Matrix& transform)
   }
 }
 
-PatchArray replicate_biquartic_sphere_faces(const PatchArray& base_patches)
-{
-  EXPECT_EQ(base_patches.size(), 1);
-
-  PatchArray all_patches;
-  all_patches.push_back(base_patches[0]);
-
-  const Matrix rot_pos_y = embed_linear_transform(axom::numerics::transforms::xRotation(M_PI / 2.0));
-  const Matrix rot_neg_y = embed_linear_transform(axom::numerics::transforms::xRotation(-M_PI / 2.0));
-  const Matrix rot_pos_z = embed_linear_transform(axom::numerics::transforms::xRotation(M_PI));
-  const Matrix rot_neg_x = embed_linear_transform(axom::numerics::transforms::yRotation(M_PI / 2.0));
-  const Matrix rot_pos_x = embed_linear_transform(axom::numerics::transforms::yRotation(-M_PI / 2.0));
-
-  const Matrix rotations[] = {rot_pos_y, rot_neg_y, rot_pos_z, rot_neg_x, rot_pos_x};
-  for(const Matrix& rotation : rotations)
-  {
-    PatchArray rotated(1);
-    rotated[0] = base_patches[0];
-    transform_patches(rotated, rotation);
-    all_patches.push_back(rotated[0]);
-  }
-
-  return all_patches;
-}
-
 template <typename Lambda>
 void expect_matching_surface_values(const PatchArray& patches_a,
                                     const PatchArray& patches_b,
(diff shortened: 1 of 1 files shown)


src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -403,6 +408,195 @@ inline RetType evaluate_area_integral_component(const NURBSCurveGWNCache<T>& nc,
...
+  LambdaRetType total_integral = LambdaRetType {};
+  for(const auto& split_patch : n.extractTrimmedBezier())
+  {
+    const auto& curves = split_patch.getTrimmingCurves();
+    if(curves.empty())
+    {
+      continue;
+    }
+
+    const auto lower_bound_y = detail::curve_array_lower_bound_y(curves);
```
jcs15c: For a trimmed patch, I think the lower bound should be taken as the maximum of this value and the minimum v parameter for the patch. Since this lower bound is determined from the control points of the curves, it could be totally outside the parameter space, and when the quadrature nodes are evaluated on the patch, they'll be clamped to the parameter space. It's better to just clamp the bounds than the individual nodes, since I think that could seriously mess with the accuracy of the method.

kennyweiss: Done.

src/axom/quest/tests/quest_step_quadrature.cpp
```diff
@@ -0,0 +1,283 @@
...
+  EXPECT_NEAR(primal::evaluate_volume_integral(patches, const_integrand, npts), 8.0 / 3.0, abs_tol);
+}
+
+TEST(quest_step_quadrature, sphere_models_match_unit_sphere_moments)
+{
+  const std::string step_dir = std::string(AXOM_DATA_DIR) + "/quest/step/";
+
+  auto revolved = read_step_patches(step_dir + "revolved_sphere.step");
+  auto biquartic = read_step_patches(step_dir + "biquartic_sphere_surface.step");
+  biquartic = replicate_biquartic_sphere_faces(biquartic);
```
jcs15c: I just merged https://github.com/llnl/axom_data/pull/37, which adds a full `biquartic_sphere.step` file that you could use here instead of replicating the single surface!

kennyweiss:
> Looks good to me! Minor thing with the new `biquartic_sphere.step` in axom data, but otherwise the math looks sound.

Thanks -- I'll update it

**Edit:** Done.

src/axom/primal/operators/detail/evaluate_integral_impl.hpp
```diff
@@ -30,12 +30,15 @@
 #include "axom/primal/geometry/Point.hpp"
 #include "axom/primal/geometry/Vector.hpp"
 #include "axom/primal/geometry/BezierCurve.hpp"
+#include "axom/primal/geometry/BezierPatch.hpp"
 #include "axom/primal/geometry/NURBSCurve.hpp"
+#include "axom/primal/geometry/NURBSPatch.hpp"
```
jcs15c: I know I already approved these changes, but I'm wondering now if there's something inherently circular about these includes. Currently, `NURBSPatch` objects call `evaluate_area_integral` on their trimming curves, which means `NURBSPatch.hpp` should technically include `evaluate_integral.hpp`, which seems like it would conflict with this include here. 

This doesn't seem to be an issue right now, since `NURBSPatch` is somehow able to find the relevant definition for `evaluate_area_integral` without the correct include, but on my new branch https://github.com/llnl/axom/pull/1834 something changed to make the include tree more picky, and now it can't find the function anymore. ~~Still working on a fix there~~ The fix was fixing a bug where `CurvedPolygon` included `NURBSPatch`, but there still might be an issue for whoever's PR gets merged second.

There was a similar issue with the winding number methods, which we resolved by having a `winding_number.hpp` file containing the interface to 2D and 3D methods, and then separate `winding_number_2d_impl.hpp` and `winding_number_3d_impl.hpp` files. Do you anticipate that we would eventually need something here too, with separate `evaluate_integral_curve_impl.hpp` and `evaluate_integral_surface_impl.hpp` files, either here or in a later PR?

kennyweiss: Great catch! Fixed.

src/axom/quest/tests/quest_step_quadrature.cpp
kennyweiss: Fixed!

PR #1831, opened 2026-03-26
Add slic macros that log once per call site

bmhan12:
This PR:
- Adds suffixed `SLIC_*_ONCE` macros that are invoked once per call site.

Addresses #1685

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -38,6 +38,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Quest: Adds fast GWN methods for STL/Triangulated STEP input and linearized NURBS Curve input which
   leverage error-controlled approximation and a spatial index (BVH).
 - Slic: Adds new Slic macros that allow you to selectively print messages once per call-site. For example,
+  `SLIC_INFO_ONCE(msg)` and `SLIC_INFO_ROOT_IF_ONCE(EXP, msg)`.
 
 ### Changed
 - Primal: Axom's polygon clipping was modified to handle some corner cases.


src/axom/slic/interface/slic_macros.hpp
```diff
@@ -722,16 +830,72 @@
...
+    static bool once = true;                     \
+    if(EXP)                                      \
+    {                                            \
+      if(once)                                   \
+      {                                          \
+        macro(true, msg);                        \
+        once = false;                            \
+      }                                          \
+    }                                            \
+  } while(axom::slic::detail::false_value)
```
bmhan12, line 900: Helper macro that the `SLIC_*_ONCE` macros use.

src/axom/slic/tests/slic_macros.cpp
```diff
@@ -118,36 +118,74 @@ void check_tag(const std::string& msg, const std::string& expected_tag)
...
+
+  check_level(str, level);
+  check_msg(str, message);
+  check_line(str, expected_line);
+  check_file(str);
+  slic::internal::clear();
+}
+
+// Convenience test macro that checks slic macro has logged a message
+#define EXPECT_SLIC_LOG(macro_call, level, message)           \
```
bmhan12, line 151: Convenience macro for testing, mainly to alleviate the tracking of `__LINE__` in the original tests. It works pretty well for gcc, except...

src/axom/slic/tests/slic_macros.cpp
```diff
@@ -118,36 +118,74 @@ void check_tag(const std::string& msg, const std::string& expected_tag)
...
-  EXPECT_FALSE(slic::internal::is_stream_empty());
-  check_level(slic::internal::test_stream.str(), "ERROR");
-  check_msg(slic::internal::test_stream.str(), "this message is logged!");
-  check_file(slic::internal::test_stream.str());
-  check_line(slic::internal::test_stream.str(), expected_line_number);
-  slic::internal::clear();
+  // Single line - Placement of ")" matters for __LINE__ for slic call and checking
+  // clang-format off
+  EXPECT_SLIC_LOG(SLIC_ERROR_IF(true, "this message is logged!"), "ERROR", "this message is logged!");
+  // clang-format on
```
bmhan12, line 188: ... clang and intel compilers have different `__LINE__` behavior, I need to disable formatting here to keep the entire call to one line. 
Otherwise, the observed behavior for example is something like:
```
  EXPECT_SLIC_LOG(
    SLIC_ERROR_IF(true, "this message is logged!"), // <-- This is the line number that is logged
    "ERROR", 
    "this message is logged!"); // <-- This is the line number the macro is checking for
```
The line number as far as I can tell in both instances is actually determined by the line placement of the closing parantheses `")"`.

Another example for future reference:

```
  EXPECT_SLIC_LOG(SLIC_ERROR_ROOT_IF(true,
    "this message is logged!"


    ), // <-- This is the line number that is logged
                  "ERROR",
                  "this message is logged!"
                  ); // <-- This is the line number the macro is checking for
```

src/axom/slic/interface/slic_macros.hpp
```diff
@@ -70,20 +60,17 @@
...
+  if(EXP)                                                         \
+  {                                                               \
+    std::ostringstream __oss;                                     \
+    __oss << msg;                                                 \
+    axom::slic::logErrorMessage(__oss.str(), __FILE__, __LINE__); \
+    if(axom::slic::isAbortOnErrorsEnabled())                      \
+    {                                                             \
+      axom::slic::abort();                                        \
+    }                                                             \
+  }
```
kennyweiss: @bmhan12 -- the `do { } while(false)` was there to require slic macros to have a semicolon at the end. Please add it back.

src/axom/slic/interface/slic_macros.hpp
```diff
@@ -42,17 +42,7 @@
...
-  {                                                               \
-    std::ostringstream __oss;                                     \
-    __oss << msg;                                                 \
-    axom::slic::logErrorMessage(__oss.str(), __FILE__, __LINE__); \
-    if(axom::slic::isAbortOnErrorsEnabled())                      \
-    {                                                             \
-      axom::slic::abort();                                        \
-    }                                                             \
-  } while(axom::slic::detail::false_value)
+#define SLIC_ERROR(msg) SLIC_ERROR_IF(true, msg)
```
kennyweiss, line 45: Nice change!

src/axom/slic/interface/slic_macros.hpp
```diff
@@ -722,16 +830,72 @@
...
+  {                                              \
+    static bool once = true;                     \
+    if(EXP)                                      \
+    {                                            \
+      if(once)                                   \
+      {                                          \
+        macro(true, msg);                        \
+        once = false;                            \
+      }                                          \
+    }                                            \
```
kennyweiss, line 899: Gut check -- this can call the macro multiple times, until EXP evaluates as true.
Is this the semantics we're looking for?  Or should it only call the the macro once, independent of whether EXP is true/false?

src/axom/quest/tests/quest_mesh_clipper.cpp
```diff
@@ -654,6 +654,7 @@ axom::klee::Geometry createGeom_CupMesh(sidre::DataStore& ds, const std::string&
   std::string proeFile = axom::utilities::filesystem::joinPath(AXOM_DATA_DIR, "quest/cup.proe");
   reader.setFileName(proeFile);
   int readStatus = reader.read();
+  AXOM_UNUSED_VAR(readStatus);
```
kennyweiss, line 657: Minor: Since this is a test, it should probably be `EXPECT_EQ(readStatus, 0)` instead of `SLIC_ASSERT`

src/axom/slic/tests/slic_macros.cpp
kennyweiss: We can use std::source_location when we move to c++20 --- https://en.cppreference.com/w/cpp/utility/source_location.html

src/axom/slic/interface/slic_macros.hpp
bmhan12: Good catch! Fixed.

src/axom/slic/interface/slic_macros.hpp
bmhan12: > this can call the macro multiple times, until EXP evaluates as true.

I think this is the behavior we want, and is what was in the original #1683 snippet.

A once-only if-check I think would be false and print nothing most of the time, unless you knew the true condition ahead of time to guarantee the first invocation prints, in which case, your `SLIC_*_IF_ONCE` usage is more like `SLIC_*_ONCE` .

src/axom/quest/tests/quest_mesh_clipper.cpp
bmhan12: Looks like this test doesn't use gtest, uses the return value from `main()` to check for errors.

Arlie-Capps (approved):
Thanks for your careful work, @bmhan12 .

src/axom/slic/docs/sphinx/sections/wrapping_slic_in_macros.rst
```diff
@@ -111,38 +111,76 @@ functions are called.
...
-| | ``SLIC_ERROR_ROOT_IF``   | |                                              | |                                                                          |
-+----------------------------+------------------------------------------------+----------------------------------------------------------------------------+
-| | ``SLIC_WARNING``         | |   Always                                     | | Not collective by default.                                               |
-| | ``SLIC_WARNING_IF``      | |                                              | | Collective after calling ``slic::enableAbortOnWarning()``.               |
-| | ``SLIC_WARNING_ROOT``    | |                                              | | No longer collective after calling ``slic::disableAbortOnWarning()``     |
-| | ``SLIC_WARNING_ROOT_IF`` | |                                              | |                                                                          |
-+----------------------------+------------------------------------------------+----------------------------------------------------------------------------+
+.. list-table:: SLIC macro availability and collective behavior
```
rhornung67, line 114: @bmhan12 thank you for this. Much better way to make a table in sphinx -- avoid lining up all the characters across rows and columns 🥇

rhornung67 (approved):

PR #1832, opened 2026-03-27
Fix TopologyMapper crash on HIP

BradWhitlock:
`TopologyMapper` in Axom develop was reported to be crashing on GPU with HIP when built into a host code. This PR fixes that. Axom's own build and tests did not exhibit this problem so it could be related to compiler, flags, etc.

* Changed 2D polygon clipping so it uses fewer intermediate polygon objects since on GPU they have static storage and are largish. This alleviated some register pressure, improving reliability of the affected kernel.
* Changed some error checking code in the affected kernel that was using `SLIC_ASSERT()` in a few places. For HIP, we now just return early from the lambda. The `SLIC_ASSERT()` macros were indicating some kind of out of bounds but it seems more likely that using device `assert()` in this case was introducing errors.
* Fixed some newer warning messages in the code
* Made slight changes to `StaticArray` since it seemed implicated in the crash and added new tests.

NOTE: I have held off making the 0.14.0 release to create this fix and can get to that once this is merged.

### DIFF
--- a/scripts/github-actions/linux-build_and_test.sh
+++ b/scripts/github-actions/linux-build_and_test.sh
@@ -45,7 +45,7 @@ if [[ "$DO_BUILD" == "yes" ]] ; then
     fi
 
     echo "~~~~~~ RUNNING TESTS ~~~~~~~~"
-    make CTEST_OUTPUT_ON_FAILURE=1 test ARGS='-T Test -VV -j$NUM_BUILD_PROCS'
+    make CTEST_OUTPUT_ON_FAILURE=1 test ARGS='-T Test --output-on-failure -j$NUM_BUILD_PROCS'
 
     if [[ "${DO_BENCHMARKS}" == "yes" ]] ; then
         echo "~~~~~~ RUNNING BENCHMARKS ~~~~~~~~"


scripts/github-actions/linux-build_and_test.sh
```diff
@@ -45,7 +45,7 @@ if [[ "$DO_BUILD" == "yes" ]] ; then
     fi
 
     echo "~~~~~~ RUNNING TESTS ~~~~~~~~"
-    make CTEST_OUTPUT_ON_FAILURE=1 test ARGS='-T Test -VV -j$NUM_BUILD_PROCS'
+    make CTEST_OUTPUT_ON_FAILURE=1 test ARGS='-T Test --output-on-failure -j$NUM_BUILD_PROCS'
```
bmhan12, line 48: Should we keep the previous `-VV` flag here?

rhornung67: I think it should be there. Was there an issue with too much output?

src/axom/bump/TopologyMapper.hpp
```diff
@@ -26,6 +26,27 @@
 // Uncomment to emit debugging messages
 //#define AXOM_DEBUG_TOPOLOGY_MAPPER
 
+#if defined(AXOM_DEVICE_CODE) && defined(AXOM_USE_HIP)
```
rhornung67, line 29: what about the CUDA case?

BradWhitlock: At the moment, this is a HIP-specific workaround. I have not run into problems with `SLIC_ASSERT()` with CUDA.

PR #1834, opened 2026-03-27
Adds OMP support for existing GWN examples

jcs15c:
# Summary

This PR adds OMP support for the GWN examples in `quest`, which threading cache creation per surface and GWN evaluation per query. These changes are supported by adding "cache manager" classes to `winding_number_2d/3d_memoization.hpp` which provide a view to GWN caches used in execution, and a mutex in `quadrature.cpp` to ensure thread safety. For OMP examples, a separate cache is allocated for each thread. In the future, it might be worth investigating if there are benefits to collocating/binning the query points used for each thread to improve cache utilization. These changes are tested in `quest_gwn_methods.cpp`.

Many thanks to @BradWhitlock for the initial implementation of these methods, from which this PR is mostly derived.

### DIFF
--- a/src/axom/quest/examples/quest_winding_number_2d.cpp
+++ b/src/axom/quest/examples/quest_winding_number_2d.cpp
@@ -475,7 +475,7 @@ int main(int argc, char** argv)
   }
 
   axom::utilities::raii::AnnotationsWrapper annotation_raii_wrapper(input.annotationMode);
-  AXOM_ANNOTATE_SCOPE("winding number example");
+  AXOM_ANNOTATE_SCOPE("2D winding number example");
 
   if (!input.elevatedMeshFile.empty())
   {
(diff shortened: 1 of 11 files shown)


Arlie-Capps (approved):
Thanks, @jcs15c .

src/axom/quest/examples/quest_winding_number_2d.cpp
```diff
@@ -108,6 +112,16 @@ class Input
       ->capture_default_str()
       ->check(axom::utilities::ValidCaliperMode);
 #endif
+    std::stringstream pol_sstr;
+    pol_sstr << "Set MIR runtime policy method.";
```
kennyweiss:
```suggestion
    pol_sstr << "Set runtime policy method.";
```

src/axom/quest/examples/quest_winding_number_3d.cpp
```diff
@@ -160,6 +164,16 @@ class Input
       ->capture_default_str()
       ->check(axom::utilities::ValidCaliperMode);
 #endif
+    std::stringstream pol_sstr;
+    pol_sstr << "Set MIR runtime policy method.";
```
kennyweiss:
```suggestion
    pol_sstr << "Set runtime policy method.";
```

kennyweiss (approved):
Thanks @jcs15c -- please also update the RELEASE-NOTES.

PR #1838, opened 2026-04-03
Adds fast-approximation GWN methods for curved primitives

jcs15c:
# Summary

This PR extends the fast-approximation GWN methods in `quest` from linear elements to full NURBS support in 2D and 3D. To properly accommodate curved input, this requires a few adjustments to the algorithm:
 - New classes that support `NURBSCurveGWNCache` and `NURBSPatchGWNCache` during OMP execution. Currently the behavior is for each thread to get it's own copy of the memoization caches, since we found this to be more performant than using a single cache with mutexes (Thanks @BradWhitlock for investigating!)
 - New functions in `FastApproximateGWN.hpp` which subdivide NURBS curves/surfaces to improve the partition of the BVH.

These changes are also supported by incremental improvements to the direct evaluation methods in `primal`, including improvements to thread stability, removal of magic numbers, and streamlining the code for improved readability.

Altogether, these changes produce significant speedups in the evaluation of the GWN for collections of curved objects, particularly trimmed NURBS surfaces. Evaluated on large-scale datasets shown below (~20k examples of 2D SVG files and ~8k examples of 3D STEP files), we see orders of magnitude improvements, with some examples that took over 10^3s with the old algorithm now resolving in under a single second. Many thanks to @kennyweiss for this large-scale performance analysis!

image


src/axom/primal/operators/winding_number.hpp
```diff
@@ -874,6 +874,27 @@ axom::Array<double> winding_number(const axom::Array<Point<T, 3>>& query_arr,
...
+  {
+    const auto& normal = nurbs.getNormal();
+    if(normal.norm() >= EPS)
+    {
+      const auto unit_normal = normal.unitVector();
+      cast_direction[0] += std::abs(unit_normal[0]);
+      cast_direction[1] += std::abs(unit_normal[1]);
+      cast_direction[2] += std::abs(unit_normal[2]);
+    }
+  }
```
kennyweiss: This fixed the teardrop test, but is definitely the wrong approach! 
Let's revert this and find a better fix.

src/axom/primal/tests/primal_solid_angle.cpp
```diff
@@ -639,8 +639,20 @@ TEST(primal_solid_angle, teardrop_regression_test)
     {
       // The vertical cross-section of the tip is given by 3sin(1/3 * asin(1-2x)) + x - 0.5
       double fun = 3 * std::sin(1.0 / 3.0 * std::asin(1 - 2 * radius)) + radius - 0.5;
-      return (x[2] <= fun) &&
-        ((x[2] > -1.0) || std::sqrt(x[0] * x[0] + x[1] * x[1] + (x[2] + 1.0) * (x[2] + 1.0) <= 1.0));
```
kennyweiss, line 643: Note: The right parenthesis in the sqrt function was wrong -- it was doing a sqrt of a bool, not checking the result of the sqrt

src/axom/primal/operators/winding_number.hpp
```diff
@@ -874,26 +903,7 @@ axom::Array<double> winding_number(const axom::Array<Point<T, 3>>& query_arr,
...
-  }
-
-  // Use one deterministic cast direction for the full surface assembly so patch
-  // contributions remain additive, but derive it from the patch normals rather
-  // than hard-coding a single direction.
-  const auto bias_direction = Vector<T, 3> {1.0, 2.0, 3.0}.unitVector();
-  cast_direction = (cast_direction.norm() < EPS)
-    ? bias_direction
-    : (cast_direction.unitVector() + 0.1 * bias_direction).unitVector();
+  const auto cast_direction = detail::surface_winding_number_cast_direction<T>(nurbs_arr.view(), EPS);
```
kennyweiss: This needs to be reverted -- sorry for the noise.

src/axom/primal/operators/winding_number.hpp
jcs15c: I did a lot of digging, and it appears that the teardrop test was failing because cast-rays were intersecting with the surface exactly at the parameter points for which the patch is split along during the recursion. It was super sensitive too, with differences in the 10th decimal place being enough to go from a good answer to a bad answer: 
```
primal::Vector<double, 3> good_cast_dir {0.61817375459421131,
                                         0.61817375459421042,
                                         0.48551253151879892};
primal::Vector<double, 3> bad_cast_dir {0.61817375459421164,
                                        0.61817375459421164,
                                        0.48551253151879675};
```
The short term fix is likely to essentially do what you originally suggested, slightly perturbing the cast direction away from the true average normal vector, but I think that should be done on a per-patch basis, not trying to fit all the surfaces to the same cast direction. The obvious more long-term solution is to use a more robust intersection algorithm, probably Bezier clipping, but I think that's better suited to a separate PR.

src/axom/primal/tests/primal_solid_angle.cpp
jcs15c: Fixed!

src/axom/quest/GWNMethods.hpp
```diff
@@ -149,17 +155,86 @@ class DirectGWN2D
...
+
+    AXOM_ANNOTATE_SCOPE("preprocessing");
+
+    if(!use_direct_eval)
+    {
+      stage_timer.reset();
+      stage_timer.start();
+      {
+        AXOM_ANNOTATE_SCOPE("subdivision");
+        constexpr double subdivision_bbox_threshold = 0.1;
```
kennyweiss: Should this be exposed as a user-facing parameter, w/ a default of 10%?

src/axom/quest/GWNMethods.hpp
```diff
@@ -149,17 +155,86 @@ class DirectGWN2D
...
+        m_subdivided_curves = subdivide_curves(m_input_curves_view, subdivision_bbox_threshold);
+        m_processed_curves_view = m_subdivided_curves.view();
+      }
+      stage_timer.stop();
+      SLIC_INFO(axom::fmt::format("  Preprocessing stage (subdivision {} -> {}): {} s",
```
kennyweiss: We have two forms of performance capture here -- logging via axom's timers and the caliper annotations.

These logs can get noisy in an application. I think we should remove the direct logging from this file, in favor of the user application deciding whether/when/what to log. 

Perhaps store the timings as class member variables if you'd like the user to be able to report this.

src/axom/quest/GWNMethods.hpp
```diff
@@ -128,18 +128,24 @@ void generate_gwn_query_mesh(mfem::DataCollection& dc,
 ///@{
 /// \name Query methods for 2D GWN applications
 
-template <typename ExecSpace>
-class DirectGWN2D
+template <typename ExecSpace, int ORDER = 2>
```
kennyweiss, line 145: Please add a short doxygen comment here describing the class

src/axom/quest/GWNMethods.hpp
```diff
@@ -248,12 +363,21 @@ class DirectGWN2D
...
+  axom::ArrayView<const CurveType> m_processed_curves_view;
+
+  // Only needed for fast approximation method
+  axom::Array<GWNMoments> m_internal_moments;
+  axom::spin::BVH<2, ExecSpace> m_bvh;
 };
 
-template <typename ExecSpace, int ORDER>
-class PolylineGWN2D
+template <typename ExecSpace, int ORDER = 2>
```
kennyweiss, line 385: Please add a short doxygen comment here describing the class

src/axom/quest/GWNMethods.hpp
```diff
@@ -447,19 +575,24 @@ class PolylineGWN2D
 ///@{
 /// \name Query methods for 3D GWN applications
```
kennyweiss, line 549: Please expand this a bit, including the template paramters

src/axom/quest/GWNMethods.hpp
```diff
@@ -468,18 +601,116 @@ class DirectGWN3D
...
+
+    if(!use_direct_eval)
     {
-      AXOM_ANNOTATE_SCOPE("preprocessing");
-      if(use_memoization)
+      stage_timer.reset();
+      stage_timer.start();
+      {
+        AXOM_ANNOTATE_SCOPE("subdivision");
+        constexpr double subdivision_bbox_threshold = 0.1;
```
kennyweiss: Similarly to 2D -- this should be exposed as a user-settable parameter w/ a default of 10%

src/axom/quest/GWNMethods.hpp
```diff
@@ -468,18 +601,116 @@ class DirectGWN3D
...
+        m_internal_moments = traverser.template reduce_tree<ExecSpace, GWNMoments>(compute_moments);
       }
+      stage_timer.stop();
+      SLIC_INFO(axom::fmt::format("  Preprocessing stage (moment precomputation): {} s",
+                                  stage_timer.elapsedTimeInSec()));
     }
+    else
+    {
+      // Without fast-approximation, processing is unnecessary, but we still clean the
+      //  the trimmed representation for more precise moment calculation
```
kennyweiss, line 675: Please briefly expand on what cleaning does

src/axom/quest/GWNMethods.hpp
```diff
@@ -524,46 +753,88 @@ class DirectGWN3D
     {
       AXOM_ANNOTATE_SCOPE("query");
       const primal::WindingTolerances tol_copy = tol;
-      const auto input_patches_view = m_input_patches_view;
 
-      // Use non-memoized form
-      if(m_nurbs_cache_mgr.empty())
+      // Use fast approximation
+      if(m_bvh.isInitialized())
```
kennyweiss, line 740: This might read a bit more easily if the `m_bvh.isInitialized()` was behind a function like  
`this->useFastApproximation()`, or perhaps if we used an enum for the processing method instead of a switchyard.

src/axom/quest/GWNMethods.hpp
```diff
@@ -524,46 +753,88 @@ class DirectGWN3D
...
-          inout[static_cast<int>(nidx)] = std::lround(wn);
-        });
+        const auto traverser = m_bvh.getTraverser();
+        auto internal_moments_view = m_internal_moments.view();
+
+        // Use fast-approximate, non-memoized form
+        if(m_nurbs_cache_mgr.empty())
+        {
+          SLIC_WARNING(
+            "Quest warning: Patch GWN evaluation is prohibitively slow without memoization.");
```
kennyweiss, line 749: 👍

kennyweiss: Is there a good reason to support the non-memoized patch GWN?  Perhaps for debugging?

src/axom/quest/GWNMethods.hpp
```diff
@@ -584,20 +856,29 @@ class DirectGWN3D
...
+  axom::ArrayView<const PatchType> m_processed_patches_view;
+
+  // Only needed for fast approximation method
+  axom::Array<GWNMoments> m_internal_moments;
+  axom::spin::BVH<3, ExecSpace> m_bvh;
 };
 
-template <typename ExecSpace, int ORDER>
-class TriangleGWN3D
+template <typename ExecSpace, int ORDER = 2>
```
kennyweiss, line 856: Please add a brief class doxygen comment

src/axom/quest/FastApproximateGWN.hpp
```diff
@@ -424,11 +460,220 @@ double fast_approximate_winding_number(const primal::Point<T, NDIMS>& query,
...
+    candidates.push_back(the_patch);
+    total_bbox.addBox(the_patch.boundingBox());
+  }
+
+  // Iterate over all the surfaces until no patch has a bounding box
+  //  bigger than threshold * (total_bbox's size)
+  for(int i = 0; i < npasses; ++i)
+  {
+    axom::Array<NURBSType> subdivisions;
+    subdivisions.reserve(candidates.size());
```
kennyweiss: Suggestion: it might be cheaper for the subdivisions array to be outside the loop and clear it at the beginning of each pass.

src/axom/quest/FastApproximateGWN.hpp
```diff
@@ -424,11 +460,220 @@ double fast_approximate_winding_number(const primal::Point<T, NDIMS>& query,
...
+    // Break if over 100,000 surfaces
+    if(candidates.size() > 1e5)
+    {
+      std::cout << axom::fmt::format("Excessive subdivision recorded after {} passes! ({} -> {})",
+                                     i,
+                                     input_patches_view.size(),
+                                     candidates.size())
+                << std::endl;
+      break;
+    }
```
kennyweiss: Is this debugging code or a heuristic that should stay in the code.
We might want the upper threshold to be user-settable, as a class variable, with a default of 1e5.

Please convert to `SLIC_WARNING` instead of `cout`

src/axom/quest/tests/quest_gwn_methods.cpp
```diff
@@ -203,45 +203,75 @@ void check_mfem_mesh_linearization()
...
+
+  mfem::VisItDataCollection windingDC_2("winding_2", dc[2].GetMesh());
+  windingDC_2.RegisterField("winding", dc[2].GetField("winding"));
+  windingDC_2.RegisterField("inout", dc[2].GetField("inout"));
+  windingDC_2.Save();
+
+  mfem::VisItDataCollection windingDC_3("winding_3", dc[3].GetMesh());
+  windingDC_3.RegisterField("winding", dc[3].GetField("winding"));
+  windingDC_3.RegisterField("inout", dc[3].GetField("inout"));
+  windingDC_3.Save();
```
kennyweiss: Suggestion: Perhaps the writes should be off by default for standard testing

kennyweiss: Also:
* Might was well give the output the name of the method, e.g. "gwn_curves_memoized", instead of `winding_1`
* Any reason not to output the evaluated linearized WN ?

src/axom/quest/examples/quest_winding_number_2d.cpp
```diff
@@ -329,6 +329,19 @@ class Input
       ->check(axom::CLI::PositiveNumber)
       ->capture_default_str();
 
+    app.add_option("--algorithm", algorithm)
+      ->description(
+        "Use direct evaluation instead of fast, heirarchical approximation? (significantly "
```
kennyweiss, line 334: Typo:

```suggestion
        "Use direct evaluation instead of fast, hierarchical approximation? (significantly "
```

src/axom/primal/operators/detail/winding_number_3d_memoization.hpp
```diff
@@ -155,35 +155,34 @@ class NURBSPatchGWNCache
...
-    }
-    else
+    if(computeNormal)
     {
-      m_averageNormal = m_alteredPatch.calculateTrimmedPatchNormal();
-    }
+      // In a GWN context, the surface has been sufficiently subdivided to
+      //  make additional bezier extraction redundant
+      const auto normal_area =
+        m_alteredPatch.calculateTrimmedPatchNormalArea(/*npts*/ 20, /*useBezierExtraction*/ false);
```
kennyweiss: Should `npts` be a class parameter w/ a default of 20?

src/axom/primal/operators/detail/winding_number_3d_memoization.hpp
```diff
@@ -358,18 +411,39 @@ class NURBSPatchCacheManagerOMP
...
     nurbs_caches_view[0].resize(patches.size());
     axom::for_all<axom::OMP_EXEC>(
       patches.size(),
-      AXOM_HOST_LAMBDA(axom::IndexType i) { nurbs_caches_view[0][i] = NURBSCache(patches[i]); });
-    SLIC_INFO("Finished the first construction");
+      AXOM_LAMBDA(axom::IndexType i) {
+        nurbs_caches_view[0][i] = NURBSCache(patches[i], mustComputeNormal);
+      });
+
+    // If we didn't comptue normals in NURBSCache constructor,
```
kennyweiss, line 436: Typo:

```suggestion
    // If we didn't compute normals in NURBSCache constructor,
```

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -3714,6 +3934,65 @@ class NURBSPatch
     return true;
   }
 
+  /*!
+   * \brief Split the patch in the longer parametric direction, determined via maximum control-net polyline length.
+   *
+   * Rather than calculating It measures the maximum polyline length of the control net in each parametric direction:
```
kennyweiss: >    Rather than calculating It measures the maximum polyline length of the control net in each parametric direction

Please clean up this doxygen line

kennyweiss (approved):
Thanks @jcs15c !

My review mostly focused on reduced logging in library code and improving comments.

src/axom/quest/GWNMethods.hpp
jcs15c: I don't think so, we found pretty solid empirical evidence that 10% gets as much benefit as a person could want. I could maybe see some value to adding a flag for "subdivision (at 10%)" vs "no subdivision" for debugging purposes, but I don't think it's helpful to have more granularity to that choice.

src/axom/quest/GWNMethods.hpp
jcs15c: I'm fine using only the caliper annotations. The timer ones were useful during debugging/analysis, but I don't think they need to be there for the final version of this code.

src/axom/quest/GWNMethods.hpp
jcs15c: Optimistically, it's possible that we could improve the performance of our direct evaluation so that it isn't prohibitively expensive without memoization. I could see it being useful in a non-memoized + agglomerated context that can afford the extra memory footprint for the BVH, but not memoization, but you're right that it's not very useful right now. Still, it doesn't introduce a lot of extra stuff that has to be maintained, so I think it's worth keeping for now.

src/axom/quest/FastApproximateGWN.hpp
jcs15c: I like the idea of "maximum subdivision" as a user-set parameter more than AABB threshold, it has a much more intuitive implication for the actual code, and is a bit more algorithm agnostic. Thanks!

src/axom/quest/examples/quest_winding_number_2d.cpp
jcs15c: One day I will spell that word right, I need to get in the habit of sounding it out.

src/axom/quest/GWNMethods.hpp
jcs15c: I expect that as we go down this path and add more options to each of these individual classes (e.g. different agglomeration strategies, different direct evaluation methods, etc), we're going to need a more robust way of selecting these things in general. I think while we only have two disjoint settings (agglom vs direct, memoized vs nonmemoized) this switchyard is as efficient as it can get, but we will almost certainly want to revisit the overall structure in the future.

src/axom/quest/GWNMethods.hpp
jcs15c: Actually, maybe there's a compromise in the form of class getters/setters for this and other constants, rather than it being passed in as an argument to `NURBSPatchGWNQuery::preprocess`? In my mind, that really cements them as sort of "expert level parameters" that you normally shouldn't have to change, but you technically do have access to if you want them.

kennyweiss: Thanks -- that sounds like a good compromise

src/axom/quest/tests/quest_gwn_methods.cpp
jcs15c: Oh whoops! These were debug statements I had when I was testing things. They shouldn't be in the testing at all, optional or otherwise, properly named or otherwise.

src/axom/primal/operators/detail/winding_number_3d_memoization.hpp
jcs15c: I've been convinced about why AABB threshold can be effectively made user-facing, but I still think this should be a bit of a magic number. We've seen that 10 points (updated from 20) is not only way more than necessary, but the additional evaluation cost is essentially negligible. I think making it a class parameter just muddies the usage more than it's worth.

This did remind me to make sure the evaluation of `calculateSurfaceMoments` in `GWNMomentData` matches the number of points as here though! Again, not that 10 vs 20 points makes any difference at all.

src/axom/quest/GWNMethods.hpp
```diff
@@ -128,18 +128,61 @@ void generate_gwn_query_mesh(mfem::DataCollection& dc,
...
 
-template <typename ExecSpace>
-class DirectGWN2D
+/*!
+ * \class NURBSCurveGWNQuery
+ *
+ * \tparam ExecSpace The execution space for the algorithm.
+ * \tparam ORDER If agglomeration is used, this is the order of the Taylor expansion.
+ *
+ * \brief Preprocesses NURBSCurve geoemtry for GWN evaluation,
```
Arlie-Capps: Typo:
```suggestion
 * \brief Preprocesses NURBSCurve geometry for GWN evaluation, 
```

Arlie-Capps (approved):
Thank you for the hard work!  Approved (pending typo fix).

PR #1841, opened 2026-04-09
[WIP] Attempt to simplify core memory management interface

rhornung67:
# Summary

- This PR is intended to start a discussion about potential changes to Axom's memory management interfaces.
- Specifically, this PR refactors the Axom memory management API in core.
- The most notable change is that the `AXOM_USE_UMPIRE` macro is used only in implementation files in the core component, except for a couple of instances to temporarily maintain the existing interfaces, rather than being peppered across both header and implementation files.
- The idea is to simplify and reduce the degree to which the API is different when Axom is configured with Umpire enabled and when it is not.
- These changes were developed in an iterative process exploring ideas with codex.
- Based on team discussion, **this would be the first of a sequence of PRs with two main goals:**
     - Unify and simplify the use of memory management operations across Axom components primarily sidre, mir, and quest
     - Have Axom's malloc routine be the default across Axom for CPU/host memory allocation, but allow users to choose the Umpire HOST allocator if desired. Currently, the Umpire HOST allocator is the default. It is significantly slower than the Axom malloc allocator.

- **The main concern I have is that changes like these may break application user codes initially.**
    - However, I think that they would benefit in the longer term.
    - I'm not sure how applications use Axom memory management operations directly, if at all. Maybe they just assume the default host memory behavior. 
    - **If you work on an application and know of any potential issues changes like the ones contained in this PR may cause, please post comments on this PR.**


rhornung67:
Closing in favor of https://github.com/llnl/axom/pull/1842 which is more complete

PR #1842, opened 2026-04-10
Rework host memory interface and defaults

rhornung67:
# Summary

This PR is intended to help discuss potential changes to Axom's host/cpu memory management API in core.  There is more work to do, including a follow-on PR.

It reworks Axom core memory-management semantics to separate the global default allocator from Axom’s default host allocator, and updates host-facing code paths to use that distinction consistently.

Specifically, when Umpire is enabled:
- the default host allocator is Axom malloc
- the default global allocator is Umpire's default allocator

When Umpire is disabled:
- the default host allocator and the default global allocator are both Axom malloc

**Note: This makes the Axom default host allocator use Axom malloc in either case for performance reasons!**

The biggest functional change is in axom::reallocate(), where cross-backend migration between malloc and Umpire is now much more clear in the code and handled more robustly. In particular, malloc -> Umpire migration is more robust and more clear since sizes for Axom-owned malloc allocations are tracked and allocate/copy/deallocate migration is explicit. Attempts to migrate foreign raw malloc pointers to Umpire now fail loudly rather than relying on brittle behavior that could silently do something that users are are not expecting or aware of. Please see: https://github.com/llnl/axom/blob/task/rhornung67/rework-mem-mgmt_part2/src/axom/core/memory_management.hpp#L623

The PR also formalizes MemorySpace::Host as “Axom’s configured host allocator”, which is distinct from MemorySpace::Dynamic. New APIs and logic in memory_management.{hpp, cpp} expose host allocator configuration, allocator-space compatibility checks, and memory-space availability queries. Array, ArrayView, execution-space traits, and related tests/examples were updated to check the new host-allocator semantics. In particular, tests were expanded where they previously assumed CPU execution spaces should always follow the global default allocator. Documentation in the GPU porting guide was also updated to clarify the Host vs Dynamic distinction.

The main conceptual idea here is:
- The Axom global default allocator is the default for `MemorySpace::Dynamic` and for APIs that do not specify an allocator (answers question: what allocator should Axom use by default?)
- The Axom host allocator is the default for `MemorySpace::Host` (answers question: what allocator should Axom use when someone explicitly asks for host-allocated CPU memory?)

The separation of these concepts is intentional b/c, in Umpire builds, the global default may be device, unified, or some other Umpire allocator, when the host allocator can still be Axom malloc or Umpire Host.

Axom host allocator choice can now be made at run time. For example:
```
  // set Axom host allocator to Axom malloc
  axom::setDefaultHostAllocator(axom::MemorySpace::Malloc); 

  // set Axom host allocator to Umpire Host allocator
  axom::setDefaultHostAllocator(axom::MemorySpace::Host);

  // set Axom host allocator to an explicitly chosen Umpire allocator
  int hostId =
    axom::getUmpireResourceAllocatorID(umpire::resource::MemoryResourceType::Host);
  axom::setDefaultHostAllocator(hostId);
```

Similarly, to change the Axom global  host allocator:
```
  // set Axom global allocator to Umpire Host
  axom::setDefaultAllocator(axom::MemorySpace::Host); 

  // set Axom global allocator to Umpire Unified memory allocator
  axom::setDefaultAllocator(axom::MemorySpace::Unified);

  // set Axom global allocator to an explicitly chosen Umpire allocator
  int allocId =
    axom::getUmpireResourceAllocatorID(umpire::resource::MemoryResourceType::Pinned);
  axom::setDefaultAllocator(allocId);
```

One important restriction in this branch is that, when Umpire is enabled, you cannot set the global default allocator to Axom malloc; e.g., this will abort:
```
  // currently, can't do this -- code will abort
  axom::setDefaultAllocator(axom::MALLOC_ALLOCATOR_ID); 
```
**(1) Should this restriction be reversed?**

Other notable considerations:
- The default allocation behavior in Sidre is tied to Axom's global default allocator, which is different than the default Memory::Space::Host behavior used in Axom core. But, Sidre group allocators are just mapped to allocatorsIDs and so can be explicitly set to anything.  **(2) Should the default allocation behavior in Sidre be the same as in Axom core?**
- I tried to make the use of the`AXOM_USE_UMPIRE` macro into function implementations and to guard methods that are clearly different when Umpire is/is not available. In particular, I tried to remove instances of function signature types changing based on that macro definition. One goal for me is to simplify and reduce the degree to which an Axom API is different when Axom is configured with Umpire enabled and when it is not.

Disclaimer: These changes were developed in an iterative process using codex exploring what it could do and how to make it explain itself when alternatives are available.


rhornung67:
@publixsubfan I especially would like your thoughts on the changes to the array stuff in core in this PR. Those things were the most tightly constrained with respect to always using the Umpire HOST allocator when Umpire is enabled and always using Axom's MALLOC "allocator" otherwise. I unraveled some of that, but I am not clear on whether the constraints are required for some use cases we don't test.

src/axom/core/Array.hpp
```diff
@@ -1149,7 +1149,7 @@ Array<T, DIM, SPACE, StoragePolicy>::Array(IndexType num_elements, IndexType cap
 {
   // If a memory space has been explicitly set for the Array object, check that
   // the space of the user-provided allocator matches the explicit space.
-  if(SPACE != MemorySpace::Dynamic && SPACE != axom::detail::getAllocatorSpace(m_allocator_id))
+  if(!axom::isAllocatorCompatibleWithMemorySpace(m_allocator_id, SPACE))
```
publixsubfan, line 1152: It looks like the `SPACE != MemorySpace::Dynamic` check got pulled into this method. I think that's fine from a correctness standpoint, but wondering from a design standpoint about overloading the meaning of "dynamic": in the memory management code it's being used as "user-specified default allocator," but here in `Array/ArrayView` it's being used as "works with any allocator."

rhornung67: Thanks @publixsubfan. There were some things that were not entirely clear to me w.r.t. yhr meaning of `Dynamic`.  I'll take a closer look tomorrow and, if I push a change, I will ask you for another look.

publixsubfan: Yeah, I think the `MemorySpace` enum was first introduced solely as a helper for Array, so I don't mind a design change that makes it more broadly usable in the code. Just thinking about if it might be also worth revisiting how Array uses it for "dynamic allocators."

rhornung67: @publixsubfan I think a good goal is to have it be very clear what `MemorySpace` enums `Malloc`, `Host`, and `Dynamic` mean. It seems that `Malloc` should always refer to Axom malloc allocator, `Host` should always refer to the host default allocator. My goal is to have the default host allocator be Axom malloc, whether Axom is configured with Umpire or not, and be the Umpire HOST allocator if a user sets it to that.
Do you agree with that? Or, do you have a preference for something different?

That leaves `Dynamic`, which is a bit blurry to me. What are your thoughts on that? Is it even needed?

src/axom/core/memory_management.cpp
```diff
@@ -15,9 +15,185 @@
...
+
+  int get() const noexcept { return allocatorId.load(std::memory_order_relaxed); }
+
+  void set(int allocId) noexcept { allocatorId.store(allocId, std::memory_order_relaxed); }
+
+private:
+  std::atomic<int> allocatorId;
+};
+
+HostAllocatorConfig defaultHostAllocatorConfig {platformHostAllocatorID()};
```
publixsubfan, line 69: Would we run into static initialization order issues with this variable?

rhornung67: Fixed

rhornung67:
Folks, I think this is just about ready to merge. Please review it and provide feedback, comments, questions, etc.

rhornung67:
Ping @publixsubfan @kennyweiss @bmhan12 @Arlie-Capps @nselliott @BradWhitlock @jcs15c 

This is ready to scrutinize and review. Please look it over when you have time.

src/axom/core/docs/sphinx/core_memory_management.rst
```diff
@@ -0,0 +1,79 @@
...
+.. ## Axom Project Contributors. See top-level LICENSE and COPYRIGHT
+.. ## files for dates and other details.
+.. ##
+.. ## SPDX-License-Identifier: (BSD-3-Clause)
+
+******************************************************
+Core memory management
+******************************************************
+
+The Axom Core component provides mechanisms to control which memory spaces
```
white238, line 11: https://axom.readthedocs.io/en/task-rhornung67-rework-mem-mgmt_part2/axom/core/docs/sphinx/core_memory_management.html

src/axom/core/tests/core_memory_management.hpp
```diff
@@ -254,78 +273,214 @@ void check_alloc_realloc_free(bool hostAccessible = true)
...
+
+  #if defined(UMPIRE_ENABLE_DEVICE)
+  EXPECT_TRUE(axom::isMemorySpaceAvailable(axom::MemorySpace::Device));
+  #endif
+
+  #if defined(UMPIRE_ENABLE_UM)
+  EXPECT_TRUE(axom::isMemorySpaceAvailable(axom::MemorySpace::Unified));
+  #endif
+
+  #if defined(UMPIRE_ENABLE_PINNED)
```
white238, line 292: These specific compiler defines aren't guarding all the places that they are used.. for example src/axom/core/memory_management.cpp

RELEASE-NOTES.md
```diff
@@ -36,6 +36,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   removed in a future version of Axom.
 - Core: Adds Durand-Kerner polynomial solver which returns the complex roots of a univariate polynomial
 - Core: Adds `axom::Array::pop_back` for API compatibility with `std::vector`
+- Core: Axom core memory allocation semantics and defaults have changed to be more consistent. Specifically, the separation of the Axom default global allocator and default host allocator (both used for CPU-accessible allocations) is more clear. Host-facing code paths have been updated to use that distinction consistently. Please see Core component user documentation for more inofrmaiton.
```
white238:
```suggestion
- Core: Axom core memory allocation semantics and defaults have changed to be more consistent. Specifically, the separation of the Axom default global allocator and default host allocator (both used for CPU-accessible allocations) is more clear. Host-facing code paths have been updated to use that distinction consistently. Please see Core component user documentation for more information.
```

src/axom/spin/tests/spin_bvh.cpp
```diff
@@ -7,6 +7,8 @@
 // axom includes
 #include "axom/config.hpp"
 #include "axom/core.hpp"
+#include "axom/core/utilities/MemoryTesting.hpp"
+
```
white238:
```suggestion
#include "axom/core/utilities/MemoryTesting.hpp"
```

src/axom/core/docs/sphinx/core_memory_management.rst
rhornung67: Did you post the link so folks could see the generated documentation, or do you you have a comment, question, etc.?

RELEASE-NOTES.md
rhornung67: Fixed.

src/axom/spin/tests/spin_bvh.cpp
rhornung67: This suggested change is no change at all.  Did you want something in particular here?

src/axom/core/tests/core_memory_management.hpp
rhornung67: I think that's OK. In the `core/memory_management.{hpp,cpp}` files, we only guard to the extent that the enum values are available. In the tests here, we want do want to do things that aren't possible.

src/axom/spin/tests/spin_bvh.cpp
white238: this was a giant nitpick that there was a extra newline. this file had all the axom headers together but this was breaking it apart in the middle. feel free to just ignore.

rhornung67: Ah, I see. I thought I was adding a newline between includes from different components and that we did that. I guess not. I will remove it.

rhornung67: Newline removed

src/axom/core/docs/sphinx/core_memory_management.rst
white238: just activating the docs and posting the link so people can see it.

white238 (approved):

src/axom/core/memory_management.hpp
```diff
@@ -78,12 +78,12 @@ enum class MemorySpace
 {
   Malloc,   //!< Host memory using malloc, free and realloc
   Dynamic,  //!< Refers to Umpire's current default allocator
-#ifdef AXOM_USE_UMPIRE
-  Host,     //!< Umpire's host memory space
-  Device,   //!< Umpire's device memory space
-  Unified,  //!< Umpire's unified memory space
-  Pinned,   //!< Umpire's pinned memory space
-  Constant  //!< Umpire's constant memory space
+  Host,     //!< Host memory space
```
kennyweiss: Since `Host` is now defined outside of `AXOM_USE_UMPIRE`, should the doxygen comment directly address `Host`?

src/axom/core/tests/core_array.hpp
```diff
@@ -2306,6 +2315,103 @@ TEST(core_array, checkUninitialized)
   }
 }
 
+//------------------------------------------------------------------------------
+TEST(core_array, host_space_accepts_malloc_allocator)
+{
+  ScopedDefaultHostAllocatorStateForArray scopedState;
+  axom::setDefaultHostAllocator(axom::MemorySpace::Malloc);
```
kennyweiss, line 2322: Does this test need to change the default host allocator?
It looks like each entity (Array/Arrayview) is getting an explicit allocator.

(and similarly for several tests below)

src/axom/core/memory_management.hpp
```diff
@@ -318,12 +391,160 @@ struct Allocator
 //                        IMPLEMENTATION
 //------------------------------------------------------------------------------
 
+namespace detail
+{
+
+void registerMallocAllocation(const void* pointer, std::size_t numbytes) noexcept;
```
kennyweiss, line 397: Please add brief comments to the functions in this detail namespace.

src/axom/core/memory_management.cpp
kennyweiss, line 1: I know the functions/structs in this file won't get doxygened, but please add some brief comments to some of the functions/struct, e.g. `platformHostAllocatorID()`, `MallocAllocationRegistry`, `HostAllocatorConfig`.

src/axom/core/memory_management.cpp
```diff
@@ -15,9 +15,296 @@
...
+
+  int get() const noexcept { return allocatorId.load(std::memory_order_relaxed); }
+
+  void set(int allocId) noexcept { allocatorId.store(allocId, std::memory_order_relaxed); }
+
+private:
+  std::atomic<int> allocatorId;
+};
+
+struct MallocAllocationRegistry
```
kennyweiss, line 75: Are we concerned about the performance cost of tracking malloc allocations? 
One reason to use the malloc allocator was to avoid bookkeeping costs from Umpire.
Should tracking be opt-in, e.g.  via a config variable?

src/axom/core/memory_management.cpp
```diff
@@ -15,9 +15,296 @@
...
+{
+  static HostAllocatorConfig config {initialDefaultHostAllocatorID()};
+  return config;
+}
+
+MallocAllocationRegistry& mallocAllocationRegistry() noexcept
+{
+  static MallocAllocationRegistry registry {};
+  return registry;
+}
```
kennyweiss, line 147: Please mention that these are singletons.
Is there a concern that we could have different host/malloc allocators in different translation units?

src/axom/core/memory_management.hpp
kennyweiss, line 190: Consider adding a comment like the following here -- 
```
// no-op -- in non Umpire configs, the default allocator is always the Malloc allocator
```

src/axom/core/memory_management.hpp
```diff
@@ -318,12 +391,160 @@ struct Allocator
...
+{
+  std::cerr << "Cannot migrate malloc-backed pointer " << pointer << " to allocator id " << allocID
+            << " because its allocation size is not tracked by Axom. "
+               "Pointers migrated from malloc to Umpire must be allocated through "
+               "axom::allocate()/axom::reallocate()."
+            << std::endl;
+  axom::utilities::processAbort();
+}
+
+template <typename T>
```
kennyweiss, line 477: This function can likely benefit from a comment about the behavior/rationale.

src/axom/core/memory_management.hpp
```diff
@@ -399,69 +625,62 @@ inline T* reallocate(T* pointer, std::size_t n, int allocID) noexcept
   assert(allocID != INVALID_ALLOCATOR_ID);
 
   const std::size_t numbytes = n * sizeof(T);
+  const detail::AllocationBackend dst = detail::getAllocatorBackend(allocID);
 
-#if defined(AXOM_USE_UMPIRE)
+  if(dst == detail::AllocationBackend::Invalid)
+  {
```
kennyweiss, line 631: We worked through a lot compatibility issues/bugs related to realloc for 0-sized arrays.  
I think we have good test coverage for this, but it'd be good to double check.

RELEASE-NOTES.md
```diff
@@ -36,6 +36,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
   removed in a future version of Axom.
 - Core: Adds Durand-Kerner polynomial solver which returns the complex roots of a univariate polynomial
 - Core: Adds `axom::Array::pop_back` for API compatibility with `std::vector`
+- Core: Axom core memory allocation semantics and defaults have changed to be more consistent. Specifically, the default allocation stractegy for host allocations is now Axom's malloc routine; before it was the Umpire default HOST allocator. Also, the separation of the Axom default global allocator and default host allocator (both used for CPU-accessible allocations) is more clear. Host-facing code paths have been updated to use that distinction consistently. Please see Core component user documentation for more information.
```
kennyweiss, line 40: Typo: 

```suggestion
- Core: Axom core memory allocation semantics and defaults have changed to be more consistent. Specifically, the default allocation strategy for host allocations is now Axom's malloc routine; before it was the Umpire default HOST allocator. Also, the separation of the Axom default global allocator and default host allocator (both used for CPU-accessible allocations) is more clear. Host-facing code paths have been updated to use that distinction consistently. Please see Core component user documentation for more information.
```

kennyweiss (approved):
This is a really nice set of improvements @rhornung67!
I have two concerns:
* The default global and host allocators are singletons that get statically linked into our applications. It is possible that we can get different default allocators in different statically linked executables. To get around some of this w/ Umpire, we had to dynamically link.  I am not sure that there is a better way though (other than to avoid default allocators)
* Tracking malloc allocations/deallocations can add overhead (in this cases, hash map lookups w/ mutexes), especially in cases where we have a lot of small allocations. Should we allow this to be opt-in? If so, are there implications associated w/ copies to/from Umpire?

src/axom/core/docs/sphinx/core_memory_management.rst
rhornung67: Thank you!

rhornung67:
> This is a really nice set of improvements @rhornung67! I have two concerns:
> 
>     * The default global and host allocators are singletons that get statically linked into our applications. It is possible that we can get different default allocators in different statically linked executables. To get around some of this w/ Umpire, we had to dynamically link.  I am not sure that there is a better way though (other than to avoid default allocators)
> 
>     * Tracking malloc allocations/deallocations can add overhead (in this cases, hash map lookups w/ mutexes), especially in cases where we have a lot of small allocations. Should we allow this to be opt-in? If so, are there implications associated w/ copies to/from Umpire?

Regarding default allocators, I will look into alternative implementations to try to avoid static linkage. What is a scenario in which we would have "different default allocators in different statically linked executables"?

I am also concerned with the overhead of tracking malloc allocation sizes, since that's the sort of thing we are trying to avoid by makin Axom malloc the default host allocation scheme. However, I don't see any other way to support run time switching from Axom malloc to Umpire HOST allocators. I'm not sure why anyone would do that, but the previous implementation was brittle. Should we disallow switching the default allocators after an allocation from them occurs?

src/axom/core/memory_management.hpp
Arlie-Capps: Why is `Host` now defined outside `AXOM_USE_UMPIRE`?  This change puts `Host` on the same level as `Malloc`.  The discussion in the Sphinx document makes more sense if I understand that `Host` means Umpire's host allocator.

I think I agree with Kenny: if `Host` needs to move outside the guard, please do address it in the doxygen.

src/axom/core/docs/sphinx/core_memory_management.rst
```diff
@@ -0,0 +1,79 @@
...
+When Umpire is disabled:
+
+  * the default host allocator and the default global allocator are both refer to ``MemorySpace::Malloc``
+
+.. note:: The Axom default host allocator is ``axom::MemorySpace::Malloc`` regardless of whether Axom is configured with Umpire.
+
+The separation of host and global allocation in Axom is intentional because,
+in Umpire-enabled builds, the global default may be device, unified, or some
+other Umpire allocator, and the host allocator can still be Axom's malloc or
+Umpire's Host allocator.
```
Arlie-Capps, line 45: Thanks for this--- it makes sense, but only if `MemorySpace::Host` refers to an Umpire idea.

Edit: this was from an incomplete understanding of `Host` vs. `Malloc`.  See below.

src/axom/quest/tests/quest_signed_distance.cpp
```diff
@@ -297,9 +297,17 @@ void run_vectorized_sphere_test()
 {
   using PointType = primal::Point<double, 3>;
 
-  int host_allocator = axom::execution_space<axom::SEQ_EXEC>::allocatorID();
+  const int current_allocator = axom::getDefaultAllocatorID();
+  int host_allocator = axom::getDefaultHostAllocatorID();
   int kernel_allocator = axom::execution_space<ExecSpace>::allocatorID();
```
Arlie-Capps, line 302: Unless there's a specific reason to use `axom::execution_space<ExecSpace>::allocatorID()`, consider changing to `int kernel_allocator = axom::getDefaultAllocatorID();`.

Arlie-Capps: See also (new) line 460.

src/docs/sphinx/dev_guide/gpu_porting.rst
Arlie-Capps, line 194: Suggestion: rephrase to "Pass ArrayView by value, not by reference."  That concept gets clearly explained in the paragraph!  But I had unfortunately mis-read the first sentence as a warning against passing by value.  Total goof on my (the reader's) part, but adding a bit of redundant information can really help people who are tired and happen to read carelessly.

By the way, the "Do this: / Do NOT do this:" was what helped me catch my mis-read.  Thank you for writing these examples.

Arlie-Capps: Un-double the back-ticks on "auto": try changing ` ``auto`` ` to have just one pair.  I don't think ``code`` mixes well with **bold.**

Arlie-Capps, line 222: Maybe add

```
/* Here, dataView is a reference. This causes the memory-space error described in the for_all body. */
```

Arlie-Capps (approved):
Thanks for this work, Rich.  Approved.

I feel like I'm blanking on why `MemorySpace::Host` got moved out of the guard on using Umpire.  Inside the guard, it makes sense to me: this is the Umpire Host allocator!  Outside the guard, I don't see the difference between `MemorySpace::Host` and `MemorySpace::Malloc`.

src/axom/core/memory_management.hpp
rhornung67: @kennyweiss @Arlie-Capps thanks for asking about this, which probably means it needs clearer explanation. Note that the comment for the enum value has changed.

Does the following make sense?

With the changes I propose in this PR, when Axom *is not* configured with Umpire, `MemorySpace::Host` means CPU (i.e., host) memory backed by Axom malloc. Specifically,
- `MemorySpace::Host` is always an valid enum value regardless of whether Umpire is available.
- `isMemorySpaceAvailable(MemorySpace::Host)` returns true with or without Umpire
- `getAllocatorIDFromMemorySpace(MemorySpace::Host)` calls `getDefaultHostAllocatorID()`
- The fault host allocator is `MALLOC_ALLOCATOR_ID` when Umpire is not available (i.e., Host --> Axom malloc)

In any case (with or without Umpire), `MemorySpace::Host` is used to explicitly refer to host memory.

Also, there are places in the implementation where `Host` use is guarded with `AXOM_USE_UMPIRE`. This is done only where the code really(!) depends on Umpire. For example, consider the method `setDefaultAllocator()` at line 263 in `memory_management.cpp`. Calling that method with the arg `MemorySpace::Host` when Umpire is enabled means set global default allocator to Umpire Host, not Axom current default host allocator b/c that could be `MALLOC_ALLOCATOR_ID` (which doesn't make sense for Umpire).

Believe me, this stuff was subtle and somewhat opaque and not consistent everywhere. I am trying to make it clear and consistent. I poured over some of these changes with pen and paper for a while.

src/axom/core/tests/core_array.hpp
rhornung67: It may not be necessary. But I wanted to make sure in each test that `Host` and `Malloc` worked together properly. For this particular test, the test name is `host_space_accepts_malloc_allocator` and the `ArrayView` object constructed from the `Array` object is defined to use `Host`. The "scoped state" thing at the top of the routine resets the default.

rhornung67:
> Thanks for this work, Rich. Approved.
> 
> I feel like I'm blanking on why `MemorySpace::Host` got moved out of the guard on using Umpire. Inside the guard, it makes sense to me: this is the Umpire Host allocator! Outside the guard, I don't see the difference between `MemorySpace::Host` and `MemorySpace::Malloc`.

@Arlie-Capps Thank you for the thorough review. I'm working the review comments and will attempt to address them all.  Regarding the `MemorySpace::Host` change, please see the explanation I added after your original comment above. Hopefully, that helps.

src/axom/core/docs/sphinx/core_memory_management.rst
```diff
@@ -0,0 +1,79 @@
...
+
```
Arlie-Capps, line 46: Perhaps insert some commentary right here on the difference between `MemorySpace::Host` and `MemorySpace::Malloc`.  Maybe something like this, which reflects my understanding (and probably needs further correction and extension):

```
Axom's `MemorySpace` enum has both `Malloc` and `Host` values which are near-synonyms.  
Memory management using `MemorySpace::Malloc` is done with malloc, free, and realloc.  
This requires CPU memory, on the host, not an accelerator.  With `MemorySpace::Host`, if 
Axom is built without the Umpire library, memory management is done with malloc, free, 
and realloc (the same as with `MemorySpace::Malloc`).  If Axom is built with Umpire, use of 
`MemorySpace::Host` will use Umpire's default host allocator: very likely malloc.  But `Host` 
also lets Axom be controlled by Umpire configuration options that change the Umpire host 
allocator's default behavior.

In summary, `Host` is always valid, uses host memory, and causes Axom to use Umpire's 
Host allocator when available.  `Malloc` is always valid, uses host memory, and causes Axom 
to use malloc, free, and realloc no matter what.
```

rhornung67: As you point out, and as I re-read it, I realize that the documentation needs to be more clear.

Regarding your comment here, the real thing to have in mind is the distinction between **host** allocator and **global** allocator.

The default host allocator can only be Axom malloc or Umpire Host allocator:

- The default host allocator is Axom malloc (i.e., getDefaultHostAllocatorID() returns `MemorySpace::Malloc`) in all cases (whether or not Axom is configured with Umpire) unless host allocator is changed
- When Axom is configured with Umpire, one can change the default Axom host allocator to Umpire Host.

** `MemorySpace::Host` is intended to be used when a user wants to be explicit about requesting a host allocation, for example.

The default global allocator is set to a default as follows:
- The default global allocator is Umpire's Host allocator when Axom is configured with Umpire (i.e., getDefaultAllocatorID() returns `MemorySpace::Host`).
- The default global allocator is Axom malloc when Axom is not configured with Umpire (i.e., getDefaultAllocatorID() returns `MemorySpace::Malloc`).
- The default global allocator **cannot** be set to Axom malloc when Axom is configured with Umpire.
- When Axom is configured with Umpire, a user may set the default global allocator to something else (for example, Umpire Pinned memory or Umpire Unified memory).

** `MemorySpace::Dynamic` always refers to the Axom global allocator ** It is used in cases where APIs do not force an allocatorID to be given, such as `Axom::Array`.

src/axom/core/tests/core_memory_management.hpp
```diff
@@ -254,78 +273,214 @@ void check_alloc_realloc_free(bool hostAccessible = true)
...
-    #ifdef UMPIRE_ENABLE_PINNED
-  const int PinnedAllocatorID = axom::getUmpireResourceAllocatorID(umpire::resource::Pinned);
+    #if defined(UMPIRE_ENABLE_PINNED)
+  const int PinnedAllocatorID = axom::getAllocatorIDFromMemorySpace(axom::MemorySpace::Pinned);
   check_alloc_and_free(PinnedAllocatorID, HOST_ACCESSIBLE);
     #endif
 
-    #ifdef UMPIRE_ENABLE_DEVICE
-  const int DeviceAllocatorID = axom::getUmpireResourceAllocatorID(umpire::resource::Device);
+    #if defineed(UMPIRE_ENABLE_DEVICE)
```
BradWhitlock: typo: "defineed"

src/axom/core/memory_management.cpp
BradWhitlock: This might be worth running with some of the GWN examples that make a lot of small primal objects that happen to create axom::Array instances. With Umpire, that caused quite a slowdown. I grepped Axom for `reallocate` and I didn't see much outside objects that should already know their current length such as `axom::Array`. Is it worth even having a `reallocate` function if it requires sizes to be tracked?

src/axom/primal/tests/primal_zip.cpp
```diff
@@ -38,7 +40,7 @@ void check_zip_points_3d()
   using ZipType = primal::ZipIndexable<PrimitiveType>;
 
   const int current_allocator = axom::getDefaultAllocatorID();
-  axom::setDefaultAllocator(axom::execution_space<ExecSpace>::allocatorID());
+  axom::setDefaultAllocator(axom::utilities::globalDefaultAllocatorForExecSpace<ExecSpace>());
```
BradWhitlock, line 43: Does `utilities::globalDefaultAllocatorForExecSpace<ExecSpace>()` mean one could have set a particular allocator for `HIP_EXEC` and then a different one for `OMP_EXEC` for example (assuming Umpire enabled)? That would be nice.

src/axom/quest/tests/quest_signed_distance.cpp
BradWhitlock: I find `axom::execution_space<ExecSpace>::allocatorID()` more clear -- I ask for it when I *really* want memory allocated in a space suitable for `ExecSpace`. If this example is just for `axom::SEQ_EXEC`, fine.

src/axom/core/tests/core_memory_management.hpp
Arlie-Capps: GOOD CATCH

src/axom/quest/tests/quest_signed_distance.cpp
Arlie-Capps: That's a good point, Brad.

src/docs/sphinx/dev_guide/gpu_porting.rst
rhornung67: That was there before. I changed it.

rhornung67: Again, not mine. But, I added a comment.

src/axom/core/tests/core_memory_management.hpp
rhornung67: Fixed!. Thanks @BradWhitlock

src/axom/primal/tests/primal_zip.cpp
rhornung67: The allocator choice is hard-coded in the template specialization for each back-end in the `*_exec.hpp` files in axom/src/axom/core/execution/internal

rhornung67:
Thanks folks! I am closing this and will start a new PR based on helpful discussion on this PR. The new PR will constrain and simplify switching between Axom malloc and Umpire Host for default host allocator. Stay tuned!

src/axom/core/memory_management.cpp
rhornung67: That's the $1M question. We had the reallocate functionality before, but it was brittle in the case when a reallocation was called after switching the host allocator from Axom malloc to Umpire Host. This PR makes it robust at the cost of tracking allocation sizes. I will make a new PR where the switch between Axom malloc and Umpire Host is more constrained and the tracking will not be needed.

PR #1845, opened 2026-04-23
Fix more cuda warnings

### LINKED ISSUE #1790: nvcc warnings
I was able to build the develop branch of Axom successfully with nvcc 12.9.1. However, there are a number of warnings like the following:

```
axom/src/axom/core/../../axom/core/FlatMapView.hpp(49): warning #20012-D: __host__ annotation is ignored on a function("FlatMapView") that is explicitly defaulted on its first declaration
    __attribute__((host)) __attribute__((device)) FlatMapView() = default;
                   ^

Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"
```

I think this can be fixed by changing `AXOM_HOST_DEVICE FlatMapView() = default;` to just `FlatMapView() = default;`. The compiler can still generate both host and device versions of a defaulted constructor.

BradWhitlock:
This PR resolves #1790 by removing additional CUDA warnings.

* The warning mentioned in the issue was already resolved.
* A number of tests caused warnings by enclosing `TEST()` functions in namespaces so this was changed

NOTE: A few "host function called from host/device function" warnings remain, mainly in primal, that would need more changes in Axom to fix properly.

### DIFF
--- a/src/axom/klee/tests/klee_io.cpp
+++ b/src/axom/klee/tests/klee_io.cpp
@@ -17,23 +17,34 @@
 #include <fstream>
 #include <sstream>
 
-namespace axom
-{
-namespace klee
-{
-namespace
-{
+namespace klee = axom::klee;
+namespace inlet = axom::inlet;
+namespace test = axom::klee::test;
+namespace primal = axom::primal;
+
+using klee::CompositeOperator;
+using klee::Dimensions;
+using klee::KleeError;
+using klee::LengthUnit;
+using klee::Rotation;
+using klee::ShapeSet;
+using klee::SliceOperator;
+using klee::TransformableGeometryProperties;
+using klee::Translation;
 using primal::Vector3D;
 using test::AlmostEqVector;
 using ::testing::Contains;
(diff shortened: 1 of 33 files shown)


rhornung67 (approved):

src/axom/klee/tests/klee_io.cpp
```diff
@@ -17,23 +17,34 @@
...
+using klee::TransformableGeometryProperties;
+using klee::Translation;
 using primal::Vector3D;
 using test::AlmostEqVector;
 using ::testing::Contains;
 using ::testing::HasSubstr;
 using ::testing::Truly;
 
+namespace
+{
```
Arlie-Capps, line 41: Maybe put this above (new) line 20, so that all the `namespace` and `using` statements are inside the anonymous namespace, the way you did it in klee_geometry_operators_io.cpp. 

Edit: Never mind, I see this is a test executable.  So I don't think this is important.

Arlie-Capps (approved):
Thanks, Brad.

PR #1847, opened 2026-04-27
Fix broken build on CUDA

BradWhitlock:
This PR addresses a broken Axom build on CUDA from bug report "1846" (not mentioning it so as to prevent closure).

* Fixed CUDA compilation errors in GWN about taking a function's address by using `AXOM_HOST_LAMBDA`. This is ok since these classes use OpenMP for the `axom::for_all`.
* Fixed CUDA compilation error in LinearBVH.


rhornung67:
@BradWhitlock will mentioning the issue in this PR close it when the PR is merged? We need to preserve the warning stuff so we can resolve.

rhornung67 (approved):
Thanks @BradWhitlock This resolves all the build errors. I will update the CUDA warning issue.

white238 (approved):

PR #1850, opened 2026-04-28
Klee scaling enhancements

### LINKED ISSUE #1628: Improve tranformation support for Klee's Geometry operators

BradWhitlock:
This PR resolves #1628.

* Move repeated `AffineMatrixVisitor` class into its own file in klee component
* Enhance transforms.hpp with `scale()` functions that can take a center point
* Enhance klee Scale operator so it can accept center point, if given
* Move repeated code to determine matrix transform for geometry into new `Geometry::getTransform()` method
* Added new tests
* Updated docs and RELEASE-NOTES

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -26,6 +26,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Primal: Adds `NURBSPatch::isTriviallyTrimmed()` to check if the trimming curves for a patch lie on the patch boundaries
 - Quest: Adds support for reading mfem files with variable order NURBS curves (requires mfem>4.9).
 - Quest: Adds OMP support for fast GWN methods for STL/Triangulated STEP input and linearized NURBS Curve input.
+- Klee: Adds an optional "center" parameter in scale operators that permits scaling relative to a custom center point.
 
 ### Removed
 
--- a/src/axom/klee/Geometry.cpp
+++ b/src/axom/klee/Geometry.cpp
@@ -228,5 +228,47 @@ const std::string& Geometry::getBlueprintTopology() const
   return m_topology;
 }
 
+numerics::Matrix<double> Geometry::getTransform() const
+{
+  const auto identity4x4 = numerics::Matrix<double>::identity(4);
+  numerics::Matrix<double> transformation(identity4x4);
+  if(m_operator)
+  {
+    auto composite = std::dynamic_pointer_cast<const CompositeOperator>(m_operator);
+    if(composite)
+    {
+      // Concatenate the transformations
+
+      // Why don't we multiply the matrices in CompositeOperator::addOperator()?
+      // Why keep the matrices factored and multiply them here repeatedly?
+      // Combining them would also avoid this if-else logic.  BTNG
+      for(auto op : composite->getOperators())
+      {
+        // Use visitor pattern to extract the affine matrix from supported operators
+        AffineMatrixVisitor visitor;
+        op->accept(visitor);
+        if(!visitor.isValid())
+        {
+          continue;
+        }
+        const auto& matrix = visitor.getMatrix();
+        numerics::Matrix<double> res(identity4x4);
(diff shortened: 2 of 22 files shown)


RELEASE-NOTES.md
```diff
@@ -26,6 +26,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Primal: Adds `NURBSPatch::isTriviallyTrimmed()` to check if the trimming curves for a patch lie on the patch boundaries
 - Quest: Adds support for reading mfem files with variable order NURBS curves (requires mfem>4.9).
 - Quest: Adds OMP support for fast GWN methods for STL/Triangulated STEP input and linearized NURBS Curve input.
+- Klee: Adds an optional "center" parameter in scale operators that permits scaling relative to a custom center point.
```
kennyweiss, line 29: 👍 - yes!!

src/axom/klee/GeometryOperators.hpp
```diff
@@ -217,6 +217,22 @@ class Scale : public MatrixOperator
...
+   * Create a new Scale operator.
+   *
+   * \param xFactor the amount by which to scale in the x direction
+   * \param yFactor the amount by which to scale in the y direction
+   * \param zFactor the amount by which to scale in the z direction
+   * \param center The center relative to which the scaling is performed.
+   * \param startProperties the initial properties, as in the parent class. 
+   * If the number of dimensions is 2, zFactor should be 1.0, but this is not enforced.
+   */
+  Scale(double xFactor,
```
kennyweiss, line 258: Would it make sense to have an overload without a Z-coordinate?

src/axom/klee/Geometry.cpp
```diff
@@ -228,5 +229,47 @@ const std::string& Geometry::getBlueprintTopology() const
...
+  if(m_operator)
+  {
+    auto composite = std::dynamic_pointer_cast<const CompositeOperator>(m_operator);
+    if(composite)
+    {
+      // Concatenate the transformations
+
+      // Why don't we multiply the matrices in CompositeOperator::addOperator()?
+      // Why keep the matrices factored and multiply them here repeatedly?
+      // Combining them would also avoid this if-else logic.  BTNG
```
kennyweiss, line 245: Seems like a good idea for a future branch!

src/axom/klee/tests/klee_io.cpp
```diff
@@ -349,6 +352,38 @@ TEST(IOTest, readShapeSet_geometryOperators)
...
+      shapes:
+        - name: wheel
+          material: steel
+          geometry:
+            format: test_format
+            path: path/to/file.format
+            units: m
+            operators:
+              - scale: [1.5, 2.5]
+                center: [10, 20]
```
kennyweiss, line 368: Great usability feature!

kennyweiss (approved):
Thanks @BradWhitlock -- looking forward to using this!

src/axom/klee/GeometryOperators.hpp
BradWhitlock: I just added 2 constructor overloads for 2D Scale.

src/axom/klee/Geometry.cpp
BradWhitlock: I think this was Brian Gunney's comment.

kennyweiss: (I saw the B at the start of the initials, but didn't pay attention to the end letters :) )

Arlie-Capps (approved):
Excellent feature.  Thanks, Brad.

PR #1851, opened 2026-04-28
Initial primal slice operator

### LINKED ISSUE #1579: Add a `slice` operator to primal
PR https://github.com/LLNL/axom/pull/1560 has a locally defined operator that slices a `Tetrahedron` with a `Plane`, resulting in a (possibly empty) polygon with 0, 3, or 4 vertices.

We should move this `primal`

See: https://github.com/LLNL/axom/pull/1560#discussion_r2089973499

BradWhitlock:
This PR resolves #1579 by relocating code from a quest test into primal as a slice operator.

* Added segment-segment intersection
* Added tests
* Updated RELEASE-NOTES.md

### DIFF
--- a/src/axom/primal/operators/detail/intersect_impl.hpp
+++ b/src/axom/primal/operators/detail/intersect_impl.hpp
@@ -142,8 +142,14 @@ bool TriangleIntersection2D(const Triangle2& t1, const Triangle2& t2, bool inclu
  * \param[out] intersection The intersection point where the segments intersect.
  *   For collinear overlapping segments, this stores the overlap endpoint with
  *   the smaller parameter on \a P.
- * \param EPS Tolerance used to detect degeneracy, collinearity, and
- *   coincidence of the candidate intersection points.
+ * \param EPS Tolerance used in several geometric comparisons in this routine.
+ *   It is used to:
+ *   - detect whether either segment is degenerate,
+ *   - detect whether the segment directions are parallel,
+ *   - test whether parallel segments are collinear,
+ *   - allow a small tolerance when checking whether collinear overlap exists,
+ *   - decide whether the closest points computed on the two segments should be
+ *     treated as the same intersection point.
  *
  * \return True if the line segments intersect; False otherwise.
  */
--- a/src/axom/primal/operators/detail/intersect_impl.hpp
+++ b/src/axom/primal/operators/detail/intersect_impl.hpp
@@ -134,6 +134,172 @@ bool TriangleIntersection2D(const Triangle2& t1, const Triangle2& t2, bool inclu
 
 //------------------------------ IMPLEMENTATIONS ------------------------------
 
+/*!
+ * \brief Intersect 2 3D line segments and return true if they intersect.
+ *
+ * \param P The first line segment.
+ * \param Q The second line segment.
+ * \param[out] intersection The intersection point where the segments intersect.
+ *
+ * \return True if the line segments intersect; False otherwise.
+ */
+template <typename T>
+AXOM_HOST_DEVICE bool intersect_segment_segment(const Segment<T, 3>& P,
+                                                const Segment<T, 3>& Q,
+                                                Point<T, 3>& intersection,
+                                                const T EPS = static_cast<T>(1e-08))
+{
+  using Vector3D = primal::Vector<T, 3>;
+
+  // Use the standard segment-segment closest-points formulation on the
+  // segment directions u and v. We only report an intersection when the
+  // clamped closest points coincide within the supplied tolerance.
+  const auto u = P.target() - P.source();
+  const auto v = Q.target() - Q.source();
+  const auto w = P.source() - Q.source();
+
+  const T a = u.dot(u);
+  const T b = u.dot(v);
+  const T c = v.dot(v);
+  const T d = u.dot(w);
+  const T e = v.dot(w);
+  const T D = a * c - b * b;
+
+  intersection = Point<T, 3>();
+
+  // Handle degenerate segments first so the general path can assume both
+  // segments have nonzero length.
+  if(axom::utilities::isNearlyEqual(a, T {0}, EPS) &&
+     axom::utilities::isNearlyEqual(c, T {0}, EPS))
+  {
+    intersection = P.source();
+    return P.source().isNearlyEqual(Q.source(), EPS);
+  }
+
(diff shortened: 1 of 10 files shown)


src/axom/primal/operators/intersect.hpp
```diff
@@ -44,6 +44,23 @@ namespace axom
 {
 namespace primal
 {
+
+/*!
+ * \brief Determines if two 3D segments intersect.
+ * \param [in] P A 3D line segment
+ * \param [in] Q A 3D line segment
+ * \param [out] intersection Intersection point of P and Q.
+ * \return true iff P intersects with Q, otherwise, false.
```
kennyweiss, line 57: Please add a note about what happens if P and Q overlap in a line segment (i.e. they are parallel and overlapping).

Please also mention that the `intersection` point should be ignored when `intersect` returns `false`

src/axom/primal/tests/primal_intersect.cpp
```diff
@@ -281,6 +281,42 @@ TEST(primal_intersect, more_ray_segment_intersection)
   }
 }
 
+TEST(primal_intersect, segment_segment_intersection)
+{
+  using Point3D = primal::Point<double, 3>;
+  using Segment3D = primal::Segment<double, 3>;
+
+  Point3D intersection;
+
```
kennyweiss, line 291: Please add a test for the two segments intersecting at:
* an endpoint of segment 1 that coincides with an endpoint of segment 2
* an endpoint of segment 1 that coincides with an internal point of segment 2

And also tests for overlapping segments (once you've decided how to handle this case):
* both segments are the same
* both segments are the same, but the second swaps endpoints
* partial overlap b/w the segments -- e.g.  if points A, B, C and D are collinear, segment 1 is (A,C) and segment 2 is (B,D)

src/axom/primal/tests/primal_slice.cpp
kennyweiss, line 1: Please add tests on the "edge" cases, where the plane intersects the tet on:
* a vertex of the tet
* an edge of the tet
* a face of the tet

(apologies if they're already there -- doesn't look like it, but I could be missing something

src/axom/primal/operators/slice.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+/*!
+ * \brief Slices a 3D tetrahedron with a plane and returns the resulting
+ *        polygon.
+ *
+ * \param [in] tet The tetrahedron to slice
+ * \param [in] plane The slicing plane
+ *
+ * \return The polygon obtained from slicing a tetrahedron with a plane.
+ */
```
kennyweiss, line 41: Please discuss what the result should be for edge cases -- where the plane-tet intersection is along a vertex of edge of the tet.  (i.e. a degenerate polygon).

Please also mention the orientation of the resulting polygon -- presumably, it has the same normal as the plane.

src/axom/primal/operators/detail/intersect_impl.hpp
```diff
@@ -134,6 +134,171 @@ bool TriangleIntersection2D(const Triangle2& t1, const Triangle2& t2, bool inclu
 
 //------------------------------ IMPLEMENTATIONS ------------------------------
 
+/*!
+ * \brief Intersect 2 3D line segments and return true if they intersect.
+ *
+ * \param P The first line segment.
+ * \param Q The second line segment.
+ * \param[out] intersection The intersection point where the segments intersect.
```
kennyweiss, line 142: Please mention how `EPS` is used

src/axom/primal/operators/detail/slice_impl.hpp
```diff
@@ -0,0 +1,65 @@
...
+    Segment<T, 3> seg1(intersectionPolygon[0], intersectionPolygon[1]);
+    Segment<T, 3> seg2(intersectionPolygon[2], intersectionPolygon[3]);
+    Point<T, 3> sp;
+    if(!primal::intersect(seg1, seg2, sp))
+    {
+      axom::utilities::swap(intersectionPolygon[2], intersectionPolygon[3]);
+    }
+  }
+  return intersectionPolygon;
+}
```
kennyweiss, line 132: Should we also ensure that the orientation of the Polygon matches that of the Plane?

kennyweiss (changes requested):
Thanks @BradWhitlock -- I asked for a bunch of edge-case tests and checks.
Probably the most important is that the orientation of the polygon should match the plane, since, presumably, that would be needed for applications.

src/axom/primal/operators/intersect.hpp
BradWhitlock: The current interface isn't good for P and Q segments overlapping in that way. Would it make more sense to have the intersection output be another Segment that could represent that and in the general cases where P and Q intersect at a point, just make the output segment points the same?

kennyweiss: Seems like this case does not come up in the plane-tet slicing.

I guess it depends on what we're using it for. Could be that `intersection` is an arbitrary point in the intersection when they overlap (?).

I think it's probably fine as long as you document the expected behavior.

BradWhitlock: I kept the interface the same and documented what happens if P,Q overlap as a line segment.

src/axom/primal/tests/primal_intersect.cpp
BradWhitlock: I had codex increase the test coverage for these intersections.

src/axom/primal/tests/primal_slice.cpp
BradWhitlock: Added tests for these edge cases.

src/axom/primal/operators/slice.hpp
BradWhitlock: The comments were improved.

src/axom/primal/operators/detail/intersect_impl.hpp
BradWhitlock: Done

src/axom/primal/operators/detail/slice_impl.hpp
BradWhitlock: Yes. Done.

Arlie-Capps (approved):
Thank you, Brad!

bmhan12 (approved):

kennyweiss (approved):
Thanks for the improvements @BradWhitlock !

PR #1854, opened 2026-05-04
python-ize convert_sidre_protocol

bmhan12:
This PR:

- Adds a python version of the `convert_sidre_protocol.cpp` script, and added nanobind bindings to `IOManager` to facilitate this.
  - Updates host-configs accordingly. 
- Adds `mpi4py` as a dependency when building `+python`.
  -  This dependency is also necessary for the current python quest interface through shroud. This allowed me to compile the python interface (was failing), but am segfaulting when trying to run the test the python interface through `quest_test.py`...


src/axom/quest/interface/python/pyQUESTmodule.cpp
```diff
@@ -863,9 +863,9 @@ initquest(void)
   struct module_state *st = GETSTATE(m);
 
   // enum axom::quest::SignedDistExec
-  PyModule_AddIntConstant(m, "CPU", axom::quest::CPU);
-  PyModule_AddIntConstant(m, "OpenMP", axom::quest::OpenMP);
-  PyModule_AddIntConstant(m, "GPU", axom::quest::GPU);
+  PyModule_AddIntConstant(m, "CPU", static_cast<int>(axom::quest::SignedDistExec::CPU));
+  PyModule_AddIntConstant(m, "OpenMP", static_cast<int>(axom::quest::SignedDistExec::OpenMP));
+  PyModule_AddIntConstant(m, "GPU", static_cast<int>(axom::quest::SignedDistExec::GPU));
```
bmhan12, line 868: I tried compiling and running the shroud python quest interface to see if it still worked (was only able to get it compiling, segfaults when ran).
These above changes are different from what shroud generates, but necessary to get the code to compile. 

The shroud deck does look to have the right namespace, but it gets lost somehow in the translation/generation: https://github.com/llnl/axom/blob/21504d400d3bea543da4d4a180345a2dc3a2ce65/src/axom/quest/interface/quest_shroud.yaml#L73-L80

scripts/spack/configs/toss_4_x86_64_ib/spack.yaml
```diff
@@ -83,6 +87,8 @@ spack:
             c: /usr/tce/packages/clang/clang-19.1.3-magic/bin/clang
             cxx: /usr/tce/packages/clang/clang-19.1.3-magic/bin/clang++
           flags:
+            cflags: -fPIC
```
kennyweiss, line 90: Were the `-fPIC` flags necessary for the branch, or an unrelated improvement?
(Not a problem either way, just curious)

scripts/spack/packages/axom/package.py
```diff
@@ -345,6 +346,9 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
 
     conflicts("^blt@:0.3.6", when="+rocm")
 
+    # python interface requires mpi
+    conflicts("~mpi", when="+python")
```
kennyweiss: This doesn't seem right to me -- I think our `python` bindings should only require `mpi` when Axom is configured with `mpi`.

We should have some other guard on our python scripts that require mpi -- e.g. only use MPI when it's available.

src/axom/quest/interface/python/pyQUESTmodule.cpp
kennyweiss: Thanks for checking.

I think the shroud-based python bindings require a shared library build.

We should remove them in favor of nanobind python bindings in a follow-up PR.

src/cmake/thirdparty/SetupAxomThirdParty.cmake
```diff
@@ -353,13 +353,14 @@ if((NOT PY_ENV_IMPORT_CODE EQUAL 0)
...
-   OR NOT PY_INICONFIG_DIR))
+   OR NOT PY_INICONFIG_DIR
+   OR NOT PY_MPI4PY_DIR))
     message(FATAL_ERROR
       "Axom's python extensions require nanobind, numpy, pytest, and conduit."
       "\nThe python library installation paths "
       "(and pytest's dependencies pluggy and iniconfig) "
       "can be specified with CMake variables: "
-      "PY_NANOBIND_DIR, CONDUIT_PYTHON_MODULE_DIR, PY_NUMPY_DIR, PY_PYTEST_DIR, PY_PLUGGY_DIR, PY_INICONFIG_DIR ")
+      "PY_NANOBIND_DIR, CONDUIT_PYTHON_MODULE_DIR, PY_NUMPY_DIR, PY_PYTEST_DIR, PY_PLUGGY_DIR, PY_INICONFIG_DIR, PY_MPI4PY_DIR ")
```
kennyweiss: MPI4PY is required by this check, but the error message does not indicate that it is required.

src/tools/convert_sidre_protocol.py
```diff
@@ -0,0 +1,257 @@
...
+    while pysidre.indexIsValid(idx):
+        yield group.getView(idx)
+        idx = group.getNextValidViewIndex(idx)
+
+
+def iter_groups(group: pysidre.Group):
+    idx = group.getFirstValidGroupIndex()
+    while pysidre.indexIsValid(idx):
+        yield group.getGroup(idx)
+        idx = group.getNextValidGroupIndex(idx)
```
kennyweiss: It's nice to see that we can get iteration support this way.

Is there a way to directly bind the C++ group/view iterators from Sidre so user applications don't need to do this themselves?

src/tools/convert_sidre_protocol.py
```diff
@@ -0,0 +1,257 @@
...
+    root = datastore.getRoot()
+
+    print(f"Loading datastore from {input_path}")
+    manager.read(root, str(input_path))
+    num_files = manager.getNumFilesFromRoot(str(input_path))
+
+    print("Loading external data from datastore")
+    external_holders: list[np.ndarray] = []
+    allocate_external_data(root, external_holders, args.verbose)
+    manager.loadExternalData(root, str(input_path))
```
kennyweiss, line 252: Does our new test exercise the external data option? If not, can we add a test that does? (could be in a follow-up PR).

src/tools/convert_sidre_protocol.py
```diff
@@ -0,0 +1,257 @@
...
+        MPI.Init()
+        initialized_mpi = True
+
+    input_path = Path(args.input)
+    manager = pysidre.IOManager()
+    datastore = pysidre.DataStore()
+    root = datastore.getRoot()
+
+    print(f"Loading datastore from {input_path}")
+    manager.read(root, str(input_path))
```
kennyweiss, line 247: Do we need a check that the current number of ranks matches the number of ranks in the file?
Or, ideally, is there a way to run this application w/ a different number of ranks than the original?

At the very least, it would be good to print the number of ranks in the file and in the calling application.

Tag: @nselliott

src/tools/convert_sidre_protocol.py
kennyweiss, line 1: This is really nice!

kennyweiss (changes requested):
Thanks @bmhan12 -- this is a really nice addition!

My only concern is that this seems to *require* MPI for our python bindings (but correct me if I am wrong). I think it MPI should be optional -- and should be enabled when the Axom config has MPI.

scripts/spack/configs/toss_4_x86_64_ib/spack.yaml
bmhan12: > Were the -fPIC flags necessary for the branch

Yes, I saw an error while linking otherwise related to mpi (`libdtcmp`).

scripts/spack/packages/axom/package.py
bmhan12: Fixed! Added guards to make mpi optional instead of a hard requirement for python.

src/cmake/thirdparty/SetupAxomThirdParty.cmake
bmhan12: Fixed - I made the `mpi4py` check separate and conditional on if Axom is configured with mpi.

src/tools/convert_sidre_protocol.py
bmhan12: Fixed - woops, this was already possible with `.groups()` and `.views()` (codex didn't find it the first time).

src/tools/convert_sidre_protocol.py
bmhan12: I added a check for the number of ranks + warning message!

It looks like a different number of ranks is supported? : https://github.com/llnl/axom/blob/61efb829e23256e4a002fc561fc5dd0fca0d60ae/src/axom/sidre/spio/IOManager.cpp#L835-L891

At least for the added unit test (and compared against the `cpp` implementation) , I was able to read 3 data files from one rank.

kennyweiss: Awesome -- I think that was added to spio after the `convert_sidre_protocol` tool was developed and we never incorporated it

kennyweiss: It would be nice to add this to the python implementation, i.e. running the python tool with `n` ranks on a problem that was originally created with `m` ranks.  

We should still log the number of input domains and the number of ranks in the calling app.

kennyweiss (approved):
Thanks for the updates @bmhan12

src/tools/convert_sidre_protocol.py
bmhan12: Sort of - the `box_2D_r3.root` data file has an external Group, but it's empty.
I can add a test in a follow-up PR - it doesn't look like any of our `axom/data` files specify external data (please correct me if that is not the case).

rhornung67 (approved):
Thank you @bmhan12

PR #1857, opened 2026-05-05
Enhance SamplingShaper with support for uniform point sampling

### LINKED ISSUE #1693: Add support for uniform grid sampling in sampling shaper
The query points in our sampling-based workflow are currently defined by Gauss-Legendre quadrature node locations within each cell.

A user has requested that we add support for sampling on a uniform grid within each cell.
Their use-case might also require different numbers of samples in each dimension within hexahedral cells.

BradWhitlock:
This PR resolves #1693 by enhancing the `SamplingShaper` so it can compute sample points at uniform grid locations within elements.

* Enhanced `SamplingShaper` with `setQuadratureType()` method that accepts quadrature enum values defined in `mfem::Quadrature1D`. Pass mfem's "Invalid" value to select the default quadrature behavior (GaussLegendre). For uniform spacing that spans the whole element, ClosedUniform is a good value. There are 7 enum values that determine valid quadrature point locations.
* Enhanced `SamplingShaper` with `setSamplingResolution()` methods that can take a single value or up to 3 values, allowing a different number of samples in each logical dimension.
* Added functionality to _shaping_helpers.cpp_ to compute the "custom" quadratures and QuadratureSpaces.
* Enhanced pathways around sample point generation, initial volume fraction copying, and final volume fraction generation so they can handle the quadrature with varying samples in each dimension (some MFEM functionality we were using does not yet support it - submitted an [issue](https://github.com/mfem/mfem/issues/5321) )
* Enhanced _shaping_driver_ so it supports these features. Quadrature types can be selected by name and the sampling resolution can be set for each dimension.
* Enhanced _shaping_driver_ so in --verbose mode it writes the quadrature sample points to a Blueprint file so we can see the sample point locations in VisIt
* Enhanced test cases so we test this functionality in 2D/3D
* Converted some `SLIC_ASSERT_MSG` to `SLIC_ERROR_IF` so checks are done in Release builds of the shaping_driver
* Fixed a few weird issues related to sampling, etc.
* Enhanced some doxygen comments and tutorial materials to mention this functionality
* Updated RELEASE-NOTES.md

Examples:
image
image


src/axom/quest/detail/shaping/shaping_helpers.cpp
```diff
@@ -25,6 +27,43 @@ namespace shaping
...
+bool usesAnisotropicCustomTensorQuadrature(const mfem::Mesh& mesh,
+                                           const int sampleResolution[3],
+                                           int quadratureType)
+{
+  if(quadratureType == static_cast<int>(mfem::Quadrature1D::Invalid))
+  {
+    return false;
+  }
+
+  switch(mesh.GetTypicalElementGeometry())
```
kennyweiss: We might need to be careful with `GetTypicalElementGeometry` on a mesh with 0 elements. 
I think I had to create a workaround for that in the past.  I'll post a link if I can find it.

src/axom/quest/detail/shaping/shaping_helpers.cpp
```diff
@@ -25,6 +27,43 @@ namespace shaping
...
+    : mfem::QuadratureSpace(mesh, *ir)
+    , m_ir(std::move(ir))
+  { }
+
+private:
+  std::unique_ptr<mfem::IntegrationRule> m_ir;
+};
+
+bool usesAnisotropicCustomTensorQuadrature(const mfem::Mesh& mesh,
+                                           const int sampleResolution[3],
```
kennyweiss: Should this be an ArrayView instead of a C-array?

src/axom/quest/detail/shaping/shaping_helpers.cpp
```diff
@@ -135,62 +174,164 @@ void copyShapeIntoMaterial(const mfem::QuadratureFunction* shapeQFunc,
...
 
   // convert requested samples into a compatible polynomial order
   // that will use that many samples: 2n-1 and 2n-2 will work
   // NOTE: Might be different for simplices
   const int sampleOrder = 2 * sampleRes - 1;
-  mfem::QuadratureSpace* sp = new mfem::QuadratureSpace(mesh, sampleOrder);
+  return new mfem::QuadratureSpace(mesh, sampleOrder);
+}
+
+mfem::QuadratureSpace* makeCustomQuadratureSpace(mfem::Mesh* mesh, int sampleRes[3], int quadratureType)
```
kennyweiss: Minor: 
Same question -- should the interface take an `ArrayView` instead of a C-array?  
If so, we wouldn't need to bake `3` into the signature.

src/axom/quest/detail/shaping/shaping_helpers.cpp
```diff
@@ -135,62 +174,164 @@ void copyShapeIntoMaterial(const mfem::QuadratureFunction* shapeQFunc,
...
-  // Delete the geometric factors associated w/ our custom quadrature rule
-  mesh->DeleteGeometricFactors();
+      for(int j = 0; j < dim; ++j)
+      {
+        for(int k = 0; k < nq; ++k)
+        {
+          pos(j, k, i) = pointMat(j, k);
+        }
+      }
+    }
```
kennyweiss, line 345: Not sure how much it matters, but is there a reason not to use the `geomFactors` construct in the non-anisotropic as well?

src/axom/quest/examples/shaping_driver.cpp
```diff
@@ -300,11 +311,11 @@ struct Input
...
         ->check(axom::CLI::NonNegativeNumber);
 
-      sampling_options->add_option("-q,--quadrature-order", quadratureOrder)
+      sampling_options->add_option("--sampling-resolution", samplingResolution)
         ->description(
-          "Quadrature order for sampling the inout field. \n"
+          "Sampling resolution per element for the inout field (x,y,[z]). \n"
           "Determines number of samples per element in determining volume fraction field")
-        ->capture_default_str()
+        ->expected(2, 3)
```
kennyweiss: Should we allow this to be a single value that gets replicated in the isotropic case?
E.g.
```suggestion
        ->expected(1, 3)
```

src/axom/quest/examples/shaping_driver.cpp
```diff
@@ -460,6 +486,42 @@ void finalizeLogger()
   }
 }
 
+//------------------------------------------------------------------------------
+/// Write the quadrature points as a Blueprint mesh.
```
kennyweiss, line 490: Instead of writing out a point mesh, can this use the new QuadratureFunction support in VisIt@3.5 ?

src/axom/quest/tests/quest_sampling_shaper.cpp
```diff
@@ -2117,6 +2307,293 @@ piece = line(end=start)
...
+  int sampleRes[3] = {3, 5, 1};
+
+  for(const auto& quadrature : supported_quadrature_types)
+  {
+    this->validateShapeFile(shape_file.getPath());
+    this->initializeShaping(shape_file.getPath());
+
+    this->m_shaper->setSamplingResolution(sampleRes);
+    this->m_shaper->setQuadratureType(quadrature.second);
+    this->m_shaper->setVolumeFractionOrder(0);
```
kennyweiss: Please document the API changes in the `RELEASE_NOTES`

scripts/github-actions/linux-build_and_test.sh
```diff
@@ -45,7 +45,8 @@ if [[ "$DO_BUILD" == "yes" ]] ; then
     fi
 
     echo "~~~~~~ RUNNING TESTS ~~~~~~~~"
-    make CTEST_OUTPUT_ON_FAILURE=1 test ARGS='-T Test --output-on-failure -j$NUM_BUILD_PROCS'
+    #make CTEST_OUTPUT_ON_FAILURE=1 test ARGS='-T Test --output-on-failure -j$NUM_BUILD_PROCS'
+    make CTEST_OUTPUT_ON_FAILURE=1 test ARGS='-T Test --output-on-failure'
```
kennyweiss: Is this a debugging change, or are you planning to commit this?

src/axom/quest/SamplingShaper.hpp
```diff
@@ -159,7 +159,77 @@ class SamplingShaper : public Shaper
...
+    SLIC_ASSERT(sampleRes[0] > 0);
+    SLIC_ASSERT(sampleRes[1] > 0);
+    SLIC_ASSERT(sampleRes[2] > 0);
+    m_sampleResolution[0] = sampleRes[0];
+    m_sampleResolution[1] = sampleRes[1];
+    m_sampleResolution[2] = sampleRes[2];
+  }
+
+  // Deprecated backward compatibility method
+  [[deprecated]] void setQuadratureOrder(int order) { setSamplingResolution(order); }
```
kennyweiss, line 246: Please note the deprecated function in the `RELEASE_NOTES`

src/axom/quest/SamplingShaper.hpp
```diff
@@ -833,17 +939,39 @@ class SamplingShaper : public Shaper
...
+                                                elemMat);
+          for(int j = 0; j < dofs; ++j)
+          {
+            for(int i = 0; i < dofs; ++i)
+            {
+              (*mass_mat)(i, j, elem) = elemMat(i, j);
+            }
+          }
+        }
+      }
```
kennyweiss: It looks like there will be performance implications for using the anisotropic quadrature vs. the isotropic?  
The latter runs on the device, but the former is on the host.

kennyweiss (approved):
Thanks @BradWhitlock -- this is really nice!

Can you please make sure there are no significant performance implications for device-based runs?  We don't yet have the sampling on the device, but the assembly and FCT can be expensive and are currently on the device.

src/axom/quest/examples/shaping_driver.cpp
BradWhitlock: I don't think that lets me easily plot the quadrature positions as dots.

src/axom/quest/examples/shaping_driver.cpp
BradWhitlock: Done.

src/axom/quest/detail/shaping/shaping_helpers.cpp
BradWhitlock: Changed this function and some others to use ArrayView<int>.

src/axom/quest/detail/shaping/shaping_helpers.cpp
BradWhitlock: Changed it.

src/axom/quest/detail/shaping/shaping_helpers.cpp
BradWhitlock: `geomFactors` is used for the isotropic and default quadrature case. It could not be used for anisotropic because MFEM was returning incorrect values (I submitted an MFEM issue).

BradWhitlock:
I did some timings with isotropic sampling 4x4x4 and anisotropic sampling 4x8x2, which should be about the same work.

batch_lu_factor
batch_lu_factor

mass_integrator_assemble - has anisotropic path, not too much performance difference in these runs
mass_integrator_assemble

batch_lu_solve - This shows some host/device differences but not too bad. Actually the 4x8x2 anisotropic case is faster in this run.
batch_lu_solve

fct_project - This is actually serial and does not show much difference.
fct_project

src/axom/quest/SamplingShaper.hpp
BradWhitlock: I added some plots in the comments further down. The timings are not too different for most phases of the algorithm in these runs.

src/axom/quest/SamplingShaper.hpp
BradWhitlock: Done

scripts/github-actions/linux-build_and_test.sh
BradWhitlock: I reverted it.

src/axom/quest/tests/quest_sampling_shaper.cpp
BradWhitlock: Done

bmhan12 (approved):

PR #1861, opened 2026-05-09
Improves performance and robustness of Delaunay triangulation and ScatteredInterpolations

kennyweiss:
# Summary
- This PR includes robustness and performance improvements for the Delaunay triangulation and scattered interpolation queries
  - Apologies for the large PR. There wasn't a great way to separate the robustness improvements from the performance improvements

- There were a few robustness issues previously
  - Our `in_sphere` and `orientation` predicates could not handle points that were very nearly on the boundary/plane
  - This caused point insertion to fail in some common cases, especially when the input points were on a regular grid
  - This PR fixes the predicates, but also exposes the underlying `orientation_determinant` and `in_sphere_determinant` results (in the axom::primal::detail` namespace) for cases where the caller needs custom handling for on/near boundary cases (or the raw determinant)
- The new functionality caused the main `Delaunay.hpp` file to grow too large, so I factored out common functionality into helper classes/files: `DelaunayElementFinder`, `DelauanayInsertionHelper`, `DelaunayPointLocation`, `DelaunayValidation`, `DelaunayImpl`
- This PR also improves the performance of the incremental 2D and 3D Delaunay construction (see table below)
  - Some of the performance improvements were in the undelying slam classes, especially in the triangle/tet mesh topological data structure class -- `mesh_struct/IA.hpp`, and in eliding some virtual function calls in the slam `Dynamic{Set,Relation,Map}` classes

## Performance improvements

There are two examples that exercise Delaunay triangulations. 
* `quest_delaunay_triangulation_ex` builds a 2D or 3D Delaunay complex and then validates it.
   * This application does not sort the input points, it just uses the points that were provided
   * This application also does not test input points on a regular grid
* `quest_scattered_interpolation_ex` builds a Delaunay complex, add scalar fields and then for a given set of query points, looks up the simplex containing the point and interpolates fields using the scalar values at the vertices of the simplex
   * This application sorts the input points using BRIO (biased randomized insertion order) to prevent worst case insertion performance.
   * I have added the ability to this application to generate a Delaunay complex from input points along a regular grid in 2D or 3D
* For the analysis, I will highlight the insertion rates -- the number of points we can insert per second in 2D and 3D

### quest_delaunay_triangulation_ex

| Build   | Dim | N         | Build sec (med) | Insertion rate pts/s (med) | Validation sec (med) | Wall sec (med) | Max RSS kB (med) |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| branch  | 2   | 1,000,000 | 2.998           | 333,523.5                  | 12.610               | 16.550         | 561,052          |
| branch  | 3   | 1,000,000 | 14.608          | 68,457.2                   | 103.862              | 121.000        | 2,251,292        |
| develop | 2   | 1,000,000 | 5.837           | 171,308.6                  | 0.359                | 7.040          | 422,836          |
| develop | 3   | 1,000,000 | 22.451          | 44,540.5                   | 30.896               | 56.030         | 1,341,032        |

Analysis for unsorted insertion:
* In 2D, our insertion rate is about twice as fast (333K pts/sec). In 3D, our insertion rate is about 50% faster (68K pts/sec)
* The validation is considerably more expensive, but that's because we're doing a lot more validation than previously
* The new code path appears to be using more memory. This is due to the validation structures, and won't show up for users. (It doesn't show up for the scattered_interpolation, which in fact is using less memory in the new branch).

### quest_scattered_interpolation_ex

| Build   | Dim | Input            | N         |  Build sec (med) | Insertion rate pts/s (med)  | Max RSS kB (med) |
| ---       | ---:  | ---:                | ---:      | ---:                      | ---:                                     | ---: |  
| branch  | 2   | grid 1000x1000   | 1,000,000 | 1.179           | 848,532.4              | 259,664          |
| branch  | 2   | random           | 1,000,000 | 1.166           | 857,268.4                  | 259,664          |
| branch  | 3   | grid 100x100x100 | 1,000,000 | 7.602           | 131,547.9            | 628,136          |
| branch  | 3   | random           | 1,000,000  | 6.129           | 163,166.3                 | 692,024          |
| develop | 2   | random           | 1,000,000 | 2.557           | 391,093.0                 | 327,588          |
| develop | 3   | random           | 1,000,000  | 14.524          | 68,852.1                 | 793,804          |

Analysis for sorted insertion (with BRIO):
* In 2D, our insertion rate is more than twice as fast (850K pts/sec). In 3D, our insertion rate is about 2-2.5x faster (130-160K pts/sec)
* Performance for creating Delaunay complexes from regular grid points is roughly the same as random points (give or take)
* Our previous performance chart for the scattered interpolation query was in https://github.com/llnl/axom/pull/859, which showed insertion rates of about 270K pts/sec in 2D and 50K pts/sec in 3D for the sorted (BRIO) insertion, so our starting baselines in develop were already faster.  (Those numbers might have been on a different machine, i.e. LLNL's CTS-1 vs. CTS-2 architectures)


src/axom/quest/detail/DelaunayInsertionHelper.hpp
```diff
@@ -0,0 +1,285 @@
...
+ * be cleared without reallocation on every point.
+ *
+ * Public members track insertion state for validation:
+ * - cavity_elems: Elements whose circumspheres contain the new point
+ * - boundary_facets: Faces between cavity and non-cavity elements
+ * - inserted_elems: New simplices created by connecting new point to boundary
+ * - containing_element, containing_bary, seed_elements_debug: For diagnostics
+ */
+template <int DIM, typename PointType, typename BaryCoordType, typename IndexType, typename IndexArray, typename IAMeshType>
+class DelaunayInsertionHelper
```
bmhan12, line 59: Should there be documentation for these individual functions?
Also for the template parameters here?

EDIT: Same comment for the rest of the header files in `quest/detail`

src/axom/quest/benchmarks/quest_delaunay_circumsphere.cpp
```diff
@@ -0,0 +1,622 @@
...
+
+double unitInterval(std::uint64_t bits)
+{
+  return static_cast<double>(bits & 0xFFFFFFFFu) / static_cast<double>(0x100000000ULL);
+}
+
+std::uint64_t stepLcg(std::uint64_t state)
+{
+  return state * 6364136223846793005ULL + 1442695040888963407ULL;
+}
```
bmhan12, line 72: Is this doing random number generation?
Any particular reason why these values?

src/axom/quest/detail/DelaunayImpl.hpp
```diff
@@ -0,0 +1,430 @@
...
+  IndexType element_i = findContainingElement(new_pt);
+
+  if(element_i == INVALID_INDEX)
+  {
+    SLIC_WARNING(
+      fmt::format("Could not insert point {} into Delaunay triangulation: "
+                  "Element containing that point was not found",
+                  new_pt));
+    return;
+  }
```
bmhan12, line 355: Should be documented operation is a no-op for invalid element/point.

src/axom/quest/detail/DelaunayValidation.hpp
```diff
@@ -0,0 +1,924 @@
...
+  const auto min_pt = m_bounding_box.getMin();
+  const auto max_pt = m_bounding_box.getMax();
+
+  double max_extent = 1.;
+  for(int dim = 0; dim < DIM; ++dim)
+  {
+    max_extent = axom::utilities::max(max_extent, max_pt[dim] - min_pt[dim]);
+  }
+
+  return 64. * std::numeric_limits<double>::epsilon() * max_extent;
```
bmhan12: why/what is 64 here?

src/axom/quest/detail/DelaunayValidation.hpp
```diff
@@ -0,0 +1,924 @@
...
+  {
+    return getElement(element_idx).signedVolume();
+  }
+}
+
+template <int DIM>
+inline double Delaunay<DIM>::getElementMeasureTolerance() const
+{
+  const double scale = axom::utilities::max(1., m_bounding_box.range().norm());
+  return 256. * std::numeric_limits<double>::epsilon() * std::pow(scale, static_cast<double>(DIM));
```
bmhan12: ditto same comment as 64.

src/axom/quest/examples/scattered_interpolation.cpp
```diff
@@ -546,23 +588,85 @@ axom::Array<primal::Point<double, DIM>> generatePts(int numPts,
...
-  using axom::utilities::random_real;
-
   using PointType = typename primal::Point<double, DIM>;
   using BoundingBox = typename primal::BoundingBox<double, DIM>;
 
   axom::Array<PointType> pts(numPts, numPts);
 
   BoundingBox bbox {PointType(bb_min.data()), PointType(bb_max.data())};
 
+  std::mt19937_64 mt;
```
bmhan12, line 601: Probably want to keep using `random_real` here?

src/axom/quest/tests/quest_delaunay.cpp
```diff
@@ -0,0 +1,588 @@
...
+                                 PointType {0.2, 0.8},
+                                 PointType {0.7, 0.7}};
+
+  DelaunayType<2> original;
+  original.initializeBoundary(BoundingBox(PointType {0., 0.}, PointType {1., 1.}));
+  insertPoints(original, std::vector<PointType>(points.begin(), points.end() - 1));
+
+  DelaunayType<2> moved(std::move(original));
+  moved.insertPoint(points.back());
+  expectValidDelaunay(moved, points);
```
bmhan12, line 302: Should this `move` test check the original Delaunay post-move?

src/axom/quest/tests/quest_delaunay.cpp
```diff
@@ -0,0 +1,588 @@
...
+{
+  EXPECT_TRUE(dt.getMeshData()->isValid(true));
+  EXPECT_TRUE(dt.getMeshData()->isConforming(true));
+  EXPECT_TRUE(dt.isConforming(true));
+  EXPECT_TRUE(dt.isValid(true));
+}
+
+}  // namespace
+
+TEST(quest_delaunay, local_bad_geometry_pool_2d)
```
bmhan12: What makes this test and the one below "bad"?

src/axom/quest/Delaunay.hpp
```diff
@@ -171,710 +393,258 @@ class Delaunay
...
+  static int signWithTolerance(double value, double tolerance)
+  {
+    return value > tolerance ? 1 : (value < -tolerance ? -1 : 0);
+  }
 
-        // check insphere condition
-        if(circumspheres[element_idx].getOrientation(vertex) == primal::ON_NEGATIVE_SIDE)
-        {
-          valid = false;
+  static CircumsphereEval evaluateCircumsphereOnMesh(const IAMeshType& mesh, IndexType element_idx);
```
bmhan12, line 544: Should these public member functions be documented?

src/axom/quest/ScatteredInterpolation.hpp
```diff
@@ -288,20 +315,64 @@ class ScatteredInterpolation
...
+    //
+    // Implementation note:
+    // The original implementation used repeated calls to `random_real()` to
+    // simulate coin flips. At large N, this becomes costly (millions of calls
+    // into `std::uniform_real_distribution`). We can generate the same level
+    // distribution using a single 64-bit random integer per point:
+    // - Take the top `nlevels` bits.
+    // - The BRIO level is the position of the highest set bit (1..nlevels), or 0 if all are 0.
+    std::mt19937_64 mt;
+    std::uint64_t seed = 0;
```
bmhan12, line 347: Just double-checking this is preferred over `random_real`, and if so, is this something that would be useful/generalizable to be part of `axom::utilities`?

src/axom/quest/tests/quest_delaunay.cpp
kennyweiss: Good call -- it was a regression test for a previous bug. I'll rename it to regression

src/axom/quest/detail/DelaunayInsertionHelper.hpp
kennyweiss: Definitely.  I'll update it.

src/axom/quest/benchmarks/quest_delaunay_circumsphere.cpp
kennyweiss: Thanks. I'll add a comment.

That is a linear congruential generator (lcg) step:
> $x_{n+1}=a \cdot x_{n}+c(mod\ 2^{64})$

which cycles through all 64-bit states when the increment $c$ is odd and the multiplier $c = 1(mod\ 4)$.

src/axom/slam/tests/slam_detail_FacetPairingMap.cpp
```diff
@@ -0,0 +1,570 @@
...
+  KeyType key1;
+  EXPECT_EQ(key1.key0, KeyType::INVALID);
+  EXPECT_EQ(key1.key1, KeyType::INVALID);
+
+  // Construction with single key (2D)
+  KeyType key2(42);
+  EXPECT_EQ(key2.key0, 42);
+  EXPECT_EQ(key2.key1, KeyType::INVALID);
+
+  // key1 parameter should be ignored in 2D
```
Arlie-Capps: Minor typo: Here and at line 37, do you mean `key3` ?

Arlie-Capps: woops, got confused by naming.  (just re-read FacetPairingMap.hpp lines 80 and 91)

kennyweiss: @Arlie-Capps -- I renamed the Key and Value types for the `FacetPairingMap` and some internal variables to better convey the intent.

That data structure is used to repair the connectivity for the adjacent faces of the simplices (triangles in 2D, tetrahedra in 3D) in the *star* of the inserted vertex -- i.e. the simplices that are incident in the inserted vertex.  
* The keys for this map correspond to the facets of the *link* of the vertex (the boundary faces of the star), which are points in 2D and edges in 3D.
* The values for this map are the simplices incident in the Key and the local index of the facet that is incident in the key

src/axom/quest/detail/DelaunayInsertionHelper.hpp
kennyweiss: Done.

src/axom/quest/benchmarks/quest_delaunay_circumsphere.cpp
kennyweiss: Documented.

src/axom/quest/detail/DelaunayImpl.hpp
kennyweiss: Done.

src/axom/quest/detail/DelaunayValidation.hpp
kennyweiss: I documented all the scaling factors, and why they are necessary. Documentation is in the Delaunay implementation and in the predicate files (orientation and in_sphere).

src/axom/quest/examples/scattered_interpolation.cpp
kennyweiss: I documented why we're using a custom RNG for the scattered interpolation

src/axom/quest/detail/DelaunayValidation.hpp
kennyweiss: I documented why the scaling factor is necessary and why it's 64

src/axom/quest/tests/quest_delaunay.cpp
kennyweiss: Thanks!!! I added a test for this and it caught a bug in `axom::Array` -- a moved-from Array should be valid, but we were setting the allocator to `INVALID_ALLOCATOR_ID`, which made the Array invalid. 

I fixed it by removing the assignment to `INVALID_ALLOCATOR_ID` and added tests there and here.

src/axom/quest/tests/quest_delaunay.cpp
kennyweiss: I renamed the test and added an explanation of the regression that we're protecting against.

src/axom/quest/ScatteredInterpolation.hpp
kennyweiss: Might be useful as a general utility, but I'm holding off on it for now since it's a bit custom and this is already a huge PR. Let's reconsider if the need arises in the future.

src/axom/quest/Delaunay.hpp
kennyweiss: Documented.

kennyweiss:
Thanks for the great comments/suggestions about documentation for this PR @bmhan12 and @Arlie-Capps (offline) -- I have added/improved the documentation significantly.

bmhan12 (approved):

src/axom/core/Array.hpp
```diff
@@ -136,6 +136,11 @@ struct DefaultStoragePolicy
  *  its memory at allocation time and we use axom's memory_management
  *  and allocator ID abstractions rather than std::allocator.
  *
+ *  Move semantics follow standard container conventions: moved-from Arrays are
+ *  left in a valid-but-unspecified state and may be safely reused
+ *  (e.g. assigned to, cleared, or appended to). 
+  * The allocator id of a moved-from Array remains valid.
```
Arlie-Capps, line 142: (fix indentation)

src/axom/quest/Delaunay.hpp
```diff
@@ -171,710 +408,331 @@ class Delaunay
...
-        {
-          continue;
-        }
+  /**
+   * \brief Returns a human-readable name for a primal::OrientationResult value
+   *  (for diagnostics and warning messages).
+   * 
+   * \param result An orientation result code (ON_POSITIVE_SIDE/ON_BOUNDARY/ON_NEGATIVE_SIDE)
+   */
+  static const char* orientationResultName(int result);
```
Arlie-Capps, line 591: If possible, please make this a `const` method 
(i.e. `static const char* orientationResultName(int result) const;`).

(If it's not possible, then I'm probably missing something.  Sorry.)

To expand, many of these static methods and even some of the non-static getters should probably be made `const`, the way `validateCavityBoundary()` already is.  I think it's of medium importance and low urgency.  This won't hold up my approval.

Arlie-Capps (approved):
Thanks, Kenny.  I approved.

src/axom/quest/Delaunay.hpp
kennyweiss: Thanks -- it's a static class function, so can't be marked `const`

src/axom/core/Array.hpp
kennyweiss: Thanks -- I'll fix that in a follow-up PR

PR #1865, opened 2026-05-13
Adds a Durand-Kerner polynomial solver to core and improves solver API

kennyweiss:
# Summary

- This PR improves Axom's polynomial solver capabilities
- It adds a [Durand-Kerner polynomial solver](https://en.wikipedia.org/wiki/Durand%E2%80%93Kerner_method), which returns the complex roots for a univariate polynomial
- It also modernizes the API for the linear, quadratic and cubic solvers to use `axom::ArrayView`s instead of pointers to C-style arrays. The latter have been deprecated
- Misc: It also adds `axom::Array::pop_back()` for compatibility with `std::vector`

# Durand-Kerner solver
Here's a visualization (courtesy of claude) of how the solver works for the polynomial: $z^3-1$, initialized with three seed points near the unit circle -- powers of $.4+.9i$

durand_kerner_z3_minus_1

### DIFF
--- a/src/axom/core/numerics/polynomial_solvers.cpp
+++ b/src/axom/core/numerics/polynomial_solvers.cpp
@@ -69,7 +69,8 @@ int effective_polynomial_degree(ArrayView<const double> coeffs_ascending, double
 //------------------------------------------------------------------------------
 PolynomialRootResult solve_polynomial_durand_kerner_checked(ArrayView<const double> coeffs_ascending,
                                                             double tol,
-                                                            int max_iters)
+                                                            int max_iters,
+                                                            Complex seed)
 {
   PolynomialRootResult result;
   const int degree = static_cast<int>(coeffs_ascending.size()) - 1;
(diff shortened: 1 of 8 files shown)


src/axom/core/numerics/polynomial_solvers.hpp
```diff
@@ -18,14 +23,105 @@
...
+ *
+ * \tparam ScalarType The scalar type used for evaluation, e.g. `double` or `std::complex<double>`.
+ * \param [in] coeffs_descending Polynomial coefficients in descending power order.
+ * \param [in] x The evaluation point.
+ *
+ * \return The polynomial value at `x`.
+ */
+template <typename ScalarType>
+ScalarType evaluate_polynomial(ArrayView<const double> coeffs_descending, const ScalarType& x)
+{
```
BradWhitlock, line 47: I noticed that the tparam doc says ScalarType is double or Complex. Is a static_assert() for that in order here, or might be useful to pass Vector<double,3> or other things in down the road?

BradWhitlock (approved):

src/axom/core/numerics/polynomial_solvers.cpp
```diff
@@ -4,25 +4,191 @@
...
+  axom::Array<double> coeffs_descending(effective_degree + 1, effective_degree + 1);
+  for(int i = 0; i <= effective_degree; ++i)
+  {
+    coeffs_descending[i] =
+      coeffs_ascending[effective_degree - i] / coeffs_ascending[effective_degree];
+  }
+
+  result.roots.resize(effective_degree);
+  // Powers of this non-real base with magnitude around 1 give deterministic, distinct initial guesses.
+  const Complex seed_center {0.4, 0.9};
```
jcs15c: Is there anything special about this particular seed value? Could there ever be utility to letting a user pick the seed value? I know that can help a lot in regular Newton solvers.

kennyweiss: It was the value on the wikipedia page and they explicitly say it's not special :)

Sure -- we can pass update the API to optionally pass in a seed.

jcs15c (approved):
Looks good to me!

src/axom/core/numerics/polynomial_solvers.hpp
kennyweiss: I added a `static_assert`. If we need more flexibility later, we can modify the function/signature.

src/axom/core/numerics/polynomial_solvers.cpp
kennyweiss: Done.

PR #1866, opened 2026-05-15
Fixes sign convention for primal's compute_moments

kennyweiss:
# Summary

- This PR fixes the sign convention for `primal::sector_area()` and `primal::sector_centroid()` to match the convention for our quadrature scheme in `primal::evaluate_area_integral()` as well as the sign convention from the original approach that our implementation is based on.
- It also adds some new tests

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -39,6 +39,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Updates CMake code check targets to only use checked in files (via `git ls-files`, when available)
 
 ### Fixed
+- Primal: Fixes signs of `compute_moments` to match orientation convention in `primal::evaluate_area_integral`
 
 ## [Version 0.14.0] - Release date 2026-03-31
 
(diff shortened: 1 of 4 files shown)


RELEASE-NOTES.md
```diff
@@ -36,6 +36,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Updates CMake code check targets to only use checked in files (via `git ls-files`, when available)
 
 ### Fixed
+- Primal: Fixes signs of `compute_moments` to match orientation convention in `primal::evaluate_area_integral`
```
kennyweiss, line 42: I could also see this being `Changed` (instead of `Fixed`) since it wasn't wrong before, just a different convention.

rhornung67 (approved):
Wow. How long has that issue been in the code?

Arlie-Capps (approved):
Thank you, Kenny.  Mathematical consistency is important.  I think "Fixed" is quite appropriate (in your changelog entry).

BradWhitlock:
The failures on dane are the same as I was seeing in my PR. I'm testing whether PR #1868 fixes them.

PR #1872, opened 2026-05-20
 Docker image updates

bmhan12:
This PR:
- Adds `+python` to Github Actions CI Docker images for python unit testing.
- Updates/Fixes `dockerfile_gcc-12_cuda-12` to use Spack 1.1
- Adds Dockerfile `dockerfile_gcc-13_vscode`, which is essentially a non-cuda version of the tutorial `dockerfile_gcc-12_cuda-12` image.
- Updates axom spack recipe with upstream spack-packages changes: spack/spack-packages#4853

Relates to #1609

### DIFF
--- a/scripts/spack/configs/docker/ubuntu22_cuda/spack.yaml
+++ b/scripts/spack/configs/docker/ubuntu22_cuda/spack.yaml
@@ -126,12 +126,14 @@ spack:
       - spec: "+shared~static"
 
     # Globally lock in version of devtools
-    cmake:
-      version: [3.23.1]
-      buildable: false
-      externals:
-      - spec: cmake@3.23.1
-        prefix: /usr/local
+
+    # RAJA needs cmake 3.24:
+    # cmake:
+    #   version: [3.23.1]
+    #   buildable: false
+    #   externals:
+    #   - spec: cmake@3.23.1
+    #     prefix: /usr/local
     doxygen:
       version: [1.8.17]
       buildable: false
--- a/scripts/spack/configs/docker/ubuntu22_cuda/spack.yaml
+++ b/scripts/spack/configs/docker/ubuntu22_cuda/spack.yaml
@@ -22,48 +22,37 @@ spack:
   - ../defaults.yaml
   - ../versions.yaml
 
-  compilers::
-  - compiler:
-      environment: {}
-      extra_rpaths: []
-      flags:
-        cflags: -pthread
-        cxxflags: -pthread
-      modules: []
-      operating_system: ubuntu22.04
-      paths:
-        cc: /usr/bin/gcc
-        cxx: /usr/bin/g++
-        f77: /usr/bin/gfortran
-        fc: /usr/bin/gfortran
-      spec: gcc@12.3.0
-      target: x86_64
+  toolchains:
+    gcc_12:
+    - spec: '%c=gcc'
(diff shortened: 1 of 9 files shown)


rhornung67 (approved):

scripts/spack/packages/axom/package.py
```diff
@@ -568,18 +576,16 @@ def initconfig_hardware_entries(self):
                 cmake_cache_string("BLT_OPENMP_LINK_FLAGS", openmp_gen_exp, description)
             )
 
-        if (
-            spec.satisfies("+openmp")
-            and spec.satisfies("+rocm")
-            and self.spec.satisfies("%cce")
-        ):
+        if spec.satisfies("+openmp") and spec.satisfies("+rocm") and self.spec.satisfies("%cce"):
```
kennyweiss: Consistency (and perhaps out of scope for this PR):  

Is there a reason some of these are `spec.satisfies()` and others are `self.spec.satisfies()`?
Are these different checks? I see `spec = self.spec` above.

scripts/spack/configs/docker/ubuntu22_cuda/spack.yaml
```diff
@@ -137,12 +126,14 @@ spack:
...
-      externals:
-      - spec: cmake@3.23.1
-        prefix: /usr/local
+
+    # RAJA needs cmake 3.24:
+    # cmake:
+    #   version: [3.23.1]
+    #   buildable: false
+    #   externals:
+    #   - spec: cmake@3.23.1
```
kennyweiss: How does commenting out this `external` for CMake lead to RAJA getting cmake@3.24? 
Is that provided by the container, or expressed as a dependency somewhere else?

kennyweiss (approved):
Thanks @bmhan12 !

scripts/spack/configs/docker/ubuntu22_cuda/spack.yaml
bmhan12: > How does commenting out this external for CMake lead to RAJA getting cmake@3.24?

spack becomes the one that builds the correct cmake dependency version. 
Useful to keep the path around commented, just in case the external cmake version is updated in the future and we want/can to use it.

rhornung67: RAJA has a min required version of CMake 3.24 in its top-level CMakeLists.txt file. I think this is also reflected in the RAJA Spack package. We would like to bump it higher because C++20 requires newer CMake. This is a compromise on what's available on LC machines and in the radiuss docker compiler images.

scripts/spack/packages/axom/package.py
bmhan12: Fixed! They are the same checks - using `spec.satisfies()` for clarity.

scripts/spack/configs/docker/ubuntu22_cuda/spack.yaml
kennyweiss: Thanks -- since we're likely not going to remember to update this line, please update the comment on line 130 to say something to the effect of 
> RAJA's spack package depends on a newer version of CMake for CUDA than 3.23.1.

i.e. let's not explicitly pin down the exact dependency, and let's point people to the RAJA package for details

PR #1874, opened 2026-05-26
Adds a BezierTriangle class to `primal`

jcs15c:
# Summary

This PR is adds the class `primal::BezierTriangle`, which manages the control points and weights for a possibly rational Bezier triangle class. The implementation is meant to mirror `BezierPatch` where possible.
- Control points and weights for an order `m` triangle are stored internally in a linear array, and are accessed through `BezierTriangle::operator()(int i, int j)`, with the restriction that `i + j <= m`.
- Evaluation methods for the surface and derivatives. Evaluation is parameterized by `(u0, v0)`, which are almost (but not quite) the first two elements of Barycentric coordinates. These coordinates are permuted so that `tri.evaluate(0.0, 0.0) == tri(0, 0)`, `tri.evaluate(0.0, 1.0) == tri(0, m)`, and `tri.evaluate(1.0, 0.0) == tri(m, 0)`. If it were real Barycentric coordinates, then `tri.evaluate(0.0, 0.0 \*, 1.0*\)` would be the same as the `(1, 0)` control point, and that seems confusing. 
- Various `BezierTriangle::split` methods, which split a triangle into 2, 3, and 4 patches respectively. There is also a dedicated `uniformSplit` method which uses a faster, more specialized method based on degenerate tetrahedon. Some cursory profiling says that it is about 4x the speed of the generic method.

I've also added some minor readability improvements to `BezierPatch`.

### DIFF
--- a/src/axom/primal/geometry/BezierTriangle.hpp
+++ b/src/axom/primal/geometry/BezierTriangle.hpp
@@ -1528,10 +1528,7 @@ public:
    *
    * \param [in] ord Triangle order
    */
-  static constexpr int triSize(int ord)
-  {
-    return (ord >= 0) ? ((ord + 1) * (ord + 2) / 2) : 0;
-  }
+  static constexpr int triSize(int ord) { return (ord >= 0) ? ((ord + 1) * (ord + 2) / 2) : 0; }
 
   /*!
    * \brief Maps triangular indices \a (i,j) to the linear storage index
--- a/src/axom/primal/geometry/BezierTriangle.hpp
+++ b/src/axom/primal/geometry/BezierTriangle.hpp
@@ -40,8 +40,7 @@ std::ostream& operator<<(std::ostream& os, const BezierTriangle<T, NDIMS>& bTri)
  *
  * \brief Represents a Bezier triangle defined by a triangular array of control points
  * \tparam T the coordinate type, e.g., double, float, etc.
- * \tparam NDIMS The dimension of each control point, e.g. 4 for homogeneous surfaces 
- *                 or 1 for rational weights.
+ * \tparam NDIMS The dimension of each control point, e.g. 1 for rational weights
  *
  * A Bezier triangle of order \a N has \f$ (N+1)(N+2)/2 \f$ control points.
  * It is parametrized over the domain \f$ u0 \ge 0, v0 \ge 0, u0+v0 \le 1 \f$.
@@ -54,6 +53,9 @@ std::ostream& operator<<(std::ostream& os, const BezierTriangle<T, NDIMS>& bTri)
  * Rational triangles are represented by an additional set of positive weights.
  * Polynomial (nonrational) Bezier triangles are identified by an empty weights array.
  *
+ * A default-constructed triangle will have order -1, and is "invalid".
+ * Arrays of nodes and weights will be empty, and most methods are invalid 
+ * 
  * \note This triangle uses permuted barycentric coordinates (u0, v0) for evaluation such that, when
  * `getOrder()==1`, the parameter values correspond to the triangle vertices:
  * - `evaluate(0,0) == (*this)(0,0)`
@@ -118,10 +120,8 @@ public:
    * All weights must be greater than 0 in a rational triangle.
    *
    * For 1D control point/weight arrays, the expected layout corresponds to the indexing
-   * used by `operator()(i,j)`:
-     \verbatim
-      pts[ triIndex(N,i,j) ]  <->  (*this)(i,j)   for 0<=i<=N and 0<=j<=N-i
-     \endverbatim
+   * used by `operator()(i,j)`: 
+   *    pts[ triIndex(N,i,j) ]  <->  (*this)(i,j)   for 0<=i<=N and 0<=j<=N-i
    */
 
   /**
--- a/src/axom/primal/geometry/BezierTriangle.hpp
+++ b/src/axom/primal/geometry/BezierTriangle.hpp
@@ -27,8 +27,6 @@
 
 #include <vector>
 #include <ostream>
-#include <unordered_map>
-#include <cstdint>
 
 namespace axom
 {
@@ -479,36 +477,41 @@ public:
       const int degPrev = n - (p - 1);
       std::vector<axom::Array<PointType>> currNets(triSize(p));
 
-      for(int a = 0; a <= p; ++a)
+      for(int fixedVcCount = 0; fixedVcCount <= p; ++fixedVcCount)
       {
-        for(int b = 0; b <= p - a; ++b)
+        for(int fixedVbCount = 0; fixedVbCount <= p - fixedVcCount; ++fixedVbCount)
         {
-          const int idx = triIndex(p, a, b);
-          const int c = p - a - b;
+          const int fixedVaCount = p - fixedVcCount - fixedVbCount;
+          const int idx = triIndex(p, fixedVcCount, fixedVbCount);
 
-          // Choose a unique predecessor (and thus a unique evaluation order) to avoid
-          // redundant reductions; the blossom symmetry ensures the final values are
-          // order-independent.
-          const Barycentric* Q = nullptr;
+          // Each net in this "tetrahedral" construction corresponds to a blossom value
+          // b(Vc^fixedVcCount, Vb^fixedVbCount, Va^fixedVaCount) at degree (n-p).
+          //
+          // There are multiple equivalent ways to compute each blossom value (blossom is
+          // symmetric in its arguments). We pick one deterministic predecessor to keep
+          // the recurrence simple and compute each net exactly once.
+          //
+          // Convention: consume Va arguments first, then Vb, then Vc.
           int predIdx = -1;
+          const Barycentric& splitPoint =
+            (fixedVaCount > 0) ? Va : (fixedVbCount > 0) ? Vb : Vc;
 
-          if(c > 0)
+          if(fixedVaCount > 0)
           {
-            predIdx = triIndex(p - 1, a, b);
-            Q = &Va;
+            predIdx = triIndex(p - 1, fixedVcCount, fixedVbCount);
           }
-          else if(b > 0)
+          else if(fixedVbCount > 0)
           {
-            predIdx = triIndex(p - 1, a, b - 1);
-            Q = &Vb;
+            predIdx = triIndex(p - 1, fixedVcCount, fixedVbCount - 1);
           }
           else
           {
-            predIdx = triIndex(p - 1, a - 1, b);
-            Q = &Vc;
+            SLIC_ASSERT(fixedVcCount > 0);
+            predIdx = triIndex(p - 1, fixedVcCount - 1, fixedVbCount);
           }
 
-          currNets[idx] = reduce_once(prevNets[predIdx], degPrev, *Q);
+          SLIC_ASSERT(predIdx >= 0);
+          currNets[idx] = reduce_once(prevNets[predIdx], degPrev, splitPoint);
         }
       }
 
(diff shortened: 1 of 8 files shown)


src/axom/primal/geometry/BezierTriangle.hpp
```diff
@@ -0,0 +1,1828 @@
...
+  using CoordsVec = axom::Array<PointType, 1>;
+  using WeightsVec = axom::Array<T, 1>;
+
+  using BoundingBoxType = BoundingBox<T, NDIMS>;
+  using OrientedBoundingBoxType = OrientedBoundingBox<T, NDIMS>;
+  using BezierCurveType = primal::BezierCurve<T, NDIMS>;
+  using EdgesVec = axom::Array<BezierCurveType>;
+
+  AXOM_STATIC_ASSERT_MSG((NDIMS == 1) || (NDIMS == 2) || (NDIMS == 3),
+                         "A Bezier Triangle object may be defined in 1-, 2-, or 3-D");
```
kennyweiss, line 95: (I was wondering what a 1D triangle might be)

It might be helpful to mention that `NDIMS` is the number of coordinates per control point, and that `NDIMS==1` can be used to represent a `BezierTriangle` of the rational weights.

src/axom/primal/geometry/BezierTriangle.hpp
```diff
@@ -0,0 +1,1828 @@
...
+   * \param [in] j Second control net index
+   * \pre Requires that the triangle be rational
+   */
+  const T& getWeight(int i, int j) const
+  {
+    SLIC_ASSERT(isRational());
+    SLIC_ASSERT(m_weights.size() == m_controlPoints.size());
+    SLIC_ASSERT(i >= 0);
+    SLIC_ASSERT(j >= 0);
+    SLIC_ASSERT(i + j <= m_ord);
```
kennyweiss: Should we have add an `isValidIndex(i,j)` utility function?

src/axom/primal/geometry/BezierTriangle.hpp
```diff
@@ -0,0 +1,1828 @@
...
+   * \param [in] edgeIdx Index of the requested edge in \a [0,2]
+   *
+   * The edges are returned in counter-clockwise order with respect to the
+   * parameter domain (u,v), with edge 0 across from corner 0 (i.e. evaluate(0,0)):
+   * - \a edgeIdx = 0: u+v = 1 from `evaluate(1,0)` to `evaluate(0,1)`
+   * - \a edgeIdx = 1: u = 0 from `evaluate(0,1)` to `evaluate(0,0)`
+   * - \a edgeIdx = 2: v = 0 from `evaluate(0,0)` to `evaluate(1,0)`
+   *
+   * \return A Bezier curve representing the requested edge.
+   */
```
kennyweiss, line 372: Minor: Should we mention that this constructs a copy of the edge (rather than (a reference to) the actual edge)?  It's probably obvious from the return type, but it might help to explicitly mention this.

src/axom/primal/geometry/BezierTriangle.hpp
```diff
@@ -0,0 +1,1828 @@
...
+    case 0:
+      for(int k = 0; k <= m_ord; ++k)
+      {
+        const int i = k;
+        const int j = m_ord - k;
+        pts[k] = (*this)(i, j);
+        if(isRational())
+        {
+          wts[k] = m_weights[triIndex(m_ord, i, j)];
+        }
```
kennyweiss: Instead of explicit indexing for the pts and triIndex for the weights, would the following be clearer for this (and similarly for the other loops)?

```suggestion
        const int idx = triIndex(m_ord, k, m_ord - k);
        pts[k] = m_controlPoints[ids];
        if(isRational())
        {
          wts[k] = m_weights[idx];
        }
```

src/axom/primal/geometry/BezierTriangle.hpp
```diff
@@ -0,0 +1,1828 @@
...
+   *          / | \
+   *         /t2|t1\
+   *        /  /Q\  \
+   *       / /  t0 \ \
+   *      //_________\\
+   *     B             C
+   *
+   * \return t0 = Tri(B,C,Q), t1 = Tri(C,A,Q), t2 = Tri(A,B,Q)
+   */
+  void split(T u0, T v0, BezierTriangle& t0, BezierTriangle& t1, BezierTriangle& t2) const
```
kennyweiss, line 597: Design: I think it might be somewhat confusing to use a `Point<T,3>` in some cases for barycentric coords (as in `restrictToSubtriangle`) and to use `u0, v0` in other cases for the first two coordinates of a barycentric triple.

Would it make sense to have a type alias for the two-coordinate version (perhaps packed as a std::pair/std::tuple) to make it clearer through the type system ?

src/axom/primal/geometry/BezierTriangle.hpp
```diff
@@ -0,0 +1,1828 @@
...
+ *
+ * Rational triangles are represented by an additional set of positive weights.
+ * Polynomial (nonrational) Bezier triangles are identified by an empty weights array.
+ *
+ * \note This triangle uses permuted barycentric coordinates (u0, v0) for evaluation such that, when
+ * `getOrder()==1`, the parameter values correspond to the triangle vertices:
+ * - `evaluate(0,0) == (*this)(0,0)`
+ * - `evaluate(0,1) == (*this)(0,1)`
+ * - `evaluate(1,0) == (*this)(1,0)`
+ * These are mapped to standard Barycentric coordinates through (u0, v0) = {1 - u0 - v0, v0, u0}
```
kennyweiss: Should we add ascii art here?

src/axom/primal/geometry/BezierTriangle.hpp
```diff
@@ -0,0 +1,1828 @@
...
+        {
+          const int end = m_ord - p + 1;
+          for(int i = 0; i < end; ++i)
+          {
+            for(int j = 0; j < end - i; ++j)
+            {
+              const auto& A = dCarray[triIndex(end, i, j)];
+              const auto& B = dCarray[triIndex(end, i, j + 1)];
+              const auto& C = dCarray[triIndex(end, i + 1, j)];
+              dCarray[triIndex(end - 1, i, j)] = A + u0 * (C - A) + v0 * (B - A);
```
kennyweiss: Minor: It seems like we are using/redefining linear interpolation on three points and weights a few times, and it might make sense to to directly define a lerp function for them -- either in this file or in Point.hpp

src/axom/primal/tests/primal_bezier_triangle.cpp
```diff
@@ -0,0 +1,1038 @@
...
+  // Vertex mapping checks
+  const auto pA = tri.evaluate(0.0, 0.0);
+  const auto pB = tri.evaluate(0.0, 1.0);
+  const auto pC = tri.evaluate(1.0, 0.0);
+  const auto pQ = tri.evaluate(u, v);
+
+  for(int i = 0; i < DIM; ++i)
+  {
+    // t0 -> (B, C, Q)
+    EXPECT_NEAR(pB[i], t0.evaluate(0.0, 0.0)[i], 1e-10);
```
kennyweiss: Minor: It might be better to define the EPS parameter as const variable for the function and use it for all these tests.

kennyweiss (approved):
Thanks @jcs15c!

src/axom/primal/geometry/BezierTriangle.hpp
jcs15c: Good suggestion! I think it makes sense as a static function alongside `triIndex(ord, i, j)`.

src/axom/primal/geometry/BezierTriangle.hpp
jcs15c: I tried to fix this in as many places as it made sense (which it turns out is a good amount!)

src/axom/primal/geometry/BezierTriangle.hpp
jcs15c: I did this, and I really like the way it turned out. In particular, it really drives home the difference between the conventions of "parametric coordinates" `(u0, v0)` and Barycentric coordiantes `{u,v,w}` (which I explain in a bit more detail in the other comment).

src/axom/primal/geometry/BezierTriangle.hpp
jcs15c: I've gone back and forth on this a lot, but I think the most straightforward way is to have the primary interface type be what I think of as parameter/lexicographical coordinates `(u0, v0)`, where these coordinates match up with the natural ordering of the control points in the array, i.e. `eval(0, 0) == m_controlPoints(0, 0)`, `eval(1, 0) == m_controlPoints(ord, 0)`, and so on. I think in a lot of ways this is a more natural parameterization of the domain than, say, the actual first two coordinates of the barycentric triple (which would have `eval(0,0) == m_controlPoints(ord, 0)`, and that feels off to me). It also makes it much, much easier to handle derivative methods, since now `BezierTriangle::du/dv` actually reflect the same parameterization as is used for evaluation. 

That being said, there are contexts within the various evaluation/derivative/subdivision methods where it's convenient to work internally in terms of real Barycentric coordinates. Similarly, the `restrictToSubtriangle` function would have a really ugly interface if you needed to pass in 6 doubles for `u0_1`, `v0_1`, `v0_2`, etc. In that case, it's more readable to pass in Barycentric coordinates directly. 

I've tried to make this convention as clear as possible in the implementation and documentation, such as referring to parametric coordinates exclusively as `(u0, v0)` to distinguish from the barycentric coordinates `{u,v,w}`. I've also added the typedef `Barycentric = primal::Point<T, 3>` to make the distinction explicit, and down the line there's no reason we couldn't overload the various evaluate methods to have `evaluate(Barycentric Q)` in addition to `evaluate(T u0, T v0)`. 

What do you think, do you think the convention is clear enough now?

src/axom/primal/geometry/BezierTriangle.hpp
```diff
@@ -0,0 +1,1862 @@
...
+ *   Parametric (u0, v0):       Barycentric {u,v,w}:
+ *        (1, 0)                      {0,0,1}
+ *          /\                          /\
+ *         /  \                        /  \
+ *   ^    /    \         <--->        /    \
+ *   |   /      \                    /      \
+ *   |  /        \                  /        \
+ *  v0 /__________\                /__________\
+ * (0, 0) u0 ---> (0, 1)       {1,0,0}        {0,1,0}
+ *
```
Arlie-Capps, line 76: Please add a note here describing order -1.  From reading the code, this is a flag that signals a default-constructed BezierTriangle for which most methods are invalid.

Arlie-Capps: I feel the Doxygen for the class is the perfect place for such a note because it's describing the software construct that implements the mathematical ideal (including such techniques as `m_ord == -1`).

Arlie-Capps (commented):
Thanks for this work.  Only one request.

src/axom/primal/geometry/BezierTriangle.hpp
jcs15c: This makes a lot of sense, thank you for the suggestion! This is the same convention that was used in the other Bezier objects (`BezierCurve` and `BezierPatch`), but was undocumented there too. I've added this comment in all of the places, and adjusted a few other places where the word "invalid" or "empty" was used improperly.

Arlie-Capps: Thank you.

Arlie-Capps (approved):

rhornung67 (approved):

PR #1878, opened 2026-06-03
FlatMap fixups for CUDA

publixsubfan:
# Summary

- Works around an issue in batched insertion with a FlatMap allocated in pinned memory. Pinned memory is not coherent with respect to the GPU when using the default device-scoped atomics; to work around this, we allocate a temporary map in the device memory space.
- Clarifies a skipped `core_flatmap` test case with additional comments.


rhornung67 (approved):
Thank you @publixsubfan

bmhan12 (approved):

kennyweiss (approved):
Thanks @publixsubfan !

PR #1880, opened 2026-06-03
Improves error handling/reported for invalid c2c contours

### LINKED ISSUE #1879: Improve error handling for degenerate c2c contour files
A user reported problems shaping in contour files.

It turned out that a linear spline had a duplicate control point, e.g.
```
point = spline_start
piece = rz(units=cm, interpolation=linear, 
     rz=  0 5
		  5 0
		  5 0  # <-- duplicate!
		  0 -5)
point = spline_end
piece = line(end=spline_start)
```
C2C converts this a NURBS curve of degree 1 with `C^-1` continuity (i.e. discontinuous), which our `NURBSCurve` class (intentionally) does not support.

Loading this input triggers a validity check in our `KnotVector` class via a `SLIC_ASSERT` and crashes in Debug configs without a good error message.
More significantly, this *does not* get triggered at construction time in Release configs and manifests as an out-of-bounds array indexing error (which is much harder to pinpoint).

We should fix this by improving the error messages when loading invalid contour files and ensuring that it gets reported in all configs (debug and release).

I don't think we need to directly support this invalid mesh in our reader since it is easy enough for users to fix their input data.
On the other hand, this case (linear spline with duplicated control point) might be common enough to try to fix on our side.

kennyweiss:
# Summary

- This PR improves error handling and reporting when loading invalid c2c contour files
- It adds constructors to `primal::KnotVector` to skip validity assertions during construction.
   - This allows users to call `isValid` and handle invalid inputs appropriately.
   - Previously, the construction would assert, and fail without error messages in Debug configs, and silently chug along in Release configs
 - It also improves our handling/reporting of errors in the shaping application and other examples that load c2c files
- Resolves https://github.com/llnl/axom/issues/1879
- Depends on https://github.com/llnl/axom_data/pull/38

### DIFF
--- a/src/axom/quest/examples/containment_driver.cpp
+++ b/src/axom/quest/examples/containment_driver.cpp
@@ -69,12 +69,17 @@ public:
   }
 
 #ifdef AXOM_USE_C2C
-  void loadContourMesh(const std::string& inputFile, int segmentsPerKnotSpan)
+  bool loadContourMesh(const std::string& inputFile, int segmentsPerKnotSpan)
   {
     AXOM_ANNOTATE_SCOPE("load c2c");
     quest::C2CReader reader;
     reader.setFileName(inputFile);
-    reader.read();
+    const int rc = reader.read();
+    if(rc != 0)
+    {
+      SLIC_WARNING(axom::fmt::format("Failed to load contour file '{}'", inputFile));
+      return false;
+    }
 
     // Create surface mesh
     m_surfaceMesh.reset(new UMesh(2, mint::SEGMENT));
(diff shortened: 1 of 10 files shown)


src/axom/quest/examples/containment_driver.cpp
```diff
@@ -69,28 +69,33 @@ class ContainmentDriver
...
     m_surfaceMesh.reset(new UMesh(2, mint::SEGMENT));
     quest::LinearizeCurves lin;
     lin.getLinearMeshUniform(reader.getCurvesView(),
                              static_cast<UMesh*>(m_surfaceMesh.get()),
                              segmentsPerKnotSpan);
+    return true;
   }
 #else
-  void loadContourMesh(const std::string& inputFile, int segmentsPerKnotSpan)
+  bool loadContourMesh(const std::string&, int))
```
kennyweiss:
```suggestion
  bool loadContourMesh(const std::string&, int)
```

rhornung67 (approved):
Looks like a good improvement. Thanks for judiciously applying the const qualifer in ctor args to make it clear that the ctor will not modify them.

src/axom/quest/io/C2CReader.cpp
```diff
@@ -57,23 +61,123 @@ int C2CReader::readContour()
 
   SLIC_INFO(fmt::format("Loading contour with {} pieces", contour.getPieces().size()));
 
+  // Build results transactionally so we don't retain partial curves on error
+  CurveArray nurbs_data;
```
jcs15c, line 65: This is maybe a nitpick, but I think it's probably unnecessarily confusing to call the array of curve objects `nurbs_data` here, and then the individual curve `nurbsData` a few lines down.

Otherwise, it looks great! I like the idea of not always forcing validity at construction time, and giving people the option to do it themselves later. I'm thinking about if that's a design pattern worth implementing in the other NURBS/Bezier classes down the line.

jcs15c (approved):

src/axom/quest/io/C2CReader.cpp
kennyweiss: Good call.  I'll change this in a follow-up.

PR #1881, opened 2026-06-09
Temper randomness in 3D GWN method 

jcs15c:
I've noticed that the text `primal_solid_angle_test` fails a bit more frequently than I'd like, as the underlying 3D GWN method has a few random components that fail with probability 0 mathematically, but probability >0 in floating point (by my estimate, about 0.1% of the time). 

In particular, this PR addresses:
 - Within the GWN algorithm, a re-cast ray was taken to be a random unit vector. This has been reworked to use a Halton sequence projected onto the unit sphere, which also has the advantage of more reliably distributing the cast attempts over the unit sphere.
 - The per-surface cast-direction is taken to be the surface's average normal, plus a small perturbation. This has been changed so that the perturbation is seeded deterministically based on a (possibly too-simple) hashed function of the surface bounding box.

I've also added a simple test to verify this in `primal_solid_angle.cpp`.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -50,6 +50,7 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 ### Fixed
 - Primal: Fixes signs of `compute_moments` to match orientation convention in `primal::evaluate_area_integral`
 - Quest: Improves error handling/reporting when loading an invalid c2c contour
+- Primal: Improves reproducibility of 3D GWN methods by removing some sources of randomness
 
 ## [Version 0.14.0] - Release date 2026-03-31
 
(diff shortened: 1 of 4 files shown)


kennyweiss (approved):
Thanks! 
Would it make sense for `halton` to be a core utility function w/ unit tests?  
Perhaps not this time, but the next time we need it (?)

Please update the RELEASE_NOTES about this change.

jcs15c:
> Thanks! Would it make sense for `halton` to be a core utility function w/ unit tests? Perhaps not this time, but the next time we need it (?)

It could get put it in `axom::utilities` alongside `binomial_coefficient`, but yes, since it's just used in this single lambda function for now, I think it'd be overkill. However, I do like Halton sequences on an aesthetic level, and will be eagerly looking for other places they could be useful.

bmhan12 (approved):

PR #1882, opened 2026-06-11
Bugfixes, benchmarks and improvements to FlatMap

kennyweiss:
# Summary

- This PR adds some bugfixes and performance improvements to axom::FlatMap
- It also adds an initial benchmark suite for FlatMap against `std::unordered_map`, google `sparsehash` and `std::map`
- Bugfixes:
   - There were some bugs related to truncating hashes to 32 bits (when IndexType is 32 bits), and in casting from float to int, and in using operator[] on const maps and in the copy-assign operator.
- Optimizations
  - Since the hashes are powers of 2, we can use bitmasks rather than mod (%)
  - Specialized batch insertion for sequential exec policy, where we don't need to worry about synchronization

### Serial benchmark results using a RelWithDebInfo config (lower is better)
We now get roughly comparable or better results in serial -- compared to `std::unordered_map`, `std::map` and our vendored google sparsehash.  `FlatMap` is our default hash function, `FlatMapFastHash` is a different hash function that appears to be somewhat faster.

#### Hashing 32K pairs ($2^{15}$)
image

#### Hashing 1M pairs ($2^{20}$)
image

### Serial vs. OMP vs GPU

This branch has some modest speedups vs. develop (showing serial and omp for this branch against axom@develop)
Showing SEQ and OMP with {1,2,4,8,16,32,64} threads and run with 
```
OMP_NUM_THREADS=<n> OMP_PLACES=cores OMP_PROC_BIND=close
```

#### Hashing 32K pairs ($2^{15}$)
core_flatmap_speedup_wall_N32768

#### Hashing 1M pairs ($2^{20}$)
core_flatmap_speedup_wall_N1048576

#### Hashing 32M pairs ($2^{25}$)
core_flatmap_speedup_wall_N33554432

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -59,6 +59,8 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 - Primal: Improves reproducibility of 3D GWN methods by removing some sources of randomness
 - Core: ArrayView assigments/copies now copy the stride
 - Core: Array construction from strided ArrayView now correctly copies the strided elements
+- Core: Improved `axom::FlatMap` insertion performance by fusing duplicate-key lookup with empty-slot probing.
+- Core: Updated DeviceHash to use 64-bit hash results and improved coverage for integer and floating-point hashing.
 
 ## [Version 0.14.0] - Release date 2026-03-31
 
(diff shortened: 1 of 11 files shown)


rhornung67 (approved):

BradWhitlock (approved):
Thanks @kennyweiss .

PR #1883, opened 2026-06-12
Inlet: add ability to have collections with variant values

white238:
@adayton1 

Thought we had this ability already but only had dictionaries with variant keys. Also added documentation and examples for both variant and homogeneous collections

### DIFF
--- a/src/axom/inlet/ConduitReader.cpp
+++ b/src/axom/inlet/ConduitReader.cpp
@@ -176,6 +176,16 @@ void arrayToMap(const conduit::DataArray<ConduitType>& array,
   }
 }
 
+void boolArrayToMap(const conduit::DataArray<conduit::uint8>& array,
+                    std::unordered_map<int, VariantValue>& map)
+{
+  map.clear();
+  for(conduit::index_t i = 0; i < array.number_of_elements(); i++)
+  {
+    map[i] = VariantValue {static_cast<bool>(array[i])};
+  }
+}
+
 /*!
  *******************************************************************************
  * \brief Recursive name retrieval function - adds the names of all descendents
@@ -382,7 +392,7 @@ ReaderResult ConduitReader::getStringMap(const std::string& id,
 ReaderResult ConduitReader::getVariantMap(const std::string& id,
                                           std::unordered_map<int, VariantValue>& values)
 {
-  return getArray(id, values);
+  return getVariantArray(id, values);
 }
 
 ReaderResult ConduitReader::getVariantMap(const std::string& id,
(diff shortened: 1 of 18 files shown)


src/axom/inlet/docs/sphinx/simple_types.rst
```diff
@@ -127,30 +127,128 @@ can get the Container instance first, then access it with the relative name.
 Arrays
 ******
 
-Coming soon!
```
white238, line 130: We won't discuss how much of a lie this was.......

src/axom/inlet/ConduitReader.cpp
```diff
@@ -155,6 +155,27 @@ void arrayToMap(const conduit::DataArray<ConduitType>& array,
...
+                std::unordered_map<int, VariantValue>& map)
+{
+  map.clear();
+  for(conduit::index_t i = 0; i < array.number_of_elements(); i++)
+  {
+    if(std::is_floating_point<ConduitType>::value)
+    {
+      const double double_value = array[i];
+      const int int_value = static_cast<int>(double_value);
+      map[i] = (static_cast<double>(int_value) == double_value) ? VariantValue {int_value}
```
adayton1, line 169: Is handling missing for bools or strings?

adayton1 (approved):

src/axom/inlet/ConduitReader.cpp
white238: This section is only for dense numeric arrays. Bools and strings are handled before it gets to this. Honestly this part of the code i am not super familiar with before this PR. Looking this over i think we can iterate and clean up cases.

src/axom/inlet/docs/sphinx/simple_types.rst
```diff
@@ -127,30 +127,128 @@ can get the Container instance first, then access it with the relative name.
...
+   :end-before: _inlet_simple_types_variant_collections_add_end
+   :language: C++
 
 Accessing
 ---------
 
-Coming soon!
+Contiguous homogeneous arrays can be retrieved as ``std::vector<T>``.  Integer-keyed
+homogeneous arrays can be retrieved as ``std::unordered_map<int, T>`` when the
+original indices are needed.
```
Arlie-Capps: The phrase "when the original indices are needed" here and at (new) lines 184 and 185 implies that you can get the integer-keyed homogeneous or variant array values in some other way, when you don't need the original indices.  Is there another to get the values?  If so, please document.  If there isn't another way to get that data, please rephrase to something like

"Integer-keyed variant arrays can be retrieved as ``std::unordered_map<int, T>``.  The keys of the ``std::unordered_map`` provide the original indices."

Arlie-Capps: OK, I see at inlet_object.cpp (new) line 847 how you can get the values as a std::array<T>.  Please mention this, perhaps rephrasing the sentence to something like

"Integer-keyed variant arrays can be retrieved as std::unordered_map<int, T>. The keys of the std::unordered_map provide the original indices.  The values (without the keys) of an integer-keyed variant array can be retrieved as a std::vector<T>."

Arlie-Capps (approved):

src/axom/inlet/docs/sphinx/simple_types.rst
adayton1: Ha!

src/axom/inlet/docs/sphinx/simple_types.rst
white238: Good catch. I split the example into smaller examples. I think its more explicit this way.

PR #1884, opened 2026-06-13
Bugfix and optimization for 1D Array/ArrayView

kennyweiss:
# Summary

- This PR is a bugfix and an optimization
- It fixes a bug in the `ArrayView` copy/assign operator where the stride was not copied
- It fixes a bug when copying strided ArrayViews into (unstrided) 1D Arrays
- It optimizes the access operator for `axom::Arrray<T,1>`. Since this always has a unit stride, we should compile away the stride multiplication.
  - The new benchmarks show the following improvement:
      - `Array<double>::address_bracket/1024`: develop ~385 ns → branch ~283 ns (~26% faster)
      - `Array<double>::address_flatIndex/1024`: develop ~344 ns → branch ~283 ns (~18% faster)
- I found this as a possible speedup for #1882, but it should benefit all `axom::Array` users

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -48,11 +48,13 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 
 ### Changed
 - Updates CMake code check targets to only use checked in files (via `git ls-files`, when available)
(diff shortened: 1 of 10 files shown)


rhornung67 (approved):
I thought we preferred `#if defined(...)` over `#ifdef`

BradWhitlock (approved):
Nice speedup.

PR #1885, opened 2026-06-13
Inlet: Add ability to have containers that hold variant user defined structs

white238:
Allows you to have a dictionary or an array that holds multiple different user-defined structs that are differentiated with a discriminator. For example:

```  
shapes = {
    { kind = "circle", radius = 2.5 },
    { kind = "box", width = 3.0, height = 4.0 }
}
```


rhornung67 (approved):

adayton1:
Is there a way to iterate over the objects in the variant array without first converting to a vector?

white238:
> Is there a way to iterate over the objects in the variant array without first converting to a vector?

Not as far as I know without either accessing the sidre tree that holds the information or reworking a bunch of Inlet.

Arlie-Capps (approved):

adayton1 (approved):

PR #1888, opened 2026-06-18
Improves lifetime management for wrapped sidre entities in Python interface

kennyweiss:
# Summary

- This PR further improves support and interoperability for sidre's Python bindings 
- It improves lifetime management for sidre entities in Python interface. 
  - Specifically, the wrapped sidre entities (views, groups, attributes, datastore, iterators) stay alive as long as their python object
  - This allows, e.g. external sidre views into numpy arrays
- It also allows wrapped IOManager to be initialized with MPI communicators other than `MPI_COMM_WORLD`
- It exports the python interface stubs and types for pysidre, allowing for better tool support and IDE integration
- It updates Axom's python test macro `axom_add_python_test` to support `MPI_NUM_TASKS`

### DIFF
--- a/src/axom/sidre/tests/sidre_spio_Py.py
+++ b/src/axom/sidre/tests/sidre_spio_Py.py
@@ -100,6 +100,32 @@ def test_iomanager_split_communicator(tmp_path):
         sub.Free()
 
 
+def test_distributed_generate_blueprint_index(tmp_path):
+    # The distributed generateBlueprintIndex overload is built only under
+    # nanobind >= 2.10; skip cleanly if this build omitted it.
+    if not pysidre.AXOM_HAS_DISTRIBUTED_BLUEPRINT_INDEX_BINDING:
+        pytest.skip("generateBlueprintIndex not bound")
+
+    ds = pysidre.DataStore()
+    root = ds.getRoot()
+    mesh = root.createGroup("mesh")
+    coords = mesh.createGroup("coordsets/coords")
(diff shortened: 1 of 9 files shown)


src/axom/sidre/tests/sidre_spio_Py.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def test_iomanager_legacy_use_scr_constructor():
+    # Preserve the previous IOManager(use_scr=False) positional API.
+    pysidre.IOManager(False)
+
+
+def test_iomanager_rejects_non_communicator():
+    with pytest.raises(AttributeError):
+        pysidre.IOManager(object())
```
bmhan12, line 51: Should these unit tests have an `assert()` ?

bmhan12 (approved):
Thanks @kennyweiss!

src/axom/sidre/tests/sidre_spio_Py.py
kennyweiss: This test is expected to throw an `AttributeError` since `IOManager` is being constructed without a valid MPI communicator. If it does not throw the exception, the test fails.

Did you have something else in mind for this test?

bmhan12: That works!

nselliott (approved):
I took a look specifically at the IOManager stuff and it looks correct.

PR #1892, opened 2026-06-24
Modernizes slam for c++17

kennyweiss:
# Summary

- This PR includes feature + refactoring changes to slam as well as some support utilities in core
- It modernizes several parts of slam with c++17 constructs (e.g. `if constexpr` instead of SFINAE; avoiding some no-longer-necessary tempate metaprogramming, like `slam::traits::void_t`),
  consolidates duplicated policy code (`RuntimeValue`/`CompileTimeValue`) and replaces several pre-C++17 idioms
- It also adds a few utilities to core
  - ~~`axom::Optional<T>` as a trivially copyable host/device analog for `std::optional` that is usable in device code (where std::optional is not available)~~  _(This turned out not to be necessary. Thanks @publixsubfan )_
  - The `AXOM_CONSTEXPR_ASSERT` macro can be used for assertions relating to constexpr. Failures during contant evaluations are compiler errors, runtime errors abort in Debug configs and are ignored in release configs
    - I moved `axom::processAbort()` to its own file in support of this change
  - `axom::type_identity` is a `c++17` equivalent to the `c++20` `std::type_identity`. Its purpose is to suppress template argument deduction. We can remove it when we upgrade to `c++20`

- Key slam changes/improvements
  - Adds utility functions to make it easier to construct slam sets (`make_*_set`), relations (`make_*_relation`) and maps (`make_map`).
  - Improves slam's storage helpers (`FieldRegistry`) to also back fields/buffers based on `axom::Array` and `axom::ArrayView`
  - Adds some overloads to slam queries that return `std::optional` when an entry might be missing
  - Tags some functions as `[[nodiscard]]`


src/axom/core/utilities/Utilities.cpp
```diff
@@ -24,7 +24,7 @@ namespace axom
 {
 namespace utilities
 {
-void processAbort()
+[[noreturn]] void processAbort()
```
BradWhitlock: It looks like the prototype for `processAbort()` is now in `Abort.hpp`. I'd move the definition to a new `Abort.cpp` to match.

src/axom/core/Macros.hpp
```diff
@@ -416,4 +416,13 @@
...
   void GTEST_TEST_CLASS_NAME_(CaseName, TestName)<gtest_TypeParam_>::TestBody()
 
+// Provides the definition for `axom::detail::constexprAssert` used by AXOM_CONSTEXPR_ASSERT.
+#include "axom/core/utilities/ConstexprAssert.hpp"
+
+/*!
+ * \def AXOM_CONSTEXPR_ASSERT(EXP)
+ * \brief Assert \a EXP in a way that is valid inside constexpr functions.
+ */
+#define AXOM_CONSTEXPR_ASSERT(EXP) ::axom::detail::constexprAssert((EXP), #EXP, __FILE__, __LINE__)
```
BradWhitlock, line 426: Could this be used elsewhere in `axom::core` where we use `assert()` because we can't use `SLIC_ASSERT`? I suppose `assert()` is still fine.

src/axom/slam/policies/ValuePolicies.hpp
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * \tparam Tag a value-policy tag (SizeTag / StrideTag / OffsetTag)
+ *  carrying the IntType, default value and validity predicate.
+ */
+template <typename Tag>
+struct RuntimeValue
+{
+public:
+  using TagType = Tag;
+
```
BradWhitlock, line 105: Maybe add a `using IntType = decltype(Tag::defaultValue())` or make the tags remember their `IntType` so you could write `Tag::IntType`.

src/axom/slam/FieldRegistry.hpp
```diff
@@ -8,131 +8,656 @@
...
+   * \brief Adds (or replaces) a view-backed field from an `axom::ArrayView`.
+   *
+   * \param key The field name.
+   * \param theSet Pointer to the set associated with the field (must outlive the map).
+   * \param data View of externally-owned storage (must outlive the map).
+   * \return A mutable reference to the stored field.
+   *
+   * \note If an entry with the same key already exists, it is overwritten.
+   */
+  ViewMapType& addField(KeyType key, const SetType* theSet, axom::ArrayView<DataType> data)
```
BradWhitlock: Is KeyType pass by value here because you later move from it? KeyType here is `std::string`. Some other places you use `std::string_view`

src/axom/slam/FieldRegistry.hpp
```diff
@@ -8,131 +8,656 @@
...
+   * \param key Field name.
+   * \return A const reference to the stored field.
+   *
+   * \pre `findField(key)` has a value (the key exists and is owning, not view-backed).
+   * \note In debug builds, this asserts if the key is missing or is view-backed.
+   * \note For view-backed fields, use `getFieldView()`. Requesting an owning field for a
+   *  view-backed key is a precondition violation: it asserts in debug builds, and in release
+   *  builds the underlying `std::get` throws `std::bad_variant_access`.
+   */
+  const MapType& getField(std::string_view key) const
```
BradWhitlock, line 206: Here `std::string_view` is used. That might be better than using pass by value `KeyType` as is done in some other methods.

src/axom/slam/RelationBuilders.hpp
```diff
@@ -0,0 +1,561 @@
...
+ *   Rel r(Rel::RelationBuilder()
+ *           .fromSet(&from).toSet(&to)
+ *           .begins (Rel::RelationBuilder::BeginsSetBuilder ().size(off.size()).data(&off))
+ *           .indices(Rel::RelationBuilder::IndicesSetBuilder().size(idx.size()).data(&idx)));
+ * \endcode
+ *
+ * make_variable_relation collapses that to one call, deducing FromSet and ToSet from the set pointers:
+ *
+ * \code
+ *   auto r = slam::make_variable_relation(&from, &to, offsets, indices);
```
BradWhitlock, line 30: This style is easier to understand.

BradWhitlock (approved):
Thanks @kennyweiss . I had a couple minor questions/comments. Overall fine - lots of formatting and comment changes!

rhornung67 (approved):

src/axom/core/Optional.hpp
```diff
@@ -0,0 +1,88 @@
...
+{
+/*!
+ * \class Optional
+ * \brief Host/device "maybe a value of type T".
+ *
+ * \tparam T The value type. Intended for small, trivially-copyable types
+ *  that are safe to capture and use in device kernels.
+ */
+template <typename T>
+struct Optional
```
publixsubfan: IIRC, we can use `std::optional` in CUDA device code with the `--expt-relaxed-constexpr` flag, and I think that behavior is the default in HIP as well. Would that be preferable from a maintainability standpoint?

src/axom/slam/policies/IndirectionPolicies.hpp
```diff
@@ -81,68 +77,64 @@ struct IndexedIndirection : public BasePolicy
...
-    PositionType pos = 0)
-  {
-    return &buf[pos];
-  }
-
-  template <bool DeviceEnable = BasePolicy::DeviceAccessible>
-  AXOM_HOST_DEVICE static inline std::enable_if_t<!DeviceEnable, ResultPtr> getIndirection(
-    IndirectionRefType buf,
-    PositionType pos = 0)
+  AXOM_HOST_DEVICE static inline ResultPtr getIndirection(IndirectionRefType buf, PositionType pos = 0)
```
publixsubfan, line 80: I think I was the original author of this SFINAE atrocity, very nice that we can now get rid of it 👍

src/axom/slam/FieldRegistry.hpp
```diff
@@ -8,131 +8,656 @@
...
+
+  /// \brief Storage for one named field, either owning or view-backed.
+  using FieldStorageType = std::variant<MapType, ViewMapType>;
+
+  // Transparent comparator (std::less<>) enables heterogeneous lookup: a
+  // std::string_view (or const char*) can be used as a key without allocating.
+  using DataVecMap = std::map<KeyType, FieldStorageType, std::less<>>;
+  using DataBufferMap = std::map<KeyType, BufferType, std::less<>>;
+  using DataViewBufferMap = std::map<KeyType, ViewBufferType, std::less<>>;
+  using DataAttrMap = std::map<KeyType, DataType, std::less<>>;
```
publixsubfan, line 83: While we're revisiting this, thoughts on making these `std::unordered_map`, either in this PR or in a separate one?

src/axom/slam/FieldRegistry.hpp
```diff
@@ -8,131 +8,656 @@
...
+    verifyFieldViewKey(key);
+    return std::get<ViewMapType>(m_maps.find(key)->second);
+  }
+
+  /*!
+   * \brief Finds a view-backed field by name without inserting or asserting.
+   *
+   * \return An optional referencing the field if present, else empty.
+   */
+  [[nodiscard]] std::optional<std::reference_wrapper<ViewMapType>> findFieldView(std::string_view key)
```
publixsubfan: I'm a little leery of that `std::optional<std::reference_wrapper<...>>` return type. From my understanding, `slam::Map`s are already view-like types. Could we just return `std::optional<ViewMapType>` here instead?

src/axom/slam/FieldRegistry.hpp
```diff
@@ -8,131 +8,656 @@
...
-    return m_buff[key];
+    return m_buff.find(key)->second;
+  }
+
+  /*!
+   * \brief Finds an owning buffer by name without inserting or asserting.
+   *
+   * \return An optional referencing the buffer if present, else empty.
+   */
+  [[nodiscard]] std::optional<std::reference_wrapper<BufferType>> findBuffer(std::string_view key)
```
publixsubfan, line 435: Similar question here for this type.

src/axom/slam/RelationBuilders.hpp
```diff
@@ -0,0 +1,561 @@
...
+ * \param indices the flat to-set indices
+ * \return a StaticRelation with VariableCardinality and STLVector indirection
+ *
+ * \pre begins.size() == fromSet->size() + 1
+ */
+template <typename FromSet,
+          typename ToSet,
+          typename PosType = typename FromSet::PositionType,
+          typename ElemType = typename ToSet::PositionType>
+auto make_variable_relation(FromSet* fromSet,
```
publixsubfan, line 137: Nit: I thought our naming convention for free functions was camelCase as well?

src/axom/slam/SetBuilders.hpp
```diff
@@ -0,0 +1,157 @@
...
+///  supplied explicitly as the leading template argument.
+/// \{
+
+/*!
+ * \brief Make a contiguous range set \f$[0, size)\f$.
+ * \param size the number of elements
+ * \return a RangeSet<PosType, ElemType>
+ */
+template <typename PosType = DefaultPositionType, typename ElemType = DefaultElementType>
+RangeSet<PosType, ElemType> make_range_set(axom::type_identity_t<PosType> size)
```
publixsubfan, line 64: Naming convention here as well?

publixsubfan (approved):
LGTM 👍

src/axom/core/Optional.hpp
kennyweiss: I'll try to make that work

src/axom/slam/FieldRegistry.hpp
kennyweiss: I'll try to make that work

src/axom/slam/RelationBuilders.hpp
kennyweiss: I felt that this was akin to `std::make_pair`  or `std::make_unique`, 
so I was going for something along the lines of the exception from the note here: https://axom.readthedocs.io/en/develop/docs/sphinx/coding_guide/sec02_names.html#function-name-format

src/axom/slam/FieldRegistry.hpp
kennyweiss: Thanks -- I agree that this should also be a `string_view` instead of a `KeyType`

src/axom/core/Macros.hpp
kennyweiss: I think so -- it should work in `constexpr` environments -- so either compile time or runtime depending on how it's called

src/axom/slam/RelationBuilders.hpp
publixsubfan: Fair enough, that sounds good to me.

src/axom/core/utilities/Utilities.cpp
kennyweiss: Addressed.

src/axom/slam/policies/ValuePolicies.hpp
kennyweiss: Addressed.

src/axom/slam/FieldRegistry.hpp
kennyweiss: Done.

src/axom/slam/FieldRegistry.hpp
kennyweiss: I looked into it. `std::map<string,V,std::less<>>::find(string_view)` works without allocating in C++17, but `std::unordered_map<string,V>::find(string_view)` does not work until C++20, and we would need `find(string(key))` which allocates. Let's revisit this when we require C++20.

src/axom/slam/FieldRegistry.hpp
kennyweiss: Thanks -- I was able to remove the reference_wrappers from the view types

src/axom/slam/FieldRegistry.hpp
kennyweiss: Unfortunately, the buffers and maps would copy without the reference_wrapper

src/axom/slam/FieldRegistry.hpp
kennyweiss: Fixed throughout.

src/axom/core/Optional.hpp
kennyweiss: Thanks for the suggestion @publixsubfan -- I was able to remove `axom::Optional` in favor of `std::optional` in device contexts!

src/axom/slam/SetBuilders.hpp
kennyweiss: discussed and resolved in a separate comment -- these are in the style of functions like `std::make_unique`

PR #1875, opened 2026-05-27
Enhance SamplingShaper so it can work on low order Blueprint meshes

BradWhitlock:
This PR resolves #1692 so the `quest::SamplingShaper` can operate on Blueprint low order meshes.

* Moved `SamplingShaper` MFEM state into a state object that can be used among helper routines
* Added a `BlueprintState` object to support Blueprint state
* Enhanced various methods to operate on both types of state, templating as needed
* Separated _"shaping_helpers.{cpp,hpp}"_ into MFEM and Blueprint files
* Quadrature enhancements: an enum and some more methods to generate quadrature points
* Blueprint sampling will generate sample points at quadrature as with HO and then sample shapes at those points, ultimately converting to material fields on the Blueprint mesh
* Changed CMake logic so `SamplingShaper` can be built when either MFEM or Conduit (or both) are present -- enables a shaper that works without MFEM
* When MFEM is present, we can shape MFEM contours onto Blueprint mesh in 2D
* Enhanced _shaping_driver_ so it can work with blueprint only mode, currently selected via inline blueprint mesh
* Added a test of Blueprint-only mode for STL spheres
* Made IntersectionShaper use BlueprintState where needed
* Enhancements that allow shaping on Blueprint meshes via `conduit::Node` or via `sidre::Group`.

PR #1895, opened 2026-06-26
Avoid overflow for volume fraction calculation

### LINKED ISSUE #1871: int32 overflow leads to crash in shaping_driver.
I ran the shaping_driver and it crashed on a relatively small mesh. It seems that the problem is caused by an int32 overflow that causes a negative size. `mfem::DenseTensor` uses int32 for computing sizes and it computes its `TotalSize` using `dofs*dofs*NE`.

* Can we work around it for large sizes?
* Or, is there an MFEM that uses an "IndexType" rather than int?
* Does the `mass_integrator.AssembleEA()` call really need to work on `dofs*dofs*NE`? Or should the size dimension really be pow(dofs,1/2) or pow(dofs, 1/3)?

```
  examples/quest_shaping_driver_ex -i $AXOM_DIR/data/shaping/spheres.yaml \
    --method sampling --sampling inout inline_mesh --min -1.1 -1.1 -1.1 --max 1.1 1.1 1.1 --dimension 3 \
    --background-material void --caliper report --order 4 --policy seq --res 64 64 64 --quadrature-order 4

[INFO]: Creating inline box mesh of resolution [ 64 64 64] and bounding box { min:(-1.1,-1.1,-1.1); max:(1.1,1.1,1.1); range:<2.2,2.2,2.2> }
[INFO]: After loading mesh has 262,144 elements and (approximate) bounding box { min:(-1.1,-1.1,-1.1); max:(1.1,1.1,1.1); range:<2.2,2.2,2.2> }
[INFO]: ========================Sampling InOut fields for shapes========================
[INFO]: --------Processing shape 'background' of material 'void' (format 'none')--------
[INFO]: -------Processing shape 'outer_shell' of material 'steel' (format 'stl')--------
[INFO]: --------Processing shape 'inner_ball' of material 'void' (format 'stl')---------
[INFO]: ================Generating volume fraction fields for materials=================
NE=262144
dofs=125
sz=-198967296
```

Image

BradWhitlock:
This is a follow-on for #1875 _(merged into develop now)_.

This PR resolves #1871 by:

* Changing MFEM shaping helper `shaping::computeVolumeFractionsForMaterial` so it works on groups of elements rather than all elements at once when it looks like the number of values will overflow int32
* Passes execution policy to `shaping::computeVolumeFractionsForMaterial` so we can parallelize for OpenMP some of the more expensive algorithm phases

Shaping driver varying mesh size for spheres.yaml input with order 4 output
image


src/axom/quest/detail/shaping/shaping_helpers_mfem.hpp
kennyweiss, line 228: Minor: Please update the doxygen for this function

src/axom/quest/detail/shaping/shaping_helpers_mfem.cpp
```diff
@@ -33,6 +38,628 @@ class OwnedQuadratureSpace : public mfem::QuadratureSpace
...
+}
+
+/*!
+ * \brief Compute the cached and chunked mass-solve settings for this volume fraction field.
+ */
+VolumeFractionMassConfig makeVolumeFractionMassConfig(const mfem::FiniteElementSpace& fes,
+                                                      axom::ArrayView<int> sampleResolution,
+                                                      axom::numerics::QuadratureType quadratureType)
+{
+  constexpr std::int64_t MAX_CACHED_MASS_BYTES = 1LL << 30;
```
kennyweiss, line 139: Please add a comment to this line. 
Presumably, it is sized to avoid overflow w/ `int32`.

Should we upstream a concern/fix to mfem about overflow w.r.t. `int32` vs `int64` indices?

src/axom/quest/detail/shaping/shaping_helpers_mfem.cpp
```diff
@@ -33,6 +38,628 @@ class OwnedQuadratureSpace : public mfem::QuadratureSpace
...
+  case RuntimePolicy::cuda:
+  #endif
+  #if defined(AXOM_RUNTIME_POLICY_USE_HIP)
+  case RuntimePolicy::hip:
+  #endif
+  #if defined(AXOM_RUNTIME_POLICY_USE_CUDA) || defined(AXOM_RUNTIME_POLICY_USE_HIP)
+    SLIC_WARNING_ROOT(
+      axom::fmt::format("MFEM volume fraction processing for '{}' uses host data and currently "
+                        "falls back to sequential execution for device runtime policies.",
+                        vfName));
```
kennyweiss, line 77: *Should* we warn the user about this?  
Our implementation (in Axom/MFEM) is only on the CPU, so there's nothing they can do about it.

If we do want to post a warning, perhaps it would be sufficient to only warn once in a given run?
E.g. using `SLIC_WARNING_ROOT_ONCE`
https://github.com/llnl/axom/blob/12695a3a9731a968aba51f069a8e0866257081fb/src/axom/slic/interface/slic_macros.hpp#L237-L255

src/axom/quest/detail/shaping/shaping_helpers_mfem.cpp
```diff
@@ -33,6 +38,628 @@ class OwnedQuadratureSpace : public mfem::QuadratureSpace
...
+  {
+  case RuntimePolicy::seq:
+    applyFCTProjectionImpl<seq_exec>(massMat, rhs, dofs, numElements, vfData, scratchBuffer);
+    break;
+  #if defined(AXOM_RUNTIME_POLICY_USE_OPENMP)
+  case RuntimePolicy::omp:
+    applyFCTProjectionImpl<omp_exec>(massMat, rhs, dofs, numElements, vfData, scratchBuffer);
+    break;
+  #endif
+  default:
```
kennyweiss, line 341: ~~I'm reasonably sure that the FCT ran on device.~~

**Update:** Nevermind -- this never ran on device
https://github.com/llnl/axom/blob/12695a3a9731a968aba51f069a8e0866257081fb/src/axom/quest/detail/shaping/shaping_helpers_mfem.cpp#L694-L699

kennyweiss (commented):
Thanks @BradWhitlock 

It's a bit difficult to tell from the github review page, but it looks like the shaping_helpers_mfem no longer runs on device, where previously several parts (but not all of it) ran on device, when using a `hip` or `cuda` policy.
Can you please run this in a GPU config/policy and confirm that we're not losing functionality? The speedups were significant in some of my runs, and getting the memory transfers correct/working was challenging.

**Update:**  I don't think any of our hip/cuda configs are currently using `mfem::Device` functionality.  Outside of this PR, we should enable/exercise mfem on-device for our shaping problems in our hip/cuda configs.
I created this issue to track this: https://github.com/llnl/axom/issues/1966

src/axom/quest/detail/shaping/shaping_helpers_mfem.cpp
kennyweiss: Perhaps it would be sufficient (or at least an improvement) for mfem to check for and give a warning about overflow?

PR #1898, opened 2026-06-29
Convert to `#pragma once`

white238:
Converts all headers to `#pragma once`. There were a couple places that were missing guards. I also didn't touch `src/thirdparty` and `src/cmake`.


rhornung67:
@white238 `#pragma once` isn't an C++ standard feature. Granted, it's cleaner than the macro guard approach, but could this cause a problem in some circumstances; e.g., sym links, moving files during the build-install process?

src/axom/core/numerics/Matrix.hpp
```diff
@@ -15,8 +15,7 @@
 #include <cstring>
 #include <iostream>
 
-#ifndef AXOM_MATRIX_HPP_
-  #define AXOM_MATRIX_HPP_
+#pragma once
```
kennyweiss: This predates the current PR, but the `#pragma once` should be above the includes in this file.

src/axom/inlet/inlet_utils.hpp
```diff
@@ -12,8 +12,7 @@
 #include "axom/core/utilities/StringUtilities.hpp"
 #include "axom/core/Path.hpp"
 
-#ifndef INLET_UTILS_HPP
-  #define INLET_UTILS_HPP
+#pragma once
```
kennyweiss: Same here -- please move the `#pragma once` to the top of the file

src/axom/sidre/core/AttrValues.hpp
```diff
@@ -27,8 +27,7 @@
 #include "axom/sidre/core/Attribute.hpp"
 #include "axom/sidre/core/SidreTypes.hpp"
 
-#ifndef SIDRE_ATTRVALUES_HPP_
-  #define SIDRE_ATTRVALUES_HPP_
+#pragma once
```
kennyweiss: Another `#pragma once` that should be moved to the top line

src/axom/slam/examples/tinyHydro/VectorXY.hpp
```diff
@@ -8,8 +8,7 @@
 // Fri Nov 21 10:50:53 PST 2014
 #include <math.h>
 
-#ifndef	_VECTORXY_H
-#define	_VECTORXY_H 1
+#pragma once
```
kennyweiss: Fourth case of four where the `#pragma once` should be moved to the first line of code in the file (also predates this PR).

src/axom/slic/interface/c_fortran/wrapSLIC.h
```diff
@@ -12,8 +12,7 @@
  */
 // For C users and C++ implementation
 
-#ifndef WRAPSLIC_H
-#define WRAPSLIC_H
+#pragma once
```
kennyweiss: Note: These changes on shroud-generated files will be lost the next time we generate the shroud files unless we modify shroud.

kennyweiss (approved):
Thanks @white238 

This should improve consistency (as much fun as it is coming up w/ slightly inconsistent header guards for each file).

There are a few cases where the guard was a a few lines lower than it should have been. These predate your PR, but this would be a good time to fix them.

white238:
`pragma once` is supported by all major compilers at this point and removes the possibility of copy/paste issue. There is a possibility that if you mix symlinked and non-symlinked header it would include it and add a bit to the compile time. But personally i think its more likely we mess it up.

It is a very common feature at this point, more concise, and possibly speeds up builds.

Arlie-Capps (approved):
Thanks, Chris.

rhornung67 (approved):

src/axom/slic/interface/c_fortran/wrapSLIC.h
white238: yea ill revert this.

src/axom/slam/examples/tinyHydro/VectorXY.hpp
white238: fixed all these.

PR #1901, opened 2026-07-01
Add checksums to Sidre Group, View.

### LINKED ISSUE #1639: Add checksum capabilities to sidre
A user has requested that we add checksums for sidre groups and views, as this will make it easier to pinpoint when the underlying data has changed.

Presumably, we'd want to:
* compute the checksum of a `sidre::View`
* compute the checksume of a `sidre::Group` based on its contained views and groups
* output a conduit node for checksum metadata. This could be diff'd against a previous conduit node of checksums

@rhornung67 suggested that we adapt our checksums from the Kahan sums in the RAJA perf suite: 
https://github.com/LLNL/RAJAPerf/blob/cda42470851fff2b7c8e6a9b5b11ab83f33a5a07/src/common/DataUtils.cpp#L598-L622

BradWhitlock:
This PR resolves #1639 

* Adapts RAJAPerf kahan sum code into `axom::utilities`
* Adds `checksum` function for `conduit::Node`
* Adds `sidre::Group::checksum` method that makes a checksum that combines the name, groups, views.
* Adds `sidre::Group::checksum` overload that can output a `conduit::Node` tree of checksum values for the group and its descendants
* Adds `sidre::View::checksum` method that makes a checksum from the view name, data, and attributes.
* Exposes `View::checksum` and `Group::checksum` in sidre Python bindings
* Adds tests for checksum


src/axom/core/utilities/Checksum.hpp
```diff
@@ -0,0 +1,81 @@
...
+    const auto value = static_cast<CheckSum>(data(j));
+    CheckSum x = (std::abs(std::sin(j + 1.0)) + 0.5) * value;
+    CheckSum y = x - ckahan;
+    volatile CheckSum t = tchk + y;
+    volatile CheckSum z = t - tchk;
+    ckahan = z - y;
+    tchk = t;
+  }
+  return tchk;
+}
```
kennyweiss, line 51: Please add a `\tparam` note about `DataGetter` -- looks like it needs to have 
```cpp
operator[](IndexType) -> <castable to CheckSum>`
```
 and be valid for indices 0 through `len-1`

src/axom/core/utilities/Checksum.hpp
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * \return A CheckSum value for the array view.
+ */
+template <typename T>
+inline CheckSum checksum(axom::ArrayView<T> view, const ScaleFactor scaleFactor = ScaleFactor {1})
+{
+  return calculateChecksum([=](axom::IndexType i) { return static_cast<CheckSum>(view[i]); },
+                           view.size()) *
+    scaleFactor;
+}
```
kennyweiss, line 83: Please say something about `scaleFactor` -- is there a range of valid/good values for it? How is it meant to be used?

src/axom/sidre/core/ConduitMemory.cpp
```diff
@@ -5,11 +5,21 @@
...
 {
+namespace
+{
+axom::utilities::CheckSum checksumNodeMetadata(const conduit::Node& n)
+{
+  auto cs = axom::utilities::CheckSum {0};
+  cs += axom::utilities::checksum(static_cast<conduit::index_t>(n.dtype().id()), 2.0);
+  cs += axom::utilities::checksum(n.number_of_children(), 3.0);
+  cs += axom::utilities::checksum(n.dtype().number_of_elements(), 5.0);
+  return cs;
```
kennyweiss: Can you please say something about why 2, 3 and 5 (first few primes?) are used as scale factors here?

src/axom/sidre/core/ConduitMemory.cpp
```diff
@@ -285,5 +295,107 @@ const ConduitMemory& ConduitMemory::instanceForConduitId(conduit::index_t condui
...
+{
+  return axom::utilities::calculateChecksum(
+           [=](axom::IndexType i) { return static_cast<axom::utilities::CheckSum>(arr[i]); },
+           arr.number_of_elements()) *
+    scaleFactor;
+}
+
+axom::utilities::CheckSum checksum(const conduit::Node& n, bool include_name)
+{
+  auto cs = axom::utilities::CheckSum {0};
```
kennyweiss, line 311: More questions about the scale factor -- should this function take a scaleFactor parameter  (and modify it in the recursive calls)?

src/axom/sidre/core/Group.cpp
```diff
@@ -2879,6 +2879,52 @@ bool Group::importConduitTreeExternal(conduit::Node& node, bool preserve_content
...
+  if(getNumGroups() > 0)
+  {
+    conduit::Node& n_groups = n_checksum["groups"];
+    for(const auto& group : this->groups())
+    {
+      conduit::Node& n_group = n_groups[group.getName()];
+      group.checksum(n_group);
+    }
+  }
+}
```
kennyweiss: Performance concern: 

If I'm understanding this correctly, there is a lot of extra computation here:

The first call 
>   n_checksum["checksum"] = static_cast<double>(checksum());

will compute the hierarchical checksum of this group and its child views and groups, and will then recurse and do the same for each child view and each child group.  I.e. we're computing the (hierarchical) checksum of each group multiple times.

Could this be reorganized to compute the checksum of the children first and then aggregate those into the parent checksum via the cached conduit value from the child?

src/axom/sidre/core/View.cpp
```diff
@@ -2070,5 +2082,27 @@ int View::getValidAllocatorId(int allocId)
...
+
+  for(IndexType attrIdx = getFirstValidAttrValueIndex(); attrIdx != InvalidIndex;
+      attrIdx = getNextValidAttrValueIndex(attrIdx))
+  {
+    const Attribute* attr = getAttribute(attrIdx);
+    if(attr != nullptr)
+    {
+      cs += checksumNamedNode(attr->getName(), getAttributeNodeRef(attr));
+    }
+  }
```
kennyweiss: Aside/Discussion: 
Should we have an iterator for the valid/activated attributes of a view?

kennyweiss: Separate question -- should the (always) attributes affect the checksum? Would it make sense to add a flag to control whether attributes contribute to the checksum of a view?

src/axom/sidre/tests/sidre_checksum.cpp
```diff
@@ -0,0 +1,364 @@
...
+  populateConduitNode(scalarChanged, "nested", "alpha", 7);
+  EXPECT_NE(checksum, axom::sidre::checksum(scalarChanged));
+
+  conduit::Node arrayChanged;
+  populateConduitNode(arrayChanged, "nested", "alpha", 42, 9.25);
+  EXPECT_NE(checksum, axom::sidre::checksum(arrayChanged));
+
+  conduit::Node renamedChild;
+  populateConduitNode(renamedChild, "renamed_nested");
+  EXPECT_NE(checksum, axom::sidre::checksum(renamedChild));
```
kennyweiss, line 96: Would it make sense to have a defensive test that shows that equal hierarchical nodes produce the same checksum?

src/axom/sidre/tests/sidre_checksum.cpp
kennyweiss, line 1: Nice tests!

kennyweiss (approved):
Thanks for working on this @BradWhitlock 

Other than some improved documentation requests, I have concerns about the efficiency of the `Group::checksum(node)`

src/axom/sidre/core/View.cpp
BradWhitlock: For the first comment, do you mean something like this?
```
for(const Attribute* attr : view->activeAttributes()) { ... }
```

For the second comment, I added a `bool includeAttributes` argument to the checksum() methods on Group, View so there is now a mode where attributes are not considered.

src/axom/sidre/core/Group.cpp
BradWhitlock: You're right. I changed the `Group::checksum(Node &) method so it will use the intermediate view/group checksums to make the final group checksum. This saves work.

src/axom/sidre/core/ConduitMemory.cpp
BradWhitlock: I added an overload that takes a `scaleFactor`, mostly to just multiply at the end rather than to fold the scaleFactor in at each level.

src/axom/sidre/core/ConduitMemory.cpp
BradWhitlock: It was an arbitrary decision to scale the values so they contribute unequally. I added a comment.

src/axom/sidre/core/View.cpp
kennyweiss: The attribute iterator is outside the scope of this PR.

Yes, I had something like that in mind.
Whenever I've used it, I've found the attribute API to be confusing and difficult to use.
I think it can use some improvements, and iterator support would be nice.

src/axom/core/utilities/Checksum.hpp
BradWhitlock: I added comments about `scaleFactor` that describe it. It was mostly inherited from the RAJAPerf implementation. It does add a good way though to combine multiple checksums with varying weights.

src/axom/core/utilities/Checksum.hpp
BradWhitlock: I updated the docs.

src/axom/sidre/tests/sidre_checksum.cpp
BradWhitlock: Done.

rhornung67 (approved):

PR #1902, opened 2026-07-01
Simplify cmake execution policy logic

### LINKED ISSUE #1750: Centralize CMake logic about available policies for tests
@BradWhitlock pointed out in #1749 that we have a lot of repeated CMake logic in our tests related to which policies are available. We should centralize this into a utility function/macro.

BradWhitlock:
This PR resolves #1750

* Define `AXOM_EXECUTION_POLICIES` list higher up in CMake build instead of repeating logic to build lists of execution policies
* Changed `mesh_tester` example in quest so it uses the usual `seq`,`omp`,`cuda`,`hip` execution policy names to be more consistent and make it compatible with updated CMake logic
* Changed some conditional compilation logic to use `AXOM_RUNTIME_POLICY_USE_*` macros for consistency


src/axom/core/examples/CMakeLists.txt
```diff
@@ -72,12 +72,7 @@ axom_add_executable(
 
 if(AXOM_ENABLE_TESTS)
     # Run the core array performance example on N ranks for each enabled policy
-    set(_policies "seq")
-    if(RAJA_FOUND)
-        blt_list_append(TO _policies ELEMENTS "omp"  IF AXOM_ENABLE_OPENMP)
-        blt_list_append(TO _policies ELEMENTS "cuda" IF AXOM_ENABLE_CUDA)
-        blt_list_append(TO _policies ELEMENTS "hip"  IF AXOM_ENABLE_HIP)
-    endif()
+    axom_execution_policies(_policies)
```
white238: This could be called once at the base of the project then the list could be named something more useful like `AXOM_EXECUTION_POLICIES`. This shouldn't be different across components.

BradWhitlock: Good point.

BradWhitlock: I changed it

src/CMakeLists.txt
```diff
@@ -132,6 +132,16 @@ if(AXOM_ENABLE_CUDA AND ${CMAKE_VERSION} VERSION_LESS 3.18.0)
     message(FATAL_ERROR "Axom requires CMake version 3.18.0+ when CUDA is enabled.")
 endif()
 
+# Build the execution policy list once so all subdirectories can reuse it.
+set(AXOM_EXECUTION_POLICIES "seq")
```
white238: Sorry can you move this to `cmake/CMakeBasics.cmake`.... i should have been clearer

white238 (approved):

src/CMakeLists.txt
BradWhitlock: I moved it. Thanks!

kennyweiss (approved):
Thanks @BradWhitlock -- this is really clean!

PR #1903, opened 2026-07-06
Centralize units code in axom::core

### LINKED ISSUE #1689: Centralize logic related to units and conversion
We have logic in Klee, in the Quest Open Cascade reader and in the shaping tutorial related to units and conversions between them.
We should centralize this logic in core and add some tests.

See:
* https://github.com/LLNL/axom/pull/1486/files#r1907564996
* https://github.com/LLNL/axom/blob/develop/src/axom/klee/Units.hpp
* https://github.com/LLNL/axom/pull/1656#discussion_r2409014477

BradWhitlock:
This PR resolves #1689, centralizing code that deals with units.

NOTE: This PR only centralizes code for dealing with length units to avoid having multiple unit-handling schemes. The PR does not add support for other types of units _(that is future work)_.

* Moves some unit definitions and helper routines from `axom::klee` to `axom::core`
* Adapts klee so it uses the moved functions, keeping compatibility with previous logic
* Moves/adapts some unit conversion code from `quest::STEPReader` into `axom::core` and updates class
* Updates `axom::C2CReader` so it uses unit functions in `axom::core`


src/axom/klee/Units.hpp
```diff
@@ -16,87 +18,15 @@ class Proxy;
...
- * \param path the Path where the length units were found in the document.
- * Used for error reporting.
- * \return the parsed units
- * \throws KleeError if the string does not represent known
- * units
- */
+using utilities::getLengthUnitName;
+using utilities::LengthUnit;
+
 LengthUnit parseLengthUnits(const std::string &unitsAsString, const std::string &path);
```
kennyweiss: It looks like the doxygen comments were removed from this header. Please add them back.

src/axom/core/utilities/Units.cpp
```diff
@@ -0,0 +1,212 @@
...
+  message += unitsAsString;
+  return message;
+}
+}  // namespace
+
+LengthUnit getLengthUnit(const std::string &unit)
+{
+  const std::string lowerUnit = toLower(unit);
+
+  static const std::unordered_map<std::string, LengthUnit> UNITS_BY_NAME {
```
kennyweiss, line 47: It looks like this performs the equivalent of the STEP reader's `getCanonicalUnit`. Is it worth mentioning that in the doxygen comment, or would you consider that an artifact not worth preserving.
(Probably the latter).

kennyweiss (approved):
Thanks @BradWhitlock -- this is really clean!

src/axom/klee/Units.hpp
BradWhitlock: I moved the original file into `src/axom/core/utilities` _(before refactoring)_ and then this one came back when a compatibility function was needed to preserve the exception interface. I added doxygen comments for this `parseLengthUnits` function.

src/axom/core/utilities/Units.cpp
BradWhitlock: I added a comment in the cpp file indicating that the code was moved from elsewhere in Axom. This code was moved from `STEPReader::getCanonicalUnit`.

rhornung67 (approved):
Thank you @BradWhitlock Nice code consolidation improvements!

PR #1904, opened 2026-07-07
More slam modernization, including traits classes and aliases

kennyweiss:
# Summary

- This PR is a follow-up to #1892 and continues to modernize slam in preparation for C++20
- It adds traits classes to slam -- ``is_set_like``, ``is_relation_like`` and ``is_map_like`` -- which will map to concepts
- It makes `StaticRelation` and `VariableRelation` trivially copyable by inheriting from `ConcreteInterface` instead of `VirtualInterface`
- It makes `axom::Array` and `axom::ArrayView` the default backing buffer type for slam's sets, relations and maps. The default was previously `std::vector`, so this might require some minor changes from users. The `std::vector` backing is still available through an explicit `STLVectorIndirection`  policy
- It adds alias types for common set and relation configurations. This greatly simplifies the setup for common cases. I have updated many examples and usages of slam throughout Axom to this type
- This branch contains lots of documentation updates, and reworks the basic example in the user docs (`UserDocs.cpp`)

### DIFF
--- a/src/axom/bump/views/dispatch_material.hpp
+++ b/src/axom/bump/views/dispatch_material.hpp
@@ -12,6 +12,8 @@
 
 #include <conduit/conduit_blueprint.hpp>
 
+#include <type_traits>
+
 namespace axom
 {
 namespace bump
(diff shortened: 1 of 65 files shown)


src/axom/bump/views/dispatch_material.hpp
kennyweiss, line 1: The changes to bump are unrelated to this branch and were required to fix the MSVC compilation.

src/axom/mir/reference/MIRMesh.cpp
kennyweiss, line 1: Note: The changes to MIR in this branch are to the unused "reference" subdirectory, which we're keeping until we can incorporate some missing features.

BradWhitlock (approved):
Thanks @kennyweiss. It looks like types are easier to write out now. Nice doc improvements too.

jcs15c (approved):
Looks good to me! As someone who wasn't super familiar with Slam before, I appreciate the simplified interface and the thorough documentation.

PR #1906, opened 2026-07-08
Adds efficient point containment predicate to primal::Sphere

### LINKED ISSUE #1815: primal: Add method for checking if a sphere contains a point
The current method is to check if the signed distance is less than (or equal to?) zero. But there's probably a faster way avoiding square roots.

kennyweiss:
# Summary

- This PR is a feature with a bonus bugfix
- It adds an efficient point containment predicate `primal::Sphere::contains(Point, includedBoundary)`, which avoids the unnecessary sqrt call from `Sphere::computeSignedDistance()` and `Sphere::getOrientation()`
  - It uses this function within a function in NURBSPatch that was using the (less efficient) `computeSignedDistance()` to test for containment
- Bonus bugfix: `Sphere<T,DIM>::getVolume()` was hard-coded for the 3D volume.
  - This PR fixes this with the appropriate 2D and 3D volume.

- Resolves https://github.com/llnl/axom/issues/1815


adayton1 (approved):

adayton1:
Thanks, @kennyweiss !

rhornung67 (approved):

PR #1907, opened 2026-07-08
Adds const overloads to data access functions in sidre::View and sidre::Buffer

### LINKED ISSUE #1695: Const-correctness: `sidre::View::getData<T>` should be marked `const`
This function returns a copy of the data in the sidre view and does not modify the data, so it should be marked `const`

I tried to call `getData()` from a `const View` and could not since the `const` was missing.

See: https://github.com/LLNL/axom/blob/9fb22cd277211d8a2092558c915aa574b709a722/src/axom/sidre/core/View.hpp#L929-L940

kennyweiss:
# Summary

- This PR addresses const-correctness issues in `sidre::View` and `sidre::Buffer` allowing usage from a `const View` or `const Buffer`
  - It adds `const` overloads to `sidre::View::getData()`  (and fixes Python bindings)
  - It marks `sidre::View::getAttributeScalar()` as const
  - It adds `const` overloads to `sidre::Buffer::getData()` and `sidre::Buffer::getVoidPtr()`
  - The const overloads all use `const_cast` (as did `View::getVoidPtr()`) since the const is on the container (`View` or `Buffer`) not on the wrapped data.
- Resolves #1695


rhornung67 (approved):
Thanks @kennyweiss Nice improvements!

BradWhitlock (approved):
Looks good to me.

nselliott (approved):

PR #1908, opened 2026-07-08
Non-Collective Root Communicator Managed MPI Buffer

gberg617:
This PR fixes a bug for the NonCollectiveRootCommunicator class where messages it sends may be deleted prior to the send being complete.  The NonCollectiveRootCommunicator is used to try to send non-blocking point-to-point messages to the root rank when we know that the job will abort due to an error.  In Lumberjack, these sorts of lines in `pushMessagesOnce()` and `pushMessagesFully()` may lead to cases where the messages are deleted before the send completes:
``` 
    m_communicator->push(packedMessagesToBeSent, receivedPackedMessages);

    if(!m_communicator->isOutputNode() && !isPackedMessagesEmpty(packedMessagesToBeSent))
    {
      delete[] packedMessagesToBeSent;
    }
```
It's important to keep the MPI sends as non-blocking for this communicator in case the root rank is not able to receive the message.  However, we want to prevent cases where the message finishes sending after it's deleted because this will lead to the root rank not receiving a proper message.

The fix being proposed here only applies to the NonCollectiveCommunicator's `push()` method, and involves copying the message contents into a buffer owned by the NonCollectiveCommunicator stored in a vector of pending messages.  Messages will only be sent by this owned communicator, and will be removed from the list of pending messages once we can confirm that the message sending completed.  Note that this assumes that buffer memory is intentionally leaked in `finalize()` in order to ensure received messages have not been deleted.

NOTE: PR is Co-authored by Codex


src/axom/lumberjack/NonCollectiveRootCommunicator.hpp
```diff
@@ -140,11 +143,21 @@ class NonCollectiveRootCommunicator : public axom::lumberjack::Communicator
...
 
 private:
+  struct PendingSend
+  {
+    MPI_Request request;
+    std::unique_ptr<char[]> buffer;
+  };
+
+  void drainCompletedSends();
+  void releasePendingSends();
```
bmhan12, line 168: Please add descriptions of what these functions will do.

src/axom/lumberjack/NonCollectiveRootCommunicator.cpp
```diff
@@ -85,11 +92,61 @@ void NonCollectiveRootCommunicator::push(const char* packedMessagesToBeSent,
...
+      PendingSend pendingSend;
+      pendingSend.request = MPI_REQUEST_NULL;
+      pendingSend.buffer.reset(new char[messageSize + 1]);
+      std::memcpy(pendingSend.buffer.get(), packedMessagesToBeSent, messageSize + 1);
+
+      pendingSend.request =
+        mpiNonBlockingSendMessagesWithRequest(m_mpiComm, 0, pendingSend.buffer.get());
+      m_pendingSends.push_back(std::move(pendingSend));
+    }
+    drainCompletedSends();
```
bmhan12: Are two drain calls necessary here?

src/axom/lumberjack/NonCollectiveRootCommunicator.cpp
```diff
@@ -26,6 +28,7 @@ namespace lumberjack
 {
 void NonCollectiveRootCommunicator::initialize(MPI_Comm comm, int ranksLimit)
 {
+  m_pendingSends.clear();
```
bmhan12: Is this necessary? Shouldn't the pending send buffer start in a valid state?

bmhan12 (commented):
I think this change makes sense, only concern might be if there is significant overhead you've observed with maintaining a buffer of messages.

src/axom/lumberjack/NonCollectiveRootCommunicator.cpp
gberg617: It's not strictly necessary. I think the second one is more crucial than the first, so I removed the initial call to drain the complete sends.

src/axom/lumberjack/NonCollectiveRootCommunicator.hpp
gberg617: I added some descriptions.

src/axom/lumberjack/NonCollectiveRootCommunicator.cpp
gberg617: Yup, you're right.  I removed this line.

gberg617:
> I think this change makes sense, only concern might be if there is significant overhead you've observed with maintaining a buffer of messages.

I'm not particularly concerned about added overhead.  This buffer is only used for abort scenarios where we want to try to pass some messages to the root rank prior to the process being terminated.  Usually those messages are not very large, and there aren't very many of them, so there's not much overhead.

bmhan12 (approved):

rhornung67 (approved):

bmhan12 (commented):
@lc-hubcast approve

PR #1909, opened 2026-07-09
Fixes InOutOctree samping on surface boundary

### LINKED ISSUE #611: Improve handling of points on surface for InOutOctree
The following simple case for In/Out  querying on a unit square in 2D had some ambiguous query points:

```
>./examples/quest_containment_driver_ex                               \
                     -i ../data/contours/unit_square.contour          \
                     --min -.234 -.43                                 \
                     --max 1.58 1.715                                 \
                     -v -n3 -l3
```
This is a unit square with 3 edges per side (`-n3`), and the query points lie within 
the bounding box `Box( min(-.234,-.43), max(1.58, 1.715))`

When querying up to resolution level 3 (`-l3`) in debug mode, we get the following warning:
```
[DEBUG] Could not determine inside/outside for point (0.370667,1) on block {grid pt: (1,3); level: 2} 
```

The problem is that the query point (gray sphere at the top of the following image) lies *on* the surface (orange edge)

image

It looks like we need better handling of such cases, e.g. via an orientation test.

kennyweiss:
# Summary

- This PR is a bugfix with some minor refactoring
- It fixes the InOutOctree for samples on (or very close to) the surface boundary, which was previously inconsistent about whether such points are considered inside or outside the surface.
  - By convention, we consider these points to be inside the contained volume, and this is consistent with our winding number treatment 
  - In addition to fixing the tests, I added some regression tests for this, and compared the results to those of the winding number for 2D points on the unit square and 3D points on the unit cube
  - This also make our sample-based shaping results for inout-based sampling and winding-based sampling more consistent on the "heroic roses" dataset, so I updated the REGEX check on our regression test
- I also added a runtime switch to output debug/visualization files of the generated octrees, and wired those through the `containment_driver` and `shaping_driver` examples
- Misc: I slightly refactored the `containment_driver` and `InOutOctree` to replace SFINAE w/ `if constexpr`, and to use a variant for the 2D and 3D octrees, since we can only have one of these within the containment driver application
- Resolves https://github.com/llnl/axom/issues/611
- Companion PR to axom_data repository: https://github.com/llnl/axom_data/pull/39


src/axom/quest/InOutOctree.hpp
```diff
@@ -1253,6 +1307,19 @@ typename std::enable_if<TDIM == 2, bool>::type InOutOctree<DIM>::withinGrayBlock
...
+  // one of the block's segments. Such points are 'within' the surface by convention,
+  // but the ray-orientation test below is degenerate for them.
+  for(int i = 0; i < numSegments; ++i)
+  {
+    if(primal::squared_distance(queryPt, m_meshWrapper.cellPositions(segmentSet[i])) <=
+       m_vertexWeldThresholdSquared)
+    {
+      return true;
+    }
+  }
```
kennyweiss, line 1344: This is the 2D bugfix

src/axom/quest/InOutOctree.hpp
```diff
@@ -1126,6 +1166,19 @@ typename std::enable_if<TDIM == 3, bool>::type InOutOctree<DIM>::withinGrayBlock
...
+  // one of the block's triangles. Such points are 'within' the surface by convention,
+  // but the ray-orientation test below is degenerate for them.
+  for(int i = 0; i < numTris; ++i)
+  {
+    if(primal::squared_distance(queryPt, m_meshWrapper.cellPositions(triSet[i])) <=
+       m_vertexWeldThresholdSquared)
+    {
+      return true;
+    }
+  }
```
kennyweiss, line 1205: This is the 3D bugfix

src/axom/quest/examples/containment_driver.cpp
```diff
@@ -603,8 +610,12 @@ int main(int argc, char** argv)
 
   const bool is2D = params.isInput2D();
 
-  ContainmentDriver<2> driver2D;
-  ContainmentDriver<3> driver3D;
+  using DriverVariant = std::variant<ContainmentDriver<2>, ContainmentDriver<3>>;
+  DriverVariant driver;
+  if(!is2D)
+  {
+    driver.emplace<ContainmentDriver<3>>();
```
adayton1, line 617: Where does the 2D driver get initialized?

src/axom/quest/InOutOctree.hpp
```diff
@@ -1105,13 +1128,30 @@ typename InOutOctree<DIM>::CellIndexSet InOutOctree<DIM>::leafCells(const BlockI
   return m_grayLeafToElementRelationLevelMap[leafBlk.level()][leafData.dataIndex()];
 }
 
+template <int DIM>
+bool InOutOctree<DIM>::withinGrayBlock(const SpacePt& queryPt,
```
adayton1, line 1158: Can't you just have two specializations of withinGrayBlock instead of separately named functions?

```
template <>
bool InOutOctree<2>::withinGrayBlock(...)
{
  // Implementation
}

template <>
bool InOutOctree<3>::withinGrayBlock(...)
{
  // Implementation
}
```

src/axom/quest/tests/quest_inout_octree.cpp
```diff
@@ -234,6 +234,146 @@ TEST(quest_inout_octree, tetrahedron_mesh)
...
+  axom::Array<Point3D> V {Point3D {0., 0., 0.},
+                          Point3D {1., 0., 0.},
+                          Point3D {1., 1., 0.},
+                          Point3D {0., 1., 0.},
+                          Point3D {0., 0., 1.},
+                          Point3D {1., 0., 1.},
+                          Point3D {1., 1., 1.},
+                          Point3D {0., 1., 1.}};
+
+  // 12 triangles (2 per face), wound CCW as seen from outside (outward normals).
```
adayton1: I would spell out counter clockwise instead of using CCW since that's not really a standard acronym.

src/axom/quest/tests/quest_inout_octree.cpp
```diff
@@ -234,6 +234,146 @@ TEST(quest_inout_octree, tetrahedron_mesh)
...
+      }
+    }
+    return std::lround(wn) != 0;
+  };
+
+  // Adds some hand-picked on-surface points: face centers, edge midpoints, corners,
+  // and points on the shared diagonal edge between the two triangles of a face.
+  axom::Array<SpacePt> onSurface {
+    SpacePt {0.5, 0.5, 0.0},
+    SpacePt {0.5, 0.5, 1.0},  // bottom/top face centers
```
adayton1: I think this comment needs to be moved up one line.

adayton1: Or better yet each point should get its own comment about where it lies. That will make any maintenance updates in the future much easier.

src/axom/quest/tests/quest_inout_octree.cpp
```diff
@@ -234,6 +234,146 @@ TEST(quest_inout_octree, tetrahedron_mesh)
...
+                          Point3D {1., 1., 0.},
+                          Point3D {0., 1., 0.},
+                          Point3D {0., 0., 1.},
+                          Point3D {1., 0., 1.},
+                          Point3D {1., 1., 1.},
+                          Point3D {0., 1., 1.}};
+
+  // 12 triangles (2 per face), wound CCW as seen from outside (outward normals).
+  axom::Array<std::tuple<int, int, int>> TRI {{0, 3, 2},
+                                              {0, 2, 1},  // z=0 bottom
```
adayton1: Should all these comments be moved up one line? Generally comments apply to the current line and below, but it seems like these apply to the line above?

src/axom/quest/tests/quest_inout_octree.cpp
```diff
@@ -234,6 +234,146 @@ TEST(quest_inout_octree, tetrahedron_mesh)
...
+  }
+
+  // Run the on-surface comparisons
+  for(const auto& q : onSurface)
+  {
+    EXPECT_TRUE(expectedWithin(q)) << "Oracle: point " << q << " should be on/within the surface";
+    EXPECT_TRUE(octree.within(q)) << "On-surface point " << q << " should be within the surface";
+  }
+
+  // Sanity check for several interior and exterior query points
```
adayton1: Can you move the portion of the comment about exterior query points down to the loop over exterior query points?

src/axom/quest/tests/quest_inout_quadtree.cpp
```diff
@@ -189,12 +189,122 @@ TEST(quest_inout_quadtree, circle_mesh)
...
+    SLIC_ASSERT(t >= 0. && t <= 1.);
+    return SpacePt::lerp(poly[edge], poly[(edge + 1) == poly.numVertices() ? 0 : edge + 1], t);
+  };
+
+  const Polygon2D unitSquare({SpacePt {0., 0.}, SpacePt {1., 0.}, SpacePt {1., 1.}, SpacePt {0., 1.}});
+
+  // Create mesh of unit square, with several edge refinements
+  for(int segsPerSide : {1, 2, 3, 5})
+  {
+    // Build the perimeter vertices (CCW; corners not duplicated).
```
adayton1: Spell out counter-clockwise

src/axom/quest/tests/quest_inout_quadtree.cpp
```diff
@@ -189,12 +189,122 @@ TEST(quest_inout_quadtree, circle_mesh)
...
+
+    // We will use the winding number over the polygon as an oracle
+    Polygon2D poly;
+    for(const auto& v : verts)
+    {
+      poly.addVertex(v);
+    }
+
+    // Create a set of query points on the boundary, starting from unit square vertices and edge midpoints
+    axom::Array<SpacePt> queryPoints {SpacePt {0.5, 1.0},
```
adayton1: Can you label where each query point lies?

src/axom/quest/examples/containment_driver.cpp
kennyweiss: Line 614 -- std::variant initializes the first template by default.
(you can use `std::monostate` if you want to avoid default initializations -- https://en.cppreference.com/cpp/utility/variant/monostate )

adayton1: Good to know, thanks!

src/axom/quest/InOutOctree.hpp
kennyweiss: I don't think the above would work directly since it would be specializing a member function of a templated class (which is illegal).  

A different spelling of the current idea, would use a tag parameter, e.g.:
```cpp
withinGrayBlockImpl(queryPt, blk, data, std::integral_constant<int, 2>{}) 
```
but this would need to be called from `withinGrayBlock()` using the same `if constexpr`, so it's not necessarily clearer.

I was able to remove the extra `TDIM` parameter, so it looks like this now:
```cpp
template <int DIM>
bool InOutOctree<DIM>::withinGrayBlock2D(const SpacePt& queryPt,
                                         const BlockIndex& leafBlk,
                                         const InOutBlockData& leafData) const
{
  static_assert(DIM == 2, "withinGrayBlock2D is only valid for 2D InOutOctrees");
  ...
}
```

src/axom/quest/tests/quest_inout_octree.cpp
kennyweiss: done

src/axom/quest/tests/quest_inout_octree.cpp
kennyweiss: done.

src/axom/quest/tests/quest_inout_octree.cpp
kennyweiss: done

src/axom/quest/tests/quest_inout_quadtree.cpp
kennyweiss: done

src/axom/quest/tests/quest_inout_quadtree.cpp
kennyweiss: done.

src/axom/quest/tests/quest_inout_octree.cpp
kennyweiss: done

src/axom/quest/InOutOctree.hpp
adayton1: That's better - I like it without the TDIM parameter.

adayton1 (approved):

bmhan12 (approved):

PR #1910, opened 2026-07-09
add basic skills and sandbox config

white238:
tired of copying these from Smith... 

They have been used in Smith for a while and removed any Smith specific stuff. This is for the new LC sandbox.


skills/slirp.out
kennyweiss, line 1: I think this is a temporary file and should be removed

skills/execution_policy.yaml
```diff
@@ -0,0 +1,104 @@
...
+      - "scp "
+      - "rsync "
+    action: conditional
+    enforcement:
+      - condition: "request.explicit.allow_network == true"
+      - if_true:
+          - action: allow
+      - if_false:
+          - action: block
+          - reply_message: "Blocked: network access is restricted. Add allow_network flag with justification."
```
kennyweiss, line 47: I'm not sure that everything in this file applies to external developers, and seems unnecessary to users already running within a sandbox.

We should probably whittle this down somewhat.

skills/cpp-unit-testing/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+                    DEPENDS_ON gtest blt::mpi my_component
+                    FOLDER     axom/my_component/tests)
+axom_add_test(NAME          my_parallel
+              COMMAND       my_parallel_test
+              NUM_MPI_TASKS 4)
+```
+
+If a test requires OpenMP, set `NUM_OMP_THREADS` accordingly.
+
+### CUDA-enabled unit tests
```
kennyweiss: We should have a HIP section as well.

skills/cpp-style/SKILL.md
```diff
@@ -0,0 +1,58 @@
...
+cmake --build <build_dir> --target style
+# example:
+cmake --build build --target style
+```
+
+## File structure
+
+- **License header**: Keep the existing copyright + SPDX block at the top of C++ files.
+- **Header guards**: Use `#pragma once` for headers.
+- **Doxygen file header**: Use a `/** ... */` block with `@file` and `@brief`.
```
kennyweiss: This should probably say to use the doxygen convention that's in the file `/** **/` or `/*! **/`  
(unless we want to switch to one convention globally)

skills/cpp-style/SKILL.md
```diff
@@ -0,0 +1,58 @@
...
+- **Data members**: `snake_case` preceded with `m_` (e.g., `m_foo`, `m_use_warm_start`).
+- **Macros / compile options**: `SCREAMING_SNAKE_CASE` (e.g., `AXOM_ENABLE_MPI`, `AXOM_DEBUG_PARAM`).
+- **Constants**: prefer `SCREAMING_SNAKE_CASE` when it’s part of a type’s API (e.g., `NUM_STATE_VARS`); for small local `constexpr` values, use a descriptive name that reads well at the callsite.
+
+## Comments & documentation
+
+- Prefer Doxygen-style docstrings for public APIs:
+  - `@brief` for summary
+  - `@param[in]` / `@param[in,out]` for parameters
+  - `@tparam` for template parameters
```
kennyweiss: Here too -- I think most (but not all) of axom uses `\brief` rather than `@brief`

skills/building/SKILL.md
kennyweiss, line 1: We should probably extend this to cover development outside LC, and with Windows/MSVC.  That could be in a follow-up PR though.

kennyweiss (approved):
Thanks @white238 !

I had some concerns about this being too limiting for external developers, e.g. developing outside LC, outside the sandbox and/or on windows, but we can iterate on those as needed.

Should there be an AGENTS.md? or are you assuming that people will have/use their own?

skills/execution_policy.yaml
white238: If you feel strongly about this please do this after this PR. I'm paranoid.

skills/cpp-style/SKILL.md
white238: I vote to chose one way if it is generating and move on. I am not sure the value in letting it decide. Unless we say something like follow the convention in the file?

skills/building/SKILL.md
white238: follow up pr if someone hits it.. these will evolve as they are used more.

bmhan12 (approved):

rhornung67 (approved):

PR #1911, opened 2026-07-09
improve error message when collection not registered as variant struct

chapman39:
improve error message by providing how to resolve it

This error occurs when you define regular struct collection (e.g. with `addStructArray()`) rather than a variant struct collection, but you're attempting to get the struct generically as a variant. This was previously allowed in Axom but forcing utilization of addVariantStructArray() or addVariantStructDictionary() will probably lead to cleaner results.

As far as I know, it's not possible to get a discriminator or other information in the error message because they haven't been set at that point. The discriminator is only set by the addVariant...() functions.

```cpp
using Shape = std::variant<Circle, Box>;

...

std::string testString = R"(
  shapes = {
    { kind = 'circle', radius = 2.5 }
  }
)";
Inlet inlet = createBasicInlet<axom::inlet::YAMLReader>(testString);
inlet.addStructArray("shapes", "shape collection");
inlet["shapes"].get<std::vector<Shape>>();
// error
```

solution:

```cpp
// replace addStructArray with:
inlet.addVariantStructArray<Shape>("shapes", "kind", "shape collection");
```


adayton1 (approved):

kennyweiss (approved):

chapman39:
This PR is ready to merge. I do not have permissions to merge it though.

white238 (approved):

Issue #1855, opened 2026-05-04
Require C++20 for Axom


kennyweiss:
RAJA and Umpire will require C++20 in their next release, so we'll need to require C++20 as well.
This will require:

- [x] Updating our host-configs to build against C++20
- [x] Updating our BLT_CXX standard
- [x] Updating our Axom spack package
- [x] Update BLT to latest release
- [x] Update to RAJA and Umpire releases

kennyweiss:
Resolved in #1912 by updating to release candidates. 
@bmhan12 has an outstanding PR to move to the official RAJA/Umpire/Camp releases -- #1949

PR #1912, opened 2026-07-13
Update to C++20, spack 1.2.0 and package updates

### LINKED ISSUE #1741: Update the cxx standard variant in Axom's spack package
Our spack package has a `cpp14` variant: 
https://github.com/LLNL/axom/blob/df7fef005ffb2c40284ef22d0be789304ab51935/scripts/spack/packages/axom/package.py#L95

Since Axom now requires `c++17` and will soon move to `c++20`, we should update this variant.
I think the convention is to use a `cxxstd=<year>` variant.

If we move to that option, do we need to keep the old `cpp14` variant and update the `when` logic, or can we just remove it.

bmhan12:
This PR:
- Updates default C++ standard to 20.
- Updates blt to 0.7.2
- Updates spack to 1.2.0
- Updates various package versions (see `versions.yaml` for list of changes)
  - **Note**: In order to test C++20 changes with umpire/camp/raja, the versions are tied to commits off of develop/main. Once releases are done for these packages and upstreamed to `spack-packages`, axom should update these to release versions (these changes should also be mirrored for Windows vcpkg build).
- Compiler updates
  - `rocm 6.4.3` (same), `rocm 7.2.1`, `cce@21.0.0` for toss 4 cray
  - `cuda@13.1.1` for toss 4

Relates to #1855  (still require updated umpire/raja/camp releases)
Closes #1741

Relevant spack-packages PR:
spack/spack-packages#5791


src/thirdparty/axom/sparsehash/internal/hashtable-common.h
```diff
@@ -50,8 +50,13 @@
 _START_GOOGLE_NAMESPACE_
 
 template <bool> struct SparsehashCompileAssert { };
+#if defined(_MSC_VER)
+#define SPARSEHASH_COMPILE_ASSERT(expr, msg) \
+  typedef SparsehashCompileAssert<(bool(expr))> msg[bool(expr) ? 1 : -1]
+#else
 #define SPARSEHASH_COMPILE_ASSERT(expr, msg) \
   __attribute__((unused)) typedef SparsehashCompileAssert<(bool(expr))> msg[bool(expr) ? 1 : -1]
+#endif
```
bmhan12, line 59: @kennyweiss I cherry-picked this change off of your Windows branch, was needed to get C++20 happy.

src/axom/core/FlatMap.hpp
```diff
@@ -634,7 +634,16 @@ class FlatMap : detail::flat_map::SequentialLookupPolicy<typename Hash::result_t
...
+  #if defined(AXOM_USE_CUDA)
+          int host_allocator_id = axom::detail::getAllocatorID<MemorySpace::Host>();
+          FlatMap host_map(*this, axom::Allocator {host_allocator_id});
+          host_map.rehash(count);
+
+          FlatMap rehashed_device(host_map, m_allocator);
+          this->swap(rehashed_device);
+          return;
+  #elif defined(AXOM_USE_HIP)
+    #if !defined(AXOM_GPUCC)
```
bmhan12: I needed to switch this code to host due to the following unit test failures with CUDA builds:

```
The following tests FAILED:
	 44 - core_flatmap_serial (Subprocess aborted)
	 45 - core_flatmap_for_all_serial (Subprocess aborted)

44: [ RUN      ] core_flatmap/1.copy_host_device
44: /usr/WS1/han12/axom_codex/src/axom/core/tests/core_flatmap.hpp:1144: Skipped
44: std::string is not supported in device-only memory on GCC's libstdc++
44: 
44: [  SKIPPED ] core_flatmap/1.copy_host_device (0 ms)


45: [ RUN      ] core_flatmap_for_all/2.insert_batched_with_existing
45: terminate called after throwing an instance of 'std::runtime_error'
45:   what():  CUDA error: invalid resource handle cudaLaunchKernel(func=0x5555560a3740, gridDim={1,1,1}, blockDim={256,1,1}, args=0x7fffffff8b20, sharedMem=0, stream=0x15531391afe0) /usr/WS1/han12/axom_codex/uberenv_cuda_libs/llvm-19.1.3/raja-c3e4576eccff2435c358b691e4fcad4e601a7441_develop-nptremsbtvcio3z3nbjyulymqizsbpmb/include/RAJA/policy/cuda/MemUtils_CUDA.hpp:274
2/2 Test #45: core_flatmap_for_all_serial ......Subprocess aborted***Exception:   0.51 sec

```

src/axom/sidre/tests/sidre_attribute.cpp
```diff
@@ -33,7 +33,6 @@ const std::string g_color_blue("blue");
 
 const std::string g_name_animal("animal");
 const std::string g_animal_none("human");
-const std::string g_animal_cat("cat");
```
bmhan12, line 36: C++20 unused change.

src/axom/quest/GWNMethods.hpp
```diff
@@ -209,21 +209,21 @@ class NURBSCurveGWNQuery
         const int ncurves = m_processed_curves_view.size();
         axom::Array<BoxType> aabbs(ncurves, ncurves);
         auto aabbs_view = aabbs.view();
+        const auto processed_curves_view = m_processed_curves_view;
```
bmhan12, line 211: C++20 lambda capture change

scripts/spack/packages/cce/package.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2013-2025 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+import os
+from spack.package import *
+from spack_repo.builtin.packages.cce.package import Cce as BuiltinCce
+
+class Cce(BuiltinCce):
```
bmhan12, line 10: Associated spack-packages PR: https://github.com/spack/spack-packages/pull/5468

scripts/spack/packages/py-nanobind/package.py
```diff
@@ -0,0 +1,10 @@
+import os
+
+from spack.package import *
+from spack_repo.builtin.packages.py_nanobind.package import PyNanobind as BuiltinPyNanobind
+
+class PyNanobind(BuiltinPyNanobind):
+
+    # Workaround for "gmake: *** internal error: invalid --jobserver-auth string":
+    # https://github.com/spack/spack-packages/issues/5106#issuecomment-4679593276
+    depends_on("gmake", type="build")
```
bmhan12: Associated spack-packages PR: https://github.com/spack/spack-packages/pull/5318

scripts/spack/specs.json
```diff
@@ -20,27 +20,27 @@
...
     [ "+python+devtools+mfem+c2c+adiak+caliper+cuda cuda_arch=90 %gcc_13",
-      "+python+devtools+mfem+c2c+adiak+caliper+cuda cuda_arch=90 %clang_19"],
+      "+python+devtools+mfem+c2c+adiak+caliper+cuda cuda_arch=90 %%clang_19"],
 
     "__comment__":"# Use amdgpu_target=gfx942 for rzadams",
     "__comment__":"# Use amdgpu_target=gfx90a for tioga/rzvernal",
     "__comment__":"# Use amdgpu_target=gfx908 for rznevada",
     "__comment__":"# -Wno-int-conversion flag needed for building HDF5",
-    "__comment__":"# caliper disabled for rocm@6.3.1, fails to compile",
+    "__comment__":"# caliper~shared required for mfem/caliper/mpi(?) to prevent rocprof runtime deadlock",
```
bmhan12, line 39: This was required to move to newer compilers + the newer mpi that the compilers required.

Without this change, I was observing runtime hangs when executables try to load in `rocprofiler-sdk` through caliper or mfem (yet, if I disabled one of either mfem or caliper during axom configuration, I don't see the hangs).

src/axom/core/FlatMap.hpp
bmhan12: @publixsubfan

scripts/spack/packages/axom/package.py
```diff
@@ -162,6 +159,19 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
...
+        default="20",
+        values=("11", "14", "17", "20"),
+        description="C++ standard to build with",
+    )
+    # C++14 required as of 0.7.0
+    conflicts("cxxstd=11", when="@0.7.0:")
+    # C++17 required as of 0.12.0
+    conflicts("cxxstd=14", when="@0.12.0:")
+    # C++20 required as of unreleased 0.15.0 (Should be 0.15.0)
+    conflicts("cxxstd=17", when="@0.14.0:")
```
bmhan12: This logic will throw errors if you are using an older C++ standard configuration with the wrong axom version. 
I pulled these versions from the release notes.

The C++17 requirement is set to 0.14.0 here for testing, but should be 0.15.0 for the upcoming release.

rhornung67 (approved):
Thank you @bmhan12

src/CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ if ("${PROJECT_SOURCE_DIR}" STREQUAL "${CMAKE_SOURCE_DIR}")
     # Set some default BLT options before loading BLT only if not included in
     # another project
     if (NOT BLT_CXX_STD)
-        set(BLT_CXX_STD "c++17" CACHE STRING "")
+        set(BLT_CXX_STD "c++20" CACHE STRING "")
```
kennyweiss, line 49: 🎉

src/axom/core/FlatMap.hpp
kennyweiss: This should be fixed/addressed before merging. 

We don't want to silently force our `FlatMap::rehash()` to be on host for CUDA.

scripts/vcpkg_ports/umpire/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "umpire",
-  "version-string": "2025.12.0",
+  "version-string": "develop-commit",
```
kennyweiss: Is it a specific non-release commit b/c Umpire hasn't released yet?

If so, please add a comment that we're planning to update to release `2026.0?.*` when it is available

scripts/vcpkg_ports/raja/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "raja",
-  "version-string": "2025.12.0",
+  "version-string": "develop-commit",
```
kennyweiss: Same question/comment as for the Umpire port about version.

scripts/vcpkg_ports/camp/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "camp",
-  "version-string": "2025.12.0",
+  "version-string": "main-commit",
```
kennyweiss: (same comment/question about version as for Umpire/RAJA)

scripts/spack/configs/versions.yaml
kennyweiss, line 1: Can we please also explicitly set our dependency to `nanobind@2.12` in this file?

The new host-configs are already using 2.12, but it would be good to set it in this file so we can be deliberate about when we upgrade.

scripts/spack/configs/versions.yaml
```diff
@@ -5,37 +5,38 @@
...
     require:
-      - spec: "@0.9.5"
+      - spec: "@0.9.7"
   hypre:
     require:
-      - spec: "@2.27.0"
+      - spec: "@3.1.0"
   mfem:
     require:
       - spec: "@4.9"
```
kennyweiss: mfem is getting ready to release mfem@4.10. 

Since we're updating to pre-releases for camp, raja and umpire, can we also update to a prerelease of mfem?
The newer mfem has some improvements to 1D NURBS that are helpful for winding numbers, so it would be good to have that in our upcoming release.

scripts/spack/packages/axom/package.py
```diff
@@ -113,9 +113,6 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
     variant("tools", default=True, description="Build tools")
     variant("tutorials", default=True, description="Build tutorials")
 
-    # Hard requirement after Axom 0.6.1
-    variant("cpp14", default=True, description="Build with C++14 support")
```
kennyweiss, line 118: Will removing this break anyone's build/workflow? 

Would it be better to have it do nothing for a short amount of time, or to keep it here, but add an explicit error message so people know to remove it?

Another bump for deprecated variants in spack: https://github.com/spack/spack/issues/38829

.github/workflows/test_windows_tpls.yml
```diff
@@ -13,7 +13,7 @@ jobs:
   run_uberenv:
     name: Runs ${{ matrix.triplet }} ${{ matrix.cfg }} uberenv with vcpkg
     # The type of runner that the job will run on
-    runs-on: windows-latest
+    runs-on: windows-2022
```
kennyweiss, line 16: 👍

host-configs/rzvernal-toss_4_x86_64_ib_cray-llvm-amdgpu@6.4.3_hip.cmake
```diff
@@ -4,28 +4,28 @@
...
...s/toss_4_x86_64_ib_cray/2026_07_07_09_58_29/llvm-amdgpu-6.4.3/fmt-12.1.0-po24stdzkwj4obatcqfus4uszhdbfd7m;/usr/WS1/axom/libs/toss_4_x86_64_ib_cray/2026_07_07_09_58_29/none-none/compiler-wrapper-1.1.0-qqipzhe4rz5fmoiuvlqgarjeevl2kzuh;/collab/usr/gapps/axom/devtools/toss_4_x86_64_ib_cray/2026_02_17_15_20_25/view/python-3.13.11;/collab/usr/gapps/shroud/public/toss_4_x86_64_ib_cray/shroud-0.14.0;/collab/usr/gapps/axom/devtools/toss_4_x86_64_ib_cray/2026_02_17_15_20_25/view/python-3.13.11;/collab/usr/gapps/axom/devtools/toss_4_x86_64_ib_cray/2026_02_17_15_20_25/view/python-3.13.11;/usr/tce/packages/cmake/cmake-3.29.2;/collab/usr/gapps/axom/devtools/toss_4_x86_64_ib_cray/2026_02_17_15_20_25/gcc-13.3.1/cppcheck-2.18.0-n6kdcwtwlrc3u3t47t7gokpyd4h6mc27;/usr/tce/packages/cray-mpich-tce/cray-mpich-8.1.29-rocmcc-6.4.3;/collab/usr/gapps/axom/devtools/toss_4_x86_64_ib_cray/2026_02_17_15_20_25/view/doxygen-1.15.0;/opt/rocm-6.4.3;/opt/rocm-6.4.3;/usr/tce/packages/rocmcc/rocmcc-6.4.3-magic/llvm;/opt/rocm-6.4.3;/collab/usr/gapps/axom/devtools/toss_4_x86_64_ib_cray/2026_02_17_15_20_25/view/python-3.13.11;/opt/rocm-6.4.3;/opt/rocm-7.2.1;/opt/rocm-6.4.3;/opt/rocm-7.2.1;/opt/rocm-6.4.3" CACHE STRING "")
```
kennyweiss: @white238 -- should we reconsider whether we really need the `CMAKE_PREFIX_PATH`/`CMAKE_BUILD_RPATH`/`CMAKE_INSTALL_RPATH` ?  They add a lot of churn to TPL updates.

src/thirdparty/axom/sparsehash/internal/hashtable-common.h
kennyweiss: Thanks @bmhan12 

Our of scope for this PR: We should probably investigate/reconsider whether we still need to vendor `sparsehash` now that we have `std::unordered_map` and `axom::FlatMap`.

**Update:** https://github.com/llnl/axom/issues/1916

kennyweiss (changes requested):
Thanks @bmhan12 for this massive update.

The only parts I'm concerned about are:
* removing the `cpp14` variant from our spack package. It needs to go away, but we might need to deprecate it for a bit, or at least cause it to give a reasonable error message before removing it outright.
* Making `FlatMap::rehash()` run w/ the `SEQ` (host) policy for `CUDA`

src/axom/core/FlatMap.hpp
bmhan12: Interesting - reproducing this error does not require all of the changes of this branch.
It can be reproduced on develop by changing `BLT_CXX_STD` to `c++20`.https://github.com/llnl/axom/blob/4839900850b7b43bcc2accf7041e3479cea59515/src/CMakeLists.txt#L48-L50

adayton1:
Instead of removing the cpp14 variant, can't you just make it conflict with the version that requires c++17 and any version newer than it?

scripts/vcpkg_ports/umpire/vcpkg.json
bmhan12: Added a "todo" comment here and elsewhere across the changed `vcpkg.json` files.

scripts/spack/configs/versions.yaml
bmhan12: Added to `versions.yaml`!

scripts/spack/configs/versions.yaml
bmhan12: > can we also update to a prerelease of mfem?

@kennyweiss 
I updated to a commit off of mfem master (c8ef9676e83c93a9037dbdf3bf0f2d741321e632 , 7/21/26).
This did require changes to the `MFEMReader.cpp` to get unit tests passing for the nurbs feature, so a look over that would be appreciated.

scripts/spack/packages/axom/package.py
bmhan12: > Will removing this break anyone's build/workflow?

I took a closer look at `cpp14`, and it is being explicitly defined `+cpp14` for some environments in the shared recipes, so removing it would break some builds.
I've change the axom recipe to keep the variant in, and removed `cpp14` as a hard requirement like it currently is in develop: https://github.com/spack/spack-packages/blob/2523a73ecf7fb6e467a6fd9585b1caed86bc72ee/repos/spack_repo/builtin/packages/axom/package.py#L332-L333

On develop, `cpp14` is used in a narrow case where you want to build with C++14 with axom <= 0.6.1 & with blt <= 0.5.1 (with or without CUDA):
https://github.com/spack/spack-packages/blob/2523a73ecf7fb6e467a6fd9585b1caed86bc72ee/repos/spack_repo/builtin/packages/axom/package.py#L412-L413

https://github.com/spack/spack-packages/blob/2523a73ecf7fb6e467a6fd9585b1caed86bc72ee/repos/spack_repo/builtin/packages/axom/package.py#L452-L453

Other than that, the variant is a no-op, so I feel comfortable removing the hard requirement.

scripts/vcpkg_ports/hdf5/vcpkg.json
```diff
@@ -0,0 +1,77 @@
...
+  "homepage": "https://www.hdfgroup.org/downloads/hdf5/",
+  "license": "BSD-3-Clause",
+  "supports": "!uwp",
+  "$comment": [
+    "Overlay of the pinned upstream hdf5 manifest used by uberenv's in-place vcpkg ports copy.",
+    "Upstream hdf5 defaults are [\"szip\", \"zlib\"]. We intentionally drop \"szip\" from default-features here",
+    "but keep the explicit \"szip\" feature below so users can still opt in when needed.",
+    "This works around vcpkg pulling libaec through hdf5's default szip feature in Axom's Windows GHA builds.",
+    "The problem was not libaec compatibility; the relevant issue was that the libaec source download repeatedly",
+    "failed from its upstream archive URL with HTTP 429 in GitHub Actions, making Windows dependency bootstrap flaky."
```
bmhan12, line 15: Needed to add a `vcpkg.json` file for hdf5, was hitting a 429 Error when trying to pull one of hdf5's dependencies.
Error was happening on this branch, as well as develop.
It doesn't look like the dependency was needed for axom, here's the Windows CI run output:
https://github.com/llnl/axom/actions/runs/30129087714

scripts/spack/packages/py-nanobind/package.py
bmhan12: Associated spack PR has been merged, and spack packages commit updated to get the updated recipe.

scripts/spack/configs/versions.yaml
bmhan12: Here was the log of the CUDA failure in `read_patches_format_1d_nurbs` unit test:
[ RUN      ] quest_mfem_reader.read_patches_format_1d_nurbs

```
[ERROR in line 185 of file /usr/WS1/han12/axom_codex/src/axom/core/../../axom/primal/geometry/NURBSCurve.hpp]
MESSAGE=Failed Assert: isValidNURBS()

** StackTrace of 19 frames **
Frame 2: axom::primal::NURBSCurve<double, 2>::NURBSCurve(axom::ArrayView<axom::primal::Point<double, 2> const, 1, (axom::MemorySpace)1>, axom::ArrayView<double const, 1, (axom::MemorySpace)1>, axom::primal::KnotVector<double> const&)
Frame 3: axom::primal::NURBSCurve<double, 2>::NURBSCurve(axom::Array<axom::primal::Point<double, 2>, 1, (axom::MemorySpace)1, axom::detail::DefaultStoragePolicy<axom::primal::Point<double, 2> > > const&, axom::Array<double, 1, (axom::MemorySpace)1, axom::detail::DefaultStoragePolicy<double> > const&, axom::primal::KnotVector<double> const&)
Frame 4: axom::quest::internal::read_mfem(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::map<int, axom::Array<axom::primal::NURBSCurve<double, 2>, 1, (axom::MemorySpace)1, axom::detail::DefaultStoragePolicy<axom::primal::NURBSCurve<double, 2> > >, std::less<int>, std::allocator<std::pair<int const, axom::Array<axom::primal::NURBSCurve<double, 2>, 1, (axom::MemorySpace)1, axom::detail::DefaultStoragePolicy<axom::primal::NURBSCurve<double, 2> > > > > >&)
Frame 5: axom::quest::MFEMReader::read(axom::Array<axom::primal::NURBSCurve<double, 2>, 1, (axom::MemorySpace)1, axom::detail::DefaultStoragePolicy<axom::primal::NURBSCurve<double, 2> > >&, axom::Array<int, 1, (axom::MemorySpace)1, axom::detail::DefaultStoragePolicy<int> >&)
Frame 6: quest_mfem_reader_read_patches_format_1d_nurbs_Test::TestBody()
Frame 7: void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*)
Frame 8: void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*)
Frame 9: testing::Test::Run()
Frame 10: testing::TestInfo::Run()
Frame 11: testing::TestSuite::Run()
Frame 12: testing::internal::UnitTestImpl::RunAllTests()
Frame 13: bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*)
Frame 14: bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*)
Frame 15: testing::UnitTest::Run()
Frame 16: RUN_ALL_TESTS()
Frame 17: main
Frame 18: __libc_start_main
Frame 19: _start
=====


Aborted (core dumped)
```

kennyweiss: **Update:**  Fixed! It turns out that we had a comment in the mfem data file used for the test, that was not supported by the reader. Fixed by removing the comment.

src/axom/quest/io/MFEMReader.cpp
```diff
@@ -219,8 +218,15 @@ int read_mfem(const std::string &fileName,
...
+
+      using SkipTag = primal::KnotVector<double>::SkipValidityChecks;
       axom::ArrayView<const double> knots_view(&kv0[0], kv0.Size());
-      const primal::KnotVector<double> kv(knots_view, kv0.GetOrder());
+      const primal::KnotVector<double> kv(knots_view, kv0.GetOrder(), SkipTag {});
+      if(!kv.isValid(true))
+      {
+        SLIC_WARNING(axom::fmt::format("MFEM patch {} has an invalid knot vector: {}", patchId, kv));
+        return MFEMReader::READ_FAILED;
+      }
```
kennyweiss, line 229: I rolled back the changes to `MFEMReader` but added this extra check against a bad knot vector in the input file.

src/axom/quest/tests/quest_mfem_reader.cpp
```diff
@@ -453,7 +453,6 @@ dimension
 2
 
 controlpoints
-# points (0,0), (0.5,0), (1,0) with weights (1,2,1) in homogeneous form (x*w, y*w, w)
```
kennyweiss, line 456: This comment in an mfem mesh test file was the cause of the CUDA config failure -- the mfem reader does not support comments here!

kennyweiss (approved):
Thanks for this comprehensive update @bmhan12 !

PR #1918, opened 2026-07-15
Add warning as error to release build Gitlab CI jobs

bmhan12:
This PR:
- Adds warning as errors to release build type jobs for dane on Gitlab CI
- Code changes (mainly `mint`) to eliminate warnings.


src/axom/mint/execution/internal/for_all_cells.hpp
```diff
@@ -421,17 +425,18 @@ inline void for_all_cells_impl(xargs::coords, const UniformMesh& m, KernelType&&
   const double z0 = origin[2];
   const double dz = spacing[2];
 
-  // Note: gcc@10.3.1 emits a '-Warray-bounds' warning in callers of this function
-  // about the sizes of nodeIds and coords not matching due to the runtime switch on dimension
```
bmhan12, line 425: This is the primary issue these `mint` changes are resolving.

src/axom/mint/execution/internal/for_all_faces.hpp
kennyweiss: Thanks for fixing these -- they have been causing warnings for years.

.gitlab/build_dane.yml
```diff
@@ -37,34 +37,35 @@
 
 ####
 # PR Build jobs
-dane-llvm_19_1_3-werror-debug-src:
+dane-llvm_19_1_3-debug-werror-src:
```
rhornung67, line 40: Would it be a good idea to make the `BUILD_TYPE` explicit for each job?  It seems odd to me that we make the build type explicit when that is the default build type in CMake and we don't specify the build type for debug builds.

Arlie-Capps (approved):
Thank you for these fixes, @bmhan12

PR #1919, opened 2026-07-16
Adds lua support to Klee

kennyweiss:
# Summary

- This PR is a feature
- It adds Lua support to Klee and to our shaping application
  - This first PR provides feature parity with the YAML reader
  - A follow-up PR will add the ability to pass variables into the deck and to define callbacks, e.g. to have different operators/behavior in 2D vs. 3D
- Misc Klee: 
  - Fixes the slice operator
  - Adds a check that each operator can be converted to a matrix (instead of silently failing)
- Misc Core:
   - Adds a utility function to core to extract the file extension from a string and uses this throughout Axom

### Example
For this YAML input:
```yaml
    dimensions: 3
    shapes:
      - name: windshield
        material: glass
        geometry:
          format: stl
          path: windshield.stl
          units: cm
          operators:
            - rotate: 90
              axis: [0, 1, 0]
              center: [0, 0, -10]
            - translate: [10, 20, 30]
```

... we now support the equivalent Lua:
```lua
    dimensions = 3
    shapes = {
      {
        name = "windshield",
        material = "glass",
        geometry = {
          format = "stl",
          path = "windshield.stl",
          units = "cm",
          operators = {
            { rotate = 90, axis = {0, 1, 0}, center = {0, 0, -10} },
            { translate = {10, 20, 30} }
          }
        }
      }
    }
```

### DIFF
--- a/src/axom/core/utilities/FileUtilities.cpp
+++ b/src/axom/core/utilities/FileUtilities.cpp
@@ -88,6 +88,21 @@ std::string joinPath(const std::string& fileDir,
 }
 
 //-----------------------------------------------------------------------------
+std::string getFileExtension(const std::string& path)
+{
+  const auto separator = path.find_last_of("/\\");
+  const auto filename = path.substr(separator == std::string::npos ? 0 : separator + 1);
+
+  if(filename.empty() || filename == "." || filename == "..")
+  {
+    return "";
+  }
+
+  const auto dot = filename.find_last_of('.');
+  return dot == std::string::npos || dot == 0 ? "" : filename.substr(dot);
+}
+
+//-----------------------------------------------------------------------------
 int makeDirsForPath(const std::string& path)
 {
   char separator = '/';
--- a/src/axom/core/utilities/FileUtilities.cpp
+++ b/src/axom/core/utilities/FileUtilities.cpp
@@ -8,6 +8,7 @@
 #include "axom/core/utilities/StringUtilities.hpp"
 #include "axom/fmt.hpp"
 
+#include <filesystem>
 #include <vector>
 #include <string>
 #include <fstream>
@@ -90,16 +91,13 @@ std::string joinPath(const std::string& fileDir,
 //-----------------------------------------------------------------------------
 std::string getFileExtension(const std::string& path)
 {
-  const auto separator = path.find_last_of("/\\");
-  const auto filename = path.substr(separator == std::string::npos ? 0 : separator + 1);
-
-  if(filename.empty() || filename == "." || filename == "..")
-  {
-    return "";
-  }
-
-  const auto dot = filename.find_last_of('.');
-  return dot == std::string::npos || dot == 0 ? "" : filename.substr(dot);
+#if defined(WIN32)
+  return std::filesystem::path(path).extension().string();
+#else
+  std::string native_path = path;
+  std::replace(native_path.begin(), native_path.end(), '\\', '/');
+  return std::filesystem::path(native_path).extension().string();
+#endif
 }
 
 //-----------------------------------------------------------------------------
--- a/src/axom/core/utilities/FileUtilities.hpp
+++ b/src/axom/core/utilities/FileUtilities.hpp
@@ -58,6 +58,20 @@ std::string joinPath(const std::string& fileDir,
                      const std::string& separator = "/");
 
 /*!
+ * \brief Gets the extension from the final component of a filesystem path.
+ *
+ * \param [in] path an absolute or relative filesystem path
+ *
+ * \return the extension, including its leading period, or an empty string when
+ * the path has no extension
+ *
+ * The returned extension preserves its original case. Dots in directory names
+ * are ignored. An initial dot in the filename is not treated as an extension
+ * separator, so ".gitignore" has no extension.
+ */
+std::string getFileExtension(const std::string& path);
+
+/*!
  * \brief Make directories for a given path string
  *
  * \param [in] path  string representing an absolute or relative directory path
--- a/src/axom/klee/Geometry.cpp
+++ b/src/axom/klee/Geometry.cpp
@@ -7,6 +7,7 @@
 #include "axom/klee/Geometry.hpp"
 #include "axom/klee/GeometryOperators.hpp"
 #include "axom/klee/AffineMatrixVisitor.hpp"
+#include "axom/klee/KleeError.hpp"
 
 #include "conduit_blueprint_mesh.hpp"
 
@@ -16,6 +17,50 @@ namespace axom
 {
 namespace klee
 {
+namespace
+{
+std::string operatorName(const GeometryOperator& op)
+{
+  if(dynamic_cast<const Translation*>(&op))
+  {
+    return "translate";
+  }
+  if(dynamic_cast<const Rotation*>(&op))
+  {
+    return "rotate";
+  }
+  if(dynamic_cast<const Scale*>(&op))
+  {
+    return "scale";
+  }
+  if(dynamic_cast<const UnitConverter*>(&op))
+  {
+    return "convert_units_to";
+  }
+  if(dynamic_cast<const SliceOperator*>(&op))
+  {
+    return "slice";
+  }
+  return "unknown";
+}
+
+void requireMatrixOperator(const std::shared_ptr<const GeometryOperator>& op, int index)
+{
+  if(std::dynamic_pointer_cast<const MatrixOperator>(op))
+  {
+    return;
+  }
+
+  const auto ordinal =
+    index >= 0 ? axom::fmt::format("operator {}", index) : std::string {"operator"};
+  throw KleeError({Path {"geometry/operators"},
+                   axom::fmt::format("Cannot convert geometry to matrix: {} ({}) is not a "
+                                     "matrix operator.",
+                                     ordinal,
+                                     operatorName(*op))});
+}
+}  // namespace
+
 bool operator==(const TransformableGeometryProperties& lhs, const TransformableGeometryProperties& rhs)
 {
   return lhs.dimensions == rhs.dimensions && lhs.units == rhs.units;
@@ -243,14 +288,21 @@ numerics::Matrix<double> Geometry::getTransform() const
       // Why don't we multiply the matrices in CompositeOperator::addOperator()?
       // Why keep the matrices factored and multiply them here repeatedly?
       // Combining them would also avoid this if-else logic.  BTNG
+      int operatorIndex = 0;
       for(auto op : composite->getOperators())
       {
+        ++operatorIndex;
+        requireMatrixOperator(op, operatorIndex);
         // Use visitor pattern to extract the affine matrix from supported operators
         AffineMatrixVisitor visitor;
         op->accept(visitor);
         if(!visitor.isValid())
         {
-          continue;
+          throw KleeError({Path {"geometry/operators"},
+                           axom::fmt::format("Cannot convert geometry to matrix: operator {} ({}) "
+                                             "is not supported by matrix extraction.",
+                                             operatorIndex,
+                                             operatorName(*op))});
         }
         const auto& matrix = visitor.getMatrix();
         numerics::Matrix<double> res(identity4x4);
@@ -260,11 +312,19 @@ numerics::Matrix<double> Geometry::getTransform() const
     }
     else
     {
+      requireMatrixOperator(m_operator, -1);
       AffineMatrixVisitor visitor;
       m_operator->accept(visitor);
       if(visitor.isValid())
       {
         transformation = visitor.getMatrix();
+      }
+      else
+      {
+        throw KleeError({Path {"geometry/operators"},
+                         axom::fmt::format("Cannot convert geometry to matrix: operator ({}) is "
+                                           "not supported by matrix extraction.",
+                                           operatorName(*m_operator))});
       }
     }
   }
--- a/src/axom/klee/Geometry.cpp
+++ b/src/axom/klee/Geometry.cpp
@@ -19,6 +19,13 @@ namespace klee
 {
 namespace
 {
+/**
+ * Require that an operator can be represented as an affine matrix.
+ *
+ * \param op the operator to check
+ * \param index the one-based operator index, or a negative value for an unnamed single operator
+ * \throws KleeError if \a op is not a MatrixOperator
+ */
 void requireMatrixOperator(const std::shared_ptr<const GeometryOperator>& op, int index)
 {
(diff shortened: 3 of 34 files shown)


src/axom/klee/io/IO.cpp
```diff
@@ -297,20 +304,119 @@ internal::NamedOperatorMap getNamedOperators(const inlet::Inlet &doc, Dimensions
...
+{
+  switch(format)
+  {
+  case InputFormat::YAML:
+    return std::make_unique<inlet::YAMLReader>();
+  case InputFormat::Lua:
+#ifdef AXOM_USE_LUA
+    return std::make_unique<inlet::LuaReader>();
+#else
+    throw KleeError(
```
BradWhitlock, line 362: I'm pretty sure there is a JSONReader too. Is that being deprecated?

src/axom/klee/Geometry.cpp
```diff
@@ -16,6 +17,50 @@ namespace axom
 {
 namespace klee
 {
+namespace
+{
+std::string operatorName(const GeometryOperator& op)
+{
+  if(dynamic_cast<const Translation*>(&op))
```
BradWhitlock: I would eliminate this function and add a virtual `name()` method on the `op`.

src/axom/klee/io/IO.cpp
white238: nothing is being deprecated. he is just adding the ability to use Lua, which is enabled by an external library while the json/yaml readers are enabled via Conduit which is a hard requirement of Inlet

BradWhitlock (approved):
I had a couple small comments. Overall very nice. Thanks @kennyweiss!

src/axom/klee/io/IO.cpp
kennyweiss: To clarify a bit -- 
* `inlet` has readers for YAML, JSON and Lua.
* `klee` is built on `inlet`, but before this PR, it only supported YAML (by design)
* This PR add support for Lua to `klee`
* If we wanted to add support for `JSON` in the future, it would be relatively easy to do so, but we haven't had requests for that.

src/axom/core/utilities/FileUtilities.hpp
```diff
@@ -57,6 +57,20 @@ std::string joinPath(const std::string& fileDir,
                      const std::string& fileName,
                      const std::string& separator = "/");
 
+/*!
+ * \brief Gets the extension from the final component of a filesystem path.
+ *
+ * \param [in] path an absolute or relative filesystem path
+ *
+ * \return the extension, including its leading period, or an empty string when
+ * the path has no extension
```
Arlie-Capps, line 66: I note that keeping the leading period matches the behavior of .Net's System.IO.Path.GetExtension() method.  Thank you!

src/axom/klee/docs/sphinx/specifying_shapes.rst
```diff
@@ -44,12 +43,106 @@ is specified in :code:`windshield.stl`. Note that Klee does not specify
 what a particular material means. A material is simply a label which can
 be used by a host code to apply properties to part of a mesh.
 
+Lua Input Decks
+***************
+Klee can also read Lua input decks when Axom is configured with
+:code:`AXOM_ENABLE_LUA=ON` and Sol support. Lua decks use the same Klee schema
+as YAML decks. The Lua is evaluated and then read into Inlet.
```
Arlie-Capps: This comment and the discussion below imply that the Lua is evaluated one time, at read time, and then passed to Inlet.  If it makes sense to do so, please state this explicitly.  I want to address a scenario where a user thinks the Lua gets evaluated each time through some loop.

Arlie-Capps: Also, if callbacks are provided, how or when are they evaluated?  I don't see examples.  (Sorry if I just missed it.)

src/axom/klee/Geometry.cpp
```diff
@@ -16,6 +17,50 @@ namespace axom
...
+    return;
+  }
+
+  const auto ordinal =
+    index >= 0 ? axom::fmt::format("operator {}", index) : std::string {"operator"};
+  throw KleeError({Path {"geometry/operators"},
+                   axom::fmt::format("Cannot convert geometry to matrix: {} ({}) is not a "
+                                     "matrix operator.",
+                                     ordinal,
+                                     operatorName(*op))});
```
Arlie-Capps: This code snippet is illustrative of a pattern throughout Klee (at least) of throwing exceptions on error.  In this case, the error is that `op` (pointing to a GeometryOperator) cannot be made into a (pointer to a) MatrixOperator.  We need a discussion of this design choice somewhere in our documentation, because users of Axom are developers on codes that may not share "throw on error" as a design choice.  At least document the pattern so they know what might be going on.  Also, Doxygen of a function should reflect that the function throws an exception on error.

I'm not asking for a sweeping Doxygen update in this PR.  If the Sphinx docs already have discussion of "throw on error," that is great; otherwise, at some point we should write about it.

src/axom/core/utilities/FileUtilities.cpp
```diff
@@ -87,6 +87,21 @@ std::string joinPath(const std::string& fileDir,
                            fileName);
 }
 
+//-----------------------------------------------------------------------------
+std::string getFileExtension(const std::string& path)
+{
```
white238, line 93: why not use `filesystem`?

```
#include <filesystem>
#include <iostream>
#include <string>

int main()
{
    const std::filesystem::path file_path{"archive.tar.gz"};

    const std::string extension = file_path.extension().string();

    std::cout << extension << '\n';  // .gz
}
```

src/axom/klee/docs/sphinx/specifying_shapes.rst
```diff
@@ -44,12 +43,106 @@ is specified in :code:`windshield.stl`. Note that Klee does not specify
...
+or syntax errors are reported during Inlet verification or Lua evaluation.
+Unknown nested fields follow the same Inlet schema strictness rules as YAML input.
+
 Paths
 *****
 The paths specified in shapes are specified either as absolute paths
 (begin with a :code:`/`), or as relative paths. Absolute paths are evaluated
 as absolute paths in the file systems. Relative paths are evaluated relative
-to the YAML file (not the current working directory). For example, in the
+to the Klee input deck (not the current working directory). For example, in the
```
white238: I personally prefer "input file" over "input deck"

src/axom/klee/tests/klee_io.cpp
```diff
@@ -290,6 +299,525 @@ TEST(IOTest, readShapeSet_file)
...
+        material = "steel",
+        replaces = {"mat1", "mat2"},
+        geometry = {
+          format = "test_format",
+          path = "path/to/file.format"
+        }
+      }
+    }
+  )",
+                                         InputFormat::Lua);
```
white238, line 545: This formatting is weird.

src/examples/shaping_tutorial/CMakeLists.txt
```diff
@@ -104,6 +104,11 @@ if(ENABLE_TESTS)
     blt_add_test(NAME   lesson_03_klee_operators_and_validation
                 COMMAND lesson_03_klee_operators_and_validation ../lesson_03/ice_cream.yaml)
 
+    if(AXOM_USE_LUA)
```
white238, line 107:
```suggestion
    if(LUA_FOUND)
```
nitpick consistency

white238: or change the one below

white238 (approved):

src/axom/sina/core/Document.cpp
```diff
@@ -1116,15 +1116,12 @@ namespace internal
 
 Protocol detectOutputProtocol(const std::string &filepath)
 {
-  size_t dotPos = filepath.find_last_of('.');
-
-  if(dotPos == std::string::npos)
+  std::string ext = axom::utilities::filesystem::getFileExtension(filepath);
+  if(ext.empty())
```
Arlie-Capps, line 1120: What about `stoopidfilename.` which has a trailing dot?  Yes, I see it will eventually (and correctly) get to the `throw` at the end of the routine, but the "no extension" message is probably more helpful than "unknown extension ." from that line.

Arlie-Capps (approved):
Thank you, Kenny!  A few minor comments, and perhaps some added documentation (possibly in another PR).

src/axom/sina/core/Document.cpp
rhornung67: "stoopid is as stoopid does"  -- Forrest Gump

rhornung67 (approved):

src/axom/klee/docs/sphinx/specifying_shapes.rst
kennyweiss: Thanks -- I'll update the docs.

This PR does not add support for callbacks (I was trying to limit the scope).
I am planning to add callback support for Lua-based Klee inputs in an immediate follow-up branch/PR.

src/axom/klee/Geometry.cpp
kennyweiss: Great suggestion -- done.

src/axom/core/utilities/FileUtilities.cpp
kennyweiss: Done.

I kept the implementation in `getFileExtension()` since we had unit tests w/ Windows paths that did not work with converting `\\` to `/`: 

```cpp
  const std::pair<const char*, const char*> testCases[] = {
    {"", ""},
    {"file", ""},
    {"file.txt", ".txt"},
    {"file.tar.gz", ".gz"},
    {"file.", "."},
    {".gitignore", ""},
    {".config.json", ".json"},
    {".", ""},
    {"..", ""},
    {"directory/", ""},
    {"directory.with.dot/file", ""},
    {"directory.with.dot/file.yaml", ".yaml"},
    {R"(directory.with.dot\file)", ""},    // <-- Returned ".dot\\file" instead of "" before fix of converting to native on POSIX
    {R"(directory.with.dot\file.lua)", ".lua"},  
    {R"(C:\directory\FILE.YML)", ".YML"}};
```
See: https://godbolt.org/z/sW1jW3fYK

(A different approach would have been to not test against Windows paths on POSIX, but this seemed more general).

src/axom/klee/Geometry.cpp
kennyweiss: Thanks -- I'll update the documentation.

Klee uses exceptions since it parses user-provided input, and catching errors allows us to provide better error messages.

src/axom/sina/core/Document.cpp
kennyweiss: @Arlie-Capps -- happy to fix this if you think it's warranted.

Are you suggesting that:
(a) the implement of `getFileExtension()` should return an empty string if the size is 1 (i.e. the last character in `filepath` is `'.'`)?
(b) the check here should be `if ext.size() < 2` ?
(c) < something else >

src/axom/klee/tests/klee_io.cpp
kennyweiss: Agreed, but apparently that's what our clang-format is happy with.

I decided not to wrap it with `clang-format off/on` when I was developing it, but can add it if you think it needs to be fixed.

src/axom/klee/docs/sphinx/specifying_shapes.rst
kennyweiss: I switched all usages to "input file"

src/examples/shaping_tutorial/CMakeLists.txt
kennyweiss: I think it might need to stay this way.

Using     
```cmake
blt_print_variables(NAME_REGEX LUA IGNORE_CASE)
```
it seems that:
* In `src/examples/CMakeLists.txt`, the `LUA_FOUND` variable exists, but not `AXOM_USE_LUA`
* In `src/examplex/shaping_tutorial/CMakeLists.txt`  (which is configured separately in the install directory), the `LUA_FOUND` variable is not defined, but `AXOM_USE_LUA` is defined.


Output from the former:
```cmake
-- [blt_print_variables] The following variables are defined at the calling site in '<axom>/src/examples/CMakeLists.txt' -- 
-- [blt_print_variables] matching NAME_REGEX 'LUA' (case insensitive)
-- [blt_print_variables]   CACHE{FIND_PACKAGE_MESSAGE_DETAILS_Lua}:INTERNAL := [/usr/lib64/liblua-5.3.so;/usr/lib64/libm.so][/usr/include][v5.3.4()]
-- [blt_print_variables]   CACHE{LUA_DIR}:PATH := /usr
-- [blt_print_variables]   LUA_FOUND := TRUE
-- [blt_print_variables]   CACHE{LUA_INCLUDE_DIR}:PATH := /usr/include
-- [blt_print_variables]   LUA_LIBRARIES := /usr/lib64/liblua-5.3.so;/usr/lib64/libm.so
-- [blt_print_variables]   CACHE{LUA_LIBRARY}:FILEPATH := /usr/lib64/liblua-5.3.so
-- [blt_print_variables]   LUA_LIB_EXT := .3.so
-- [blt_print_variables]   CACHE{LUA_MATH_LIBRARY}:FILEPATH := /usr/lib64/libm.so
-- [blt_print_variables]   LUA_VERSION_MAJOR := 5
-- [blt_print_variables]   LUA_VERSION_MINOR := 3
-- [blt_print_variables]   LUA_VERSION_PATCH := 4
-- [blt_print_variables]   LUA_VERSION_STRING := 5.3.4
-- [blt_print_variables]   Lua_FOUND := TRUE
-- [blt_print_variables]   _is_lua_5_4 := -1
-- [blt_print_variables] ----------------------------------------------------------
```

Output from the latter:
```cmake
-- [blt_print_variables] The following variables are defined at the calling site in '<install-axom>/examples/axom/shaping_tutorial/CMakeLists.txt' -- 
-- [blt_print_variables] matching NAME_REGEX 'LUA' (case insensitive)
-- [blt_print_variables]   AXOM_LUA_DIR := /usr
-- [blt_print_variables]   AXOM_USE_LUA := TRUE
-- [blt_print_variables] ----------------------------------------------------------
```

src/axom/klee/Geometry.cpp
kennyweiss: Done.  I added a section to the sphinx docs and marked every throwing function.

src/axom/klee/docs/sphinx/specifying_shapes.rst
kennyweiss: I updated the docs about when Lua input is processed.

PR #1921, opened 2026-07-16
Feature/sina/support partial append

HaluskaR:
Adds one new flag to Sina's Document append() functionality. This extends the ability to selectively overwrite to curves; with the flag on, any curves included in the to-be-appended Document will overwrite existing curves at the same position. Off by default.


doutriaux1:
@HaluskaR I don't think overwrite is what we want, users can use the normal write if that's what they wanted right? I think we need to do something similar to what you did for our code e.g. remember the last index you wrote and only append from this point. Thoughts?

HaluskaR:
> @HaluskaR I don't think overwrite is what we want, users can use the normal write if that's what they wanted right? I think we need to do something similar to what you did for our code e.g. remember the last index you wrote and only append from this point. Thoughts?

So quick explanation: the difference between "append with overwriting curves" and "write" is the preservation of any other records in the document and any other values in the record. Say someone has a few snapshot records, which is the case in our code. Write()ing to the same document would wipe all those out.

That said, this could be done either way, append or overwrite (less remembering the last index we wrote since where would we store that state, but perhaps looking at the curve we have and appending the difference in length). The reason it's overwrite right now is really just to provide additional functionality. We have another case (less immediate) wherein the curve's past can be "rewritten" as part of a rewind...the advice I received there was to not even try to harmonize things, just let them send the new curve :) That might be a thing we want to write(), it might be a thing we want to append(), but I figured implementing it as an overwrite from the beginning would hopefully cover that base nicely.

That said, you might have some more insight into where that's going, and I can do this either way. I haven't checked the performance implications of an overwrite or an append (IIRC the major cost was not the write, but rather the access, given your tests; doesn't mean I couldn't take a look).

doutriaux1:
Becky ok that makes sense, I'm not entirely sure this actually costs us anything performance-wise anyway.

Thanks for the clarification

> > @HaluskaR I don't think overwrite is what we want, users can use the normal write if that's what they wanted right? I think we need to do something similar to what you did for our code e.g. remember the last index you wrote and only append from this point. Thoughts?
> 
> So quick explanation: the difference between "append with overwriting curves" and "write" is the preservation of any other records in the document and any other values in the record. Say someone has a few snapshot records, which is the case in our code. Write()ing to the same document would wipe all those out.
> 
> That said, this could be done either way, append or overwrite (less remembering the last index we wrote since where would we store that state, but perhaps looking at the curve we have and appending the difference in length). The reason it's overwrite right now is really just to provide additional functionality. We have another case (less immediate) wherein the curve's past can be "rewritten" as part of a rewind...the advice I received there was to not even try to harmonize things, just let them send the new curve :) That might be a thing we want to write(), it might be a thing we want to append(), but I figured implementing it as an overwrite from the beginning would hopefully cover that base nicely.
> 
> That said, you might have some more insight into where that's going, and I can do this either way. I haven't checked the performance implications of an overwrite or an append (IIRC the major cost was not the write, but rather the access, given your tests; doesn't mean I couldn't take a look).

src/axom/sina/tests/sina_Document.cpp
```diff
@@ -657,8 +657,9 @@ TEST(Document, test_validate_append_valid)
 
 void doEveryErrorTest(
   const std::string &protocol,
-  std::function<conduit::Node(const std::string &, const sina::Document &, int, bool)> appendDocumentFunc,
-  bool skipValidation = false)
+  std::function<conduit::Node(const std::string &, const sina::Document &, int, bool, bool)> appendDocumentFunc,
+  bool skipValidation = false,
+  bool curvesAreFullLength = false)
```
doutriaux1: different name?  i would rename to overwriteCurves

src/axom/sina/core/Document.hpp
doutriaux1: do we need the overwrite flag here as well? I believe so.

src/axom/sina/core/Document.cpp
doutriaux1: we need to overwrite flag here i believe

doutriaux1 (commented):
Becky this looks good with your clarification, you missed a couple of entry point for the new flags and I have a name nitpick in there but otherwise it looks good to me. Thanks

doutriaux1:
I merged develop back in once the test passed I'ts good to go on my end

doutriaux1 (approved):
Once the test pass (need to rerun becauseI merged develop in) I think we're good.

rhornung67 (approved):

HaluskaR:
Just to confirm what was mentioned in the meeting, I can't complete the merge on my end, not authorized to push to this branch.

bmhan12:
@lc-hubcast approve

bmhan12 (commented):
@lc-hubcast approve

PR #1922, opened 2026-07-17
Add verbose to MIR options

BradWhitlock:
The MIR algorithms output some information via `SLIC_INFO` about the number of clean and mixed zones. When the MIR algorithms are used on a lot of mesh patches or MPI ranks, this can cause very verbose output that can slow things down. This PR changes the MIR algorithms so the output is conditional on a `verbose` flag passed in the `conduit::Node` MIR options. The default is for verbose to be off so the output should be quiet.


src/axom/mir/tests/mir_elvira2d.cpp
```diff
@@ -120,6 +120,7 @@ struct braid2d_mat_test
           options["matset"] = "mat";
           options["plane"] = 1;
           options["pointmesh"] = pointMesh ? 1 : 0;
+          options["verbose"] = 1;
```
rhornung67, line 123: Doesn't this mean that verbose is on by default?

src/axom/mir/tests/mir_elvira3d.cpp
```diff
@@ -101,6 +101,7 @@ struct test_Elvira3D
     MIR m(topologyView, coordsetView, matsetView);
     conduit::Node deviceMIRMesh;
     conduit::Node options;
+    options["verbose"] = 1;
```
rhornung67, line 104: Same comment as above and more places below

rhornung67 (commented):
Question about verbose default setting.

src/axom/mir/tests/mir_elvira2d.cpp
BradWhitlock: The [option is off by default](https://github.com/llnl/axom/blob/503395d08161b86b0758342b49bb41708f8845c1/src/axom/bump/Options.hpp#L137) when someone uses the MIR algorithms (e.g. host codes). These are tests where I want the option turned on.

rhornung67: Got it. Thanks.

rhornung67 (approved):

Arlie-Capps (approved):

PR #1923, opened 2026-07-20
Reintroduce curvature methods

### LINKED ISSUE #1530: Add function to compute curature of a NURBSCurve/BezierCurve at a given parametric location
Functions to compute local curvature within a NURBS curve was removed from the C2CReader in #1526 

Let's add it back to the NURBSCurve class.
See:
https://github.com/LLNL/axom/blob/7ab265aa7f3be1d3e7fc4a7f360e96ce977c8e4a/src/axom/quest/readers/C2CReader.cpp#L561-L637

BradWhitlock:
This PR resolves #1530.

* Added `curvature` and `curvatureDerivative` methods to `BezierCurve` and `NURBSCurve`.
* Added tests


src/axom/primal/operators/curvature.hpp
```diff
@@ -0,0 +1,95 @@
...
+    const T yp = Dt[1];  // y'
+
+    const T xpp = DtDt[0];  // x''
+    const T ypp = DtDt[1];  // y''
+
+    // This is signed curvature as formulated at:
+    // https://en.wikipedia.org/wiki/Curvature#Curvature_of_a_graph
+    // k = (x'y'' - y'x'') / pow(x'x' + y'y', 3./2.)
+    const T xp2_plus_yp2 = xp * xp + yp * yp;
+    return (xp * ypp - yp * xpp) / pow(xp2_plus_yp2, 3. / 2.);
```
jcs15c: Is it worth having an extra check to make sure the denominators of these expressions is nonzero? It's technically possible for a Bezier curve to have a zero derivative at isolated parameter points, and NURBS curves might have discontinuous derivatives across knot spans, but neither is a case that really comes up except adversarially.

jcs15c (approved):
Looks good to me, thanks for adding this! Only possible concern is about curves with discontinuous derivatives, but I don't think it's a big deal either way.

src/axom/primal/operators/curvature.hpp
BradWhitlock: Thanks @jcs15c . I added some `SLIC_ASSERT()` macros in these functions to guard against division by zero.

Arlie-Capps (approved):

rhornung67 (approved):

PR #1924, opened 2026-07-21
Add c2c assembly support

### LINKED ISSUE #978: Enhance shaper so it can use C2C assembly.
I read through the C2C docs and it supports making assemblies of multiple contours. Since I was making a new dataset for Klee, I tried grouping some contours into assemblies since it could make the YAML input file smaller by making a material per assembly. Axom complained at runtime.

[WARNING]: Input was an assembly! This is not currently supported
[WARNING]: reading C2C file failed, setting mesh to NULL

If things need to be handled one contour at a time, perhaps if we encounter an assembly we can return a vector of contours and associate them all with the material that caused the assembly to be loaded. This would be nicer than having to write a large YAML input file.

* Support C2C assemblies in Klee

BradWhitlock:
This PR resolves #978 

* Enhances `C2CReader` so it can handle `.assembly` files _(all input files contribute to same list of curves)_
* This support is primarily useful for the `SamplingShaper`
* Adds tests
* Updates `data` directory (separate axom_data PR too)
* Shaping with assemblies is a little faster for heroic_roses.


rhornung67 (approved):
The code is easy to read and it looks reasonable. However, I have little experience with C2C so my approval should be confirmed by others.

Arlie-Capps (approved):
Thanks, Brad

PR #1925, opened 2026-07-21
Fix some windows warnings

BradWhitlock:
I had something break in CI on Windows and the errors did not immediately stand out due to the large number of warnings generated on Windows with MSVC. This PR takes a stab at reducing the warning count on Windows with MSVC.

* Fixed a lot of int/IndexType conversion warnings (adding static_cast or changing `int` to `IndexType`)
* Fixed a few Windows-specific warnings
* Limited changes in `quest` and `slam` because of other impending large PRs

I think I'll stop for now since the PR is getting larger. Another pass is needed later. As-is this PR reduces the MSVC warning output from about 5380 warnings to 1009 warnings.


src/axom/primal/geometry/NURBSCurve.hpp
```diff
@@ -612,6 +612,7 @@ class NURBSCurve
    * \return A reference to the control point at index \a idx
    */
   PointType& operator[](int idx) { return m_controlPoints[idx]; }
+  PointType& operator[](axom::IndexType idx) { return m_controlPoints[idx]; }
```
kennyweiss, line 618: (non-binding comment):
Instead of duplicating the implementation of `operator[]()`, `setWeight()`. `getWeight()`, it might be clearer to have an overload that static_casts the `int` to `axom::IndexType`.

I realize that each  one is 1-3 lines, but that would better express that it should be equivalent and prevent the code from drifting in the future.

src/axom/core/FlatMap.hpp
```diff
@@ -822,11 +822,11 @@ FlatMap<KeyType, ValueType, Hash>::FlatMap(IndexType bucket_count, Allocator all
   // N * GroupSize - 1 >= minBuckets
   // TODO: we should add a countl_zero overload for 64-bit integers
   {
-    std::int32_t numGroups = std::ceil((bucket_count + 1) / (double)BucketsPerGroup);
+    std::int32_t numGroups = static_cast<std::int32_t>((bucket_count + BucketsPerGroup) / BucketsPerGroup);
```
kennyweiss: This change is not obviously equivalent (to me).

Is it guaranteed to be equivalent to (or perhaps better than?) the previous version? Should we add some unit tests to make sure this is correct.

src/axom/core/utilities/System.cpp
```diff
@@ -10,11 +10,13 @@
   #include <windows.h>
   #include <tchar.h>
 #else
+  #include <cstdlib>
```
Arlie-Capps: Why is this line here, as well as (new) line 19?

Arlie-Capps (approved):
Thank you for this wide-ranging cleanup.

src/axom/primal/operators/detail/intersect_bezier_impl.hpp
kennyweiss, line 1: It looks like this one can have a real change in the (corner) case where it intersects on the boundary.
(@bmhan12 is currently investigating a very similar case for ray-segment intersections)

Do we have sufficient test coverage to check that we're handling this correctly?

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -390,16 +400,18 @@ class NURBSPatch
...
              const T* knots_u,
              int nkts_u,
              const T* knots_v,
              int nkts_v)
     : NURBSPatch(axom::ArrayView<const PointType, 2>(pts, {{npts_u, npts_v}}),
                  axom::ArrayView<const T, 2>(nullptr, {{0, 0}}),
-                 KnotVectorType(axom::ArrayView<const T>(knots_u, nkts_u), nkts_u - npts_u - 1),
-                 KnotVectorType(axom::ArrayView<const T>(knots_v, nkts_v), nkts_v - npts_v - 1))
+                 KnotVectorType(axom::ArrayView<const T>(knots_u, nkts_u),
+                                static_cast<int>(nkts_u - npts_u - 1)),
```
kennyweiss, line 412: Note: I suppose this implies that `NURBSPatch` is effectively using an `IndexType`,and `KnotVector` is using an `int` for the number of points?

rhornung67 (approved):
Thank you @BradWhitlock for the extensive house cleaning

src/axom/primal/geometry/NURBSPatch.hpp
```diff
@@ -2039,13 +2059,13 @@ class NURBSPatch
    */
   NURBSCurveType isocurve_u(T u) const
   {
-    SLIC_ASSERT_MSG(isValidParameter_u(u, 1e-5),
+    SLIC_ASSERT_MSG(isValidParameter_u(u, static_cast<T>(1e-5)),
```
kennyweiss, line 2062: Does the warning imply that we're using `NURBSCurve<float>` in some cases ?

src/axom/core/utilities/System.cpp
BradWhitlock: Good catch. I guess that was codex being "helpful". I removed line 13 and kept 19.

src/axom/primal/geometry/NURBSCurve.hpp
```diff
@@ -392,28 +394,29 @@ class NURBSCurve
 
     SLIC_ASSERT(static_cast<T>(theta_1 - theta_0) <= static_cast<T>(2.0 * M_PI));
 
-    T pi23 = 2.0 * M_PI / 3.0;
-    int n_segments = std::ceil((theta_1 - theta_0) / pi23);
+    const T pi23 = static_cast<T>(2.0 * M_PI / 3.0);
```
kennyweiss, line 397: (out of scope for this PR):
I'm guessing we don't have a ton of test coverage for `NURBSCurve<float>`, but it might be a good idea to add more.

src/axom/core/tests/core_device_hash.hpp
```diff
@@ -416,7 +416,7 @@ TEST(core_device_hash, hash_float_bit_pattern)
   const int NUM_KEYS = 1000;
   for(int i = 1; i <= NUM_KEYS; i++)
   {
-    float_hashes.insert(float_hasher(i / static_cast<float>(NUM_KEYS + 1)));
+    float_hashes.insert(float_hasher(static_cast<float>(i) / static_cast<float>(NUM_KEYS + 1)));
     double_hashes.insert(double_hasher(i / static_cast<double>(NUM_KEYS + 1)));
```
kennyweiss: It's interesting that the top numerator needs to be explicitly cast to float, but the bottom one doesn't need to be explicitly cast to double

src/axom/core/utilities/BitUtilities.hpp
```diff
@@ -264,9 +266,12 @@ template <typename FlagType, typename BitType>
...
 constexpr void setBit(FlagType &flags, BitType bit, bool value = true)
 {
+  using UnsignedFlagType = std::make_unsigned_t<FlagType>;
   assert(static_cast<int>(bit) < BitTraits<FlagType>::BITS_PER_WORD << 3);
-  const auto mask = 1 << bit;
-  flags = value ? (flags | mask) : (flags & ~mask);
+  const auto mask = static_cast<UnsignedFlagType>(UnsignedFlagType {1} << bit);
+  const auto flags_unsigned = static_cast<UnsignedFlagType>(flags);
+  flags = static_cast<FlagType>(
+    value ? (flags_unsigned | mask) : (flags_unsigned & static_cast<UnsignedFlagType>(~mask)));
```
kennyweiss: Would we expect any performance implications for this change? 

We typically use bit twiddles in performance critical sections, and this looks like this is using extra registers and/or doing extra work.

src/axom/core/utilities/System.cpp
```diff
@@ -116,5 +117,23 @@ std::locale locale(const std::string& name)
   return loc;
 }
 
+std::string getEnvironmentVariable(const std::string& name)
```
kennyweiss, line 120: :+1: -- 
I guess std::getenv doesn't work on Windows?
Does C++20 provide a way to do this?

kennyweiss (approved):
Thanks @BradWhitlock for this comprehensive set of fixes. I'm excited to have significantly fewer warnings in MSVC!

I flagged two changes to reconsider:
* The bitset might have a performance regression in performance sensitive code since it (appears to be) doing more work
* The bezier curve-ray intersection edge case changed behavior in the corner case of the ray intersecting a nearly linear segment at its endpoint, so we should make sure that we have sufficient test coverage to ensure that this is the right fix

src/axom/core/FlatMap.hpp
kennyweiss: (this one is marked "outdated", so it might no longer exist (?) )

src/axom/core/utilities/System.cpp
BradWhitlock: `std::getenv` does exist on Windows but Windows issues a warning about its safety so codex wanted to use a different function, annoying. I met it halfway by telling it to make a `getEnvironmentVariable` function that hides the differences.

src/axom/primal/geometry/NURBSCurve.hpp
BradWhitlock: In the current version of the code, only the `axom::IndexType` version of the methods exist. At some point I ran into a build where they could not co-exist since they mapped to the same type and were considered duplicates.

src/axom/core/FlatMap.hpp
BradWhitlock: I just iterated some more with codex on this and it figured out an integer-only way of doing it. I had it add comments to explain.

src/axom/primal/operators/detail/intersect_bezier_impl.hpp
BradWhitlock: I changed the code back to more closely resemble develop while preserving the variable name change that eliminated the shadowing warning. I added a test too.

src/axom/primal/geometry/NURBSPatch.hpp
BradWhitlock: yup

src/axom/primal/geometry/NURBSCurve.hpp
BradWhitlock: I had codex extend test coverage for `NURBSCurve<float>`.

src/axom/core/tests/core_device_hash.hpp
BradWhitlock: Cleaned it up.

src/axom/primal/geometry/NURBSPatch.hpp
BradWhitlock: `KnotVector` is currently using `int` for the degree. We can revisit this if it makes sense. I stopped partway through on eliminating warnings so there are likely still some things that would be addressed in a next pass.

src/axom/core/utilities/BitUtilities.hpp
BradWhitlock: I changed/simplified the code. Codex doesn't expect an optimized build of the revised code to incur performance penalties.

PR #1926, opened 2026-07-22
 C++20 cuda rehash workaround

bmhan12:
This PR:
- Attempts to address commented issue here: https://github.com/llnl/axom/pull/1912#discussion_r3571699113
- **Note**: This PR is intended to merge into `feature/han12/c_20`, not `develop`

### DIFF
--- a/src/axom/core/FlatMap.hpp
+++ b/src/axom/core/FlatMap.hpp
@@ -633,42 +633,9 @@ public:
     }
     else
     {
-      if constexpr(std::is_trivially_copyable_v<KeyValuePair>)
+      if(tryDeviceAwareRehash(count))
       {
-#if defined(AXOM_USE_UMPIRE) && defined(AXOM_USE_GPU)
-        MemorySpace space = detail::getAllocatorSpace(m_allocator.getID());
-        if(space == MemorySpace::Device || space == MemorySpace::Unified)
-        {
-  #if defined(AXOM_USE_CUDA)
-          int host_allocator_id = axom::detail::getAllocatorID<MemorySpace::Host>();
-          FlatMap host_map(*this, axom::Allocator {host_allocator_id});
-          host_map.rehash(count);
-
-          FlatMap rehashed_device(host_map, m_allocator);
-          this->swap(rehashed_device);
-          return;
-  #elif defined(AXOM_USE_HIP)
-    #if !defined(AXOM_GPUCC)
-          // Similar to the issue in ArrayBase (PR #1582), using FlatMap from
-          // a GPU-enabled Axom library but with a host-only compiler results
-          // in an ODR violation.
-
-          // HACK: this looks ugly, but is the best we can do pending a DR:
-          // https://stackoverflow.com/questions/44059557/whats-the-right-way-to-call-static-assertfalse
-          // https://cplusplus.github.io/CWG/issues/2518.html
-          static_assert(std::is_pod_v<KeyType> && !std::is_pod_v<KeyType>,
-                        "Cannot instantiate device-aware FlatMap operations when file is compiled "
-                        "with a host-only compiler. Axom was built with GPU support, so you should "
-                        "build all source files using FlatMap with a CUDA/HIP compiler.");
-          using ExecSpace = axom::SEQ_EXEC;
-    #else
-          using ExecSpace = axom::HIP_EXEC<256>;
-    #endif
-          this->parallelRehash<ExecSpace>(count);
-          return;
-  #endif
-        }
-#endif
+        return;
       }
       FlatMap rehashed(m_size,
                        std::make_move_iterator(begin()),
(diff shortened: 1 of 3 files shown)


src/axom/core/FlatMap.hpp
```diff
@@ -627,42 +627,9 @@ class FlatMap : detail::flat_map::SequentialLookupPolicy<typename Hash::result_t
     }
     else
     {
-      if constexpr(std::is_trivially_copyable_v<KeyValuePair>)
```
bmhan12, line 636: This `if constexpr` expression is the culprit for the original unit test error as I understand it. Based on `printf` debugging, the code hits a runtime error when launching any kernel (including empty, no-op kernels) inside this code block.

bmhan12: @publixsubfan Once I narrowed the error to this expression, I had codex try to come up with a refactor to avoid the compile-time code generation (and not revert to rehashing on the host). 
The rehashing algorithm is different as a result.

adayton1: The CUDA programming guide lists a few restrictions on extended (i.e. host device) lambdas that we use with RAJA:

"With Microsoft Visual Studio host compilers, an extended lambda shall not be defined within the body of an if constexpr block."

"A variable cannot be implicitly captured inside an if-constexpr block that is lexically nested inside an extended lambda unless the variable has been implicitly captured outside the if-constexpr block or appears in the extended lambda’s explicit capture list."

See https://docs.nvidia.com/cuda/cuda-programming-guide/05-appendices/cpp-language-support.html#extended-lambda-restrictions for the full list of limitations

src/axom/core/FlatMapUtil.hpp
```diff
@@ -231,27 +145,119 @@ auto FlatMap<KeyType, ValueType, Hash>::create(ArrayView<KeyType> keys,
...
+  Array<SpinLock> lock_vec(num_groups, num_groups, new_metadata.getAllocatorID());
+  const auto group_locks = lock_vec.view();
 
-  FlatMapOffsetIterator bucket_begin {m_buckets.data(), filled_bucket_idx_vec.data()};
+  const IndexType num_old_buckets = old_buckets.size();
+
+  for_all<ExecSpace>(
+    num_old_buckets,
+    AXOM_LAMBDA(IndexType bucket_idx) {
+      IndexType group_idx = bucket_idx / GroupBucket::Size;
```
publixsubfan, line 177: On first glance, this looks like a slightly modified version of the batched- insertion kernel. How easy would it be to merge this and the original kernel? I'd be a little worried about maintainability long-term.

src/axom/core/FlatMapUtil.hpp
```diff
@@ -231,27 +145,119 @@ auto FlatMap<KeyType, ValueType, Hash>::create(ArrayView<KeyType> keys,
...
+            curr_group = (curr_group + lookup.getNext(iteration)) & group_mask;
+            ++iteration;
+          }
+          else
+          {
+            new_metadata[curr_group].template setBucket<true>(empty_slot, hash_8);
+            IndexType new_bucket = curr_group * GroupBucket::Size + empty_slot;
+            // Relocate raw storage directly so device rehash does not depend on
+            // pair construction or copy-oriented batched insertion code paths.
+            new_buckets[new_bucket] = old_buckets[bucket_idx];
```
publixsubfan: This copy might be invalid for non-trivial types on the host, should we have a host-guarded construction path here?

src/axom/core/FlatMapUtil.hpp
bmhan12: > How easy would it be to merge this and the original kernel? 

I/codex gave this a shot, refactoring out the parts that are shared between the batch-insertion kernel and the added rehash one. Seemed to work for CUDA, but then the refactor was causing the batch-insertion kernel to hang with HIP... I punted on the refactor.

src/axom/core/FlatMapUtil.hpp
bmhan12: Guard added!

bmhan12:
@kennyweiss @publixsubfan 
I benchmarked this branch (bugfix/han12/cuda_rehash) against develop on 1 node (-N 1 -n 112) of rzvector.
For the most part, there was not any significant regressions - except for `find_hit_prehashed`, 42% slower. 
I traced the regression down to nvcc code generation differences with C++20 (originally I thought it could be one of the TPL changes).
You can see in the graph that benchmarking on develop with `BLT_CXX_STD` set to `c++20` (without the changes of this branch) leads to the regression.
flatmap_benchmark_three_way_cxx20

src/axom/core/FlatMapUtil.hpp
publixsubfan: Ah unfortunate. Punting on it is fine.

PR #1928, opened 2026-07-22
Proposal for reworking host memory

rhornung67:
# Summary

This PR touches a lot of files, but the changes are all closely related and limited in scope. I tried to stack this as a sequence of PRs but struggled to reconcile subsets of cherry picked commits with the develop branch due to all the changes there. 

The bulk of the changes are in Axom core. Other components are modified as the changes in core percolate through to higher level components.

The main goal of this PR is to remove implicit dependence on Axom's global default host allocator for runtime memory allocations. The most substantial changes in this PR are:
- Routines that allocate host memory now take an explicit host allocator argument.
- Internal helper code does not call `getDefaultHostAllocatorID()` or resolve `MemorySpace::Host` through global state *unless it is in a legacy compatibility path*.
- Startup behavior no longer depends on whether some earlier static constructor happened to allocate host memory.

The pre-existing global host allocator API still exists, especially in the core container classes (Array, etc.) because they are user-facing. Thus, most API changes are done via method overloads so that the pre-existing APIs still function as before. In particular, the `getDefaultHostAllocatorID()` and `setDefaultHostAllocator(...)` methods are retained as user-facing compatibility APIs, but they are no longer used in Axom internals.

Rather than thread raw `int` values (allocator ids) everywhere, a `axom::HostAllocator` wrapper was introduced to:
- Better communicate intent throughout the code.
- Prevent accidental passage of a device allocator where host staging memory is needed.
- Avoid overloading ambiguity with existing `allocator_id` parameters that refer to primary storage rather than host scratch memory.

Many APIs already accepted a primary `allocator_id` for owned storage. However, that is not enough for code paths that may also allocate temporary host ("staging") buffers. This PR takes the following approach:

- If an object allocates only one owned buffer, keep the existing explicit allocator parameter.
- If an object may allocate both primary storage and host staging/scratch, add a second explicit host allocator parameter.

Tests have been adjusted/added and documentation has been adjusted. 

**I will update the release notes if/when this PR is approved for merging.**

**If this PR is approved and merged. The global allocator dependent code and APIs would be deprecated and eventually removed in a future PR.**


src/axom/core/docs/sphinx/core_memory_management.rst
```diff
@@ -0,0 +1,175 @@
...
+  axom::fill(devicePtr, numValues, value, hostAlloc);
+
+Overloads that do not receive ``axom::HostAllocator`` remain available for
+compatibility. They forward through Axom's current default host allocator and
+should be treated as legacy convenience paths in new production code.
+
+Changing the default host and global allocators (legacy)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+THe contents of this section refer to Axom *legacy* convenience routines,
```
BradWhitlock: "THe"

rhornung67: D'oh! Thanks. Fixed.

src/axom/core/docs/sphinx/core_memory_management.rst
```diff
@@ -0,0 +1,175 @@
...
+Similarly, the Axom global allocator can be changed. For example::
+
+  // set Axom global allocator to Umpire Host
+  axom::setDefaultAllocator(axom::MemorySpace::Host);
+
+  // set Axom global allocator to Umpire Unified memory allocator
+  axom::setDefaultAllocator(axom::MemorySpace::Unified);
+
+  // set Axom global allocator to an explicitly chosen Umpire allocator
+  int allocId =
```
BradWhitlock, line 165: `axom::getAllocatorIDFromMemorySpace(axom::MemorySpace::Pinned)` seems like a better example.

src/axom/core/examples/core_array_perf.cpp
```diff
@@ -159,30 +159,44 @@ InputParams params;
...
+
+#if !defined(AXOM_USE_UMPIRE)
   AXOM_UNUSED_VAR(policy);
-#if defined(AXOM_USE_UMPIRE)
-  int allocatorID = policy == axom::runtime_policy::Policy::seq
-    ? axom::detail::getAllocatorID<axom::MemorySpace::Host>()
-    :
+#else
+
+  const axom::HostAllocator hostAllocator {axom::execution_space<axom::SEQ_EXEC>::allocatorID()};
```
BradWhitlock, line 168: This leaves me wanting syntactic sugar: `const auto hostAllocator = axom::execution_space<axom::SEQ_EXEC>::hostAllocator();`

src/axom/core/execution/internal/cuda_exec.hpp
```diff
@@ -64,7 +64,7 @@ struct execution_space<CUDA_EXEC<BLOCK_SIZE, SYNCHRONOUS>>
 
   static int allocatorID() noexcept
   {
-    return axom::getUmpireResourceAllocatorID(umpire::resource::Device);
+    return axom::getAllocatorIDFromMemorySpace(memory_space);
```
BradWhitlock: Good change.

src/axom/core/tests/core_array.hpp
```diff
@@ -14,9 +14,68 @@
 #include "gtest/gtest.h"
 
 #include <algorithm>
+#include <cstdlib>
+#include <exception>
 
 namespace
 {
+#if defined(AXOM_USE_GPU) && defined(AXOM_USE_UMPIRE)
+bool runtimeMemorySpaceAvailable(axom::MemorySpace space)
```
BradWhitlock: This looks like it belongs in the new `MemoryTesting.hpp` file.

rhornung67: Probably. I'll look into it.

src/axom/core/examples/core_array_perf.cpp
rhornung67: I agree that we can smooth this further.

src/axom/core/docs/sphinx/core_memory_management.rst
rhornung67: Yeah, I didn't think too hard about the example. As I noted in the PR summary, many of the changes involve overloads to preserve the pre-existing interfaces, but show how the new explicit host allocator API works.

If we agree that this is an improvement and the way to move forward, then we would deprecate the "implicit" host allocator API and remove all references to global host allocator state, including in the docs.

src/axom/core/tests/core_array.hpp
```diff
@@ -889,6 +948,8 @@ template <typename T, int DIM, axom::MemorySpace SPACE>
 void check_device(axom::Array<T, DIM, SPACE>& v)
 {
   const axom::IndexType size = v.size();
+  const int explicit_host_alloc =
```
BradWhitlock, line 896: Why this and not `axom::getAllocatorIDFromMemorySpace(axom::MemorySpace::Host)`?

src/axom/core/tests/core_array.hpp
```diff
@@ -903,7 +964,7 @@ void check_device(axom::Array<T, DIM, SPACE>& v)
   }
 
   // Then check the contents by assigning to an explicitly Host Array
-  axom::Array<T, 1, axom::MemorySpace::Host> check_raw_array_host = v;
+  axom::Array<T, 1, axom::MemorySpace::Host> check_raw_array_host(v, explicit_host_alloc);
```
BradWhitlock, line 912: In this statement and some others in this test it switched from assigment operator to copy constructor. Is that intentional?

src/axom/core/tests/core_array.hpp
```diff
@@ -1359,30 +1446,43 @@ TEST(core_array, checkAlloc)
...
       }
 // Then, if Umpire is available, we can use the space as an explicit template parameter
-#ifdef AXOM_USE_UMPIRE
+#if defined(AXOM_USE_UMPIRE)
   #ifdef UMPIRE_ENABLE_DEVICE
-      axom::Array<int, 1, axom::MemorySpace::Device> v_int_device(capacity, capacity);
-      ::check_alloc(v_int_device, axom::getUmpireResourceAllocatorID(umpire::resource::Device));
-      axom::Array<double, 1, axom::MemorySpace::Device> v_double_device(capacity, capacity);
-      ::check_alloc(v_double_device, axom::getUmpireResourceAllocatorID(umpire::resource::Device));
+      if(runtimeMemorySpaceAvailable(axom::MemorySpace::Device))
```
BradWhitlock: I've seen other efforts in this PR to minimize use of the Umpire enums and related functions and instead request allocators for the `axom::MemorySpace::Device` (whichever enum value). Here they are being mixed. Could `axom::getAllocatorIDFromMemorySpace(axom::MemorySpace::Device)` be used?

src/axom/core/tests/core_array_for_all.hpp
```diff
@@ -17,8 +17,70 @@
 // gtest includes
 #include "gtest/gtest.h"
 
+#include <exception>
+
 namespace testing
 {
+namespace
+{
+bool runtimeMemorySpaceAvailable(axom::MemorySpace space)
```
BradWhitlock: I think this is repeated in another test.

src/axom/core/tests/core_memory_management.hpp
```diff
@@ -24,12 +29,94 @@
 // in check_alloc_realloc_free when reallocating to 3 * ARRAY_SIZE.
 constexpr int ARRAY_SIZE = 5345;
 
+struct ScopedDefaultAllocatorState
```
BradWhitlock, line 33: This RAII pattern is repeated from a different test. Would it be useful to have in `axom::core`?

src/axom/core/docs/sphinx/core_memory_management.rst
rhornung67: @BradWhitlock I added your suggestion as an alternative example.

src/axom/core/tests/core_memory_management.hpp
```diff
@@ -204,12 +289,14 @@ void check_alloc_realloc_free(bool hostAccessible = true)
 
     // Reallocate to a larger size.
     buffer_size *= 3;
-    buffer = axom::reallocate<int>(buffer, buffer_size);
-#ifdef AXOM_USE_UMPIRE
+#if defined(AXOM_USE_UMPIRE)
```
BradWhitlock, line 238: Is the `AXOM_USE_UMPIRE` check still needed since you switched to not using the `umpire::ResourceManager`?

src/axom/core/tests/core_memory_management.hpp
```diff
@@ -24,12 +29,94 @@
...
+    axom::setDefaultAllocator(m_defaultAllocator);
+    axom::setDefaultHostAllocator(m_defaultHostAllocator);
+  }
+
+  int m_defaultAllocator;
+  int m_defaultHostAllocator;
+};
+
+#if defined(AXOM_USE_UMPIRE)
+bool coreMemoryManagementRuntimeMemorySpaceAvailable(axom::MemorySpace space)
```
BradWhitlock: This may be repeated, this time with a different name.

src/axom/core/Array.hpp
```diff
@@ -103,6 +103,32 @@ struct DefaultStoragePolicy
   }
 };
 
+inline HostAllocator hostAllocatorForPrimaryAllocator(int allocator_id)
```
BradWhitlock, line 106: It seems like these functions might be useful in other places too. Do they belong in `memory_management.hpp`?

src/axom/core/Array.hpp
```diff
@@ -1253,24 +1453,41 @@ AXOM_HOST_DEVICE Array<T, DIM, SPACE, StoragePolicy>::Array(const Array& other)
...
+AXOM_HOST_DEVICE Array<T, DIM, SPACE, StoragePolicy>::Array(Array&& other) noexcept
   : ArrayBase<T, DIM, Array<T, DIM, SPACE, StoragePolicy>>(
       static_cast<ArrayBase<T, DIM, Array<T, DIM, SPACE, StoragePolicy>>&&>(std::move(other)))
   , m_resize_ratio(0.0)
-  , m_arrayOps(other.m_arrayOps)
+  , m_allocator_id(other.m_allocator_id)
+  , m_host_allocator(other.m_host_allocator)
+  , m_arrayOps(m_allocator_id, m_executeOnGPU, m_host_allocator)
 {
+#if defined(AXOM_DEVICE_CODE)
```
BradWhitlock, line 1464: Making this `AXOM_HOST_DEVICE` and adding the trap/abort behavior suggests to me that we'll now get a runtime error if we accidentally capture `axom::Array` in device code. Nice. It might be good to test this in CI if it isn't already _(codex did not see a test)_.

src/axom/core/Array.hpp
```diff
@@ -436,6 +512,7 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM, SPACE, StoragePolicy>>, pro
 
   AXOM_HOST_DEVICE inline T* data() { return m_data; }
   AXOM_HOST_DEVICE inline const T* data() const { return m_data; }
+  inline int getHostAllocatorID() const { return m_host_allocator.getID(); }
```
BradWhitlock, line 515: It looks like some other properties can be queried from the host allocator. Would be make sense to instead provide a `const HostAllocator& getHostAllocator() const` method? It would simplify some of the `FlatMap` initializers too.

BradWhitlock:
I have questions / comments for Monday.

src/axom/core/tests/core_array.hpp
rhornung67: This is intended to be an explicit check that the default allocator and the Umpire Host allocator are consistent. In particular, `axom::MemorySpace::Host` could refer to either the Umpire Host allocator or Axom MALLOC, depending on what the default is set to.

src/axom/core/tests/core_array.hpp
rhornung67: This is testing the mechanics of the ctor overloads, that takes a host allocator id explicitly.

src/axom/core/tests/core_array.hpp
rhornung67: These tests are checking that the `axom::MemorySpace` enum values match up with the `umpire::resource::` enum values.

src/axom/core/tests/core_array.hpp
rhornung67: I eliminated the redundancy and put the method in the `MemoryTesting.hpp` file.

src/axom/core/tests/core_array_for_all.hpp
rhornung67: I eliminated the redundancy and put the method in the `MemoryTesting.hpp` file.

src/axom/core/tests/core_memory_management.hpp
rhornung67: It may be redundant here since I think (hope) that if `AXOM_USE_UMPIRE` is not defined, then `UMPIRE_ENABLE_*` will not be defined. However, I found the use of `AXOM_USE_UMPIRE` and `UMPIRE_ENABLE_*` macro constants inconsistent and confusing. It was there in the original, so I left it.

src/axom/core/tests/core_memory_management.hpp
rhornung67: Removed and used the centralized method.

src/axom/core/Array.hpp
```diff
@@ -221,6 +247,11 @@ class Array : public ArrayBase<T, DIM, Array<T, DIM, SPACE, StoragePolicy>>, pro
         IndexType capacity = 0,
         int allocator_id = axom::detail::getAllocatorID<SPACE>());
 
+  template <IndexType SFINAE_DIM = DIM,
+            MemorySpace SFINAE_SPACE = SPACE,
+            typename std::enable_if<SFINAE_DIM == 1>::type* = nullptr>
+  Array(IndexType num_elements, IndexType capacity, int allocator_id, HostAllocator host_allocator);
```
publixsubfan, line 253: Big picture: do we expect users to be frequently setting different "host allocators" between different Array instances? I'm a little worried about the expansion in the array API surface; wondering if a global interface for a "default" value for all Arrays, plus a setter interface for specific instances, might tame things a little.

src/axom/core/Array.hpp
publixsubfan: There is a test for it, but it's disabled on AMD; however, AMD's host/device checks are stricter at compile-time so it's less of a concern.

src/axom/core/Array.hpp
rhornung67: I share your concern. My PR attempt before this tried to do that, but it seemed brittle and subtle w.r.t. maintaining global host allocator state (default or user-selected) and what to do when the host allocator is implicit. 

What I think we want is a simple way to define a host allocator default and allow users to change it if/where they want. Unfortunately, trying to make this robust requires a lot of subtle checking for what had been done in the past (like host allocation, then changing the host allocator default).

I'm open to other ideas. This is the sort of input/feedback I'm looking for by posting this PR.

rhornung67: As it stands now in develop branch, the host allocator is determined based on how an Axom build is configured. If Umpire is enabled, Umpire Host allocator is used/assumed. If Umpire is NOT enabled, Axom MALLOC is used/assumed.

In this PR, Axom MALLOC is the default in either case, which is preferable IMO.

PR #1930, opened 2026-07-29
Fixes formatting of reference and pointer variables

kennyweiss:
# Summary

- This maintenance PR updates our .clang-format
  - It ensures that we consistently place the `*` and `&` for pointers and references next to the type rather than the variable
  - It ensures that line ending are LF (instead of CRLF)
  - It updates the language to C++17
     - we should change it to C++20 when #1912 merges
  - It updates the link/commit at the top to clang@19, instead of clang@14
  - It removes obsolete/deprecated attributes from our .clang-format
     - I verified that the associated commit has no formatting changes

Since this merge will affect all outstanding existing branches, it probably makes sense to merge this close to the upcoming August release.
- I added a huge commit with all of the formatting changes related to the pointer/references/crlf, but will plan to redo that commit before merging.


src/.clang-format
kennyweiss, line 1: This file contains the only hand-edited changes.
Many of the changes were related to removing obsolete/deprecated features, and updating to the new ones.
Tested using `clang-format --dump-config`

The rest of the files in this PR are automatically generated via `make style`
(Note: Reviewing with "Hide whitespace" reveals that nearly every change in this PR is a whitespace change.)

src/.clang-format
```diff
@@ -48,10 +50,12 @@ IndentPPDirectives: BeforeHash
 IndentWidth: 2
 
 PenaltyExcessCharacter: 10
+PointerAlignment: Left
 
+ReferenceAlignment: Left
```
kennyweiss, line 55: These two control the placement of  `T* ptr` and `T& ref`

The `DerivePointAligment: false` a few lines up says not to use the style found in the file.

rhornung67 (approved):

src/.clang-format
```diff
@@ -1,26 +1,26 @@
-# requires clang 14
-# https://releases.llvm.org/14.0.0/tools/clang/docs/ClangFormatStyleOptions.html
+# requires clang 19
+# https://releases.llvm.org/19.1.0/tools/clang/docs/ClangFormatStyleOptions.html
 
 BasedOnStyle: Google
 Language: Cpp
-Standard: Cpp11
+Standard: c++17
```
jcs15c: Should this not be c++20 by now? Or is that going to change a lot of formatting rules, and introduce a lot more non-whitespace changes

jcs15c (approved):
Looks good to me! It is indeed mostly whitespace changes.

src/.clang-format
kennyweiss: Good catch -- I posted this before #1912 merged 
 (I had a note about updating it in the PR description).

kennyweiss: Done!

src/.clang-format
kennyweiss, line 33: Now that we're past clang@11, do we want to set `BraceWrapping/BeforeLambdaBody: true`?

If so, we'd check changes like this:

Image

I can see both sides: Lambda curly brace wrapping on a separate line would be more consistent w/ our other choices, but somehow, I've grown used to seeing lambda curly braces at the end of a line.

We can also defer this decision to a follow-up PR if we can't reach a quick consensus.

PR #1931, opened 2026-07-29
Change to CXX flags to squash CUDA build warnings

rhornung67:
# Summary

- This PR squashes CUDA build warnings for TPL tests.

### DIFF
--- a/src/thirdparty/tests/CMakeLists.txt
+++ b/src/thirdparty/tests/CMakeLists.txt
@@ -292,7 +292,7 @@ endif()
 
 foreach(test ${BLT_SMOKE_TESTS})
     get_filename_component( test_name ${test} NAME_WE )
-    axom_add_executable(
+    blt_add_executable(
         NAME       ${test_name}_test
         SOURCES    ${test}
         OUTPUT_DIR ${TEST_OUTPUT_DIRECTORY}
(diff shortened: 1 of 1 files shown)


Arlie-Capps (approved):

src/thirdparty/tests/CMakeLists.txt
```diff
@@ -292,7 +292,7 @@ endif()
 
 foreach(test ${BLT_SMOKE_TESTS})
     get_filename_component( test_name ${test} NAME_WE )
-    axom_add_executable(
+    blt_add_executable(
```
kennyweiss, line 295: Out of curiosity, how does `axom_add_executable` -> `blt_add_executable` help fix this?

kennyweiss (approved):
Thanks @rhornung67

src/thirdparty/tests/CMakeLists.txt
rhornung67: I don't think it does. There is a mix of `axom_add_executable` and `blt_add_executable` macros in that file. I don't see a consistent pattern and don't know why one is used vs. the other. I changed it here because the tests are for BLT, no other reason.  The change to CXX flags is the substantial change.

kennyweiss: Thanks -- I suspect it's historical -- the `axom_add_executable` macro came later and we didn't change everything.

Looks like the only difference is adding the `openmp` target in `openmp` configs before calling `blt_add_executable`:
https://github.com/llnl/axom/blob/f1e1c7653da6b99eaf4f3f9842f2b10b1e122bf6/src/cmake/AxomMacros.cmake#L161-L199

Issue #1848, opened 2026-04-27
CUDA build warnings


rhornung67:
When building Axom using our host config files to configure for `llvm@19.1.3_cuda`(matrix and rzvector yield the same results), several compiler warnings are encountered:

- [x] warnings: unused variables and variables used before initialization in `thirdparty/tests`
- [x] warnings: calling `constexpr __host__` function from `__host__ __device` function in `core/detail/FlatTable.hpp`, `core/FlatMapUtil.hpp`.  Resolve by removing `__host__` decoration from constexpr function or use experimental flag `--expt-relaxed-constexpr`?
- [ ] warnings: similar to above in `core/tests/core_array_for_all.hpp`
- [x] warnings: similar to above in `primal/geometry/Polygon.hpp`
- [ ] warnings: similar to above in `primal/geometry/BoundingBox.hpp`
- [ ] warnings: similar to above in `primal/geometry/Vector.hpp`
- [x] warnings: similar to above in `primal/detail/clip_impl.hpp`
- [ ] nvlink warnings when stack size cannot be statically determined.

rhornung67:
From @adayton1 "This flag is the one I asked the nvidia folks about and they said that they plan to support it for the foreseeable future. It is less experimental and more of an indication that just because a function is constexpr doesn't mean it is necessarily device safe (e.g. if it uses host memory). But I think the vast majority of constexpr functions should be safe to use on the device."

This flag refers to the nvcc compiler flag `--expt-relaxed-constexpr`

kennyweiss:
@BradWhitlock mentioned that it might be difficult to fix the warnings on the primal primitives since they are using `axom::Array` and will be difficult to properly use on the GPU. We'd need to rethink the storage for these, e.g. in the way that `primal::Polygon` take a storage parameter.

rhornung67:
Follow-on issue https://github.com/llnl/axom/issues/1933 supersedes

PR #1932, opened 2026-07-30
Squash CUDA build warnings

rhornung67:
# Summary

- This PR eliminates ~20 warnings in our CUDA builds.
- A significant number of warnings that will require some work to remove and are not addressed here. These fall into two categories:
   - nvlink warnings about the compiler being unable to statically determine stack size
   - calling host function from host-device function where the host function is an axom::Array method. However, axom::Array is not designed to be used in device code. So removing these warnings may require some code re-design.

This PR resolves some of https://github.com/llnl/axom/issues/1848. A new issue was created to capture the remaining warnings.


white238 (approved):

Arlie-Capps (approved):

kennyweiss (approved):
Thanks @rhornung67 !

Issue #1745, opened 2025-12-15
Adds CI plan for minimal sina config


kennyweiss:
PR https://github.com/LLNL/axom/pull/1740 updates our spack recipe to allow users to selectively enable components.
Some users have expressed a desire for a minimal Axom config for just the sina component.
Let's add a CI test for the minimal sina config.

Related:
* https://github.com/LLNL/axom/issues/1698
* https://github.com/LLNL/axom/issues/1722

kennyweiss:
We can test this in a CI job that adds the following two CMake variables:
```cmake
-DAXOM_ENABLE_ALL_COMPONENTS=OFF
-DAXOM_ENABLE_SINA=ON
```

HaluskaR:
> We can test this in a CI job that adds the following two CMake variables:
> 
> -DAXOM_ENABLE_ALL_COMPONENTS=OFF
> -DAXOM_ENABLE_SINA=ON

That's huge Kenny, thanks! Think I was overcomplicating this trying to generate a host file etc. Pushing a first stab now.

kennyweiss:
Merged in #1939

PR #1939, opened 2026-08-03
add initial stab at gitlab sina-only CI; minimal build

HaluskaR:
# Summary

- This PR is to address https://github.com/llnl/axom/issues/1745#issuecomment-5097213741
- It's meant to add a minimal CI config to ensure that the "Sina only" build options are/continue working
- First push is checking how this runs as part of the CI, I imagine that ideally we can run the Sina unit tests as well.
- I templated it off dane/llvm as an arbitrary choice (it was the TPL I was using when trying to get a host config out of uberenv, that might've been overkill..), might be good to scoop up a compiler version and machine name from elsewhere to decrease the maintenance surface area if the combo doesn't matter too much.


.gitlab/build_only_sina.yml
```diff
@@ -0,0 +1,12 @@
+# As this is only to test Sina's ability to build with the defined set of
+# requirements, I set it up (initially?) to only build as one combo, dane+llvm.
+
+dane-llvm_sina-only:
```
bmhan12: You'll want to put this job inside the `build_dane.yml` file.
The job is trying to extend off of `.src_build_on_dane`, but it's not defined in this file.

bmhan12 (commented):
@lc-hubcast approve

.gitlab/build_dane.yml
```diff
@@ -108,3 +108,17 @@ dane-gcc_13_3_1_no_fortran-full:
...
 #  extends: .full_build_on_dane
+
+
+####
+# Component Build jobs
+dane-llvm_sina-only:
+  rules:
+    # Do we want it as a $TEST_COMPONENTS instead of a $TEST_SINA? Would these ever be turned off?
+    - if: '$CI_COMMIT_BRANCH =~ /_qnone/ || $ON_DANE == "OFF" || $TEST_SINA == "OFF"'
+      when: never
```
bmhan12: Hm, looks like the job is still not running. 
Maybe it's missing `- when: on_success` as the final rule? : https://docs.gitlab.com/ci/jobs/job_rules/#rules-examples

Alternatively, if you don't need the `TEST_SINA` conditional, you could remove the rules section entirely so that it extends the rules from `.on_dane`.

bmhan12 (commented):
@lc-hubcast approve

.gitlab/build_dane.yml
```diff
@@ -108,3 +108,13 @@ dane-gcc_13_3_1_no_fortran-full:
...
+
+
+####
+# Component Build jobs
+dane-llvm_sina-only:
+  variables:
+    COMPILER: "llvm@19.1.3"
+    HOST_CONFIG: "dane-toss_4_x86_64_ib-${COMPILER}.cmake"
+    EXTRA_CMAKE_OPTIONS: "-DAXOM_ENABLE_ALL_COMPONENTS=OFF -DAXOM_ENABLE_SINA=ON"
+  extends: .src_build_on_dane
```
kennyweiss, line 121: Minor: Perhaps this should be grouped with the "src" builds rather than the "full" builds

.gitlab/build_only_sina.yml
kennyweiss, line 1: Do we need this file? Or is it covered by the additions in `.gitlab/build_dane.yml`?

kennyweiss (approved):
Thanks @HaluskaR -- looks good assuming the expected tests run and pass

rhornung67 (commented):
@lc-hubcast approve

.gitlab/build_dane.yml
```diff
@@ -108,3 +108,13 @@ dane-gcc_13_3_1_no_fortran-full:
...
 #  extends: .full_build_on_dane
+
+
+####
+# Component Build jobs
+dane-llvm_sina-only:
+  variables:
+    COMPILER: "llvm@19.1.3"
+    HOST_CONFIG: "dane-toss_4_x86_64_ib-${COMPILER}.cmake"
+    EXTRA_CMAKE_OPTIONS: "-DAXOM_ENABLE_ALL_COMPONENTS=OFF -DAXOM_ENABLE_SINA=ON"
```
kennyweiss: @HaluskaR -- it seems that while slic is not required to build sina, it is required for the sina tests.
```
2026-08-10T21:47:26.461873Z 01O /usr/WS1/atk/gitlab-runner/builds/8d1k5syQs/001/gitlab/axom/axom/src/axom/sina/examples/sina_check_datum_type.cpp/usr/WS1/atk/gitlab-runner/builds/8d1k5syQs/001/gitlab/axom/axom/src/axom/sina/examples/sina_file_object_removal.cpp:/usr/WS1/atk/gitlab-runner/builds/8d1k5syQs/001/gitlab/axom/axom/src/axom/sina/examples/sina_create_datum.cpp:8:10::88 :fatal error: :'axom/slic.hpp' file not found
2026-08-10T21:47:26.461874Z 01O 1010: :fatal error:  'axom/slic.hpp' file not foundfatal error: 
```

Adding `-DAXOM_ENABLE_SLIC=ON` to this line should fix it.

kennyweiss: Note that we already encode this requirement in our spack package:
https://github.com/llnl/axom/blob/2957850014f3899363fcfddc7f83c8ab853ebe44/scripts/spack/packages/axom/package.py#L336

bmhan12 (commented):
@lc-hubcast approve

.gitlab/build_dane.yml
```diff
@@ -108,3 +108,13 @@ dane-gcc_13_3_1_no_fortran-full:
...
 #  extends: .full_build_on_dane
+
+
+####
+# Component Build jobs
+dane-llvm_sina-only:
+  variables:
+    COMPILER: "llvm@19.1.3"
+    HOST_CONFIG: "dane-toss_4_x86_64_ib-${COMPILER}.cmake"
+    EXTRA_CMAKE_OPTIONS: "-DAXOM_ENABLE_ALL_COMPONENTS=OFF -DAXOM_ENABLE_SINA=ON -DAXOM_ENABLE_SLIC=ON -DAXOM_ENABLE_TUTORIALS=OFF"
```
kennyweiss: I added a config variable to disable the tutorials since they require non-sina components

bmhan12 (approved):

.gitlab/build_dane.yml
kennyweiss: That didn't quite work (the CI script was still defaulting to testing the tutorials), 
so I added an ENV variable to explicitly disable the tests in the minimal-sina config

rhornung67: Should we also have a variant in our spack package to disable tutorials, defaulted to off?

kennyweiss: We already have a `tutorials` variant, which defaults to `True`.
If you think it should default to `False`, that seems reasonable since most spack users won't want our tutorials.

I don't think the variant defaulting to `True` will cause problems for spack users since the problem I solved wasn't in the `build`  or `install` phase w/ the minimal sina config (those both worked). The problem was in our CI script assuming that `<install>/axom/examples/{radiuss_tutorial,shaping_tutorial}` existed in all CI jobs and trying to run it.

See: https://github.com/llnl/axom/blob/8373edcda37c76b17447aee713c8379c3c6bcec9/scripts/spack/packages/axom/package.py#L115

rhornung67: We've had RAJA user complain about having to turn off building examples, exercises, etc. So we made everything off by default except tests. Whatever you think is best. I won't push on it either way.

PR #1948, opened 2026-08-12
Add Dynamic Filtering Option to Distributed Closest Point 

MrBurmark:
# Summary

Add dynamic filtering option in the distributed closest points algorithm to avoid sending data to ranks that could not improve any of the closest points already found.

Optionally capture the max of the current closest points while updating the closest points with the local mesh and use that to filter on the bounding boxes of the next ranks. If a rank couldn't contain any points that are closer than the max of the current closest points then skip that rank. Each rank that passed the static filtering still expects to get a message so send a small skip message so to ensure the receive count completion criteria is fulfilled.

- This PR is a feature
- It does the following:
  - Adds dynamic filtering at the request of me

### DIFF
--- a/src/axom/quest/detail/DistributedClosestPointImpl.hpp
+++ b/src/axom/quest/detail/DistributedClosestPointImpl.hpp
@@ -251,6 +251,7 @@ public:
     , m_rank(-1)
     , m_nranks(-1)
     , m_sqDistanceThreshold(axom::numeric_limits<double>::max())
+    , m_dynamicDistanceFiltering(true)
   { }
 
   virtual ~DistributedClosestPointImpl() { }
@@ -300,6 +301,11 @@ public:
     SLIC_ERROR_IF(sqThreshold < 0.0, "Squared distance-threshold must be non-negative.");
     m_sqDistanceThreshold = sqThreshold;
   }
+
+  /*!
+   @brief Enables dynamic filtering of object ranks using current closest distances.
+  */
+  void setDynamicDistanceFiltering(bool on) { m_dynamicDistanceFiltering = on; }
 
   /*!
     @brief Set which output data fields to generate.
@@ -528,6 +534,7 @@ protected:
   int m_nranks;
 
   double m_sqDistanceThreshold;
+  bool m_dynamicDistanceFiltering;
 
   bool m_outputRank = true;
   bool m_outputIndex = true;
@@ -747,6 +754,7 @@ public:
     constexpr int tag = 9873;
 
     int remainingRecvs = 0;
+    int fullXferRecvs = 0;
 
     std::list<conduit::relay::mpi::Request> isendRequests;
 
@@ -762,7 +770,7 @@ public:
       BoxArray allQueryBbs;
       gatherBoundingBoxes(myQueryBb, allQueryBbs);
 
-      computeLocalClosestPoints(xferNode);
+      double currentMaxSqDistance = computeLocalClosestPoints(xferNode);
 
       const auto& myObjectBb = m_objectPartitionBbs[m_rank];
       for(int r = 0; r < m_nranks; ++r)
@@ -770,8 +778,7 @@ public:
         if(r != m_rank)
         {
           const auto& otherQueryBb = allQueryBbs[r];
-          double sqDistance = axom::primal::squared_distance(otherQueryBb, myObjectBb);
-          if(sqDistance <= m_sqDistanceThreshold)
+          if(statically_eligible(otherQueryBb, myObjectBb))
           {
             ++remainingRecvs;
           }
@@ -783,7 +790,7 @@ public:
         partition, if any.  Increase remainingRecvs, because this data
         will come back.
       */
-      int firstRecipForMyQuery = next_recipient(xferNode);
+      int firstRecipForMyQuery = next_recipient(xferNode, currentMaxSqDistance, tag, isendRequests);
       if(m_nranks == 1)
       {
         SLIC_ASSERT(firstRecipForMyQuery == -1);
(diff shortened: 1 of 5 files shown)


BradWhitlock (approved):

src/axom/quest/detail/DistributedClosestPointImpl.hpp
```diff
@@ -529,6 +535,7 @@ class DistributedClosestPointImpl
   int m_nranks;
 
   double m_sqDistanceThreshold;
+  bool m_dynamicDistanceFiltering;
```
kennyweiss: Should we set the default value here instead of in the constructor?

```suggestion
  bool m_dynamicDistanceFiltering {true};
```

src/axom/quest/detail/DistributedClosestPointImpl.hpp
```diff
@@ -847,27 +873,62 @@ class DistributedClosestPointExec : public DistributedClosestPointImpl
   }
 
 private:
+  bool statically_eligible(const BoxType& queryBb,
```
kennyweiss: Please add a brief doxy comment to this function and to `send_skip_token`

src/axom/quest/detail/DistributedClosestPointImpl.hpp
```diff
@@ -847,27 +873,62 @@ class DistributedClosestPointExec : public DistributedClosestPointImpl
...
+                           const BoxType& objectBb,
+                           double* sqDistance = nullptr) const
```
kennyweiss: Would this parameter (`sqDistance`) be better as a `std::optional` than as a possibly `nullptr`?

src/axom/quest/detail/DistributedClosestPointImpl.hpp
kennyweiss: `isStaticallyEligible` (or `is_statically_eligible`) might be a better name for this function to indicate that it is a predicate.

See: 
* https://axom.readthedocs.io/en/develop/docs/sphinx/coding_guide/sec02_names.html#function-names-should-indicate-behavior
* https://axom.readthedocs.io/en/develop/docs/sphinx/coding_guide/sec02_names.html#function-name-format

src/axom/quest/detail/DistributedClosestPointImpl.hpp
```diff
@@ -974,27 +1038,76 @@ class DistributedClosestPointExec : public DistributedClosestPointImpl
...
-                                              axom::execution_space<axom::SEQ_EXEC>::allocatorID());
-        sqDistThresh_host[0] = m_sqDistanceThreshold;
-        axom::Array<double> sqDistThresh_device =
-          axom::Array<double>(sqDistThresh_host, m_allocatorID);
-        auto sqDistThresh_device_view = sqDistThresh_device.view();
-
         auto ptCoordsView = m_objectPtCoords.view();
         auto ptDomainIdsView = m_objectPtDomainIds.view();
 
+        if(m_dynamicDistanceFiltering)
```
kennyweiss, line 1068: Please add a brief comment here about the difference between using and not using dynamic distance filtering. 

Is there a reason not to use it? Or are you adding the toggle as a test/safeguard that it's working properly?

src/axom/quest/examples/CMakeLists.txt
```diff
@@ -393,6 +393,8 @@ if(AXOM_ENABLE_MPI AND AXOM_ENABLE_SIDRE AND HDF5_FOUND)
                   set(_center 0.7 0.9 0.5)
                 endif()
 
+                # Default dynamic distance filtering is tested across the full
+                # enabled policy/mesh matrix.
```
kennyweiss: This comment makes sense in the context of this PR, but might be confusing later. 

Consider removing this comment, and updating the one below to mention that it is testing the `--no-dynamic-distance-filtering` option

src/axom/quest/examples/CMakeLists.txt
kennyweiss, line 425: It might be helpful to add the explicit `--dynamic-distance-filtering` flag to this test command.

kennyweiss (approved):
Thanks @MrBurmark

src/axom/quest/detail/DistributedClosestPointImpl.hpp
MrBurmark: This is an output parameter and you can't have a `std::optional<double&>`. This defaulted param is mainly here to avoid getting an unused distance back in one of the 2 call sites.

src/axom/quest/detail/DistributedClosestPointImpl.hpp
MrBurmark: This is more of a safeguard in case this causes problems.

src/axom/quest/detail/DistributedClosestPointImpl.hpp
kennyweiss: Would `std::optional<double>&` work?
(The pointer is also ok).

MrBurmark: No, that would be difficult to default and it would require the cases where I do want that result to pass in a reference to an optional when they would rather just pass in a reference to a double.

MrBurmark: It is possible to have an interface that has two overloads, one that lacks an output arg and one that takes a reference to a double. Then those could both forward to an implementation like this that has some kind of optional output parameter, but that seemed excessive for an internal function that is called twice.

kennyweiss: I agree -- sounds like `double* = nullptr` makes sense here.

MrBurmark: Turns out it is possible to do this.
```suggestion
                           std::optional<std::reference_wrapper<double>> sqDistance = std::nullopt)const
```

src/axom/quest/detail/DistributedClosestPointImpl.hpp
MrBurmark: Added some more comments explaining what happens in the dynamic filtering path.

Issue #1927, opened 2026-07-22
Program startup slowness with HIP


BradWhitlock:
I noticed that several shaping tests with the `shaping_driver` took about 11 seconds on rzadams vs 2 seconds on rzwhippet.

I dug into this a bit and found that most of that time (~9 seconds) was spent in `axom::getDefaultAllocatorID` in the course of setting up the `sidre::DataStore`. That function just calls the Umpire `ResourceManager` to get the default allocator. `ResourceManager` queries the HIP runtime to determine the number of GPUs, etc.

I hacked the shaping_driver with a timed block to call `hipGetDeviceCount()` before the DataStore creation and the 9 seconds of time moved to this first touch of HIP.

Next, I had codex make a short standalone program that did the types of queries that Umpire was doing. The program was way faster, ~245ms instead of 9s.

Next, I had codex start linking that standalone program with additional Axom dependencies until it showed a slowdown in HIP calls. The `hipblas` library caused the highest jump in program startup time. Maybe there is some static initialization?

I believe Axom is getting this dependency from MFEM, which makes sense since MFEM is needed for the `shaping_driver`.

There might not be much we can do about this but I thought it was worth recording.

## Timings
```
================================================================
case: baseline-hip
program: hip
command: hip
start: Wed Jul 22 12:10:31 PDT 2026
HIP/hipBLAS startup reproducer
build=hip
[begin] program total
[begin] hipGetDeviceCount (baseline)
[end]   hipGetDeviceCount (baseline) : 254.035 ms
  baseline device count: 4
HIP device count: 4
  device 0: AMD Radeon Graphics, totalGlobalMem=137438953472, gcnArchName=gfx942:sramecc+:xnack-
  device 1: AMD Radeon Graphics, totalGlobalMem=137438953472, gcnArchName=gfx942:sramecc+:xnack-
  device 2: AMD Radeon Graphics, totalGlobalMem=137438953472, gcnArchName=gfx942:sramecc+:xnack-
  device 3: AMD Radeon Graphics, totalGlobalMem=137438953472, gcnArchName=gfx942:sramecc+:xnack-
[end]   program total : 254.070 ms
real 0.30
user 0.06
sys 0.20
exit_status: 0
end: Wed Jul 22 12:10:31 PDT 2026

================================================================
case: hipblas-linked
program: hip_hipblas
command: hip_hipblas
start: Wed Jul 22 12:10:31 PDT 2026
HIP/hipBLAS startup reproducer
build=hip+hipblas
[...]

PR #1950, opened 2026-08-13
Adds a bump-based Marching Cubes option

kennyweiss:
# Summary

- Our existing Marching Cubes functionality in quest only supports structured (Cartesian) grids, but we have users that want to run Marching Cubes on unstructured quad and hex meshes
- This feature PR adds a `bump`-based Marching Cubes option to quest to run on structured and unstructured meshes
   - The algorithm is policy-based, and runs on cpu, omp, cuda and hip execution spaces
   - I left the original MC algorithm in place since it is currently more performant on structured grids
   - This is likely due to the new algorithm performing more work. Specifically:
     - it welds the vertices of the MC
     - it also generates a polygonal MC mesh, but our application requires a triangle mesh, so we currently triangulate the output polygons
   - We should instead have an option to use a triangle-based clipping table, which would avoid the extra work
- Also improves the python script that generates Cartesian meshes as input to our MC example, and adds a new script to generate unstructured meshes for the MC example

### TODO:
- [ ]  Show performance numbers for CPU and GPU runs


src/cmake/AxomMacros.cmake
```diff
@@ -196,6 +196,13 @@ macro(axom_add_executable)
                        OUTPUT_NAME ${arg_OUTPUT_NAME}
                        FOLDER      ${arg_FOLDER})
 
+    # Avoid loading expensive unused ROCm solver dependencies with HIP 6.x.
```
kennyweiss: The hip@6.x tests were taking about an hour or more without this change!
See #1927 

I created a separate PR (https://github.com/llnl/axom/pull/1991) to deal with hip@6.x the runtime issues, including moving some mfem-related function calls into separate translation units to avoid unnecessary linking. 

Once that's merged, I'll remove the changes from this PR.

PR #1952, opened 2026-08-17
I/O and memory diagnostic improvements in DistributedClosestPoint driver

kennyweiss:
# Summary

* This PR improves the I/O capabilities for our DistributedClosestPoint (DCP) driver example
* The driver now takes a distributed blueprint mesh for the object mesh and for the query mesh
   * It previously used an internally generated mesh of a circle in 2D and a sphere in 3D
   * Re retained the ability to analytically verify the results for input data with the appropriate verification tag
   * The two inputs can be multi-domain or single domain
   * We added test meshes to the axom_data repo, and updated the unit tests to use these
* Also adds some memory diagnostics to the DCP driver to see how much memory is being used and the extent to which `malloc_trim` helps clear memory
   * This led to a fix to the underlying DCP implementation to release memory after copying to MPI buffers
   * This bugfix might resolve some user-reported memory issues w/ the DCP algorithm.
* This PR also improves the convert_sidre_protocol.py script to convert from conduit blueprint formats, instead of just sidre formats
* It also adds a new python script to generate distributed blueprint point meshes w/ following several point distributions (circle/sphere, annulus/torus, Gaussian, uniform)

### Misc:
* I updated our build skill to be able to build/test w/ MPI on an allocated node, and to work around issues with shroud and symlinks in sandboxed environments


src/axom/quest/detail/DistributedClosestPointImpl.hpp
```diff
@@ -802,6 +802,8 @@ class DistributedClosestPointExec : public DistributedClosestPointImpl
         isendRequests.emplace_back(conduit::relay::mpi::Request());
         auto& req = isendRequests.back();
         relay::mpi::isend_using_schema(*xferNodes[m_rank], firstRecipForMyQuery, tag, m_mpiComm, &req);
+        // isend_using_schema() deep-copies into the request buffer.
+        xferNodes.erase(m_rank);
```
kennyweiss: These three  `xferNodes.erase()` seem to save a lot of memory during `computeClosestPoint` in hip runs

| input | branch | query time | RSS peak | RSS post-query | RSS after trim
|---|---|---:|---:|---:|---:|
| `100k` | develop | 2.16 s | 3.76 GiB | 3.76 GiB | 3.57 GiB | 
| `100k` | branch | 2.14 s | 3.64 GiB | 3.64 GiB | 3.55 GiB | 
| `1M` | develop | 22.91 s | 5.97 GiB | 5.38 GiB | 4.33 GiB |
| `1M` | branch | 23.23 s | 4.77 GiB | 4.27 GiB | 4.13 GiB | 
| `10M` | develop | 286.76 s | 27.81 GiB | 9.04 GiB | 9.00 GiB | 
| `10M` | branch | 285.99 s | 15.89 GiB | 9.04 GiB | 9.00 GiB | 

Notes: 
* In the 10M case, the peak memory went down from ~28 GB to ~16GB with this change.
* `malloc_trim` only helped marginally in `develop` and even less after the changes in this file
* This was in a Debug config, so don't read too much into the query times.

rhornung67: Nice memory reduction for larger problem sizes.

kennyweiss: Thanks.  I think it might essentially be the same fix as https://github.com/llnl/axom/pull/1889, which hasn't merged yet.

Tag: @MrBurmark

kennyweiss:
Note: The failing github actions CI job (gcc@13.3.1, shared, 32 bit) is related using a 64-bit type in a 32-bit build.  Should be easy to fix.

rhornung67 (approved):
@kennyweiss thank you for this.

src/axom/quest/detail/DistributedClosestPointImpl.hpp
MrBurmark: Indeed reducing the lifespan of the transfer nodes is the essence of #1889.

MrBurmark: @kennyweiss are you planning to collect data on the post #1948 develop?

kennyweiss: @MrBurmark  -- I rebased my branch on axom@develop after merging both of your PRs and removed my `node.erase` calls. 

I reran the tests w/ and w/o dynamic filtering to confirm that your changes captured the memory improvements (w/ `amdclang++` / `rocm@7.2.1`):

| input | dynamic filtering | query time | RSS peak | RSS post-query | RSS after trim |
|---|---|---:|---:|---:|---:|
| `1M` | enabled | 2.318 s | 4.67 GiB | 4.16 GiB | 4.03 GiB |
| `1M` | disabled | 2.300 s | 4.66 GiB | 4.16 GiB | 4.00 GiB |
| `10M` | enabled | 25.588 s | 15.78 GiB | 8.93 GiB | 8.89 GiB |
| `10M` | disabled | 27.196 s | 15.78 GiB | 8.93 GiB | 8.89 GiB |

Note: the previous timings were in a `Debug (-O0)` config and these are in a `DebWithRelInfo (-O2)`, so don't read too much into the timings changes.

MrBurmark: Looks good, thanks @kennyweiss.

rhornung67: Thanks for rerunning. AFAIK, codes don't put debug builds out for users 😆

kennyweiss: (I used a Debug config for the initial run since I was focusing on the memory rather than the timings.)

publixsubfan (approved):

Issue #1853, opened 2026-04-30
Problematic find_package calls in axom-config.cmake


ldowen:
The export target logic in axom-config.cmake calls things like `find_package(MPI REQUIRED)` if conduit and MPI are both enabled. This seems excessive and poses an issue for any project that might only want to find the MPI package for certain languages. It also seems unnecessary in the event that MPI has already been found, which is likely if the project is using MPI locally.
https://github.com/llnl/axom/blob/21504d400d3bea543da4d4a180345a2dc3a2ce65/src/cmake/axom-config.cmake.in#L160

rhornung67:
This line can be removed from export logic (and tested to make sure it still works), but does not solve issue in BLT.  Maybe @white238 know what this means....

See also https://github.com/llnl/blt/pull/755

kennyweiss:
Related issue to investigate/fix when working on this:

PR #1956 added the following to support non-MPI Axom configs (`-DENABLE_MPI=OFF`) w/ MPI-based MFEM
```cmake
if(TARGET mfem AND MFEM_USE_MPI)
    blt_patch_target(NAME mfem DEPENDS_ON mpi)
endif()
```

But, BLT only creates an `mpi` target (in `BLTSetupMPI.cmake`) when `BLT_ENABLE_MPI` is set.

Since `blt_patch_target` appends `DEPENDS_ON` entries to `INTERFACE_LINK_LIBRARIES` without checking they are targets, 
this silently becomes `-lmpi` on the link line rather than a configure error, and also doesn't append an MPI include directory.

Once we have language-dependent targets (per this issue), we can change this check to something like:
```cmake
if(TARGET mfem AND MFEM_USE_MPI)
    if(NOT TARGET MPI::MPI_C)
        find_package(MPI QUIET COMPONENTS C)
    endif()
    if(TARGET MPI::MPI_C)
        blt_patch_target(NAME mfem DEPENDS_ON MPI::MPI_C)
    else()
        message(WARNING "MFEM was built with MPI but no MPI C target was found; linking against MFEM may fail")
    endif()
endif()
```

PR #1956, opened 2026-08-19
Enables Axom to be configured without MPI when (some of) its dependencies use MPI

### LINKED ISSUE #1944: error building quest w/o mpi
axom develop, build on dane using gcc 10.3 (I used this compiler b/c I have a conduit tpls built with this compiler)

```
      |                                                 ^
In file included from /usr/WS1/visit/visit/thirdparty_shared/3.5.0/toss4/hdf5/2.0.0/linux-x86_64_gcc-10.3/include/H5public.h:51,
                 from /usr/WS1/visit/visit/thirdparty_shared/3.5.0/toss4/hdf5/2.0.0/linux-x86_64_gcc-10.3/include/hdf5.h:21,
                 from /usr/workspace/cyrush/2026.08.axom/axom/src/axom/core/../../axom/sidre/core/Group.hpp:41,
                 from /usr/workspace/cyrush/2026.08.axom/build/include/axom/sidre.hpp:16,
                 from /usr/workspace/cyrush/2026.08.axom/axom/src/axom/core/../../axom/mint/mesh/FieldData.hpp:14,
                 from /usr/workspace/cyrush/2026.08.axom/axom/src/axom/core/../../axom/mint/mesh/Mesh.hpp:14,
                 from /usr/workspace/cyrush/2026.08.axom/axom/src/axom/core/../../axom/quest/interface/internal/QuestHelpers.hpp:13,
                 from /usr/workspace/cyrush/2026.08.axom/axom/src/axom/quest/interface/internal/QuestHelpers.cpp:7:
/usr/workspace/cyrush/2026.08.axom/axom/src/axom/core/../../axom/quest/interface/internal/mpicomm_wrapper.hpp:25:15: error: expected ‘)’ before numeric constant
   25 | constexpr int MPI_COMM_SELF = -1;
      |               ^~~~~~~~~~~~~
```

Maybe this is an issue with gcc 10.3's C++20 support, but it seems like constexpr should be fine?

kennyweiss:
# Summary

- This PR is a feature/bugfix
- It allows Axom to be configured without MPI (`ENABLE_MPI=OFF`) when (some of) its dependencies were configured with MPI
  - It removes Quest's fake `MPI_COMM` definitions in favor of an internal `CommHandle` definition to avoid reusing MPI symbols
  - Removes a possible footgun that would read meshes on all ranks if the MPI_Comm/CommHandle parameter was missing
- Most of this was internal Axom implementations (e.g. in the internal `QuestHelpers`), and should minimally affect users (unless they were using internal implementations, in which case the upgrade should be straightforward)
- Resolves #1944


src/axom/quest/interface/internal/QuestHelpers.hpp
```diff
@@ -11,32 +11,77 @@
...
+{
+  Serial = -1
+};
+
+/// \brief Returns a handle for the calling rank alone
+inline CommHandle commSelf() { return CommHandle::Serial; }
+
+/// \brief Returns a handle for all ranks
+inline CommHandle commWorld() { return CommHandle::Serial; }
+#endif
```
kennyweiss, line 77: We're now using `CommHandle` instead of faking `MPI_Comm` in API calls that can optionally support MPI.  Most of these were only in internal APIs  (e.g. in an `axom::*::internal` namespace).

Using a fake `MPI_Comm` was problematic when Axom was configured without MPI but some of its dependencies were configured with MPI since those could bring in MPI defines.

src/axom/quest/interface/inout.cpp
```diff
@@ -392,6 +394,24 @@ int inout_init(std::shared_ptr<mint::Mesh>& mesh, MPI_Comm comm)
...
+}
+
+#if defined(AXOM_USE_MPI)
+int inout_init(const std::string& file, MPI_Comm comm) { return inout_init_impl(file, comm); }
+
+int inout_init(std::shared_ptr<mint::Mesh>& mesh, MPI_Comm comm)
+{
+  return inout_init_impl(mesh, comm);
+}
+#endif
```
kennyweiss, line 413: Note that the public API only has an MPI comm parameter when Axom is configured w/ MPI.

src/axom/quest/DiscreteShape.hpp
```diff
@@ -12,21 +12,20 @@
...
 /*!
-  @brief Post-processed klee::Shape, with the geometry discretized and transformed
-  according to the Shape's operators.
-
-  TODO: Move this class into internal namespace.
-*/
+ * @brief Post-processed klee::Shape, with the geometry discretized and transformed
+ * according to the Shape's operators.
+ *
+ * TODO: Move this class into internal namespace.
```
kennyweiss, line 27: This TODO is interesting. 
I will look into whether this class is actually internal and if the repercussions, if any, of moving it to the `internal` namespace.

src/axom/quest/interface/internal/QuestHelpers.hpp
cyrush: In Ascent, we use the fortran mpi handle (always an integer) b/c interface can get messy due various ways MPI_Comm is implemented. 

We can always convert this to an MPI_Comm when needed

src/cmake/thirdparty/SetupAxomThirdParty.cmake
```diff
@@ -192,6 +192,11 @@ else()
     message(STATUS "MFEM support is OFF")
 endif()
 
+# MFEM's exported target does not always propagate its MPI dependency.
+if(TARGET mfem AND MFEM_USE_MPI)
+    blt_patch_target(NAME mfem DEPENDS_ON mpi)
+endif()
```
kennyweiss, line 198: I think this can be resolved more robustly once we use language-dependent MPI targets, and posted a comment to the relevant issue -- https://github.com/llnl/axom/issues/1853#issuecomment-5348492067

src/axom/core/tests/utils_config.cpp
```diff
@@ -100,12 +100,6 @@ TEST(utils_config, mfem_configuration)
     mfemHasMPI = true;
   #endif
     std::cout << "mfem is built " << (mfemHasMPI ? "with" : "without") << " MPI" << std::endl;
-
-    if(!axomHasMPI)
-    {
-      EXPECT_FALSE(mfemHasMPI) << "Axom expects mfem to be built without MPI "
-                                  "when it is not built with MPI";
-    }
```
bmhan12: Is this check being handled somewhere else or is redundant?

bmhan12 (approved):

Arlie-Capps (approved):
Thanks for this fix, Kenny.

PR #1958, opened 2026-08-19
Re-enable python testing in Github Actions

bmhan12:
This PR:
- Re-enable python testing in Github Actions
  - Fix involved an issue in spack where `spec["python"].package.platlib` in the host-configs get resolved to `*/dist-packages` instead of the actual install path of `*/site-packages`. As a result, check for nanobind fails and python interface/testing is disabled.
  - Fix was to not use an external python, and have spack build python (`spec["python"].package.platlib` resolves to `*/site-packages` after change).


rhornung67 (approved):

kennyweiss (approved):
Thanks @bmhan12

PR #1962, opened 2026-08-25
Adds callback support to Klee's Lua input

kennyweiss:
# Summary

- This feature PR adds callback support to Klee's Lua input
- It also adds the ability to pass a separate initialization context (a Lua table) into Klee (Lua only)
  - This allows users to provide their own variables/functions to override information in the Lua input
  - E.g. the Klee input might have dimension-specific logic, and this initialization context allows then to set the dimension
- The callback support allows fields to support function alternatives via the new inlet `addFunctionAsValueAlternative()`
- Docs: https://axom.readthedocs.io/en/feature-kweiss-klee-lua-callbacks/

### TODO

- [x] Update RELEASE-NOTES


BradWhitlock (approved):
Looks good to me.

src/axom/inlet/docs/sphinx/functions.rst
```diff
@@ -58,11 +63,62 @@ In Lua, the following operations on the ``Vector`` type are supported (for ``Vec
...
+   :dedent: 2
+
+For the shared input name, ``contains`` reports the concrete value, including a default,
+and ``containsFunctionValueAlternative`` reports a supplied function. Check the function
+first because a concrete default can exist alongside it. Either supplied form counts as
+user-provided input, and strict Containers recognize both. An unrelated input type fails
+verification.
+
+``addFunctionAsValueAlternative`` returns a ``Verifiable<Function>``.
+Calling ``required()`` on it requires a function, and a concrete value does not satisfy the reqquirement.
```
white238, line 105:
```suggestion
Calling ``required()`` on it requires a function, and a concrete value does not satisfy the requirement.
```

src/axom/quest/examples/CMakeLists.txt
```diff
@@ -350,6 +351,48 @@ if(AXOM_HAS_MFEM_WITH_MPI AND AXOM_ENABLE_SIDRE AND AXOM_ENABLE_KLEE)
...
         set(_nranks 3)
+
+        if(LUA_FOUND)
+            set(_testname quest_shaping_driver_ex_lua_callbacks)
+            axom_add_test(
+                NAME ${_testname}
+                COMMAND quest_shaping_driver_ex
+                        -i ${shaping_data_dir}/sphere_callbacks.lua
+                        --method sampling
+                        inline_mesh --min -6 -6 -6 --max 6 6 6 --resolution 16 16 16 -d 3
```
white238, line 362: Do these numbers need to be explained?

src/axom/klee/tests/klee_io_util.cpp
```diff
@@ -66,22 +56,6 @@ InletTestData::InletTestData(const std::string& input, DefOp defOp)
   }
 }
```
white238, line 58: What was the reason for all these deleted tests?

white238: Ah this functionality was removed.

src/axom/quest/examples/CMakeLists.txt
kennyweiss: I can add a comment -- it's defining parameters for the background mesh -- a 3D cube of resolution `16*16*16` inside a bounding box from `(-6,-6,-6)` to `(6,6,6,)`

src/axom/inlet/SphinxWriter.cpp
```diff
@@ -112,6 +112,12 @@ void SphinxWriter::documentContainer(const Container& container)
 
   for(const auto& function_entry : container.getChildFunctions())
   {
+    // A function value alternative is documented through the concrete entry
+    // that shares its input path, not under its internal schema name
```
kennyweiss, line 116: @white238 -- would the following be clearer?

```
// Function value alternative names are internal and cannot appear in an input file.
// The corresponding Field/Container is documented instead.
```

src/axom/klee/tests/klee_io_util.cpp
kennyweiss: They were internal parsing for vectors/points, and were supeceeded by the new (also internal) `getPoint()`/`getVector()` helper functions in `klee/io/GeometryOperatorsIO.cpp`

The deleted tests now map to new tests in `klee_geometry_operators_io.cpp`, e.g. `TEST(GeometryOperatorsIO, readTranslation_wrongDimensions)`

src/axom/inlet/docs/sphinx/functions.rst
```diff
@@ -1,11 +1,11 @@
 ##################
-Function Callbacks
+Function callbacks
```
white238, line 2: The rest of Inlet's documentation capitalizes all section titles. This is inconsistent.

src/axom/inlet/examples/functions.cpp
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Lawrence Livermore National Security, LLC and other
+// Axom Project Contributors. See top-level LICENSE and COPYRIGHT
+// files for dates and other details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include "axom/fmt.hpp"
+#include "axom/inlet.hpp"
+#include "axom/slic.hpp"
+#include "axom/slic/core/SimpleLogger.hpp"
```
white238, line 10:
```suggestion
```

This is included in the SLIC unified header and is not needed.

src/axom/inlet/examples/functions.cpp
```diff
@@ -0,0 +1,55 @@
...
+double readScale(const std::string& luaInput)
+{
+  auto reader = std::make_unique<inlet::LuaReader>();
+  reader->parseString(luaInput);
+  inlet::Inlet input(std::move(reader));
+
+  // _inlet_function_value_alternative_schema_start
+  // The alternative is declared before the concrete entry it applies to
+  input.addFunctionAsValueAlternative("scale", inlet::FunctionTag::Double, {});
+  input.addDouble("scale");
```
white238, line 27: I couldnt find it but do you have a test that makes sure they don't declare two concrete alternatives after the Function alternative?

```
  input.addFunctionAsValueAlternative("scale", inlet::FunctionTag::Double, {});
  input.addDouble("scale");
  input.addInt("scale");  // or whatever, would expect failure here
```

kennyweiss: It seems that inlet in axom@develop doesn't currently guard against registering a name twice:
```cpp
  input.addDouble("scale");
  input.addInt("scale");
```
and just returns the first `Field` it finds w/ the same name (regardless of type).

See:
https://github.com/llnl/axom/blob/dda9cc0cc944597519065f30b47728ae5ddfc256/src/axom/inlet/Container.cpp#L360-L369

I'll create an issue for this so we can address it separately.

PR #1963, opened 2026-08-25
Adds missing subscript operator to ArrayIteratorBase

### LINKED ISSUE #1961: Build with clang-22
Build with clang-22 (on Mac) fails. codex points to the following issue:

 Axom’s ArrayIteratorBase is an incomplete random-access iterator.
 
  - It declares std::random_access_iterator_tag but lacks required operator[]: librepos/axom/src/axom/core/ArrayIteratorBase.hpp:31.
  - Axom passes it to std::sort: librepos/axom/src/axom/core/numerics/polynomial_solvers.cpp:164.
  - libc++ 22 changed its sort implementation to use iterator[index]; libc++ 20/21 used *(iterator + index). This exposes the missing operation: build-libs-logs/build_axom_step_1.log:226.
 
This is an Axom iterator-compliance bug, not an MPI, Flang, or configuration problem. The proper fix is to add operator[] to ArrayIteratorBase; temporarily, Clang 21 would avoid exposing it.

kennyweiss:
# Summary

- This PR is a bugfix
- It adds a missing subscript operator to core's `ArrayIteratorBase`, which is necessary to satisfy the random access iterator contract
  - This missing function was responsible for the error triggered in a clang-22 build on Mac (#1961) due to an apparent library change from a dereference to a subscript operator in the implementation of `std::__sift_down`, called by `std::sort`
  - Note: I don't have access to the environment from #1961, but the new tests trigger an error for the missing `operator[]`  without the changes to `ArrayIteratorBase` and are resolved with this fix
- Fixes #1961

### DIFF
--- a/src/axom/core/ArrayIteratorBase.hpp
+++ b/src/axom/core/ArrayIteratorBase.hpp
@@ -6,7 +6,7 @@
 
 #pragma once
 
-#include "axom/core/IteratorBase.hpp"  // for Iterator
+#include "axom/core/IteratorBase.hpp"
 
 namespace axom
 {
@@ -45,14 +45,16 @@ public:
   AXOM_HOST_DEVICE
   ArrayIteratorBase(IndexType pos, ArrayPointerType arr) : BaseType(pos), m_arrayPtr(arr) { }
 
-  /**
-   * \brief Returns the current iterator value
-   */
+  /// \brief Returns the current iterator value
   AXOM_HOST_DEVICE
   ValueType& operator*() const { return m_arrayPtr->flatIndex(BaseType::m_pos); }
 
+  /// \brief Returns the value at offset \a n from the current iterator position
+  AXOM_HOST_DEVICE
+  ValueType& operator[](IndexType n) const { return m_arrayPtr->flatIndex(BaseType::m_pos + n); }
+
 protected:
-  /** Implementation of advance() as required by IteratorBase */
+  /// Implementation of advance() as required by IteratorBase
   AXOM_HOST_DEVICE
   void advance(IndexType n) { BaseType::m_pos += n; }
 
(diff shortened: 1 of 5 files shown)


src/axom/core/ArrayIteratorBase.hpp
```diff
@@ -45,14 +45,16 @@ class ArrayIteratorBase : public IteratorBase<ArrayIteratorBase<ArrayType, Value
...
-   * \brief Returns the current iterator value
-   */
+  /// \brief Returns the current iterator value
   AXOM_HOST_DEVICE
   ValueType& operator*() const { return m_arrayPtr->flatIndex(BaseType::m_pos); }
 
+  /// \brief Returns the value at offset \a n from the current iterator position
+  AXOM_HOST_DEVICE
+  ValueType& operator[](IndexType n) const { return m_arrayPtr->flatIndex(BaseType::m_pos + n); }
+
```
kennyweiss, line 55: This is the bugfix

publixsubfan (approved):

src/axom/core/ArrayIteratorBase.hpp
publixsubfan: Would it make sense to put this in `IteratorBase`?

kennyweiss: Good suggestion, but I don't think so since other iterators that are not random access iterators derive from `IteratorBase`. 

E.g. `ItemCollection` is a forward iterator, and it's advance would give the wrong answer for `operator[](-1)`
See: 
https://github.com/llnl/axom/blob/97f1e492be6c9ed0ffae3dcb45a35d237c0355d1/src/axom/core/ItemCollection.hpp#L207-L214

Arlie-Capps (approved):
Thanks, Kenny!

rhornung67 (approved):
thank you @kennyweiss for the quick fix

PR #1967, opened 2026-08-27
Reduce verbosity of quest's file readers

kennyweiss:
# Summary

- This PR reduces the verbosity of the quest readers and writers
- It also modernizes them via nested namespaces, `constexpr`, `[[nodiscard]]`
- The `[[nodiscard]]` changes the contract and requires callers to check the return values.

### DIFF
--- a/src/axom/quest/examples/containment_driver.cpp
+++ b/src/axom/quest/examples/containment_driver.cpp
@@ -110,7 +110,8 @@ public:
     AXOM_ANNOTATE_SCOPE("load stl");
     quest::STLReader reader;
     reader.setFileName(inputFile);
-    SLIC_ERROR_IF(reader.read() != 0, "Failed to load STL file '" << inputFile << "'.");
+    const int read_status = reader.read();
+    SLIC_ERROR_IF(read_status != 0, "Failed to load STL file '" << inputFile << "'.");
 
     // Create surface mesh
     m_surfaceMesh.reset(new UMesh(3, mint::TRIANGLE));
--- a/src/axom/quest/examples/containment_driver.cpp
+++ b/src/axom/quest/examples/containment_driver.cpp
@@ -110,7 +110,7 @@ public:
     AXOM_ANNOTATE_SCOPE("load stl");
     quest::STLReader reader;
     reader.setFileName(inputFile);
-    reader.read();
+    SLIC_ERROR_IF(reader.read() != 0, "Failed to load STL file '" << inputFile << "'.");
 
     // Create surface mesh
     m_surfaceMesh.reset(new UMesh(3, mint::TRIANGLE));
--- a/src/axom/quest/io/C2CReader.cpp
+++ b/src/axom/quest/io/C2CReader.cpp
@@ -22,13 +22,11 @@
 #include <string>
 #include <utility>
 
-namespace axom
-{
-namespace quest
+namespace axom::quest
 {
 namespace
 {
-c2c::LengthUnit toC2CLengthUnit(utilities::LengthUnit unit)
+constexpr c2c::LengthUnit toC2CLengthUnit(utilities::LengthUnit unit)
 {
   switch(unit)
   {
@@ -138,9 +136,9 @@ C2CReader::ResultType C2CReader::readAssembly(const std::string& filename, Curve
   SLIC_INFO_ROOT(fmt::format("Loading assembly with {} pieces", assembly.getNumEntries()));
 
   // Make an initial guess at the number of curves we may need.
-  const int contoursPerFileGuess = 6;
+  constexpr int contours_per_file_guess = 6;
   CurveArray assemblyCurves;
-  assemblyCurves.reserve(assembly.getNumEntries() * contoursPerFileGuess);
+  assemblyCurves.reserve(assembly.getNumEntries() * contours_per_file_guess);
 
   ResultType ret = ResultType::Success;
   for(auto it = assembly.begin(); it != assembly.end(); it++)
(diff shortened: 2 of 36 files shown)


src/axom/quest/examples/containment_driver.cpp
```diff
@@ -110,7 +110,7 @@ class ContainmentDriver
     AXOM_ANNOTATE_SCOPE("load stl");
     quest::STLReader reader;
     reader.setFileName(inputFile);
-    reader.read();
+    SLIC_ERROR_IF(reader.read() != 0, "Failed to load STL file '" << inputFile << "'.");
```
kennyweiss: We now require users to check the return values of several quest IO functions.
Seems like a worthwhile thing to do, but will require callers to slightly modify their code.
I can roll the `[[nodiscard]]` back if it's controversial.

src/axom/quest/io/C2CReader.cpp
```diff
@@ -172,7 +170,7 @@ C2CReader::ResultType C2CReader::readContour(const std::string& filename, CurveA
   c2c::Contour contour = c2c::parseContour(filename);
   const c2c::LengthUnit c2cLengthUnit = toC2CLengthUnit(m_lengthUnit);
 
-  SLIC_INFO(fmt::format("Loading contour with {} pieces", contour.getPieces().size()));
+  SLIC_INFO_ROOT(fmt::format("Loading contour with {} pieces", contour.getPieces().size()));
```
kennyweiss, line 173: @dylan-copeland -- this is the line that was getting in your way

src/axom/quest/io/C2CReader.hpp
```diff
@@ -77,7 +76,8 @@ class C2CReader
    * 
    * \return A view that contains the curves.
    */
-  CurveArrayView getCurvesView() { return m_nurbsData.view(); }
+  [[nodiscard]] CurveArrayView getCurvesView() { return m_nurbsData.view(); }
```
BradWhitlock: I can understand adding `[[nodiscard]]` to read/write methods where there is a chance at failing due to the file system. `[[nodiscard]]` here seems unnecessary since you would not call these methods if you were not using the curve views for something.

src/axom/quest/io/ProEReader.hpp
```diff
@@ -81,19 +79,19 @@ class ProEReader
...
    */
-  void setFileName(const std::string& fileName) { m_fileName = fileName; };
+  void setFileName(const std::string& fileName) { m_fileName = fileName; }
 
   /*!
    * \brief Returns the number of nodes of the mesh.
    * \return numNodes the number of nodes.
    */
-  int getNumNodes() const { return static_cast<int>(m_num_nodes); };
+  [[nodiscard]] int getNumNodes() const { return static_cast<int>(m_num_nodes); }
```
BradWhitlock: Same `[[nodiscard]]` complaint here.

src/axom/quest/io/STEPReader.hpp
```diff
@@ -59,43 +57,44 @@ class STEPReader
    * \param[in] validate Adds validation tests on the model, when true
    * \return 0 for a successful read; non-zero otherwise
    */
-  virtual int read(bool validate);
+  [[nodiscard]] virtual int read(bool validate);
 
-  std::string getFileUnits() const;
+  [[nodiscard]] std::string getFileUnits() const;
```
BradWhitlock: Same `[[nodiscard]]` complaint and for the next several methods.

src/axom/quest/io/STLReader.cpp
```diff
@@ -16,21 +16,19 @@
 
 namespace
 {
-const std::size_t BINARY_HEADER_SIZE = 80;  // bytes
-const std::size_t BINARY_TRI_SIZE = 50;     // bytes
+constexpr std::size_t binary_header_size = 80;  // bytes
+constexpr std::size_t binary_tri_size = 50;     // bytes
```
BradWhitlock: When I saw the capitalized variable name it was reminiscent of a macro and made me think "this is a constant". As a lower-case variable name it made me think the variable was computed dynamically somewhere close to its use. While this may conform to some coding guideline, I find it less intuitive. My 2 cents.

This comment probably makes more sense further down where the values are used.

src/axom/quest/io/STLReader.cpp
```diff
@@ -132,14 +130,14 @@ int STLReader::readAsciiSTL()
...
+  //    an 80 byte header (binary_header_size)
   //    followed by a 32 bit int encoding the number of faces
-  //    followed by the triangles, each of which is 50 bytes (BINARY_TRI_SIZE)
+  //    followed by the triangles, each of which is 50 bytes (binary_tri_size)
 
   // A local union data structure for triangles in a binary STL
   union BinarySTLTri
   {
-    std::int8_t raw[BINARY_TRI_SIZE];
+    std::int8_t raw[binary_tri_size];
```
BradWhitlock: This is another place where if I saw capital letters, I'd be more likely to think constant whereas I associate lower-case with runtime computed values.

kennyweiss: I agree -- I'll revert the capital -> lowecase constants

src/axom/quest/io/C2CReader.hpp
kennyweiss: Agreed -- I'll remove `[[nodiscard]]` from these. 

The idea was that there is no reason to call these functions if you're not using the values. But I agree they can be a bit heavy handed.

BradWhitlock (approved):
Thanks @kennyweiss !

rhornung67 (approved):
Nice changeset. Thanks @BradWhitlock

PR #1973, opened 2026-08-31
spack recipe: Add v0.15.0 and upstream components fix

bmhan12:
This PR:
- Updates spack recipe
  - Add v0.15.0  
  - Add upstream components fix (PR: https://github.com/spack/spack-packages/pull/6175)

Mirrors spack-packages PR: https://github.com/spack/spack-packages/pull/6294

### DIFF
--- a/scripts/spack/packages/axom/package.py
+++ b/scripts/spack/packages/axom/package.py
@@ -37,6 +37,28 @@ _AXOM_COMPONENTS = (
     "spin",
 )
 
+_AXOM_COMPONENT_REQUIREMENTS = {
+    "bump": ("sidre", "slic", "spin", "primal"),
+    "inlet": ("sidre", "slic", "primal"),
+    "klee": ("sidre", "slic", "inlet", "primal"),
+    "mint": ("slic", "slam"),
+    "mir": ("bump", "sidre", "slic", "slam", "primal"),
+    "multimat": ("slic", "slam"),
+    "primal": ("slic",),
+    "quest": ("slic", "slam", "primal", "mint", "spin"),
+    "sidre": ("slic",),
+    "sina": ("slic",),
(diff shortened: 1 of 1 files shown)


scripts/spack/packages/axom/package.py
```diff
@@ -37,6 +37,28 @@
     "spin",
 )
 
+_AXOM_COMPONENT_REQUIREMENTS = {
```
rhornung67, line 40: Is 'core' not included because everything depends on it? 

Please double check this. It is inconsistent with the dependency figure in the Axom docs. For example, the diagram shows only that sina depends on core (not slic). 

Also, how are optional dependencies handled here? For example quest optionally depends on sidre.

kennyweiss: Correct -- `core` is always on, and cannot be turned off.  
@bmhan12 -- A brief comment might help clarify this.

kennyweiss: This list of requirements enables the minimal configuration of the user's requested components. 
E.g. if they ask for `inlet` via `component=inlet` they will also get `sidre` since `inlet` requires `sidre`. 

If they want more than the minimum, e.g. `inlet` and `spin`  (and their hard dependencies), they can ask for `components=inlet,spin`.

bmhan12: > A brief comment might help clarify this.

Looks like a comment was already added above that section, where the components are listed out: 
https://github.com/llnl/axom/blob/76874e23e3830522250569102fbdfc50346fb2ac/scripts/spack/packages/axom/package.py#L22-L23

rhornung67 (approved):

BradWhitlock (approved):

PR #1974, opened 2026-09-01
DistributedClosestPoint GPU-related optimizations

publixsubfan:
# Summary

This is a grab-bag of various optimizations to improve DistributedClosestPoint query performance:

- DistributedClosestPoint: sort query points by Morton order to reduce divergence
- BVH:
  - Add a `BVH2Node` struct for representing internal nodes. Since HIP architectures appear to do temporal coalescing, this may be more efficient than having separate arrays for the bounding boxes and child node indices by reducing the number of memory waits (`s_waitcnt vmcnt(*)`) 
  - Adds a new method to the BVH traverser interface `traverseTreeShared()` which uses GPU shared memory to cache the most-recently used entries in the BVH stack
 - Primal: use `fmax/fmin` for clamp operation inside `squared_distance(Point, BoundingBox)`
    This seems to improve GPU codegen compared to the generic `axom::utilities::clampVal()` function.

## Performance results

I rebased this branch onto Kenny's branch in #1952, and ran a test with the following setup:
- One node of rzadams, CPX mode, 24 ranks
- Object mesh: 24 domains, 24M points total (1M points per rank)
- Query mesh: 24 domains, 48M cells total (2M cells per rank)

  | query runtime
-- | --
Pre-optimizations | 23.200s
Morton sort query points | 17.798s
BVH optimizations | 17.234s


Arlie-Capps:
@publixsubfan , the first line of your performance table is labeled "Clean."  Does this mean "before optimizations?"

Arlie-Capps (approved):

publixsubfan:
> @publixsubfan , the first line of your performance table is labeled "Clean." Does this mean "before optimizations?"

@Arlie-Capps, yes -- I think that would be a better wording.

rhornung67:
@publixsubfan does this PR address any of the memory issues (i.e., "leaked" memory) observed by user applications?

publixsubfan:
Unfortunately no, this PR won't improve that situation. There's some other potential optimizations I'm looking at that may help with the memory usage.

rhornung67:
> Unfortunately no, this PR won't improve that situation. There's some other potential optimizations I'm looking at that may help with the memory usage.

Thanks. Just checking.

rhornung67 (approved):

src/axom/primal/operators/squared_distance.hpp
```diff
@@ -85,16 +85,28 @@ AXOM_HOST_DEVICE inline double squared_distance(const Point<T, NDIMS>& P,
...
+      cp[i] = fmax(B.getMin()[i], fmin(P[i], B.getMax()[i]));
+    }
+  }
+  else
+  {
+    for(int i = 0; i < NDIMS; ++i)
+    {
+      cp[i] = clampVal(P[i], B.getMin()[i], B.getMax()[i]);
+    }
   }
```
kennyweiss, line 106: Is this optimization generally useful?
Would it make sense to move the specialization for floating_point to `axom::utilities::clampVal`

See: https://github.com/llnl/axom/blob/887b2932fce845e6820b6f9911b6a7904c442341/src/axom/core/utilities/Utilities.hpp#L149-L163

kennyweiss: Alternatively, maybe we should benchmark this change -- 
It could be that it's only beneficial in a device context, and is slower on the host (?)

Claude claims that w/ gcc@13.3, x86-64 (Cascade Lake), the ternary form in `clampVal` will vectorize better and generate fewer assembly instructions than the fmin/fmax form:

Image

src/axom/spin/internal/linear_bvh/bvh_traverse.hpp
kennyweiss, line 1: Should these internal classes be independently unit tested? Based on their assumptions, they seem somewhat brittle.

src/axom/spin/internal/linear_bvh/BVHNode.hpp
```diff
@@ -0,0 +1,39 @@
...
+{
+namespace spin
+{
+namespace internal
+{
+namespace linear_bvh
+{
+
+/*!
+ * \brief Node structure for a 2-wide BVH tree.
```
kennyweiss: What does 2-wide BVH tree mean? Please expand the doxygen here.

kennyweiss (approved):
Thanks @publixsubfan -- impressive speedups!

Do we have sufficient test coverage via the BVH tests, or should we add some tests for the new features/classes?

src/axom/primal/operators/squared_distance.hpp
publixsubfan: Let me benchmark this again on the CPU. I did avoid putting this change in `clampVal` specifically for the reason Claude mentioned: it requires `lower <= upper`, which we check for in this method by calling `bbox.valid()` before the clamp operation.

publixsubfan: Here's numbers for a 1M query/object single-domain run on 1 CPU/GPU with `-O3 -march=native`:

  | host | hip
-- | -- | --
fmin/fmax | 5.70192401 | 0.13014153
clampVal | 5.11066444 | 0.14765718

Good catch, looks like we should add the guard for the CPU side of things. (I did try and see if the compiler could do better with a `__builtin_assume(isfinite(...))` before the fmin/fmax clamp, but no dice)

src/axom/spin/internal/linear_bvh/bvh_traverse.hpp
publixsubfan: I had Codex add some tests for the internal stack classes.

I think there's also an opportunity to improve testing for the device traversal interfaces, but I wanted to clean up the existing BVH testing first, and didn't want to clutter up this PR too much.

src/axom/spin/internal/linear_bvh/BVHNode.hpp
publixsubfan: I tried to clarify the doxygen comment.

2-wide refers to the number of children in an internal node; you could also have 4-wide BVHs in hopes of reducing traversal depth. There's some recent-ish work I wanted to take a look at wrt constructing 4-wide BVHs: https://gpuopen.com/download/HPLOC.pdf

rhornung67 (commented):
@lc-hubcast approve

PR #1981, opened 2026-09-09
Update conduit & mfem dependencies

bmhan12:
This PR:
- Updates conduit to 0.9.8, mfem to 4.10
  - Updates the spack-packages and vcpkg commit hashes as a result. 

Relevant Windows CI build: https://github.com/llnl/axom/actions/runs/34278645661


scripts/spack/configs/versions.yaml
```diff
@@ -14,14 +14,14 @@ packages:
...
+      - spec: "@0.9.8"
   hypre:
     require:
       - spec: "@3.1.0"
-  # master - 7/21/26
+  # 4.10
   mfem:
     require:
-      - spec: "@git.c8ef9676e83c93a9037dbdf3bf0f2d741321e632=master"
+      - spec: "@git.d964264cdb9a13e94a201b6c236c7721e0c8765f=develop"
```
kennyweiss, line 24: I'm guessing the `=develop` is because the `4.10` version isn't in spack-packages yet?

Looks like there is a PR for it: https://github.com/spack/spack-packages/pull/6314

kennyweiss: Since we have an mfem package override, would it make sense to add the mfem@4.10 version there?
I don't think it's worth rebuilding everything just to update this, but if you have to rebuild again, please add it, since that will make it more obvious which mfem version we're using.

See: https://github.com/llnl/axom/blob/develop/scripts/spack/packages/mfem/package.py

kennyweiss (approved):
Thanks @bmhan12

scripts/spack/configs/versions.yaml
bmhan12: > would it make sense to add the mfem@4.10 version there?

I figured once the mfem spack-packages PR went in, we could update the spack-packages commit forward. 
Looks like there might be additional changes to the mfem recipe upstream, and we'll want to capture those changes in that case.

BradWhitlock (approved):
Thanks @bmhan12 !

Arlie-Capps (approved):
Thanks, Brian.

PR #1987, opened 2026-09-21
Reject a Pro/E tetrahedron that names a node the file does not define

arpitjain099:
`ProEReader::read` reads each tetrahedron's node ids from the file and uses them without checking them. The bounding-box predicate then does `Point3D p(&nodes[3 * (tet_nodes[i] - 1)], 3)`, so an id above the declared node count, or 0, indexes `m_nodes` out of range.

A four-node file whose single tet names node 5 is enough. Running the shipped example built with `-fsanitize=address`:

```
ERROR: AddressSanitizer: heap-buffer-overflow
READ of size 8
    NumericArray<double, 3>::NumericArray(double const*, int) NumericArray.hpp:406
    primal::Point<double, 3>::Point(double const*, int) Point.hpp:119
```

The same file through `mesh_converter` stops on a `SLIC_ASSERT_MSG` about element 4 in a map of size 4, but that macro compiles to `((void)0)` with `AXOM_DEBUG` off, so the check is not there in a release build, and the reader is the common path either way.

The fix validates the ids where they are read, with the warning-and-return-minus-one the function already uses for a file it cannot handle. Added a negative test next to `read_missing_file`; it fails on develop and the file's 12 tests pass with the change.


rhornung67:
@kennyweiss please take a look at this and see if this is something we want.

PR #1988, opened 2026-09-21
Add a cleanmesh option to ElviraAlgorithm.

BradWhitlock:
This PR adds a "cleanmesh" option to the `axom::mir::ElviraAlgorithm` options so we can set cleanmesh=0 to skip the phase of the algorithm where for 3D meshes we attempt to merge like vertices and like polyhedral faces. This work is not always needed if you do not care about the quality of the mesh - and it adds a bunch of time.

Other:
* Added a --cleanmesh=on|off option to the MIR heavily_mixed example program
* Enhanced testing for "coupled" tests (MIR + TopologyMapper) so we try Elvira MIR with _cleanmesh_ set both ways to show that it does not negatively impact sending "messy" meshes into the TopologyMapper.


bmhan12 (approved):

rhornung67 (approved):

Arlie-Capps (approved):
Thanks, Brad!

RELEASE-NOTES.md
```diff
@@ -18,10 +18,12 @@ The Axom project release numbers follow [Semantic Versioning](http://semver.org/
 
 ## [Unreleased] - Release date yyyy-mm-dd
 
+### Added
 - Axom's host execution-space default allocator is now a configure-time policy. The default policy is malloc, regardless
   of whether Axom is configured with Umpire enabled. Umpire builds may opt into the Umpire `HOST` resource with
   `-DAXOM_DEFAULT_HOST_ALLOCATOR=UMPIRE_HOST`. Runtime per-use selection remains available through existing explicit
   allocator-ID arguments.
```
kennyweiss, line 25: This bullet was missing a heading in axom@develop. 

I'd consider it under the `### Changed` heading, rather than `### Added`.

src/axom/mir/detail/elvira_detail.hpp
```diff
@@ -50,6 +50,12 @@ class ELVIRAOptions : public axom::bump::Options
    * \return True if the output should be a point mesh, false otherwise.
    */
   bool pointmesh() const { return flagValue("pointmesh", false); }
+
+  /**
+   * \brief Get whether like coordinates and faces should be merged to make a better mesh.
+   * \return True if we should clean the mesh, false otherwise.
+   */
+  bool cleanMesh() const { return flagValue("cleanmesh", true); }
```
kennyweiss, line 58: Is there a corresponding function to set the "cleanmesh" flag? 
Or is it implied by the (optional) user-provided the flag to override the default (of `true`)

Are the valid flags documented somewhere?

kennyweiss (approved):
Thanks @BradWhitlock

RELEASE-NOTES.md
BradWhitlock: I moved my bullet to under `### Changed` and added some of other typical sections that were missing also.

src/axom/mir/detail/elvira_detail.hpp
BradWhitlock: That class gets the options from a user-supplied `conduit::Node`. It's meant to provide default values for when the user does not give a value. In this case, `cleanmesh` defaults to true, which does not change the algorithm behavior but lets me turn off extra work if I choose.

I forgot to update the MIR sphinx docs so I did that in a recent commit. The flags are documented there.

PR #1989, opened 2026-09-22
Improve mir benchmark scripts and heavily_mixed example

BradWhitlock:
This PR renames and improves the MIR benchmark script so it works with the `mir_heavily_mixed` example program instead of just `mir_concentric_circles`.

Other:
* Changed from `int` to `conduit::index_t` in the heavily_mixed example to fix some indexing overflow for larger meshes
* Enhanced the heavily_mixed example so it can save/reload the mesh it generates, which is handy for running multiple instances in a row (usually with different execution policies) since they can share the same generated mesh


BradWhitlock (commented):
@lc-hubcast approve

rhornung67 (approved):

kennyweiss (approved):

Arlie-Capps (approved):

